PR #1, opened 2019-02-10
feat: port a lot of functionality from bull 3.x

manast:


manast:
@stansv the tests pass now, I think this could be merged into master if you approve it. A couple of tests are still missing from bull 3.x but I can add them in the following days.

PR #3, opened 2019-07-23
port sandbox processors implementation and tests from bull 3.x

stansv:
The `should process with concurrent processors` test in `test_sandboxed_process.ts` fails from time to time: I noticed that `completed` event handler sometimes has not invoked all 4 times.

Before this event is emitted in `worker.ts`, we're awaiting for `job.moveToCompleted()`, which executes the `moveToFinished-6.lua`; looks like the promise neither resolves not rejects. I've already tried to increase timeout to 60s, with no luck.


manast:
thanks for this first PR. One hint for debugging lua scripts is to use the MONITOR command in redis, you can call it like this:
```
redis-cli MONITOR
```
then you can see what happens in the script and see if moveToCompleted is behaving strangely. I find quite strange that it does not complete, unless it is the redis connection that has been disconnected without waiting for the command to complete, but I would expect a rejection then.
I also noticed that some of the files have not the standard formatting, are you using prettier? I always have the vs code prettier plugin installed to easily get the same formatting on all the files.

stansv:
I've spent some hours debugging this, still no results. I've tried to use `MONITOR` already and I can see lua script not executed at all when this happens.

I've simplified test and made it more reproducible

```
describe('test ioredis stuck', () => {

    it('test', async function() {
        this.timeout(3000);

        const queue = new Queue("any-2" + v4());

        const worker = new Worker("any", async () => {}, {});
        await worker.waitUntilReady(); // [1]

        const promises = [];
        for(let i = 0; i < 100; i++) {
            promises.push(queue.append("test", { "foo": "bar" }, {}));
        }
        await Promise.all(promises);

        console.log("calling moveToFinished...");

        await worker.waitUntilReady(); // [2]
        await (<any>worker.client).moveToFinished(
            ["", "", "", "", "", ""]
        );

        console.log("done!");

        return true;
    });

});
```

The behavior differs depending on a moment when waitUntilReady() is called (`// [1]` and `// [2]`). When it happens before adding 100 jobs to a queue, test _almost_ always fails with timeout. Otherwise, if `// [1]` is removed, test always passes.

I've tested it on both local Redis instance (running in Docker) as well as remote one deployed in cloud.

Will continue to debug.

manast:
ok. Seems like this error is related to BRPOPLPUSH. If a redis client instance calls that command the client gets busy until the command completes. It has a timeout of 5 seconds, so in theory after those 5 seconds the next command in the queue should be executed. I will test why it does not timeout correctly.

manast:
and, this is the reason we need to reserve a redis connection for the worker, since we can never know if the connection is busy waiting for a job to arrive.

manast:
ok, found the problem, the timeout is in seconds, and default is 5000, so it will wait a very long time before the connection is freed again. redis documentation is not clear about the unit used in the timeout: https://redis.io/commands/brpoplpush

manast:
I commit a fix into my branch: https://github.com/taskforcesh/bullmq/pull/1/commits/c6cfe7c0b50cabe5e5eb31f4b631a8b1d3706611
If you update your branch the test should work now. You can change the drainDelay option in the test if you want a faster test.

stansv:
I suspect that "should get all workers for this queue" test in `test_worker.ts` is unstable: it failed one time on my machine with "expected 1, actual 0" on 67d4ebe without any code changes.

@manast, could you please review this PR, I think we can refactor the code a bit

manast:
regarding lock files. I think I do not understand them so well yet. I always thought that a library with a lockfile will be used when using the library so that all the library dependencies are locked to the version they had when the package was released in NPM. But recently I learnt that that is not the case, so I am not sure if we should or should not keep the lockfile.

src/classes/sandbox.ts
```diff
@@ -0,0 +1,63 @@
+//TODO remove for node >= 10
+require('promise.prototype.finally').shim();
+
+const sandbox = (processFile: any, childPool: any) => {
+  return function process(job: any) {
```
manast: this can be simplified to something like this:
```ts
const child = await childPool.retain(processFile);

let msgHandler: any;
let exitHandler: any;

child.send({
  cmd: "start",
  job: job.toJSON()
});

const done = new Promise((resolve, reject) => {
  msgHandler = (msg: any) => {
    switch (msg.cmd) {
      case "completed":
        resolve(msg.value);
        break;
      case "failed":
      case "error": {
        const err = new Error();
        Object.assign(err, msg.value);
        reject(err);
        break;
      }
      case "progress":
        job.updateProgress(msg.value);
        break;
      case "log":
        job.log(msg.value);
        break;
    }
  };

  exitHandler = (exitCode: any, signal: any) => {
    reject(
      new Error("Unexpected exit code: " + exitCode + " signal: " + signal)
    );
  };

  child.on("message", msgHandler);
  child.on("exit", exitHandler);
});

try {
  await done;
} finally {
  child.removeListener("message", msgHandler);
  child.removeListener("exit", exitHandler);

  if (child.exitCode !== null) {
    childPool.remove(child);
  } else {
    childPool.release(child);
  }
}
```

src/classes/child-pool.ts
```diff
@@ -0,0 +1,112 @@
...
+        getPort().then((port: any) => {
+          return `${argName}=${port}`;
+        }),
+      );
+    }
+  });
+
+  return Promise.all(promises).then(convertedArgs => {
+    return standard.concat(convertedArgs);
+  });
```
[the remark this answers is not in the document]
stansv: Not exactly, because there's also inner callback used to iterate with `_.forEach`. I'll fix this a bit different.

src/classes/child-pool.ts
```diff
@@ -0,0 +1,112 @@
...
+    child.send({ cmd: 'init', value: processFile }, resolve);
+  });
+};
+
+export class ChildPool {
+  retained: any = {};
+  free: any = {};
+
+  constructor() {
+    // todo for what this check is needed? to implement singleton?
```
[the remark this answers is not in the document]
stansv: Didn't know, thanks!

src/classes/master.js
```diff
@@ -0,0 +1,121 @@
...
+      }, this);
+
+      return alt;
+    },
+    configurable: true,
+    writable: true,
+  });
+}
+
+process.on('message', msg => {
```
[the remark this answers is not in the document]
stansv: are we OK to use modern ES features in raw JS code?

src/classes/sandbox.ts
stansv: had to add `return done` after `await done` as well

stansv:
As far as I know, `package-lock.json` is needed mainly for build stability. Npm resolves dependency tree from package.json where you can use semver range expressions instead of fixed versions of any package, the same with all transitive dependencies. Actual versions of all packages is written into lock file along with package hashes, because `npm install` may give different results at a different moments of time without this lock file. But the lockfile itself not included into release package.

stansv:
Looks like we need to return that `finally`-enabling polyfill back, tests are failing on Node 8 :(

    Promise.resolve(...).then(...).finally is not a function

manast:
@stansv it must be the tsconfig for the target, ES2017 may not be compatible with node 8, but we can just change it to ES2016 :). Thats the good thing about typescript, we can support older versions of node and still use the most modern features.

stansv:
Forget to say, this is `master.js` used to bootstrap forked process. We can try to compile from TS but I'm not sure if `__dirname` would work properly in all cases to find compiled version inside `/dist` dir. Also I noticed that tsc not configured to copy js files to `/dist` as is, I can fix this.

src/classes/sandbox.ts
```diff
@@ -0,0 +1,71 @@
...
+      child.removeListener('exit', exitHandler);
+
+      if (child.exitCode !== null) {
+        childPool.remove(child);
+      } else {
+        childPool.release(child);
+      }
+    }
+
+    // return done.finally(() => {
```
manast: remove out-commented code

src/interfaces/advance-opts.ts
```diff
@@ -1,24 +1,24 @@
 export interface AdvancedOpts {
```
manast: the file should be called ```advanced-opts```

manast: (I now it was my fault :).

PR #7, opened 2019-08-13
Compatibility class implementing bull3 api in bullmq

### LINKED ISSUE #4: Create a compatibility class for bull 3.x
In order to simplify migration between 3.x and 4.0 we should have a compatibility class. 
In 4.0 we modularized by dividing responsibilities in different classes such as client, worker, events, scheduler, etc. In 3.0 on the other hand we have a unique class "Queue" that provides the same functionality.
We could create a class that mimics 3.x as close as possible (although probably we cannot require it to be 100% backwards compatible), by wrapping all the other classes. It will not be as efficient in terms of redis client usage but it will make it much easier for 3.x users to upgrade.

stansv:
Work is still in progress, there're a lot of things to improve:

review and port all tests
implement missing features in bullmq
refactor initialization of bullmq worker (there're some drawbacks in current implementation)
do some manual testing as well
Resolves #4


src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+   * Everytime the same queue is instantiated it tries to process all the old jobs
+   * that may exist from a previous unfinished session.
+   */
+  constructor(queueName: string, opts?: QueueOptions3);
+  constructor(queueName: string, url: string, opts?: QueueOptions3);
+
+  constructor(queueName: string, arg2?: any, arg3?: any) {
+    super();
+
+    Object.defineProperties(this, {
```
manast: not sure I understand the purpose of ```Object.defineProperties``` in this context :/

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+        enumerable: false,
+        writable: true,
+      },
+      workerPromiseReject: {
+        enumerable: false,
+        writable: true,
+      },
+      client: {
+        get: () => {
+          return (this.queueScheduler as any).connection.client;
```
manast: why queueScheduler and not just queue?

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+    //
+    // We cannot use ioredis keyPrefix feature since we
+    // create keys dynamically in lua scripts.
+    //
+    if (this.opts.redis && this.opts.redis.keyPrefix) {
+      delete this.opts.redis.keyPrefix;
+    }
+
+    // Forcibly create scheduler to let queue act as expected by default
+    this.getQueueScheduler();
```
manast: I think a QueueScheduler is only needed in workers. so it could be instantiated only when the process function is defined.

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+      this.workerPromiseReject = reject;
+    }).catch(e => this.worker);
+  }
+
+  /**
+   * Returns a promise that resolves when Redis is connected and the queue is ready to accept jobs.
+   * This replaces the `ready` event emitted on Queue in previous verisons.
+   */
+  async isReady(): Promise<this> {
+    await this.getQueueScheduler().waitUntilReady();
```
manast: why queueScheduler?

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+    callback: ProcessCallbackFunction3<T>,
+  ): Promise<void>;
+  process(
+    name: string,
+    concurrency: number,
+    callback: ProcessPromiseFunction3<T>,
+  ): Promise<void>;
+  process(name: string, concurrency: number, callback: string): Promise<void>;
+
+  process(arg1: any, arg2?: any, arg3?: any): Promise<void> {
```
manast: for the process function I think we can have something that matches the new signature in Bull 4, even if we are not 100% backwards compatible, because keeping this code for handling all the parameter combinations is not nice :).

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+  add(data: T, opts?: JobOptions3): Promise<Job3<T>>;
+
+  /**
+   * Creates a new named job and adds it to the queue.
+   * If the queue is empty the job will be executed directly,
+   * otherwise it will be placed in the queue and executed as soon as possible.
+   */
+  add(name: string, data: T, opts?: JobOptions3): Promise<Job3<T>>;
+
+  async add(arg1: any, arg2?: any, arg3?: any): Promise<Job3<T>> {
```
manast: as with the process case, I think we can keep the new "append" signature from Bull 4, even if it is not stricly compatible with bull 3. I see this compat layer as an easier middle migration step, it does not need to be 100% compatible with 3.

src/classes/compat.ts
```diff
@@ -0,0 +1,2621 @@
...
+  private delay: number;
+  private failedReason: string;
+
+  private job: Job;
+
+  constructor(queue: Queue3, data: any, opts?: JobOptions3);
+  constructor(queue: Queue3, name: string, data: any, opts?: JobOptions3);
+
+  constructor(queue: Queue3 | InternalJobAndQueueWrapper, arg2: any, arg3?: any, arg4?: any) {
+    Object.defineProperties(this, {
```
manast: can you provide some info regarding this pattern?, I am not used to it.

stansv:
Yes, I had to make a lot of work to learn underlying mechanisms of queues and jobs operation. My plan was to take advantage of existing unit tests to make sure I correctly use new API and everything worked as expected, also to expose what features we're missing in bullmq in comparison to v3. But unit tests are also relying on internal, hidden properties of objects a lot.

I think once we have a stable tested version of this compat class, we can refactor and simplify it then, remove extra features and reduce amount of tests.

src/classes/compat.ts
stansv: This is just to hide some new internals from users, which is not itself very important, and define getters/setters to transparently access state of these internal bullmq classes.

src/classes/compat.ts
stansv: yeah I utilizing lazy initialization a lot in order to reduce connections to Redis, and I thought QueueScheduler class is super important and should be always created with Queue3. But it makes sense to only create it in conjunction with worker, you're right

stansv: So I mean we can use the Queue class itself here, just get it via getQueue(), not directly

src/classes/compat.ts
stansv: Agreed

src/classes/compat.ts
stansv: Here is the same idea as with Queue3 class — we have internal job property which is bullmq instance, and this we define getters/setters to access the same properties on that internal job. But, maybe we get rid of Job3 class at all in favor of its bullmq counterpart.

manast: I think it is better to use the new Job class, it is quite similar anyway :).

src/classes/compat.ts
manast: yes, my idea with separating the functionality was so that we can minimize the amount of connections. In a queue with many workers maybe only two or three will have a queue scheduler to reduce unnecessary connections and performing unnecessary work.

manast:
as mentioned, it is ok if the layer is similar to bull 3, I do not see it will be possible to make it 100% compatible anyway, if it is 80% compatible I am happy, and it will help a lot for the users to move to upgrade to version 4.

src/classes/compat.ts
stansv: Didn't you want to rename `append` back to `add`? :)

stansv:
What can be simplified:
- replace Job3 wrapper with bullmq Job class
- get rid of bull3 opts interfaces in favor of its bullmq counterparts, where possible. Old & new configuration options generally look similar, but new interfaces are separated for each of 4 entities (queue, scheduler, worker, events). Queue3 constructor should accept opts of combined type.
- rm named processors at all, use the same signature as Worker.process(), but process() should fail if called twice
- fail fast with informative message if user requested to register handler for event emitted only by worker, but there wasn't process() call so no worker created. Generally, events emitted by worker have no much sense if one do not want to process any jobs locally. With this we would get rid of ugly onWorkerInit() handlers.

src/classes/compat.ts
```diff
@@ -0,0 +1,962 @@
...
+      throw new Error('Queue3.process() cannot be called twice');
+    }
+
+    this.worker = new Worker(this.name, processor, this.opts);
+    this.queueScheduler = new QueueScheduler(this.name, this.opts);
+    return this.worker.waitUntilReady();
+  }
+
+  append(jobName: string, data: any, opts?: JobsOpts): Promise<Job> {
+    if (opts && opts.repeat) {
```
manast: this is already done in the queue4 class.
Btw, maybe we should just rename append to add everywhere to avoid confusions.

src/classes/compat.ts
```diff
@@ -0,0 +1,962 @@
...
+  /**
+   * ???
+   */
+  async nextRepeatableJob(
+    name: string,
+    data: any,
+    opts?: JobsOpts,
+    skipCheckExists?: boolean,
+  ): Promise<Job> {
+    return this.queue.repeat.addNextRepeatableJob(
```
manast: ok, I need to add the equivalent method for this.

manast:
Is there anything left to do or can it be merged now?

stansv:
Just a couple of things
1) unstable test merged from development-4.0 (#16)
2) get rid of direct access to queue.repeat (#15)

Both can be fixed after merge, so you can merge now.

Issue #17, opened 2019-08-26
Refactor Options interfaces


stansv:
- names should be consistent (use either "Opts" or "Options" suffix)
- find and remove ignored options like createClient

Issue #21, opened 2019-09-01
Add option for trimming event stream


manast:
Currently all events are stored in the event stream, this will grow indefinitely until running out of memory, therefore we need an option to trim to a maximum number of events.

PR #27, opened 2019-09-08
Feat/add trim option #21

manast:
also refactors options to fix issue #17


stansv:
Don't we want to merge into development branch first?

PR #51, opened 2019-10-27
Save job stacktrace on failure

### LINKED ISSUE #31: Stacktrace is not saved (always empty array) on Job failure
Found in 3cf2617 (currently this is a feat/add-trim-option-#21)

Test code to reproduce

```
  const queue = new Queue("test", {  });
  queue.add("test", {  }, { stackTraceLimit: 100 });

  const worker = new Worker("test", async (job) => {
    throw new Error("Fail!");
  });
```

This is covered by tests ("marks the job as failed" in "test_job.ts") but I still see empty stacktrace in Redis

Screen Shot 2019-09-29 at 2 13 47 PM

stansv:
Fixes #31.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -460,6 +460,7 @@ export class Job {
     if (this.opts.stackTraceLimit) {
       this.stacktrace = this.stacktrace.slice(0, this.opts.stackTraceLimit - 1);
     }
+    this.stacktrace.push(err.stack);
 
     const params = {
       attemptsMade: this.attemptsMade,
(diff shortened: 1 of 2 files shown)


src/classes/job.ts
```diff
@@ -460,14 +460,14 @@ export class Job {
     if (this.opts.stackTraceLimit) {
       this.stacktrace = this.stacktrace.slice(0, this.opts.stackTraceLimit - 1);
     }
+    this.stacktrace.push(err.stack);
```
manast: slightly simpler to push first and then check for stackTraceLimit (without - 1 then)

PR #52, opened 2019-10-27
[Draft] Job locking mechanism

### LINKED ISSUE #47: Implement job locking mechanism
As in Bull 3 we need a locking mechanism where a worker needs to pick a lock everytime it starts processing a job. Every worker and every concurrent processor withing a worker needs to have a uuid that will be stored in the lock, by doing this we can also know which worker that is processing a given job and also if a worker has lost the lock  (which in the future should imply a cancellation).

stansv:
Fixes #47.


src/classes/job.ts
```diff
@@ -249,7 +251,8 @@ export class Job {
     //
     // Check if an automatic retry should be performed
     //
-    var moveToFailed = false;
+    let moveToFailed = false;
+    // FIXME why don't we moveToFailed with fetchNext = true?
```
manast: don't remember but there was a reason for it. Since moving to failed is a special case it does not affect performance anyway.

src/classes/scripts.ts
```diff
@@ -112,9 +125,11 @@ export class Scripts {
     propVal: string,
     shouldRemove: boolean | number,
     target: string,
+    token: string,
     fetchNext = true,
   ) {
     const queueKeys = queue.keys;
+    const opts: WorkerOptions = <WorkerOptions>queue.opts;
```
manast: Maybe queue should be a Worker instead of QueueBase, then you avoid forcing opts to be the WorkerOptions. As far as I know only workers will call this function anyway.

src/classes/scripts.ts
```diff
@@ -141,10 +156,10 @@ export class Scripts {
       target,
       remove,
       JSON.stringify({ jobId: job.id, val: val }),
-      !fetchNext || queue.closing || (<WorkerOptions>queue.opts).limiter
-        ? 0
-        : 1,
+      !fetchNext || queue.closing || opts.limiter ? 0 : 1,
       queueKeys[''],
+      token,
+      opts.lockDuration,
```
manast: why do we need lockDuration for moving to finished status?

src/classes/worker.ts
```diff
@@ -108,35 +113,55 @@ export class Worker extends QueueBase {
...
+    const tokens: string[] = Array.from({ length: opts.concurrency }, () =>
+      uuid.v4(),
+    );
 
     while (!this.closing) {
       if (processing.size < opts.concurrency) {
-        processing.add(this.getNextJob());
+        const token = tokens.pop();
+        processing.set(
+          this.getNextJob(token).catch(error => {
```
manast: I don't think this catch is necessary, since we are awaiting on the Promise.race() below which will throw an exception if the promise is rejected. This can be handled on the call to run instead.

src/classes/worker.ts
```diff
@@ -108,35 +113,55 @@ export class Worker extends QueueBase {
...
+      const token = processing.get(completed);
       processing.delete(completed);
 
       const job = await completed;
-
       if (job) {
-        processing.add(this.processJob(job));
+        // reuse same token if next job is available to process
+        processing.set(
+          this.processJob(job, token).catch(error => {
```
manast: same comment as above, should not be needed.

src/classes/worker.ts
```diff
@@ -202,16 +233,56 @@ export class Worker extends QueueBase {
...
+    // end copy-paste from Bull3
+
+    const handleCompleted = async (result: any): Promise<Job | void> => {
       const jobData = await job.moveToCompleted(
         result,
+        token,
         !(this.closing || this.paused),
       );
       this.emit('completed', job, result, 'active');
+      // FIXME should we call nextJobFromJobData here to emit drained event?
```
manast: not needed, this is handled correctly already.

src/classes/worker.ts
```diff
@@ -224,19 +295,22 @@ export class Worker extends QueueBase {
         error = (<any>error).cause; // Handle explicit rejection
       }
 
-      await job.moveToFailed(err);
+      await job.moveToFailed(err, token);
       this.emit('failed', job, error, 'active');
+      // FIXME can we also fetch next job right away as in handleCompleted?
```
manast: fail is more complex since it has to handle retries, etc, so we can skip fetching next job, its just an optimization, not so important in the case of failure.

src/classes/worker.ts
```diff
@@ -316,11 +390,57 @@ export class Worker extends QueueBase {
...
           this.childPool && this.childPool.clean();
         }
         this.emit('closed');
         resolve();
       });
     }
   }
+
+  // code from former TimerManager
+  private timers: any = {};
```
manast: we should refactor this into a separate class

src/classes/scripts.ts
stansv: this is to take next job with rpoplpush in a single script execution (to avoid extra roundtrip)

stansv: i.e. we can lock next job right away

src/classes/scripts.ts
stansv: should we also do the same in moveToCompleted/moveToFailed? this is depends on do we allow to take/finish job "manually" without worker

src/classes/worker.ts
stansv: In this case we're exiting from worker loop

src/classes/scripts.ts
manast: I think in the case of taking manually we still should use the Worker class for that.

src/classes/worker.ts
manast: that will not work, the loop will continue anyway since you are not awaiting anyway...

src/classes/scripts.ts
stansv: OK, we can make processFn optional for this later

Issue #37, opened 2019-10-12
Fix unstable tests


stansv:
This issue is to track unstable tests (collected from travis build logs https://travis-ci.org/taskforcesh/bullmq)

```
Pause
  should be able to pause a running queue and emit relevant events
```
- timeout (#159.3, #161.1, #161.2, #162.3, #162.1, #163.2, #163.3, #167.3, #170.2, #170.3)

```
Compat
  events
    emits drained and global:drained event when all jobs have been processed
```
- timeout (#134.3)

```
repeat
  should emit a waiting event when adding a repeatable job to the waiting list
```
- Error: Connection is closed. (#119.2, #123.3, #134.1, #140.2)

```
Delayed jobs
  should process a delayed job only after delayed time
```
- timeout (#196.2, #214.2)

stansv:
For now it seems all tests are fixed so build became stable enough (tested on my local env, latest CI builds are also succeeded). I'll reopen this in case if new unstable tests will appear.

stansv:
```
Delayed jobs
  should process delayed jobs in correct order
```
- timeout (#281.1)

PR #58, opened 2019-11-04
Fix #57

### LINKED ISSUE #57: Override queue default job options when options are provided to add method
I think there is a problem with queue `defaultJobOptions` handling.
Options provided to Queue constructor are not overrided by options provided during job creation.
```
const queue = new Queue(
  "my-queue",
  {
    connection,
    defaultJobOptions: {
      attempts: 2,
      backoff: {
        type: "fixed",
        delay: 2000,
      },
    },
  },
);
const job = await queue.add(
  "my-job",
  payload,
  {
    attempts: 3,
    backoff: {
      type: "fixed",
      delay: 5000,
    },
  },
);
console.log(job.opts);
```
It print defaultJobOptions instead of provided options.
`opts` parameter provided to the `add` method should be merged into `defaultJobOptions`.

HNicolas:
I updated job options merging in order to override default job options when options are provided at job level.
Maybe a deep merge between the two objects could be even better.

closes #57


manast:
thanks for the fix. One thing, we require conventiona commit messages, normally the commit prehook should warn you about this, but for some reason it did not work for you. So in this case the message should be: "fix: default job settings #58"

stansv:
Looks like there's one more unstable test (#37), I'll take a look

Issue #70, opened 2019-11-19
Job data field is doubly serialized


jbr:
[This](https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L151) seems to be the line in question. As a result of this, `JSON.parse(JSON.stringify(job)).data` is always a string, which was quite surprising. My particular use case is sending a Job over a socket serialized as json. It's easy enough to work around this, but since this is still an early release it seemed worth mentioning that this might surprise others as well.

manast:
Yes, the problem is that toJSON should be called something different, because it is not really meant to be used by JSON.stringify.

PR #74, opened 2019-11-20
fix: avoid changing Job property types through JSON serialization/deserialization #70

jbr:
renames Job#toJSON to Job#asJSON and adds a test

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -144,7 +144,7 @@ export class Job {
     }
   }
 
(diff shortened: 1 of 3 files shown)


jbr:
@manast 5752727 strips out the queue from JSON serialization, which is probably what people would want when serializing a Job. It might also make sense just to remove the test and leave the toJSON→asJSON rename, since JSON serialization of Jobs isn't necessarily part of the supported API.

PR #75, opened 2019-11-21
feat: remove dependence on Bluebird and delay #67

### LINKED ISSUE #67: Remove Bluebird
Just from a brief glance, it looks like only `delay` is being used from Bluebird.  It really is a no-starter for us to include libs with Bluebird since it seriously hinders the async guarantee in the server env.

Is this something I could help with?

PS: so excited to see a rewrite of Bull!  Thanks for all the hard work

jbr:
This approach uses a simple setTimeout-based delay in the only production
usage (src/classes/queue-events.ts), defined in src/util.ts, and since
there was already a devDependency on a library called
[delay](https://www.npmjs.com/package/delay), this commit uses that in
test contexts. It should be easy to replace that dep with a
`(ms:number) => new Promise(r => setTimeout(r, ms))` as well.

Closes #67

### DIFF
--- a/src/test/fixtures/delay.js
+++ b/src/test/fixtures/delay.js
@@ -1,0 +1,5 @@
+module.exports = function delay(ms) {
+  return new Promise(function(resolve) {
+    return setTimeout(resolve, ms);
+  });
+};
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
   },
   "dependencies": {
     "@types/ioredis": "^4.0.13",
-    "bluebird": "^3.5.3",
     "cron-parser": "^2.7.3",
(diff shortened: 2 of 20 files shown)


src/test/fixtures/delay.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function delay(ms) {
```
manast: why cant we use the delay implementation in utils?

jbr: That would be requiring ts from js, which seemed like a substantial change to the fixtures

manast: ah ok, got it.

Issue #76, opened 2019-11-21
If QueueEventsOptions lastEventId is not set to `$`, it replays the full (all-time) event history on process restart


jbr:
Possible solutions:
* Mark stream events as processed with [XACK](https://redis.io/commands/xack)
* Set the default lastEventId to `"$"` instead of `"0-0"`, only streaming new events. Downside of this approach: Events are ignored during QueueEvents listener unavailability

stansv:
Hmm, I  thought we've already replaced `"0-0"` to `"$"` (https://github.com/taskforcesh/bullmq/issues/29)..

jbr:
@stansv https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-events.ts#L29 is `0-0` and I don’t see any other redis commands than the XREAD

jbr:
Would a PR that uses XREADGROUP be welcome?

manast:
@jbr the code above regarding XTRIM is for internal delayed jobs events, 100 is a extremely safe value so that we do not need to rely on watchdogs as we had in Bull3.

Regarding QueueEVents, the workaround is to pass '$' to only receive new events. I thought I had changed that already but for some reason is not the default value yet.

PR #82, opened 2019-12-06
$ as default queue event id #76

jbr:
closes: #76

### DIFF
--- a/src/classes/compat.ts
+++ b/src/classes/compat.ts
@@ -42,11 +42,10 @@ export class Queue3<T = any> extends EventEmitter {
    * The name of the queue
    */
   name: string;
+  queueEvents: QueueEvents;
 
   private opts: CommonOptions;
-
   private readonly queue: Queue;
-  private queueEvents: QueueEvents;
   private worker: Worker;
   private queueScheduler: QueueScheduler;
 
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -25,7 +25,7 @@ export class QueueEvents extends QueueBase {
     const opts: QueueEventsOptions = this.opts;
 
     const key = this.keys.events;
-    let id = opts.lastEventId || '0-0';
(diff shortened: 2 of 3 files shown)


manast:
thanks for the PR. Why can't you use ```waitUntilReady``` from the base class? seems like the effect will be the same.

manast:
regarding consumer groups, I do not see how that would be useful in this case. Exactly once semantics is already provided by the current implementation, you just need to save the last event id and use it the next time you start listening to events.

jbr:
@manast I was hesitant to use QueueBase#waitUntilReady because of this comment: https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-base.ts#L59. Will it make whatever refactor is going to be necessary on deprecation easier to use the same interface? What will waitUntilReady be replaced with on deprecation?

jbr:
Updated to use QueueEvents#waitUntilReady

manast:
@jbr if I recall correctly, the idea was to deprecate it as public method, and just use it internally when needed.

jbr:
@manast As far as consumer groups (and [`XREADGROUP`](https://redis.io/commands/xreadgroup)), my read of the streams documentation suggested to me that the use of [`XACK`]( https://redis.io/commands/xack) would effectively mark an individual message as "done" for that consumer group, ensuring that it wouldn't be replayed. Storing the most recent id seems like it surfaces internal implementation details of the library in a way that most events users shouldn't _need_ to care about, and pushes the responsibility of storing ids in redis onto them.  If the preferred approach is to store the most recently processed event id instead of using consumer groups, it seems like _that_ behavior should be handled by bull internally

manast:
@jbr consumer groups will not work for this use case since you want all the events to be delivered to all the workers.

jbr:
@manast Is there more work needed on this PR? It uses waitUntilReady and $ and makes no other changes.

PR #91, opened 2019-12-13
fix: check of existing redis instance

### LINKED ISSUE #90: Redis connection does not duplicate correctly in worker
Hi,

I've discovered a weird behavior when using the `concurrency` option in a worker.

I've reproduced the bug [here](https://github.com/Embraser01/bullmq-issue-85/tree/concurrency) (On the `concurrency` branch).

When using a worker with concurrency set, each time jobs are added to the queue, the worker will take more time to complete.

On the repo run:

```bash
node worker.js 

# On another shell
node queue.js
```

Example of a run:
```
Running WithoutConcurrency...
Results for WithoutConcurrency run 0: creation(82ms) duration(2546ms) total(2628ms)
Results for WithoutConcurrency run 1: creation(2ms) duration(108ms) total(110ms)
Results for WithoutConcurrency run 2: creation(2ms) duration(107ms) total(109ms)
Results for WithoutConcurrency run 3: creation(1ms) duration(107ms) total(108ms)
Running WithConcurrency...
Results for WithConcurrency run 0: creation(22ms) duration(2020ms) total(2042ms)
Results for WithConcurrency run 1: creation(2ms) duration(5567ms) total(5569ms)
Results for WithConcurrency run 2: creation(2ms) duration(16022ms) total(16024ms)
Results for WithConcurrency run 3: creation(1ms) duration(25514ms) total(25515ms)
```

Edited: The test case has been updated

Embraser01:
Should fix #90.

I changed 2 things:
- In worker.ts, the blocking connection was not duplicated because the condition did not use `opts.connection` but `opts`
- In worker.ts and redis-connection.ts, the condition used to check that it was a Redis instance was using `instanceof` which is not totally reliable depending on the environnement (if multiple instance of the package `ioredis` exist through the app, it will not work)

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -3,6 +3,7 @@ import IORedis, { Redis } from 'ioredis';
 import * as semver from 'semver';
 import { load } from '../commands';
 import { ConnectionOptions, RedisOptions } from '../interfaces';
+import { isRedisInstance } from '../utils';
 
 export class RedisConnection extends EventEmitter {
   static minimumVersion = '5.0.0';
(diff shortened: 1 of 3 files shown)


manast:
thanks for the analysis and the fix, great job!

PR #99, opened 2020-01-02
[fix] Remove duplicate active entry

chrisabrams:
Noticed that `active` was in the keys list twice for events. I understand that functionally this does not affect anything but figured it was still good to clean up.


manast:
thanks for the clean up. In order to merge the commit needs to follow the conventional commits pattern, 
so in this case it would be:
```
fix: remove duplicate active entry
```

chrisabrams:
> thanks for the clean up. In order to merge the commit needs to follow the conventional commits pattern,
> so in this case it would be:
> 
> ```
> fix: remove duplicate active entry
> ```

Changed.

PR #105, opened 2020-01-05
Fix 104 improve error handling

### LINKED ISSUE #104: Fix UnhandledPromiseRejectionWarnings
There're some places in code where promises are ignored and rejections may not be properly handled, resulting in warnings, such as

- UnhandledPromiseRejectionWarning
- ioredis warnings regarding ignored 'error' events

stansv:
Fixed some cases when UnhandlerPromiseRejectionWarning's can appear, the simplest solution is to do console.log to handle rejections. Emitting `error` events in catch() sections would also produce UnhandlerPromiseRejectionWarning (it was surprising for me). Probably it'd be better to modify instance state (write error details into some field) and make all its methods non-functional by re-throwing that error in each call, but this is not very suitable for worker which can be used without calling any instance methods.

Also I've moved array of worker tokens to worker class field level, this can be helpful for monitoring.

Fixes #104


dirkgroenen:
Shouldn't the errors be emitted instead of logged, allowing end-users to catch them and act accordingly? 

```typescript
this.run().catch(this.emit.bind(this));
```

src/classes/queue-scheduler.ts
```diff
@@ -29,8 +29,9 @@ export class QueueScheduler extends QueueBase {
   constructor(protected name: string, opts: QueueSchedulerOptions = {}) {
     super(name, { maxStalledCount: 1, stalledInterval: 30000, ...opts });
 
-    // tslint:disable: no-floating-promises
-    this.run();
+    this.run().catch(error => {
+      console.error(error);
```
manast: can't we emit the error instead of console it out?

src/classes/queue.ts
```diff
@@ -21,14 +21,17 @@ export class Queue<T = any> extends QueueGetters {
...
+    this.waitUntilReady()
+      .then(client =>
+        client.hset(
+          this.keys.meta,
+          'opts.maxLenEvents',
+          get(opts, 'streams.events.maxLen', 10000),
+        ),
+      )
+      .catch(error => {
+        console.error(error);
```
manast: emit instead of console it out

src/classes/worker.ts
```diff
@@ -77,7 +81,6 @@ export class Worker<T = any> extends QueueBase {
     }
     this.timerManager = new TimerManager();
 
-    /* tslint:disable: no-floating-promises */
     this.run().catch(error => {
       console.error(error);
```
manast: emit error instead of console out

src/classes/worker.ts
```diff
@@ -251,11 +248,13 @@ export class Worker<T = any> extends QueueBase {
...
-              lockExtender();
+            if (!this.closing) {
+              const result = await Scripts.extendLock(this, job.id, token);
+              if (result && !timerStopped) {
+                lockExtender();
+              }
+              // FIXME if result = 0 (missing lock), reject processFn promise to take next job?
             }
-            // FIXME if result = 0 (missing lock), reject processFn promise to take next job?
           } catch (error) {
```
manast: emit error instead of console out

stansv:
Hi guys, yes I think suggested change is a good idea, but I'm not completely understand how to create an event listener for these errors. I've tried to `this.emit.bind(this)('error', error)` before, but this event (`error`) causes UnhandlerPromiseRejectionWarning if no listeners attached..

manast:
@stansv are you sure about that? emitting an error thrown by a promise should avoid the warning you are mentioning.

stansv:
@manast

```

class Test extends EventEmitter {
  constructor() {
    super();
    Promise.reject(new Error('test')).catch(error => {
      this.emit('error', error);
    })
  }
}

(() => {
  new Test();
})();
```

```
$ ts-node src/index.ts
(node:15135) UnhandledPromiseRejectionWarning: Error: test
    at new Test (/Users/stansv/bullmq/src/index.ts:11:20)
    at /Users/stansv/bullmq/src/index.ts:18:3
```

I feel I missing something about promises..

ccollie:
Could be `node`'s scheduling of the `catch`. In many `promise` implementations it's delayed at least until the next tick of the event loop. In your case the end of the program may have been reached before the event fired.

stansv:
No, I've found the reason — `emit('error')` itself throws an Error if there's no registered `'error'` event listener ([this is documented](https://nodejs.org/api/events.html#events_error_events)). Throwing in catch() block itself causes UnhandledPromiseRejectionWarning.
The "correct" code would be

```

class Test extends EventEmitter {
  constructor() {
    super();
    Promise.reject(new Error('test')).catch(error => {
      try {
        this.emit('error', error);
      } catch(_) {
        // ignore...
      }
    })
  }
}

(() => {
  new Test();
})();
```

stansv:
@manast thanks for approval, but I cannot agree with current behavior, namely with

```
this.run().catch(this.emit.bind(this));
```

— if emit is called then first arg must be event name, not arbitrary error object; I do not understand how this might work (@dirkgroenen maybe you can help). How about this?

```
this.run().catch(error => {
      try {
        this.emit('error', error); // emit exactly 'error' event
      } catch(error) { // if 'error' event listener was not specially attached,
        console.log(error); // just print error to stdout
      }
    })
```

dirkgroenen:
> — if emit is called then first arg must be event name, not arbitrary error object; I do not understand how this might work (@dirkgroenen maybe you can help). How about this?

You're right, I think I made a type and was supposed to write `this.run().catch(emit.bind(this, 'error'));` :thinking:

manast:
Going to close this PR as master has drifter too much away from this branch and it is not easy to merge anymore.

PR #107, opened 2020-01-07
fix: don't fail a job when the worker already lost the lock

### LINKED ISSUE #71: "Missing lock" error will cause an uncaught exception
While the worker is processing a job that is delayed (or node process is stalling the event loop > lockDuration) and attempts to call handleCompleted(), an "Error: Missing lock for job 778 finished" error is thrown on line 303, is caught, and handleFailed() is called on line 305. 

https://github.com/taskforcesh/bullmq/blob/ad1fc3c19d5c02382989a7e1cb72b807b743465c/src/classes/worker.ts#L298-L308

However, in handleFailed, it attempts to move the job to failed state on line 293.

https://github.com/taskforcesh/bullmq/blob/ad1fc3c19d5c02382989a7e1cb72b807b743465c/src/classes/worker.ts#L284-L295

moveToFailed will also throw an "Error: Missing lock for job 778 finished" exception which will not be caught and will exit the node process.

I think the logic for these types of errors could be handled more gracefully.  But, at the very least, a try catch block should be put around line 293 (await job.moveToFailed()) so that the node process does not exit and the emit('failed') will be executed.

Embraser01:
Fix #71 

Now, if the `moveToFailed` command fail (eg: because the lock is missing), it will  catch the error. The error will still be sent as a local event.

This situation happens if the lock is too short for a job. It may be nice to add a message to help users debug this error.
> The task finished after lock duration. Try increasing `lockDuration` or use sandbox to prevent this situation

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -282,8 +282,14 @@ export class Worker<T = any> extends QueueBase {
     };
 
     const handleFailed = async (err: Error) => {
-      await job.moveToFailed(err, token);
-      this.emit('failed', job, err, 'active');
+      try {
+        await job.moveToFailed(err, token);
+        this.emit('failed', job, err, 'active');
+      } catch (e) {
+        // It probably means that the job has lost the lock before completion
+        // The QueueScheduler will (or already has) move the job to the waiting list (as stalled)
+        this.emit('error', e);
+      }
     };
 
     // TODO: how to cancel the processing? (null -> job.cancel() => throw CancelError()void)
--- a/src/test/test_worker.ts
+++ b/src/test/test_worker.ts
@@ -893,6 +893,43 @@ describe('workers', function() {
     expect(countAfterEmpty).to.be.eql(0);
   });
 
+  it('emit error if lock is lost', async function() {
+    this.timeout(10000);
+
+    const worker = new Worker(
+      queueName,
+      async job => {
+        return delay(2000);
+      },
+      {
+        lockDuration: 1000,
+        lockRenewTime: 3000, // The lock will not be updated
+      },
+    );
+    await worker.waitUntilReady();
+
+    const queueScheduler = new QueueScheduler(queueName, {
+      stalledInterval: 100,
+    });
+    await queueScheduler.waitUntilReady();
+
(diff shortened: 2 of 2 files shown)


src/classes/worker.ts
```diff
@@ -282,7 +282,12 @@ export class Worker<T = any> extends QueueBase {
     };
 
     const handleFailed = async (err: Error) => {
-      await job.moveToFailed(err, token);
+      try {
+        await job.moveToFailed(err, token);
+      } catch (e) {
+        // It probably means that the job has lost the lock before completion
+        // The QueueScheduler will (or already has) move the job to the waiting list (as stalled)
```
manast: So in the catch we should emit the error like this:
```js
this.emit('error', e);
```

Embraser01: Ok, so `this.emit('failed', job, err, 'active');` should only be executed if moveToFailed works?

manast: exactly, since that event is telling the user that the job has failed and it is now in the failed set.

Embraser01:
I updated the PR to emit an error instead of a failed event

manast:
a test to verify this feature would be the icing on the cake :)

Embraser01:
> a test to verify this feature would be the icing on the cake :)

Done ✅

Embraser01:
Hi,

There is a test failing on CI (I had it locally too), but I can't restart the CI (unless I create a new commit)
```
  1) Delayed jobs
       should process delayed jobs with exact same timestamps in correct order (FIFO):
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/taskforcesh/bullmq/src/test/test_delay.ts)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)
```

Other than that, I think it should be good

dirkgroenen:
We were experiencing issues on our production environments caused by a missing lock. Due to the unhanded error the worker stopped processing new jobs, requiring a manual restart.  

Thanks @Embraser01 and @manast. Would love to see this one merged and released :)

PR #114, opened 2020-01-18
feat(types): add sandboxed job processor types

tlaziuk:
At first: I don't know what's the PR format, I just wanted to make working with this library a little less frustrating by adding processor types.

So with this change you can write your processors like:
```typescript
const processor: SandboxedJobProcessor<string> = async (job) => {
    job.progress(0);
    // do some stuff with `job.data` or smth...
    job.progress(100);
    return 'done';
};
export default processor;
```

### DIFF
--- a/src/classes/master.ts
+++ b/src/classes/master.ts
@@ -5,6 +5,8 @@
  */
 import { promisify } from 'util';
 import { toString } from 'lodash';
+import { JobJson } from './job';
+import { SandboxedJob } from '../interfaces/sandboxed-job';
 
 let status: any;
 let processor: any;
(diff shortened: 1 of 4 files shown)


manast:
The example you wrote is not correct. You need to await the async calls, like this:
```ts
const processor: SandboxedJobProcessor<string> = async (job) => {
    await job.progress(0);
    // do some stuff with `job.data` or smth...
    await job.progress(100);
    return 'done';
};
export default processor;
```

src/classes/master.ts
```diff
@@ -101,41 +103,43 @@ process.on('uncaughtException', err => {
...
+function wrapJob(job: JobJson): SandboxedJob {
   let progressValue = job.progress;
-  job.progress = function(progress: any) {
-    if (progress) {
-      // Locally store reference to new progress value
-      // so that we can return it from this process synchronously.
-      progressValue = progress;
-      // Send message to update job progress.
+
+  return {
```
manast: nice fix so that this function does not mutate the job object!

src/interfaces/sandboxed-job-processor.ts
```diff
@@ -0,0 +1,20 @@
+import { JobJson } from '../classes/job';
+import { JobsOptions } from './jobs-options';
+
+export interface SandboxedJob<T = any, R = any>
```
manast: I think it is better to create a separate file for this export, named just ```sandboxed-job.ts``

src/interfaces/sandboxed-job-processor.ts
```diff
@@ -0,0 +1,20 @@
...
+  data: T;
+  opts: JobsOptions;
+  progress:
+    | (() => object | number)
+    | ((value: object | number) => Promise<void>);
+  log: (row: any) => void;
+  returnValue: R;
+}
+
+export type SandboxedJobProcessor<T = any, R = any> =
```
manast: This export is currently unused, where did you plan to use it?

tlaziuk: nowhere in the code, but in the example provided, in fact this is the type of the function which is exported from the sandboxed job, should I use it as a type of processor function in `master,ts`?

manast:
btw, the last commit "change import path" must follow conventional commits format, like: chore: change import patch

tlaziuk:
omg, what a pain, can I squash this PR into one commit?

manast:
husky is suppose to not allow you to do commits that do not follow that pattern, I do not know why some people is not getting the errors. A squash would work fine.

PR #123, opened 2020-01-31
docs: fix QueueEvents examples in quick start guide

rhinodavid:


docs/gitbook/README.md
```diff
@@ -74,16 +74,36 @@ import { QueueEvents } from 'bullmq'
...
+queueEvents.on('completed', ({ jobId, returnvalue }) => {
+    console.log(`${jobId} has completed and returned ${returnvalue}`);
+});
+
+queueEvents.on('failed', ({ jobId, failedReason }) => {
+    console.log(`${jobId} has failed with reason ${failedReason}`);
+});
+```
+
+You may also access the timestamp of the event, which looks like "1580456039332-0".
```
rhinodavid: Does anybody know what the suffix (`-0`) is? Does it increment when multiple events get fired during the same ms?

ccollie: Exactly. See
https://redis.io/commands/xadd#specifying-a-stream-id-as-an-argument

Issue #129, opened 2020-02-07
Default Import Error with ioredis Prevents TypeScript Compilation with esModuleInterop Disabled


bobdercole:
Hello,

My TypeScript projects fails to compile due to an error in some BullMQ files. Here is the error:

```
$ tsc -p .
node_modules/bullmq/dist/classes/redis-connection.d.ts:3:8 - error TS1259: Module '"/application/node_modules/@types/ioredis/index"' can only be default-imported using the 'esModuleInterop' flag

3 import IORedis, { Redis } from 'ioredis';
         ~~~~~~~

  node_modules/@types/ioredis/index.d.ts:43:1
    43 export = IORedis;
       ~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

I found this problem in the following files:
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/classes/redis-connection.ts#L2
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/commands/index.ts#L15
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/interfaces/queue-options.ts#L3
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/interfaces/redis-options.ts#L1
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/utils.ts#L1
https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/test/test_queue.ts#L5

I do have `esModuleInterop` disabled. The issues seems to go away if I enabled `esModuleInterop`. However, I don't think that this flag should be required.

I can confirm that the error goes away if the ioredis imports within BullMQ are changed to the recommended ioredis usage:

`import * as Redis from "ioredis";`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/946a98e60e6ae4f4f2910f1c75b8403e0f676876/types/ioredis/index.d.ts#L21

Any thoughts? Thanks!

manast:
[...]

PR #132, opened 2020-02-10
fix: modified imports to work when esModuleInterop is disabled

bobdercole:
This change should enable compatibility with top-level projects that have esModuleInterop disabled.

I did not include any tests since I disabled esModuleInterop in BullMQ. This should cover any future issues.

Fixes GH-129.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1,4 +1,4 @@
-import * as Redis from 'ioredis';
+import { Redis, Pipeline } from 'ioredis';
 import { debuglog } from 'util';
 import { RetryErrors } from '../enums';
 import { BackoffOptions, JobsOptions, WorkerOptions } from '../interfaces';
@@ -92,7 +92,7 @@ export class Job<T = any, R = any> {
     const multi = client.multi();
 
     for (const job of jobInstances) {
-      job.addJob(<Redis.Redis>(multi as unknown));
+      job.addJob(<Redis>(multi as unknown));
     }
 
     const result = (await multi.exec()) as [null | Error, string][];
@@ -454,7 +454,7 @@ export class Job<T = any, R = any> {
     );
   }
 
-  private addJob(client: Redis.Redis): string {
+  private addJob(client: Redis): string {
     const queue = this.queue;
 
     const jobData = this.asJSON();
@@ -462,7 +462,7 @@ export class Job<T = any, R = any> {
     return Scripts.addJob(client, queue, jobData, this.opts, this.id);
   }
 
-  private saveAttempt(multi: Redis.Pipeline, err: Error) {
+  private saveAttempt(multi: Pipeline, err: Error) {
     this.attemptsMade++;
     this.stacktrace = this.stacktrace || [];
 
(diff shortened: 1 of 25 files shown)


src/classes/redis-connection.ts
```diff
@@ -1,14 +1,14 @@
 import { EventEmitter } from 'events';
-import IORedis, { Redis } from 'ioredis';
+import * as Redis from 'ioredis';
```
manast: Same comment as above

src/classes/job.ts
```diff
@@ -1,4 +1,4 @@
-import IORedis from 'ioredis';
+import * as Redis from 'ioredis';
```
manast: why can't we just do: 
```ts
import { Redis } from 'ioredis'
```

src/classes/worker.ts
```diff
@@ -1,14 +1,14 @@
-import fs from 'fs';
-import { Redis } from 'ioredis';
-import path from 'path';
+import * as fs from 'fs';
+import * as Redis from 'ioredis';
```
manast: same comment as above

src/commands/index.ts
```diff
@@ -12,7 +12,7 @@
  */
 'use strict';
 
-import IORedis from 'ioredis';
+import * as Redis from 'ioredis';
```
manast: same as above

src/interfaces/queue-options.ts
```diff
@@ -1,6 +1,6 @@
 import { JobsOptions } from '../interfaces';
 
-import IORedis from 'ioredis';
+import * as Redis from 'ioredis';
```
manast: same

src/test/test_clean.ts
```diff
@@ -21,7 +21,7 @@ describe('Cleaner', () => {
   afterEach(async function() {
     await queue.close();
     await queueEvents.close();
-    await removeAllQueueData(new IORedis(), queueName);
+    await removeAllQueueData(new Redis(), queueName);
```
manast: its late now, so I am confused why we can instantiate Redis here and not just import the Redis class by this (as commented above)
```ts
import { Redis } from 'ioredis'
```

src/classes/job.ts
bobdercole: That will work for all cases where it's used as an interface. However, this usage won't work (or at least I can't figure out how to make it work):

```typescript
new Redis();
```
```
'Redis' only refers to a type, but is being used as a value here.ts(2693)
```

I'm not sure why the ioredis typings are written this way.

The import in this particular file, `src/classes/job.ts`, could be changed to this since there are no instantiations here:
```typescript
import { Redis, Pipeline } from 'ioredis';
```
I used the same import style across all of the files for consistency and to account for files that need to instantiate an instance. I suppose it's a matter of opinion. I could change all of the files that don't instantiate Redis to the named import style if you'd prefer.

This reply should apply to all of your comments.

manast: ok, but in many places we are not instantiating, just using it as an interface. Besides that, I think that we should change the name of the import to IORedis to make it explicitly that it is the module name, so it will become IORedis.Redis instead of the slightly more awkward Redis.Redis.

bobdercole: Good points. I will make those changes.

Issue #130, opened 2020-02-07
How to Check if Job Exists


bobdercole:
Hi,

I am currently trying to upgrade from https://github.com/OptimalBits/bull/ and I ran into a bit of a snag. My project relies on this logic:

```typescript
const existingJob = await myQueue.getJob(jobId);

if (existingJob === null) {
	// do some stuff
}
```

The BullMQ type definition for `getJob` looks like this:

```typescript
export declare class QueueGetters extends QueueBase {
    getJob(jobId: string): Promise<Job<any, any>>;
}
```

According to that, `getJob` will always return a job, which makes it hard to check if a job exists.

I noticed `QueueGetters.getJob()` calls `Job.fromId()` internally, which appears to return null if the job is empty.

https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/classes/job.ts#L143

However, even `job.d.ts` has the same definition:

```typescript
export declare class Job<T = any, R = any> {
    static fromId(queue: QueueBase, jobId: string): Promise<Job<any, any>>;
}
```

Interestingly the v3 compatibility class has exactly what I need.

https://github.com/taskforcesh/bullmq/blob/36726bfb01430af8ec606f36423fc187e4a06fb4/src/classes/compat.ts#L201

Was this functionality purposely changed? Or is it a mistake? If the former, how should I go about getting a job that might not exist?

Thanks!

manast:
probably a mistake in the typings, but it should work the same as v3.

bobdercole:
I created a PR with a fix.

I also noticed some weirdness related to the typings here: https://github.com/bobdercole/bullmq/blob/65183fcf542d0227ec1d4d6637b46b5381331787/src/classes/job.ts#L138

My PR adds `Promise<Job | null>` to the return signature. However, if `jobId` is falsy the method will return undefined, which is not covered by the types I added. Additionally the comment says that `jobId` could be `undefined`, but parameter `jobId: string` states otherwise. Should it be `jobId?: string`?

manast:
yes you are right. Some parts are directly ported from Bull 3 and I have overlooked these types. But yes, jobId should be optional and probably the return value should be ```Promise<Job | undefined>```

bobdercole:
Alright. I will comment in the PR.

PR #133, opened 2020-02-10
fix: modified QueueGetters.getJob and Job.fromId to also return null

bobdercole:
Fixes: https://github.com/taskforcesh/bullmq/issues/130

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -135,7 +135,7 @@ export class Job<T = any, R = any> {
     return job;
   }
 
-  static async fromId(queue: QueueBase, jobId: string) {
+  static async fromId(queue: QueueBase, jobId: string): Promise<Job | null> {
     // jobId can be undefined if moveJob returns undefined
     if (jobId) {
       const client = await queue.client;
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -135,12 +135,15 @@ export class Job<T = any, R = any> {
     return job;
   }
 
-  static async fromId(queue: QueueBase, jobId: string): Promise<Job | null> {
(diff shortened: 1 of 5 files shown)


src/classes/job.ts
```diff
@@ -135,7 +135,7 @@ export class Job<T = any, R = any> {
     return job;
   }
 
-  static async fromId(queue: QueueBase, jobId: string) {
+  static async fromId(queue: QueueBase, jobId: string): Promise<Job | null> {
```
manast: the question is if it should be null or just undefined? because it will actually return undefined and not null.

bobdercole: I agree; I think it should be `Promise<Job | undefined>`. I believe the only downside is that it adds another small step in the migration path from v3 (my implementations of Bull do strict checks on `null`).

I assume this change means that this line:
```typescript
return isEmpty(jobData) ? null : Job.fromJSON(queue, jobData, jobId);
```
Needs to change to:
```typescript
return isEmpty(jobData) ? undefined : Job.fromJSON(queue, jobData, jobId);
```

Should there be a different issue/PR for the `jobId` parameter change? I am actually a little confused about this change anyway. An optional ID on a method called `fromId` does seem a little strange. Maybe the code that consumes `fromId` should just not call `fromId` if the parameter is `undefined`? If so, the method could be simplified to the following:

```typescript
static async fromId(queue: QueueBase, jobId: string): Promise<Job | undefined> {
  const jobData = await client.hgetall(queue.toKey(jobId));
  return isEmpty(jobData) ? undefined : Job.fromJSON(queue, jobData, jobId);
}
```

PR #150, opened 2020-03-03
- stacktrace limit fix

wenq1:


manast:
in order for the automatic release to work the commits need to adhere to conventional commit standard, I can squeeze and merge to fix it, but please think about it next time:
https://www.conventionalcommits.org/en/v1.0.0/

wenq1:
I will. I’ll have it checked next time. Looks to me like a linting issue

manast:
Closing since slice already takes care of ignoring the value if it is larger than length.

Issue #138, opened 2020-02-13
Timestamp handling


wenq1:
In bullmqv4, Job.fromJSON () method, job.timestamp is not recovered (in fact commented out). 

````
// job.delay = parseInt(json.delay);
// job.timestamp = parseInt(json.timestamp);
````

This creates a problem for UI when trying to read jobs from IDs. The timestamp will always be set to Date.now () in the constructor, regardless of the timestamp stashed in RedisLayer.

````
this.timestamp = opts.timestamp ? opts.timestamp : Date.now();
````

Sounds like a bug?

wenq1:
Anyone?

jamesholcomb:
Running into this as well.  Have you tried uncommenting?

wenq1:
uncommenting will work.

However I open a thread here so as to discuss the intention of this regression, or if it a feature rather than a regression.

wenq1:
Is @manast still hiding?

manast:
I am a bit confused too why these are out commented, should work to uncommented AFAICS.

onehorsetown:
@manast - Do you want a pull request for un-commenting out those two lines of code?

PR #152, opened 2020-03-03
fix: restore the Job timestamp when deserializing JSON data (#138)

onehorsetown:
Uncommented out line of code that sets/restores the job timestamp from the job's JSON data.

#138

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -113,7 +113,7 @@ export class Job<T = any, R = any> {
     job.progress = JSON.parse(json.progress || 0);
 
     // job.delay = parseInt(json.delay);
-    // job.timestamp = parseInt(json.timestamp);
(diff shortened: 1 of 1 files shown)


jamesholcomb:
Can this be uncommented as well?

`// job.delay = parseInt(json.delay);`

onehorsetown:
@jamesholcomb - I looked into uncommenting that as part of the previous PR, but I did not find anywhere that the `delay` property was actually used.   That property is not even defined on either the `JobJson` interface or `Job` class.

jamesholcomb:
Thanks, I ran into this from a dependency on `delay` here:

https://github.com/darky/bull-repl/blob/1f9053a979f7a283b6e7a9116055c0a1f6936c8b/src/utils.ts#L38

I'll open a new issue in that repo to remedy.

Issue #119, opened 2020-01-26
QueueEvents delay never exits 


seawatts:
Whenever I try to use the QueueEvents class my process hangs forever because of this line: https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-events.ts#L65
However, If i just remove the QueueEvents class then it will exit correctly. 

Example:
```ts
(async () => {

const queueName = `test-queue-1`;
const queue = new Queue(queueName);
const queueEvents = new QueueEvents(queueName)
await queueEvents.waitUntilReady();
const worker = new Worker(queueName, () => console.log('processed'));

await queue.add('hi', {});

await new Promise(resolve => queueEvents.on('completed', resolve));

await worker.close();
await queue.close();
await queueEvents.close();
})();

```

I used wtfnode in order to test this and got these results

```
[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 2 (tty) (stdio)
- Sockets:
  - 127.0.0.1:50644 -> 127.0.0.1:6379
- Timers:
  - (5000 ~ 5 s) (anonymous) @ ./node_modules/bullmq/dist/utils.js:33
```

manast:
I do not understand, the line you are refering to is a 5000 seconds delay, so how is it hanging it forever?

seawatts:
I agree, I'm not sure. That's why I'm confused.

seawatts:
Hmm, it looks like if I take `QueueEvents` out then it works fine. Maybe it's not closing the connection correctly?

onehorsetown:
I suspect I see the problem here.  `worker.close()` does not actually return a promise or await anything.

```
await worker.close();
```

Looking at https://github.com/taskforcesh/bullmq/blob/fccacee7164155ea1fdf5bee163dd56103cfb223/src/classes/worker.ts#L370 you can see that `Worker.close` constructs a promise, but it does not return or await that promise.  Basically it is creating a hanging promise that causes problems when you want to exit the process cleanly.

manast:
@onehorsetown close should wait for closing. But the promise it is not "hanging", it has a very important function.

manast:
https://github.com/taskforcesh/bullmq/pull/153

manast:
AFAIK this is resolved now.

PR #154, opened 2020-03-04
Node process not exiting cleanly

onehorsetown:
Having lots of issues trying to cleanly shut down a worker + scheduler process (similar to #119)  

The changes in this PR fix all of the issues I was having.

1) Change to Worker queue is the same as in PR #153 whereby nothing was being awaited. 

2) If Worker.close is called without the force flag, the blocking client connection is never getting closed, which prevents the Node event loop from exiting.

https://github.com/taskforcesh/bullmq/blob/b96f5ba3e9c36182174300af629b34fe43c62ecd/src/classes/worker.ts#L378

3) QueueBase.close and QueueScheduler.close, if called multiple times, causes trouble.  I added the same check you are doing in Worker.close

4) Added "closing" and "closed" events to QueueBase and QueueScheduler.   This achieves functional parity with the Worker class.

### DIFF
--- a/src/classes/queue-scheduler.ts
+++ b/src/classes/queue-scheduler.ts
@@ -160,11 +160,17 @@ export class QueueScheduler extends QueueBase {
   }
 
   async close() {
-    if (this.isBlocked) {
-      this.closing = this.disconnect();
-    } else {
-      super.close();
+    if (!this.closing) {
+      this.emit('closing');
+
+      this.closing = (this.isBlocked
+        ? this.disconnect()
+        : this.connection.close()
+      ).then(() => {
+        this.emit('closed');
+      });
     }
+
     return this.closing;
   }
 }
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -374,11 +374,9 @@ export class Worker<T = any> extends QueueBase {
           await this.resume();
           if (!force) {
             await this.whenCurrentJobsFinished(false);
-          } else {
-            await client.disconnect();
           }
-          // await this.disconnect();
-          await super.close();
+          await client.disconnect();
+          await this.connection.close();
         } catch (err) {
           reject(err);
         } finally {
@@ -389,5 +387,7 @@ export class Worker<T = any> extends QueueBase {
         resolve();
       });
     }
+
+    return this.closing;
   }
 }
--- a/src/classes/queue-base.ts
+++ b/src/classes/queue-base.ts
@@ -72,7 +72,15 @@ export class QueueBase extends EventEmitter {
   }
 
   close() {
-    return (this.closing = this.connection.close());
+    if (!this.closing) {
+      this.emit('closing');
+
+      this.closing = this.connection.close().then(() => {
+        this.emit('closed');
+      });
+    }
+
+    return this.closing;
   }
 
   disconnect() {


src/classes/queue-scheduler.ts
```diff
@@ -160,11 +160,17 @@ export class QueueScheduler extends QueueBase {
...
-    if (this.isBlocked) {
-      this.closing = this.disconnect();
-    } else {
-      super.close();
+    if (!this.closing) {
+      this.emit('closing');
+
+      this.closing = (this.isBlocked
+        ? this.disconnect()
+        : this.connection.close()
```
manast: why not call super.close() as in the old version?

src/classes/worker.ts
```diff
@@ -374,11 +374,9 @@ export class Worker<T = any> extends QueueBase {
           await this.resume();
           if (!force) {
             await this.whenCurrentJobsFinished(false);
-          } else {
-            await client.disconnect();
           }
-          // await this.disconnect();
-          await super.close();
+          await client.disconnect();
+          await this.connection.close();
```
manast: same comment as above

manast:
I also want to say that checking for this.closing in the close method is defensive programming, in reality calling close more than once is already fishy, we could instead throw an exception for the user.

src/classes/worker.ts
```diff
@@ -374,11 +374,9 @@ export class Worker<T = any> extends QueueBase {
           await this.resume();
           if (!force) {
             await this.whenCurrentJobsFinished(false);
-          } else {
-            await client.disconnect();
           }
-          // await this.disconnect();
-          await super.close();
+          await client.disconnect();
```
manast: disconnect should not be needed when not using force since the current jobs must complete and since we are "closing" no new calls to the blocking command should be performed. If we have a blocking connection after waiting fo current jobs to fail, there is a bug somewhere else.

onehorsetown: Without closing (`client.disconnect`) the blocking client disconnect, my Node event queue won't exit cleanly.  Ill put together a standalone example of this happening.

src/classes/worker.ts
onehorsetown: `this.closing` is set on line 370:

https://github.com/taskforcesh/bullmq/blob/f4b17b565e6de597e7b16179afd469ffa5640d0d/src/classes/worker.ts#L370

If you call `QueueBase.close()` inside the promise it will then:

a) in the old code it will overwrite the `this.closing` promise

b) In your proposed changes in PR #153, it won't do anything because `this.closing` is already set:

https://github.com/taskforcesh/bullmq/blob/4ea537e0d1c68d3120b7d24a91b6c5aee481d2f9/src/classes/queue-base.ts#L74-L79

c) In my proposed changes, it will also do nothing because, as in your PR, `this.closing`  is already set.

src/classes/worker.ts
onehorsetown: Standalone example of the issue in #159

https://github.com/onehorsetown/bullmq-159/blob/master/index.js

stavalfi:
can we merge this? this issue makes bullmq not production ready...

adrienduchemin:
Hi, 
We are experiencing the same issues and this is blocking for us. Can this be merge ?
Thanks !

manast:
@adrienduchemin what issue are you experience? the example code provided by the poster (https://github.com/onehorsetown/bullmq-159/blob/master/index.js) is working in the last version of BullMQ:
image

stavalfi:
> @adrienduchemin what issue are you experiencing? the example code provided by the poster (https://github.com/onehorsetown/bullmq-159/blob/master/index.js) is working in the last version of BullMQ:
> image

last time I checked (few months ago), the process hanged for 5s until the queue closed. with Jest this is a blocker because jest fails the test by default because of this.

manast:
@stavalfi it is not taking so long now, 1 second or something like that.

stavalfi:
@manast as long as there is no missing `await`, im good with that. if there is, some test-runners (jest) won't like it.

adrienduchemin:
@manast we investigated on our side and we found the problem. 
If we create our own instance of a IORedisConnection that we pass to our queues and workers, then we **have to** quit/disconnect it manually. It would be nice to precise it in the doc (https://docs.bullmq.io/guide/connections). If we don't, then the process is never killed.

manast:
@adrienduchemin I think this is common sense, if you create your own connection to redis it is also your responsibility to close it when you do not need it anymore...

PR #162, opened 2020-03-12
fix(sandbox): If the child process is killed, remove it from the pool.

xudongdev:
This fix comes from bull.

Reference: 
https://github.com/OptimalBits/bull/issues/1460
https://github.com/OptimalBits/bull/pull/1461


manast:
great, thank you very much!

PR #188, opened 2020-04-23
Second Argument for worker is missing

seigel:
I am not sure exactly the options here, but definitely this will work.  It also appears that some sort of resolution might occur if a string is passed, but I can't figure that one out.


manast:
thanks

Issue #189, opened 2020-04-27
Job.createBulk  doesnt not obey custom jobID


willoughby:
I have a set of jobs which I only ever want to be queued once at most.
``` javascript
const queue = [];

const services = await Services.find({
    uid: {
        $in: activeIds
    },
    active: true
});

for (const service of services) {
    try {
        queue.push({
            name: 'sync',
            data: {
                id: service._id.toString()
            },
            opts: {
                jobId: service._id.toString(),
                connection,
                priority: 1,
                removeOnComplete: true,
                removeOnFail: true,
                timeout: 30000,
            },
        });
    } catch (e) {
        console.log('Error Queueing services', e, e.stack);
    }
```

I have traced in createBulk in job.ts 88 its missing the 5th argument which allows the custom jobId

``` javascript
const jobInstances = jobs.map(
      job => new Job<T, R>(queue, job.name, job.data, job.opts, opts && opts.jobId),
);
```

willoughby:
Pull request #190

manast:
As I comment in the PR, it seems really strange to have job id as a fifth argument when it is already passed in the job options, I think this is what we need to improve and remove the fifth parameter alltogether, I do not have a valid explanation on to why it looks this way, maybe some copy/paste error from bull v3 but I am not sure.

roggervalf:
passing a customId should be possible now as in this test case https://github.com/taskforcesh/bullmq/blob/master/tests/test_bulk.ts#L108-L139

PR #190, opened 2020-04-27
fix: job.createBulk() did not set custom job id

willoughby:
job id argument was not passed into the job ctor

#189


manast:
thanks for the patch. When reviewing the code I was a bit confused about why jobId is passed as a last argument to the constructor when it is already passed in the job options, I think this needs to be understood and improved.

willoughby:
Yeah that's true, it could be handled better. Ill take a look at what the consequences could be by removing the 5th argument and rework my pull request

PR #202, opened 2020-05-18
Provide ESM and CJS versions

arlac77:
AS node now supports ESM and conditional exports.
We can provide CJS and ESM in one package


manast:
thanks for the patch. Seems like the option you passed to tsc is not correct though, check the build status.

arlac77:
Sorry I am not used to ts but the 4 errors where the beforehand too (at least on my local machine)

manast:
In the latest test run all the tests related to child processors are failing because they do not find the processor files.

package.json
```diff
@@ -2,14 +2,20 @@
...
+    "require": "./dist/cjs/index.js",
+    "./": "./"
+  },
   "source": "src/index.ts",
   "author": "Taskforce.sh Inc.",
   "license": "MIT",
   "scripts": {
     "docs": "typedoc --out docs src",
-    "build": "tsc && yarn copylua",
+    "build": "tsc && tsc --build tsconfig.esm.json && yarn copylua",
```
manast, line 18: do we need to build 2 times?

manast:
closing since no answer to the feedback.

PR #204, opened 2020-05-23
fix: Scheduler does not set the next record to delay stream

intpro:
Assume we have one deleayed job. 

Scheduler on start loads incorrect local nextTimestamp value 6513388584960082.
It isn't divided by 4096 in https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-scheduler.ts#L51-L58

So for every iteration the following code wouldn't be executed. It does not set the next timestamp for execution in delay stream: https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-scheduler.ts#L101-L112

I don't know how to write test suite for such case. But this fix helps to alive our frozen cron tasks in production.

BTW for what purpose do you divide & multiply timestamp by 4096? We really spent a lot of time to figure out this mistake.

### DIFF
--- a/src/classes/queue-scheduler.ts
+++ b/src/classes/queue-scheduler.ts
@@ -54,7 +54,7 @@ export class QueueScheduler extends QueueBase {
     let streamLastId = delaySet[1] || '0-0';
 
     if (nextTimestamp) {
-      this.nextTimestamp = nextTimestamp;
+      this.nextTimestamp = nextTimestamp / 4096;
     }
 
     while (!this.closing) {


intpro:
I don't know what happens with Travis tests. But my changes do not affect on that 2 failing tests.

manast:
Good catch!.
The reason we divide by 4096 is that that gives us 12 bits (2^12 = 4096) for storing the job id. The value of the combined job id and the timestamp is used as the score in the delay set, this guarantee that at least 4096 jobs in the same timestamp will be processed in the right order (only works for id generated by bullmq not for custom job ids.

PR #226, opened 2020-06-12
docs: fix clean comment

joebowbeer:
The current comment refers to a `remove` function that no longer exists.

Now, clean is similar to drain?


manast:
thanks!

PR #243, opened 2020-07-15
feat(sandbox): kill child workers gracefully

GabrielCastro:
Properly handle SIGTERM in the child workers by waiting for the current
job to exit, then exiting cleanly.

Worker#close now also kills the workers and waits for them to close

This is the same feature as https://github.com/OptimalBits/bull/pull/1802

### DIFF
--- a/src/classes/child-pool.ts
+++ b/src/classes/child-pool.ts
@@ -4,8 +4,11 @@ import { forEach, values, flatten } from 'lodash';
 import * as getPort from 'get-port';
 import * as fs from 'fs';
 import { promisify } from 'util';
+import { killAsync } from './process-utils';
 
 const stat = promisify(fs.stat);
+
+const CHILD_KILL_TIMEOUT = 30_000;
 
 export interface ChildProcessExt extends ChildProcess {
   processFile?: string;
@@ -28,11 +31,21 @@ const convertExecArgv = async (execArgv: string[]): Promise<string[]> => {
   return standard.concat(convertedArgs);
 };
 
-const initChild = function(child: ChildProcess, processFile: string) {
-  return new Promise(resolve => {
-    child.send({ cmd: 'init', value: processFile }, resolve);
+async function initChild(child: ChildProcess, processFile: string) {
+  const onComplete = new Promise(resolve => {
+    const onMessageHandler = (msg: any) => {
+      if (msg.cmd === 'init-complete') {
+        resolve();
+        child.off('message', onMessageHandler);
+      }
+    };
+    child.on('message', onMessageHandler);
   });
-};
+  await new Promise(resolve =>
+    child.send({ cmd: 'init', value: processFile }, resolve),
+  );
+  await onComplete;
+}
 
 export class ChildPool {
   retained: { [key: number]: ChildProcessExt } = {};
@@ -89,21 +102,17 @@ export class ChildPool {
     }
   }
 
-  kill(child: ChildProcess, signal?: NodeJS.Signals) {
-    child.kill(signal || 'SIGKILL');
+  async kill(child: ChildProcess, signal: 'SIGTERM' | 'SIGKILL' = 'SIGKILL') {
     this.remove(child);
+    await killAsync(child, signal, CHILD_KILL_TIMEOUT);
   }
 
-  clean() {
+  async clean() {
     const children = values(this.retained).concat(this.getAllFree());
-
-    children.forEach(child => {
-      // TODO: We may want to use SIGKILL if the process does not die after some time.
-      this.kill(child, 'SIGTERM');
-    });
-
     this.retained = {};
     this.free = {};
+
+    await Promise.all(children.map(c => this.kill(c, 'SIGTERM')));
   }
 
   getFree(id: string): ChildProcessExt[] {
--- a/src/classes/master.ts
+++ b/src/classes/master.ts
@@ -10,6 +10,15 @@ import { SandboxedJob } from '../interfaces/sandboxed-job';
 
 let status: any;
 let processor: any;
+let currentJobPromise: Promise<unknown> | undefined;
+
+// same as process.send but waits until the send is complete
+// the async version is used below because otherwise
+// the termination handler may exit before the parent
+// process has received the messages it requires
+const processSendAsync = promisify(process.send.bind(process)) as (
+  msg: any,
+) => Promise<void>;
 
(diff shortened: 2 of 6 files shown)


src/classes/process-utils.ts
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+import { ChildProcess } from 'child_process';
+
+function onExitOnce(child: ChildProcess): Promise<void> {
+  return new Promise(resolve => {
+    if (child.exitCode !== null || child.signalCode) {
```
manast, line 7: This code should not be needed since you are also checking in line 19. Since both checks are happening in the same event loop, they must give identical results.

src/classes/process-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  let onExit = onExitOnce(child);
+  child.kill(signal);
+
+  if (timeoutMs === 0 || isFinite(timeoutMs)) {
+    const timeoutHandle = setTimeout(() => {
+      if (child.exitCode !== null || child.signalCode) {
+        child.kill('SIGKILL');
+      }
+    }, timeoutMs);
+    onExit = onExit.then(() => {
```
manast, line 32: you should be able to just use "await" here.

src/classes/process-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (child.exitCode !== null || child.signalCode) {
+    return;
+  }
+
+  let onExit = onExitOnce(child);
+  child.kill(signal);
+
+  if (timeoutMs === 0 || isFinite(timeoutMs)) {
+    const timeoutHandle = setTimeout(() => {
+      if (child.exitCode !== null || child.signalCode) {
```
manast, line 28: Is this condition correct? If the child has already exited one of these two should be defined, so it seems it will kill the child only when it has already exited by itself.

src/classes/worker.ts
```diff
@@ -364,30 +364,24 @@ export class Worker<T = any> extends QueueBase {
...
-        }
-        this.emit('closed');
-        resolve();
-      });
-    }
+      const client = await this.blockingConnection.client;
+      await this.pause();
+      if (!force) {
+        // ignore most failures and continue with closing
+        await this.whenCurrentJobsFinished(false).catch(err => {});
```
manast, line 378: I think it is better to not silently ignore these errors, could lead to difficult to debug issues in the future.

src/classes/worker.ts
```diff
@@ -364,30 +364,24 @@ export class Worker<T = any> extends QueueBase {
...
+        await this.childPool?.clean()?.catch(err => {});
```
manast, line 379: why only clean the pool if "force === false" ?

src/classes/worker.ts
```diff
@@ -364,30 +364,24 @@ export class Worker<T = any> extends QueueBase {
...
+      }
+      await client.disconnect();
+      this.timerManager.clearAllTimers();
```
manast, line 382: you should wrap the code in try/catch/finally as the original code, so that timers and pool are always cleaned.

src/classes/master.ts
```diff
@@ -10,6 +10,13 @@ import { SandboxedJob } from '../interfaces/sandboxed-job';
 
 let status: any;
 let processor: any;
+let currentJobPromise: Promise<unknown> | undefined;
+
+// same as process.send but waits until the send is complete
+// the async version is used below because otherwise
+// the termination handler may exit before the parent
+// process has recived the messages it requires
```
manast, line 18: recived -> received

src/classes/master.ts
```diff
@@ -29,7 +36,19 @@ if (!('toJSON' in Error.prototype)) {
   });
 }
 
-process.on('message', async msg => {
+function terminateSelf() {
+  status = 'TERMINATING';
+  return (currentJobPromise || Promise.resolve()).finally(() => {
```
manast, line 41: use await and try/finally instead of promises

src/classes/master.ts
```diff
@@ -29,7 +36,19 @@ if (!('toJSON' in Error.prototype)) {
   });
 }
 
-process.on('message', async msg => {
+function terminateSelf() {
```
manast, line 39: I think it may be more descriptive to rename this function so that it better expresses the intent, i.e. waiting for the current job to complete.

src/classes/master.ts
```diff
@@ -60,23 +82,26 @@ process.on('message', async msg => {
...
-        process.send({
-          cmd: 'completed',
-          value: result,
-        });
-      } catch (err) {
-        if (!err.message) {
-          err = new Error(err);
+      currentJobPromise = (async () => {
+        try {
+          const result = await Promise.resolve(processor(wrapJob(msg.job)) || {});
```
manast, line 87: actually we do not need to use Promise.resolve here since await accepts a plain value too:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await

src/classes/worker.ts
```diff
@@ -364,30 +364,37 @@ export class Worker<T = any> extends QueueBase {
...
 
-          if (!force) {
-            await this.whenCurrentJobsFinished(false);
-          } else {
-            await client.disconnect();
+      const client = await this.blockingConnection.client;
+
+      this.resume();
+      await Promise.resolve()
+        .finally(() => {
```
manast, line 378: I have mixed feelings about this chain of "finally"'s. Not sure we should continue trying to close if some fatal error has happened, in any case the chain of finallys looks a bit odd... what is your opinion on this?

GabrielCastro: It looks a bit odd

at the very least the chain of `.finally()`s is much cleaner than the alternative `try {...} finally {...}` async syntax

I'm not sure there's a 100% right way to handle all the possible failures here, once there's a failure in closing it's pretty had to be in a safe defined state, so might as well close up as many things as possible. Since there's already been a failure, it's ok if we hit others during the clean up

src/classes/child-pool.ts
```diff
@@ -89,21 +100,17 @@ export class ChildPool {
     }
   }
 
-  kill(child: ChildProcess, signal?: NodeJS.Signals) {
-    child.kill(signal || 'SIGKILL');
+  async kill(child: ChildProcess, signal: 'SIGTERM' | 'SIGKILL' = 'SIGKILL') {
     this.remove(child);
+    await killAsync(child, signal, 30_000);
```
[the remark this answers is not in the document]
GabrielCastro: done

Issue #229, opened 2020-06-24
`convertExecArgv()` embedded loop not completing before function


preston-vasquez:
It looks like you're using an async function inside a for loop here:  

https://github.com/taskforcesh/bullmq/blob/master/src/classes/child-pool.ts#L18

`convertExecArgv()` completes before the loop completes.  I'm surprised this hasn't come up before, but I'm dealing with a fairly complex cluster of child processes.  What do you guys think?  It looks like the async loop function is used only for the sake of `getPort()`?  

Maybe this would work better:

```typescript
  for (let arg in execArgv) {
    if (arg.indexOf('--inspect') === -1) {
      standard.push(arg);
    } else {
      const argName = arg.split('=')[0];
      const port = await getPort();
      convertedArgs.push(`${argName}=${port}`);
    }
  };
```

manast:
yes thats definitely better, the older code is incorrect, must have been a copy paste error from Bullv3. 
If you could make this into a PR I would gladly merge it ;).

PR #249, opened 2020-08-04
Remove async for loop in child pool

preston-vasquez:
Fix for the issue discussed [here](https://github.com/taskforcesh/bullmq/issues/229)


manast:
Thanks for the PR!

PR #258, opened 2020-08-17
build: sort script targets, rename prepublish target to prepare

stieg:
If a user wishes to pull in a version of bullmq via the git/github syntax as supported by npm, the resulting code will not work before this change.  This is because npm will not invoke the `prepublish` target to compile the TS code.  However it will invoke the `prepare` target which is a superset of the `prepublish` target.  Thus it makes sense to rename the `prepublish` taret to `prepare` as that is a more desirable behavior.  More can be read on it at https://docs.npmjs.com/files/package.json. 

This change also sorts the targets because sorting is good.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,19 +8,19 @@
   "author": "Taskforce.sh Inc.",
   "license": "MIT",
   "scripts": {
-    "docs": "typedoc --out docs src",
     "build": "tsc && yarn copylua",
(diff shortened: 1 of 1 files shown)


manast:
thanks, your change makes a lot of sense!.

stieg:
Cool.  Happy to help.  Thanks for being responsive and not letting the PR languish.

PR #273, opened 2020-09-14
feat: support separate group-specific rate limits in a single queue

csingh:
Hello! This PR introduces a new feature expanding upon the existing `groupKey` feature which allowed grouped rate limiting in a single queue. My team had a requirement for those groups having separate rate limits, and this is what I came up with.

The overall workflow for this new feature:

- Support group-specific rate limit configs via the groupRate key in Queue and Worker configs
    - Note: This unfortunately needs to be added to both Queue and Worker configs, please see `test_rate_limiter.ts` for comment and examples.
- Modified addJob to pass in groupRate for the given job's group (if defined), and it will write that config to Redis
- Modified moveToActive to check if a groupRate for the selected job exists in Redis, and to use that if it exists. Otherwise fallback to queue's default rate limit config

Note:
- sorry for the messy/extra commits in the PR. If possible when merging, please squash and use the PR title, which should follow the Conventional Commits format :)


src/classes/scripts.ts
```diff
@@ -53,6 +53,23 @@ export class Scripts {
       opts.lifo ? 'RPUSH' : 'LPUSH',
     ];
 
+    // If groupKey-specific limiter is defined, pass in the rate limiter options
+    // so the script can store them in Redis. Then moveToActive can look for these
+    // rate values if they exist
+    if (
+      queue.opts &&
```
manast, line 60: here we can use optional chaining for more succinct code:
```ts
queue.opts?.limiter?.groupKey && queue.opts?.limiter?.groupRates
```

src/classes/scripts.ts
```diff
@@ -53,6 +53,23 @@ export class Scripts {
...
+      queue.opts.limiter &&
+      queue.opts.limiter.groupKey &&
+      queue.opts.limiter.groupRates
+    ) {
+      const jobData = JSON.parse(job.data);
```
manast, line 65: JSON.parse is an expensive operation, so we must find a way to avoid a stringify and a subsequent parse in order to get the groupIdentifier.

src/commands/addJob-8.lua
```diff
@@ -56,6 +58,17 @@ end
 rcall("HMSET", jobIdKey, "name", ARGV[3], "data", ARGV[4], "opts", ARGV[5],
       "timestamp", ARGV[6], "delay", ARGV[7], "priority", ARGV[9])
 
+-- Store the group rate limits if they exist
+local groupKey = string.match(jobId, "[^:]+$")
```
manast, line 62: this match operation should only be performed if we have grouped rate limits, so that standard "add" does not get this performance hit

src/commands/moveToActive-8.lua
```diff
@@ -48,6 +48,7 @@ end
 if jobId then
   -- Check if we need to perform rate limiting.
   local maxJobs = tonumber(ARGV[6])
+  local duration = tonumber(ARGV[7])
```
manast, line 51: this assignment can be moved to the if statement below

manast (changes requested):
Thanks for a great PR. I wrote some comments. I will still need to review it more deeply, I need to understand if this has other less obvious implications for other parts in the system.

src/classes/scripts.ts
csingh: Ahh, I'm a relative JS/TS noob, so thanks for this! :D Will change

src/commands/addJob-8.lua
csingh: @manast I've updated, let me know if this is okay now :)

src/classes/scripts.ts
csingh: Hmm what do you think of this?
https://github.com/wishabi/bullmq/commit/8bdbb30e08a235c054c4c4f150bfd04b7bf020be

Basically passed in the entire job object into the Scripts.addJob, and then only toJSON what is necessary, so we don't have to do the expensive JSON.parse just to access the job data.

csingh:
Hi @manast, just wondering if there is anything else I can address here? :)

manast:
I am going to study this PR once more and then I will let you know, sorry for taking so long time.

csingh:
@manast Not a problem, just wanted to followup. Thank you!

manast:
I am trying to figure out how we can have this feature avoiding to duplicate the group rate settings both in Queue and Worker.

manast:
Ok, I think this is how this should be done:
Introduce a new class in BullMQ, ```GroupLimit``` for example, although in the future it could be just Limit. This class takes as input your new Group limit structure:
```
   [key: string]: { max: number, duration }
```
This structure is stored in redis in a hash encoded like this:

```
  groupKey: `${max}:${duration}`
```
The semicolon used to easily separate the max and duration when reading from it.
You do not need a redis lua script for this, you can easily use the HMSET command. This will remove all the code form the addJob script and in moveToActive you just need to check with a HGET if there is a group rate limit for a particular group.

That do you think?

(I also added some comments to the source code, after my proposal above some of those comments will probably be irrelevant)

csingh:
@manast sorry I've taken so long to respond! Haven't had too much time to work on this lately.

I do like that idea! I updated the code to try this out, and its definitely much cleaner.

Note: I didn't update the docs yet, just wanted to get an idea what you think of the implementation first. Thanks!

csingh:
Hi @manast have you had a chance to review the updates? :)

csingh:
Hey @manast, happy new year! Hope everything is well, I was wondering if there is a chance this will still be looked at?

manast:
This is in the top of my list, I promise I will look into it ASAP. The reason it is taking more time is because I am designing a new rate limiter for grouping as a Redis module and I want it to be aligned with this one as much as possible to ensure compatibility between the two.

csingh:
Oh okay, thank you for the context! No worries, just wanted to know if this is still on the list. :) 

Thanks for all your hard work!

roggervalf:
this is better handled in bullMQ-pro https://docs.bullmq.io/bullmq-pro/groups

PR #277, opened 2020-09-22
feat: Retrieve Jobs by ID pattern

Skn0tt:
I'm using bullmq for [Quirrel](https://quirrel.dev) and need a way to fetch pending jobs. This PR adds a new function `getJobByIdPattern` that can be used like the following:

```ts
const { newCursor, jobs: jobsByCustomer1 } = await queue.getJobByIdPattern("customer1:*")

// later on ...

const { jobs: nextBatchOfJobsByCustomer1 } = await queue.getJobByIdPattern("customer1:*", newCursor)
```

### DIFF
--- a/docs/gitbook/guide/fetching-jobs.md
+++ b/docs/gitbook/guide/fetching-jobs.md
@@ -1,0 +1,58 @@
+# Fetching Jobs
+
+BullMQ allows you to iterate through existing jobs using the methods `.getJobsFromIdPattern` and `.getJobsByName`.
+
+## `.getJobsFromIdPattern`
+
+Returns all jobs with IDs matching a specific pattern.
+
+```typescript
+import { Queue } from "bullmq";
+
+const queue = new Queue('painter');
+
+await queue.add("default", ... ,{ jobId: "foo.1" })
+await queue.add("default", ..., { jobId: "foo.2" })
+await queue.add("default", ..., { jobId: "bar.1" })
+
+const result = await queue.fromIdPattern(
+    "foo.*", // pattern to search for
+    0 // cursor, initially 0
+)
+
+result.jobs // includes `foo.1` and `foo.2`, but not `bar.1`
+result.newCursor // cursor to use for next iteration, `null` if scan is finished
+```
+
+{% hint style="info" %}
+Uses Redis' `SCAN` internally and thus
+will perform badly on big instances.
+{% endhint %}
+
+## `.getJobsByName`
+
+Returns all jobs with a specific name.
+
+```typescript
+import { Queue } from "bullmq";
+
+const queue = new Queue('painter');
+
+await queue.add("foo", ...)
+await queue.add("foo", ...)
+await queue.add("bar", ...)
+
+const result = await queue.getJobsByName(
+    "foo",
+    0 // cursor, initially 0
+)
+
+result.jobs // includes the first two jobs
+result.newCursor // cursor to use for next iteration, `null` if scan is finished
+```
+
+{% hint style="info" %}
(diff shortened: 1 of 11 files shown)


manast:
thanks for a very good written patch.
A couple of comments: 
I am wondering if you could share the complete use case for this, when you write pending jobs I am not sure what you mean. Studying the code, it will return all jobs matching the pattern even jobs that have been completed, failed or active, not just delayed and waiting, which would be what I would intuitively understand for "pending".
Secondly, it may be important to consider performance here, in a redis instance with many queues it could be slow to iterate across all job keys, also consider that the scan command can return zero keys if it does not match any keys in one iteration.
Thirdly, it would be great with some documentation for this new feature too :).

Skn0tt:
Thanks for the quick response :)

**About the use case:**
Quirrel is a managed task queue for serverless deployments. In its current implementation, it's basically a HTTP adapter to `bullmq`.
- `POST /jobs` adds a new job
- `DELETE /jobs/:id` can be used to prevent delayed jobs from executing.

Every job is given an ID that adheres to the following schema: `<customer>:<queue>:<job-id>`. 

I now want to add a new route `GET /jobs` that retrieves all delayed jobs that haven't been executed. It uses `getJobFromIdPattern('<customer>:*:*', ...)` to achieve that. Since I enabled `removeOnComplete`, this works for my use case.

**About performance:**

The [Redis docs on `SCAN`](https://redis.io/commands/scan) call it production-safe, so I don't think that should be a problem.

About matching jobs during one iteration: What do you propose to do about this? We could add another set that contains the IDs of a queue's jobs, but that'd come with increased complexity.

**About docs:**

Will do ^^

manast:
hmm, ok, lets see, do you mean you want to get all the jobs from a given customer, across all his queues?

Skn0tt:
I have one big queue that contains all of the jobs, rather than one per customer. If I had one per customer, there'd be a lot of idle Redis connections, which I don't want.

Individual customers are distinguished by the unique IDs their jobs receive: `<customer>:<queue>:<custom-job-id>`.

When a customer makes a call (say `GET /jobs/my-queue`), I scan across all delayed jobs for that customer x queue: `randomcustomer:my-queue:*`.

Skn0tt:
I've added a new method `.getJobsByName` that returns all jobs with a specific name. To make this more performant, I added a new `SET` called `by-name:<job-name>`, containing the IDs of all jobs with a specific name. This can then be `SSCAN`ed easily. It's updated in the `addJob` and `removeJob` commands.

As you rightfully demanded, I've also added docs.

nemphys:
Any news on this? Would be a very nice feature.

Skn0tt:
I don't need this feature anymore, resorted to writing [my own queueing library](https://github.com/quirrel-dev/owl) at some point. But still, this would be a very nice feature indeed!

manast:
This is a fine PR but the reason I have not merged it is because it creates a new data set which many users will not need, so I have been reluctant to merge it. Also, I think this could be refactored into a separate module that works side-by-side with Bullmq so that we do not need to increase the size of the core module.

Skn0tt:
That sounds sensible. I won't be able to write that 3rd party module since I don't use BullMQ anymore (see above) - but if anybody wants to do that, go ahead and grab this PR's code :D

nemphys:
Wouldn't it be nice if the new feature was optional through a parameter (and disabled by default, so that it does not affect users that don't need it)?
It would be a shame for this to be left out like this.

manast:
@nemphys if somebody wants to pick up on this job and implement it as an optional feature then there are good chances it will be merged.

PR #279, opened 2020-09-25
Feat/grouped rate limiter

manast:


bkniffler:
This is an amazing feature, is this likely to be merged some time?

PR #280, opened 2020-09-29
Add missing TS definition for rate limit options

danbenn:
### Summary

As indicated by tests/test_rate_limiter.ts: https://github.com/taskforcesh/bullmq/blob/3a958dd30d09a049b0d761679d3b8d92709e815e/src/test/test_rate_limiter.ts

There are more limiter options available than are currently included in the Typescript definition.


danbenn:
Working on fixing failing tests...

danbenn:
My bad, this is wrong, the rate limiter options are set on the _Worker_, not the Queue. Closing

PR #285, opened 2020-10-13
fix(worker): fix memory leak on Promise.race (#282)

### LINKED ISSUE #282: Memory leak
I have noticed that the jobs are leaking on the worker side, when a worker queues jobs on that worker queues.
For clarity, if the worker object is constructed with `new Worker('oom-test')` and tries to do `new Queue('oom-test').add(...)`, these jobs will leak.

However, if the worker queues job using a different queue (e.g. `new Queue('oom-test-one').add(...)`, then it no longer leaks.

In order to reproduce, see the example project attached.
1. If not present needed a folder named '/tmp' at the root of the project (this is where the heapsnapshot will be stored)
2. Run `npm install`
3. In one terminal, run `node master.js`, in another `node worker.js`. Inspect the snapshots using chrome, and verify they are not leaking.
4. In one terminal, run `node master.js`, in another `node worker-leak.js`. By inspecting the snapshots using chrome, you can notice that the 1MB strings have never been collecting and have an increasing distance.

[bullmq-leak.zip](https://github.com/taskforcesh/bullmq/files/5365731/bullmq-leak.zip)

Embraser01:
Fix #282 (explainations are available in the issue)

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -132,11 +132,13 @@ export class Worker<T = any> extends QueueBase {
 
       /*
        * Get the first promise that completes
-       * Explanation https://stackoverflow.com/a/42898229/1848640
        */
-      const [completed] = await Promise.race(
-        [...processing.keys()].map(p => p.then(() => [p])),
+      const promises = [...processing.keys()];
+      const completedIdx = await Promise.race(
+        promises.map((p, idx) => p.then(() => idx)),
       );
+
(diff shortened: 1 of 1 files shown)


manast:
Great job man! really appreciate it.

manast (approved):

manast:
I could not get the tests run on this PR so I just created a new one and merged it.

PR #292, opened 2020-10-20
removed top level `await`

VakoNonikashvili:


manast:
top level await is supported in node 14. In this case it is also serving as an example of usage. Removing the await will make some users think that the call is not asynchronous and lead to errors down the road.

Issue #305, opened 2020-11-04
cron expression error and cron job not working as expected. 


5achinJani:
According to the [docs](https://docs.bullmq.io/guide/jobs/repeatable)
![image](https://user-images.githubusercontent.com/23077544/98113804-2fda6100-1eca-11eb-9695-ffa24a40dcac.png)

First thing is the [cron expression](https://www.freeformatter.com/cron-expression-generator-quartz.html) link that you've mentioned in the docs results in error with "Unexpected end of expression."
![image](https://user-images.githubusercontent.com/23077544/98114417-3fa67500-1ecb-11eb-81a0-4151cd0897fa.png)

I used the same cron expression in the code:
![image](https://user-images.githubusercontent.com/23077544/98115273-92ccf780-1ecc-11eb-8087-36a018d729ee.png)

And cron runs every second :
![image](https://user-images.githubusercontent.com/23077544/98115209-7630bf80-1ecc-11eb-9e84-80e3629d4f90.png)

Here's the code for queue, worker and required things : 

![image](https://user-images.githubusercontent.com/23077544/98115600-171f7a80-1ecd-11eb-92fe-c851e47c56bd.png)

![image](https://user-images.githubusercontent.com/23077544/98115673-33231c00-1ecd-11eb-8a09-41e7658afc42.png)

![image](https://user-images.githubusercontent.com/23077544/98115728-4b933680-1ecd-11eb-8ad7-384766904e94.png)

node version : v11.14.0
```json
"bullmq": "^1.9.0",
```
Redis:
![image](https://user-images.githubusercontent.com/23077544/98115892-839a7980-1ecd-11eb-968b-565588e00f46.png)

5achinJani:
This doesn't look cron expression or [parser](https://www.npmjs.com/package/cron-parser) specific to me cause I even tried this :
```ts
export const registerPointsCronJob = async () => {
  await points_queue.add(
    points_jobs.POINTS_CRON_JOB,
    {
      params: {},
    },
    {
      repeat: {
        // cron: '15 3 * * *',
        every: 30000,
        limit: 10,
      },
    },
  )
  console.log('Points cron job registered')
}
```

And it keeps triggering every second and doesn't stop too.
![image](https://user-images.githubusercontent.com/23077544/98117164-5353da80-1ecf-11eb-9df5-d3d9832c5512.png)

I restarted my redis server and node server for both the cases just to make sure but it is still the same.

manast:
[...]

PR #307, opened 2020-11-06
Replace link to crontab generator, which used the wrong format.

sbleon:
The previously linked generator uses the non-standard Quartz "cron" format. The new link is a standard unix cron entry generator.

Addresses part of #305 .


manast:
Thanks but that link is not correct either, since the cron expressions in bull supports seconds too.

sbleon:
Oh, really? That's good to know. Does it actually use Quartz format, or
something bespoke?

manast:
I think this is documented somewhere but this is the module we use: https://www.npmjs.com/package/cron-parser

sbleon:
OK. I took a stab at incorporating that information here.

manast:
Documentation: https://docs.bullmq.io/patterns/manually-fetching-jobs

PR #310, opened 2020-11-14
Fix job types

francescov1:
Was looking into how I could setup BullMQ with typed job names (to ensure no unhandled jobs are added to queues). Ended up adding a few new generics to support those new types. Big ones are adding two new generics to Queue and Worker: job return values and job names, as well as augmenting the Processor with the same Queue and Worker generics.

Didnt change any actual code other than a small fix in `src/classes/sandbox.ts`. With new types, tslint didn’t like the unhandled promises in the msgHandler.

Been using BullMQ for some time now but not super familiar with the codebase itself, let me know if there’s anything you’d like me to change with this! 

Side note: I'm a huge fan of Bull, amazing piece of technology!

### DIFF
--- a/src/classes/compat.ts
+++ b/src/classes/compat.ts
@@ -94,7 +94,7 @@ export class Queue3<T = any> extends EventEmitter {
    * If the promise is rejected, the error will be passed as a second argument to the "failed" event.
    * If it is resolved, its value will be the "completed" event's second argument.
    */
-  async process(processor: string | Processor) {
+  async process(processor: string | Processor<T>) {
     if (this.worker) {
       throw new Error('Queue3.process() cannot be called twice');
     }
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -23,7 +23,7 @@ export interface JobJson {
   returnvalue: string;
 }
 
-export class Job<T = any, R = any> {
+export class Job<T = any, R = any, N extends string = string> {
   progress: number | object = 0;
   returnvalue: R = null;
   stacktrace: string[] = null;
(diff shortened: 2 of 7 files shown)


manast:
Looks great!.
It would be nice if you could write a github comment with some examples using the new typings so that I can update the documentation with them too.

francescov1:
Sure thing! How's this:

```typescript
import { Queue, Worker } from "bullmq";

// Types
interface JobData {
  foo: string;
}
interface JobReturnData {
  success: boolean;
}
type JobName = "myJob" | "anotherJob"

// Queue
const myQueue = new Queue<JobData, JobReturnData, JobName>('myQueue');

// Worker
const myWorker = new Worker<JobData, JobReturnData, JobName>('myQueue', async job => {
  // Error: This condition will always return 'false' since the types 'JobName' and '"undefinedJobName"' have no overlap.
  if (job.name === "undefinedJobName") {
    console.log("Would never get here")
  }

  // Error: Type '{ success: number; }' is not assignable to type 'JobReturnData'.
  return { success: 1 }
});

// Adding jobs
async function addJobs(){
  // Error: Argument of type '"undefinedJobName"' is not assignable to parameter of type 'JobName'.
  await myQueue.add('undefinedJobName', { foo: 'bar' });

  // Error: Argument of type '{ foos: string; }' is not assignable to parameter of type 'JobData'.
  await myQueue.add('myJob', { foos: "bar" });
}
```

francescov1:
When can we expect this to be merged? Our team is in need of stronger typings but we've been putting off using a custom fork with the hopes of having it officially supported. 

Let me know if there's anything else you need from me for the PR to be merged!

PR #316, opened 2020-11-21
Add a remark about scheduler for retries to work

sh0oki:
Took me some time to realize I need a QueueScheduler for retries to work, thought this may help people using the manual.
Thanks!


manast:
thanks!.

PR #321, opened 2020-11-29
docs: fix redis default port

johakr:


manast:
thanks for the correction.

PR #335, opened 2020-12-11
docs: add doc for bebounce usecase

Valtena:
Hi,

This days, I have search to debounce job with same Id in queue to avoid run multiple time an heavy computation. I didn't found this function here. I have fork project to implement this before see that it's already implemented. So i have tried to add a little example to doc this functionality.

### DIFF
--- a/docs/gitbook/patterns/debounce-jobs.md
+++ b/docs/gitbook/patterns/debounce-jobs.md
@@ -1,12 +1,16 @@
 # Debounces
 
-Sometimes, you want to update data in reactions of a sequence of events instead at each event. 
+Sometimes, you want to update data in reactions to a sequence of events instead at each event. 
 You can enforce `jobId` to be unique with `JobsOptions.jobId?: string`. 
-That override the job ID - by default, the job ID is a unique
+That overrides the job ID - by default, the job ID is a unique
 integer, but you can use this setting to override it.
 If you use this option, it is up to you to ensure the
 jobId is unique. If you attempt to add a job with an id that
 already exists, it will not be added.
+
+Hint: Be careful if using removeOnComplete/removeOnFailed options, 
+since a removed job will not count as existing and a new job with 
+the same job ID would indeed be added to the queue.
 
 example: 
 ```typescript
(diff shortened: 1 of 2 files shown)


docs/gitbook/patterns/debounce-jobs.md
```diff
@@ -0,0 +1,40 @@
+# Debounces
+
+Sometimes, you want to update data in reactions of a sequence of events instead at each event.
```
manast, line 3: I am not native english but I think "of a" -> "to a"

docs/gitbook/patterns/debounce-jobs.md
```diff
@@ -0,0 +1,40 @@
+# Debounces
+
+Sometimes, you want to update data in reactions of a sequence of events instead at each event. 
+You can enforce `jobId` to be unique with `JobsOptions.jobId?: string`. 
+That override the job ID - by default, the job ID is a unique
```
manast, line 5: overrides

manast: a unique number ->an increasing counter

docs/gitbook/patterns/debounce-jobs.md
```diff
@@ -0,0 +1,40 @@
...
+integer, but you can use this setting to override it.
+If you use this option, it is up to you to ensure the
+jobId is unique. If you attempt to add a job with an id that
+already exists, it will not be added.
```
manast, line 9: Hint: Be careful if using removeOnComplete/removeOnFailed options, since a removed job will not count as existing and a new job with the same job Id would indeed be added to the queue.

manast (changes requested):
I wrote some comments

docs/gitbook/patterns/debounce-jobs.md
Valtena: For `a unique integer ->an increasing counter`, I have a question. I pick this part in comm from `src/interfaces/jobs-options.ts` l 37. The comm in file is outdated  or need the same patch?

airhorns:
Would be great to get this in -- really handy feature but you have to know the trick to use `jobId` so I think this documentation would really help!

PR #340, opened 2020-12-15
feat: add ability to get if queue is paused or not

ersoma:
Currently it is possible to pause a queue or resume it but there were no way to find out if a queue is paused or not.
Directly using Redis commands would be an anti-pattern in my opinion as the name of the keys and the way how paused state is stored should be an internal detail that the users should not be aware of or interfere with.

I created a public method in the Queue class that exposes if it is paused or not. If the implementation should be renamed or replaced, I'm happy to move it around.

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -120,6 +120,12 @@ export class Queue<
     this.emit('resumed');
   }
 
+  async isPaused() {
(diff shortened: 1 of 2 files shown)


manast:
Thanks. This was actually implemented in Bull not a while ago and not ported to BullMQ.

PR #377, opened 2021-01-25
chore(redis-connection): add explicit promise type

evolkmann:
Silences a typescript warning when viewing the file.


manast:
thanks

PR #381, opened 2021-01-27
refactor(timer-manager): use Map to store timers

evolkmann:
Typed map of timers instead of a plain object and type "any".


manast:
thanks for another fine PR.

Issue #324, opened 2020-12-05
Remove client property from QueueBaseOptions


paeolo:
[QueueBaseOptions](https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/interfaces/queue-options.ts#L11) exposes both a `client` and a `connection` property. The connection property can be either `RedisOptions` or a `Redis` client, so it seems the `client` property is redundant.

And indeed, [looking at the code](https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/classes/queue-base.ts#L22) I noticed the client property is not used (in favor of the connection property), so it can mislead developpers that tries to use the client property.

Suggestion: If the `client` property is not used, and moreover redundant, remove it.

manast:
But the client property is just a convenient getter for connection.client:
https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/classes/queue-base.ts#L57

paeolo:
> But the client property is just a convenient getter for connection.client:
> https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/classes/queue-base.ts#L57

@manast  My point is to remove the client property from [QueueBaseOptions](https://github.com/taskforcesh/bullmq/blob/5e6631e7c3f40b7857d2f73b48abfd3f9d072b35/src/interfaces/queue-options.ts#L11), not the getter from [QueueBase](https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/classes/queue-base.ts#L57) class.
https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/interfaces/queue-options.ts#L13

Indeed, if you leave it like that, then the developer as the opportunity to do that:

```ts
import Redis from 'ioredis';
import { Queue } from 'bullmq';

const redisClient = new Redis();
const queue = new Queue('my-queue', { client: redisClient });
```
[...]

PR #403, opened 2021-02-16
fix: remove "client" property of QueueBaseOptions (#324)

paeolo:
This PR resolves https://github.com/taskforcesh/bullmq/issues/324.

[QueueBaseOptions](https://github.com/taskforcesh/bullmq/blob/5e6631e7c3f40b7857d2f73b48abfd3f9d072b35/src/interfaces/queue-options.ts#L11) has a `client` property that is not used by the code:
https://github.com/taskforcesh/bullmq/blob/5e6631e7c3f40b7857d2f73b48abfd3f9d072b35/src/interfaces/queue-options.ts#L13

Indeed, instead the `connection` property is used:
https://github.com/taskforcesh/bullmq/blob/5e6631e7c3f40b7857d2f73b48abfd3f9d072b35/src/classes/queue-base.ts#L23

Note that this is not related to the [client getter](https://github.com/taskforcesh/bullmq/blob/7372ef16465fdf7d3c803a9cd9c1a25422d53373/src/classes/queue-base.ts#L57) which also uses the aforementioned `connection` property:
https://github.com/taskforcesh/bullmq/blob/5e6631e7c3f40b7857d2f73b48abfd3f9d072b35/src/classes/queue-base.ts#L59

### DIFF
--- a/src/interfaces/queue-options.ts
+++ b/src/interfaces/queue-options.ts
@@ -1,6 +1,4 @@
 import { JobsOptions } from '../interfaces';
-
-import { Redis } from 'ioredis';
 import { ConnectionOptions } from './redis-options';
 
 export enum ClientType {
@@ -10,7 +8,6 @@ export enum ClientType {
 
 export interface QueueBaseOptions {
   connection?: ConnectionOptions;
-  client?: Redis;
   /**
    * Prefix for all queue keys.
    */


manast:
Thanks!

paeolo:
> Thanks!

Thanks for merging it, and thanks for your work;)

PR #428, opened 2021-03-13
Add methods to track related job dependencies

roggervalf:
This PR tries to add a method to link job dependencies. Currently I have some jobs that call other jobs, so it would be helpful to have a way to track those dependencies. Also I add a new option when creating a Job, to pass dependencies information. Here are the new methods:
* addDependency: this method adds a job as a dependency of another job, also it generates a dependent record for the parent.
* removeDependency: this method could be useful for who wants to delete a dependency relationship, maybe because a dependency is completed or if you just want to delete it.
* dependenciesReady: this method helps you to track if your dependencies are in completed state or not (if all dependencies are completed, this function returns false, on the other hand true).

Extention:

* create: now you can pass dependencies and depends option, it will generate the relationships in the creation flow of those jobs.


manast:
Thanks for the PR. I would like to know more about the use case for this. We have been discussing for a long time about parent-child relationships, so I am quite interested in real life scenarios where this is useful.

roggervalf:
Hi @manast, sure let me explain it, currently we have some jobs that trigger other jobs in different steps, those children jobs could belongs to different queues (not the same of the parent), also we are using https://www.npmjs.com/package/bull-arena to visualize our jobs, but currently we have to look in our code to see which job trigger which one, so my idea is to handle this as we handle logs, so we could have the queue reference and the child job id in this array https://github.com/taskforcesh/bullmq/pull/428/files#diff-4bca2e77f20a6d5c913b413082d06a26f1a9ff661be71a42e26491564b2c9e66R206, so in that way we could get the queue identifier and the job id to be used for example in bull arena ui, I would like also to open a pr in bull arena to support this new feature, what do you think?

loris:
@manast here is a personal use case where support for parent-child relationships would shine: we have an ETL platform which deals with huge CSV files. When a CSV file is sent to us, a parent bull job is created, which streams/parse the CSV file and creates 1 child bull job per line (for many reasons: distribute processing of each lines between many workers, keep processing remaining lines if some line are invalid, etc). That setup is great, but we miss a native support in bull where the parent job is aware of the status/completion of its children (ie, how many lines of that CSV file have been completely processed so far). Hope this helps.

ccollie:
A few years ago, I used [qless](https://github.com/seomoz/qless-core) which supports this natively. You may find some ideas there, though their implementation is pretty involved.

evolkmann:
@manast Another use case example.
We have an app where a user can create so called "product groups". The creation process consists of multiple asynchronous steps (fetch group metadata, fetch products that belong to the group, creation of db records and relations, inform other systems that we are now the leading system, etc.). These steps have to performed in a particular order and are implemented in different jobs. Upon completion of step 1, another job is scheduled to execute step 2, and so on ([as in the "Flows" docs](https://docs.bullmq.io/patterns/flows)).

The creation process itself can be triggered in the UI for a single group or via mass import of a CSV file (similar to what @loris described).

We currently do display the progress of certain long running sub-tasks, but we have no visualization of the complete end-to-end-process. @roggervalf We also use Arena, so a visualization there would also be a great addition!

src/commands/addJob-8.lua
```diff
@@ -103,4 +104,15 @@ end
 local maxEvents = rcall("HGET", KEYS[3], "opts.maxLenEvents")
 if (maxEvents) then rcall("XTRIM", KEYS[7], "MAXLEN", "~", maxEvents) end
 
+local dependencies={}
+if (ARGV[11]~="") then
+    ARGV[11]:gsub("([^".." ".."]*)",function(c) table.insert(dependencies,c) end)
```
roggervalf, line 109: @manast what do you think about this, I'm using " " to separate dependencies strings, are we safe to use this delimiter or do I use another one?

roggervalf:
Thank you guys for your feedback, I changed a little bit the logic and add some methods that would help us to track these dependencies as in the new description of these PR. Please take a look.

evolkmann:
@roggervalf Thanks for the update, I guess the naming with `dependencies` is better than `child/parent` 👍🏻

- - - -

However, I would suggest to reverse the naming and logic of `waitForDependencies()`.

> waitForDependencies: this method helps you to track if your dependencies are in completed state or not (if all dependencies are completed, this function returns false, on the other hand true)

In my opinion, it would be easier to understand if the function was named for example `dependenciesReady()` and if it returned `true` if all dependencies are completed.

manast:
ok, This is my third attempt to write a meaningful comment :).
I am trying to wrap around my head on how this api can be used in the standard way of working with BullMQ, i.e. using workers with defined processor functions. Because as I see in the tests, when we create the parent job it could be completed directly without any chance to add the dependencies. In my idea of how this should be resolved I think that it should be possible to specify the dependencies at the same time as the parent job is added to a queue, I think it is ok to not allow adding dependencies after the parent job is in the queue. The other thing is that when all the child jobs complete, they should also "complete" the parent job inside the finish lua script. However this does not seem to be enough, we wouldn't need to ask the parent job if it has completed or not, instead it would be better if the parent job starts to get executed, i.e. its processor function called, when all its children have completed, so it will get a list of all the children and can do some final work, for example, it could be that it needs to stitch all the children into a larger result and store it in a database. A parent job as such would not exists in the queue at all until their children are completed, it would be a virtual job of sorts, only when the children are completed the job it self is created and can be processed... Not sure if I am making sense with my explanation above. But as I envision it it would be more flexible and also possible to chain jobs that include dependencies to other jobs and so on.  I am available for completing this PR so that it suits the ideas above, but I would like to get as much feedback from you guys as possible, maybe you think my idea is completely wrong and if so I would like to hear it.

roggervalf:
> @roggervalf Thanks for the update, I guess the naming with `dependencies` is better than `child/parent` 👍🏻
> 
> However, I would suggest to reverse the naming and logic of `waitForDependencies()`.
> 
> > waitForDependencies: this method helps you to track if your dependencies are in completed state or not (if all dependencies are completed, this function returns false, on the other hand true)
> 
> In my opinion, it would be easier to understand if the function was named for example `dependenciesReady()` and if it returned `true` if all dependencies are completed.

hi @evolkmann thank you for the suggestion, I updated the PR

roggervalf:
> ok, This is my third attempt to write a meaningful comment :).
> I am trying to wrap around my head on how this api can be used in the standard way of working with BullMQ, i.e. using workers with defined processor functions. Because as I see in the tests, when we create the parent job it could be completed directly without any chance to add the dependencies. In my idea of how this should be resolved I think that it should be possible to specify the dependencies at the same time as the parent job is added to a queue, I think it is ok to not allow adding dependencies after the parent job is in the queue. The other thing is that when all the child jobs complete, they should also "complete" the parent job inside the finish lua script. However this does not seem to be enough, we wouldn't need to ask the parent job if it has completed or not, instead it would be better if the parent job starts to get executed, i.e. its processor function called, when all its children have completed, so it will get a list of all the children and can do some final work, for example, it could be that it needs to stitch all the children into a larger result and store it in a database. A parent job as such would not exists in the queue at all until their children are completed, it would be a virtual job of sorts, only when the children are completed the job it self is created and can be processed... Not sure if I am making sense with my explanation above. But as I envision it it would be more flexible and also possible to chain jobs that include dependencies to other jobs and so on. I am available for completing this PR so that it suits the ideas above, but I would like to get as much feedback from you guys as possible, maybe you think my idea is completely wrong and if so I would like to hear it.

Thank you @manast for the quick response, I added in last version the feature to pass the dependents when creating a job. I think I got your point, in my team we have parent jobs that in some steps we trigger some dependency jobs, so at another step we need to wait for that dependency to be completed, then we can continue with the parent flow, so we don't complete the parent job when the dependencies are completed, we need to wait in specific steps for our dependencies to be completed, I tried to handle the other cases with the dependenciesReady function, so we can have a way for checking the completion of the dependencies. I didn't complete the job in lua since I think we want a way to check those dependencies in our parent work flow. Now you can create a dependency job passing dependents to it, so now the parent job will track those jobs when those dependencies are created, I forgot to add that in the previous version, thank you. I'll open to more suggestion and I'll glad also to have your help in this new feature too

manast:
So what about something like this, we define a new option on job options called parent:
```typescript
interface Parent<T> {
  name: string,
  queue: string,
  data: T
}
```
This option can be used when adding jobs to the queue, but if you have several children you should use addBulk instead, for example:
```typescript
myqueue.addBulk(jobs, { parent: { 
  name: 'foo',
  queue: 'bar',
  data: {qux: 'baz'} 
}});
```
The parent job will not be created until all the children have completed successfully, on the "finish" lua script we will keep track of the completed children jobs and when all are completed we will add the parent job passing the results from all the children in its processor function (we will need to add a third optional argument to the processor function).
Maybe I am missing something but I think this simple design provides a lot of flexibility and covers much of the cases exposed in this old thread: https://github.com/OptimalBits/bull/issues/340
what do you think?

roggervalf:
> So what about something like this, we define a new option on job options called parent:
> 
> ```ts
> interface Parent<T> {
>   name: string,
>   queue?: string,
>   data?: T
> }
> ```
> 
> This option can be used when adding jobs to the queue, but if you have several children you should use addBulk instead, for example:
> 
> ```ts
> myqueue.addBulk(jobs, { parent: { 
>   name: 'foo',
>   queue: 'bar',
>   data: {qux: 'baz'} 
> }});
> ```
> 
> The parent job will not be created until all the children have completed successfully, on the "finish" lua script we will keep track of the completed children jobs and when all are completed we will add the parent job passing the results from all the children in its processor function (we will need to add a third optional argument to the processor function).
> Maybe I am missing something but I think this simple design provides a lot of flexibility and covers much of the cases exposed in this old thread: [OptimalBits/bull#340](https://github.com/OptimalBits/bull/issues/340)
> what do you think?

hey @manast I tried to take advantage of the Depend type to pass the dependents and dependencies in bulk function, could you please take a look :eyes:, I still don't do the second part, because maybe would be good that the user be aware to use dependenciesReady function to wait for their dependencies and not only at the end of the parent job final step, for example when a parent job that has 5 steps, in the 2 step it triggers a dependency, in step 3 it triggers another dependency, so in step 4 we wait for those dependencies to be completed, in the 5 we use the result of those jobs getting those values from db for example, step 6 parent job continue and it completes. I guess more of our cases are in waiting in some part of the parent flow for dependency results, so it can continue with other tasks. Or maybe I don't get your idea very well :sweat:

manast:
@roggervalf no, I do not think we should wait for dependencies to complete, specially inside processors. A processor should perform whatever it does as quickly as possible so that the jobs in the queue are processed as soon as possible. Having a processor that waits for other jobs to finish is going to create queue congestion and other problems down the road. But as I mentioned, I do not think it is needed at all to achieve your use case. The case you are describing, a process that is composed on 5 steps, the most robust and efficient way to solve it is dividing it in different jobs (probably running in different queues also for more flexibility). After every step you add the next step job or jobs. So step 1 completes adding the step 2 child jobs, when all the child jobs are completed, BullMQ  will automatically add the "parent" job which will now perform step 3 with the result of all the children. Having one big job that performs many steps is also problematic in cases of failure, it is easier to retry small failed jobs, and in fact you should strive after having "idempotent" jobs, i.e. jobs that even if run twice they will produce the same result.

roggervalf:
> @roggervalf no, I do not think we should wait for dependencies to complete, specially inside processors. A processor should perform whatever it does as quickly as possible so that the jobs in the queue are processed as soon as possible. Having a processor that waits for other jobs to finish is going to create queue congestion and other problems down the road. But as I mentioned, I do not think it is needed at all to achieve your use case. The case you are describing, a process that is composed on 5 steps, the most robust and efficient way to solve it is dividing it in different jobs (probably running in different queues also for more flexibility). After every step you add the next step job or jobs. So step 1 completes adding the step 2 child jobs, when all the child jobs are completed, BullMQ will automatically add the "parent" job which will now perform step 3 with the result of all the children. Having one big job that performs many steps is also problematic in cases of failure, it is easier to retry small failed jobs, and in fact you should strive after having "idempotent" jobs, i.e. jobs that even if run twice they will produce the same result.

thank you @manast for clarify it, also in my team we had this functionality that is called waitForSubCommands, this the one that checks if children jobs are completed if it is not the case, we throw an error (agree that the job should not keep waiting, I was meaning checking maybe follow this strategy to not cause congestion) and with the exponential backoff time, at some point the parent job will retry its process at that same steps to check if children jobs are completed. We are using other queues to create the children jobs to not have queue congestion. We don't have this check natively and just using this method that someone in my team built at the past. What do you think about it? Also could you please help me with some pseudocode changes or functionalities, I can add more things in this PR with your help

manast:
@roggervalf I am planing to work on this feature tomorrow, probably use or base my solution in some of your code in this PR.

manast:
Please take a look at this WIP PR https://github.com/taskforcesh/bullmq/pull/454 which is mostly inspired by this PR, and maybe we can reuse some bits of it, not sure. This is my main activity now until the PR is complete, so please do not hesitate to give feedback if you can.

manast:
As you can see from my PR, a parent job is created but is placed in a special set "wait-children", first when all children are processed the parent is moved to "wait" and can therefor be processed by any worker. Since you can use different queues for children and parents I think you do not need the "waitForChildren" function anymore, since either having a running worker on the parents queue will just pick up the parent as soon as it is available, or if fetching it manually it will be a similar effect to "waitForChildren", just that you do not need to specify which parent you want to wait for, you will get the next one in line (which in my opinion is the way to work with queues).

loris:
@manast sounds awesome! quick question though since i'm not sure how will behave jobs in the "wait-children" set: If the parent job gets processed, creates the children jobs, waits for them to be completed and then its process gets restarted (for instance let's say the server which initially started the parent job got restarted because of a crash caused by an unrelated job), how will the recovery be handled? will the parent job gets restarted from the start (and then, creates duplicate children jobs) or not?

manast:
@loris If the children jobs have been completed, then the parent job will be moved to the wait list of its respective queue and since it is a regular job from now on (only different is that it can request the children return values), it will behave as such in case of crashes, etc. So the answer to your question is that it will be restarted (since it became stalled after the crash) but the children will not be processed again.

manast:
Btw, as it is implemented it would be possible to have children jobs in different queues too, so one parent in a given queue and each and of the children in different queues. However the current API only permits children in one queue but thats because I am reusing the addBulk functionality for this. We will need a new set of methods to allow other scenarios.

PR #450, opened 2021-04-04
Upgrade ioredis

roggervalf:
This upgrade is in order to support the new methods in redis client, since now we cannot use lpos directly for example. After this PR is approved I could update #445

### DIFF
--- a/src/test/test_repeat.ts
+++ b/src/test/test_repeat.ts
@@ -142,7 +142,7 @@ describe('repeat', function() {
     const worker = new Worker(queueName, async job => {});
 
     const date = new Date('2017-02-07 9:24:00');
-    this.clock.tick(date.getTime());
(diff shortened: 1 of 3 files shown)


src/test/test_repeat.ts
```diff
@@ -142,7 +142,7 @@ describe('repeat', function() {
     const worker = new Worker(queueName, async job => {});
 
     const date = new Date('2017-02-07 9:24:00');
-    this.clock.tick(date.getTime());
+    this.clock.setSystemTime(date);
```
roggervalf, line 145: I had to change tick method for setSystemTime because in these tests, they were getting stock. With this change, these tests are passing as expected

manast:
Great. Thanks for fixing the typos too! :).

PR #454, opened 2021-04-07
feat: add support for flows

manast:
This PR adds support for parent-child jobs. The basic idea is that a parent job is a job that exists in some queue that only is considered for execution (moved to the wait list) when all its children has been completed. Parent jobs that are waiting for children are stored in a special set, with conceptually status  of "waiting for children". Currently the mechanism is implemented only for Bulk jobs and the api is a bit rough in the corners, however the design should be robust enough to be able to support things such as trees of jobs where children for a given parent are in turn parents for other children and so on. Also would allow chains of jobs, which is a specific case of the "tree of jobs" but with only one child per level.
This PR should also cover the most important use cases discussed here: https://github.com/OptimalBits/bull/issues/340 however needs to be ported to Bull v3 (the port should be more or less trivial after this PR is completed.).


pbittyscu:
Just wanted to chime in and say thanks for all of your work on this. We are currently working on adding parent-child (dependent) jobs to Bull 3 by rolling our own. So, I've been tracking this issue very closely to see if it made sense for us to uptake it. Let me know if there's certain kinds of tests that would help and we'll be sure to run them when we uptake this feature in Bull 4.

manast:
I refactored the initial implementation into a separate class "flow". Besides being more convenient it is also more flexible, it is not possible to create a full tree of jobs and dependencies between them.

ccollie:
Looking through the commits, it does not seem that changes were made to accommodate removing dependencies, or updating parents when a child is deleted. 
 Am I missing something or was that an oversight ?

manast:
@ccollie what is the use case for removing dependencies?

ccollie:
> @ccollie what is the use case for removing dependencies?
I see may have worded that in an ambiguous way. I meant that when a job is removed, the dependent jobs should (probably?) also be removed. Also, when a child is removed, the dependency list of its parent should be updated.

In the first instance, I guess for "idempotent" processes ? If i have an order composed of multiple parts (e.g. a job representing the transcoding of a series, where every child represents an episode). In the example each dependency is also relatively expensive.

For the second (a child being removed) - mostly for completeness and to ensure internal consistency, since the delete code currently makes no distinction between regular jobs and jobs in a flow. So the code as is could lead to possibly stalled parents, since a parent is only executed based on the completion state of the child. The id would still exist in the `waiting-children` list even though the job itself would no longer exist.

manast:
> I see may have worded that in an ambiguous way. I meant that when a job is removed, the dependent jobs should (probably?) also be removed. Also, when a child is removed, the dependency list of its parent should be updated.

Yes. Definitely when removing a child it should be removed from the dependecy list of the parent, and if it was the last child it should start processing the parent. Removing the childs when removing a parent is desirable but we need to consider possible edge cases. For example, if a child job has started to be processed we should not touch that job, only the ones that are in waiting/waiting-children statuses. 

I will work on a new PR for adding support for this, thanks for the feedback.

lukepolo:
Wow amazing. This is huge for bull IMHO.

Pietrox:
OMG!

manast:
There is some initial documentation too: https://docs.bullmq.io/guide/jobs/parent-child-jobs
please keep in mind some features are missing like "remove" as mentioned above.

Pietrox:
Somehow I can't set the name of the FlowJob. Tried manually hardcoding it, via factory, classToPlain even from my Dto.
It seems the property "name: string" no matter what will set job name as "__default__".

manast:
> Somehow I can't set the name of the FlowJob. Tried manually hardcoding it, via factory, classToPlain even from my Dto.
> It seems the property "name: string" no matter what will set job name as "**default**".

please open an issue and I will look into it asap

Pietrox:
#468 

Added with small example

manast:
@ccollie https://github.com/taskforcesh/bullmq/pull/474

loris:
@manast Flows look totally awesome and a great improvements to Bull. Quick question regarding some of my use cases, not sure the current implementation of flows would work: Could it work with child jobs which are not added all at once? I have use cases where I need to create lots of children jobs using streaming (ie, stream of CSV lines, stream of mongo documents, etc) and would love to use the Flow feature to be notified when all the children jobs are successfully processed.

manast:
@loris we have a PR open exploring this functionality (https://github.com/taskforcesh/bullmq/pull/477). 
In your case, if I understand it correctly, since you are streaming you need to create the jobs dynamically. I see two cases, the first one where your number of jobs is not very big, then you can just create a new flow with all these jobs so that atomically they are added to the queue and you get the guarantee that they will eventually complete and when so the parent will be processed and everything is nice and tidy. The other case is when you have a very large amount of jobs so that you cannot add them all atomically in a practical way (due to memory constraints or whatever). And this second case is complicated with several edge cases. For instance how would you mark that all the children have now been added so that the queue will know that it is time to execute the parent?, what happens if you manage to add 20% of the child jobs and then the process dies or is restarted for some reason, you get dangling children and ugly stuff...

loris:
@manast good to know about the PR, I'll have a look!
Indeed, we are on the complicated case: we use streaming here because the number of children jobs can be huge (millions) so they are created using streaming for obvious reasons (memory). Right now, we are able to know when all children have been added (when the stream ends), but not when the children jobs have been completed themselves.
Regarding the parent process crashing, this is a good question, and it can happen in our case (some parents can run for dozens of minutes) but we ensured all our children jobs are idempotent whatever they do.

PR #473, opened 2021-04-16
Consider waiting-children state when getState

roggervalf:
We should also consider waiting-children when getting an state for a job. Also I was playing with getDependencies and I got an error so I tried to take the opportunity to fix it here.


manast:
great thanks!.

PR #477, opened 2021-04-17
Feat move to waiting children for manual processing

roggervalf:
Trying to get advantage of the new state of waiting children, so when creating a job we can relate a parent, then we can decide in a parent flow when to wait for children with this new method.
Description:
moveToWaitingChildren: allows you to move a job from active state to waiting-children. In manual processing, you can decide when to wait for those children to be completed as you need. Remember that you cannot finished a job if there are children jobs that are not completed.
Note: in order to create children jobs in a manual way, you can pass the parent option when creating one, the parent job could still be processed and decide when to wait for those children with moveToWaitingChildren.


src/classes/job.ts
```diff
@@ -528,14 +533,22 @@ export class Job<T = any, R = any, N extends string = string> {
     const queue = this.queue;
 
     const jobData = this.asJSON();
+    const parentDependenciesKey = this.opts.parent
```
manast, line 536: we can now use ```this.parentKey``` which is already the same as `${this.opts.parent.queue}:${this.opts.parent.id` we could even deprecate the parent opt soon since it is now superfluous.

manast (changes requested):
Thanks for the PR.
 I gave you some comments.
It would be great if you could provide a description of the complete usecase since I will need to document how this feature is used and I am not grasping all the details.
 I get that this method can be used within the process function of a worker, so that you can go from being active to waiting a series of children. But I have 2 questions:
1) how to do avoid that the job that is being processed and moved to "wait-children" is not moved to the "completed" set cleanly? I suspect you get an error now since the job is not in the active status when the processor ends and tries to "complete" the job.
2) When and how do you define the children? I think that moveToWaitingChildren should be an implicit operation when actually adding the children, otherwise there is a risk that you manage to move the job to waiting-children but fail to actually add the children and so you will end with a dangling parent job that can never be processed.

src/classes/job.ts
roggervalf: hey, I'm seeing that for .create the only way for now to obtain the parentKey is with opts, no sure if this should be refactored here, since await Job.create does not have this.parentKey when creating a job. What do you think about it?

roggervalf:
> Thanks for the PR.
> I gave you some comments.
> It would be great if you could provide a description of the complete usecase since I will need to document how this feature is used and I am not grasping all the details.
> I get that this method can be used within the process function of a worker, so that you can go from being active to waiting a series of children. But I have 2 questions:
> 
>     1. how to do avoid that the job that is being processed and moved to "wait-children" is not moved to the "completed" set cleanly? I suspect you get an error now since the job is not in the active status when the processor ends and tries to "complete" the job.
> 
>     2. When and how do you define the children? I think that moveToWaitingChildren should be an implicit operation when actually adding the children, otherwise there is a risk that you manage to move the job to waiting-children but fail to actually add the children and so you will end with a dangling parent job that can never be processed.



> Thanks for the PR.
> I gave you some comments.
> It would be great if you could provide a description of the complete usecase since I will need to document how this feature is used and I am not grasping all the details.
> I get that this method can be used within the process function of a worker, so that you can go from being active to waiting a series of children. But I have 2 questions:
> 
>     1. how to do avoid that the job that is being processed and moved to "wait-children" is not moved to the "completed" set cleanly? I suspect you get an error now since the job is not in the active status when the processor ends and tries to "complete" the job.
> 
>     2. When and how do you define the children? I think that moveToWaitingChildren should be an implicit operation when actually adding the children, otherwise there is a risk that you manage to move the job to waiting-children but fail to actually add the children and so you will end with a dangling parent job that can never be processed.

sure, for your questions:
1. Yes, I'm moving the job to wait-children if I find dependencies, on the other hand it is not moved to wait-children and can continue with his job. When it is moved to waiting-children it is deleted from active.
2. In this case I'm checking if there are dependencies for the parent job, so as those are not processed the job can be moved, with this check we prevent to have dangling parent job, I add an example in a test where job are processed in a manual way, so children can be processed before the parent go to waiting-children.
From my use case I'm currently working with a parent job that can trigger in some steps children jobs. i.e:
Steps:
1. Step 1 do something
2. Step 2 create a child job (could more or less time to be processed than step 3)
3. Step 3 do another thing (don't depend on children jobs)
4. Step 4 create another child job (could more or less time to be processed than step 5)
5. Step 5 do another thing (don't depend on children jobs)
6. Step 6 waiting for children
7. Step 7 we already have children finished so we can use their results (depends on children results)
8. Step 8 completed
So with with moveToWaitingChildren we can have the flexibility to wait for children and maybe have step 3 and 5 already done, sometimes we depend on jobs can take a lot of time to be completed so would be nice to save for the parent to also processed steps that don't depend on those children.

src/classes/job.ts
roggervalf: In last version I reused getParentKey to refactor this, now parentKey could be assigned in job constructor

src/commands/moveToWaitingChildren-4.lua
```diff
@@ -0,0 +1,42 @@
...
+local rcall = redis.call
+
+if ARGV[1] ~= "0" then
+  if rcall("GET", KEYS[1]) ~= ARGV[1] then
+      return -2
+  end
+end
+
+if rcall("EXISTS", KEYS[4]) == 1 then
+  if rcall("SCARD", KEYS[4] .. ":dependencies") ~= 0 then
```
manast, line 29: I do not grasp why this check is needed?. As I understand it, it only moves the job if there are some dependencies already for that job, but shouldn't we check instead if this particular dependency is part of the dependency list?

src/test/test_job.ts
```diff
@@ -61,6 +61,84 @@ describe('Job', function() {
...
+    ]);
+
+    const processingParent = new Promise<void>((resolve, reject) => [
+      (parentProcessor = async (job: Job) => {
+        try {
+          const movedToWaitingChildren = await job.moveToWaitingChildren('0');
+          const { processed } = await job.getDependencies();
+
+          if (Object.keys(processed).length !== 3) {
+            expect(movedToWaitingChildren).to.be.true;
```
manast, line 92: hmm, in this case the promise will never resolve.

src/test/test_job.ts
```diff
@@ -61,6 +61,84 @@ describe('Job', function() {
...
+        if (processedChildren === values.length) {
+          resolve();
+        }
+      }),
+    ]);
+
+    const processingParent = new Promise<void>((resolve, reject) => [
+      (parentProcessor = async (job: Job) => {
+        try {
+          const movedToWaitingChildren = await job.moveToWaitingChildren('0');
```
manast, line 88: why is '0' the token and not parentWorker.token ?

src/test/test_worker.ts
```diff
@@ -1684,6 +1684,116 @@ describe('workers', function() {
...
+
+      expect(isActive1).to.be.true;
+
+      await child1.moveToCompleted('return value1', childToken);
+      const {
+        processed: processed2,
+        unprocessed: unprocessed2,
+      } = await parent.getDependencies();
+      const movedToWaitingChildren = await parent.moveToWaitingChildren(
+        parentToken,
```
manast, line 1744: parentToken should be worker.token, and anything else different from worker.token should raise an exception.

src/test/test_job.ts
manast, line 88: I am still a bit confused that we can allow this call from inside the worker processor, when the call ends it will try to move the job to "completed" from "active", but since we manually moved the job to "waiting-children" it should raise an error somewhere.

manast (changes requested):

src/commands/moveToWaitingChildren-4.lua
roggervalf: oh, here it is to move the parent, I'm checking if there is any dependency in the dependency list of this parent job

src/test/test_job.ts
roggervalf: let me refactor this test

src/test/test_job.ts
roggervalf: tyty I was not sure where to get a token in this case

src/test/test_worker.ts
roggervalf: I'm reusing the parentToken that I passed to the worker when calling getNextJob, if I use parentWorker.token it gets an error of **Property 'token' does not exist on type 'Worker<any, any, string>'. Did you mean 'toKey'?**
Selection_024

src/test/test_job.ts
roggervalf: sure, in moveToFinished I'm also checking if there are pending dependencies, in case there are, throws an error

src/commands/moveToWaitingChildren-4.lua
roggervalf: also this gives another feature that I can add, as you mentioned we can also have a method to check if a particular dependency is part of the dependencies list. For now this method is what my team currently is kind of emulating, waiting for all the children at some point. Let me think how to refactor this to also support on waiting for a particular dependency that is passed on

src/test/test_worker.ts
manast: ah sorry, you are right, token is private and when getting jobs manually you must create the token yourself.

src/test/test_job.ts
manast: ok. However in this case we really need the token from the worker that picked the job, I am surprised this is working since moveToWaitingChildren must match the stored token in the lock and the token passed to it. For this to work we need to pass the token to the processor, maybe as an extra parameter to the processor function.

manast: yes I saw that, but the current test code, does it not trigger that  error? since the job has been moved and when the processor call returns the worker will try to move the job to completed and your check should return an error. So two things here to think about, how to handle this error (I suspect it is erroring silently now), and two, what is the proper way to do this without raising an error. Somehow we have to tell the worker that we have moved the job to waiting-children so that it does not matter anymore if the job was in active, and in this case do a no op. Or maybe "moveToWaitingChildren" should be a method that can only be used when processing jobs manually. Do you have a need for this use case where it has to be called inside the processor function?

roggervalf:
@manast the pr was updated with master, please when you get some time :eyes:

src/commands/moveToWaitingChildren-4.lua
```diff
@@ -0,0 +1,60 @@
...
+  else
+    if rcall("SCARD", KEYS[4] .. ":dependencies") ~= 0 then 
+      return move_to_waiting_children(KEYS[2], KEYS[3], ARGV[4], ARGV[3])
+    end
+
+    return 1
+  end
+end
+
+return -1
```
manast, line 60: missing newline

manast:
Can you join for a conversation here: https://bullmq.slack.com/ I would like to discuss this PR in more detail.

roggervalf:
> Can you join for a conversation here: https://bullmq.slack.com/ I would like to discuss this PR in more detail.

sure, I guess I should be invited to that workspace :sweat_smile:. rogger.valverde@uni.pe

src/test/test_job.ts
```diff
@@ -63,6 +63,80 @@ describe('Job', function() {
...
+        }
+      }),
+    ]);
+
+    const processingParent = new Promise<void>((resolve, reject) => [
+      (parentProcessor = async (job: Job, token: string) => {
+        const movedToWaitingChildren = await job.moveToWaitingChildren(token);
+        const { processed } = await job.getDependencies();
+
+        if (Object.keys(processed).length !== 3) {
```
manast, line 92: so basically here, this will be true the first time this job is processed right? and then when it is processed the second time after all children are processed it will be false. Please correct me if my assumption was wrong.

roggervalf: yes you are right, the first try it should have children, after it move waiting-children the second try will go to the else block

manast: in both this cases there must be some silent errors since the first time the job fails but BullMQ is not able to move to job to failed, and in the other case it completes and then it works. I want to figure out an API where moving the processed job from within a processor is legit without breaking current semantics, since I see other uses cases that are important, for example being able for a job to delay itself if some condition is met during the processing.

roggervalf: Sounds also like a good addition to support polling

src/classes/worker.ts
```diff
@@ -326,12 +326,20 @@ export class Worker<
...
+      this.emit('waiting-children', job, 'active');
+    };
+
     // TODO: how to cancel the processing? (null -> job.cancel() => throw CancelError()void)
     this.emit('active', job, null, 'waiting');
 
     lockExtender();
     try {
       const result = await this.processFn(job, token);
+      if (typeof result === 'string' && result === 'waiting-children') {
```
roggervalf, line 340: @manast, what do you think about this change?, we could send a reserved word to handle the moveToWaitingChildren in a worker processJob

manast: I thought about this but I think it is a bit hackish. Either enhance the "result" object so that it is richer, like:
```ts
interface Result {
   data?: T,
   type: 'data' | 'waiting-children' | 'delay' | 'etc '
}
```
or having an internal job result type that is set up implicitly when calling job.moveToWaitingChildren, etc.
The problem with the first approach (that is probably the cleaner one), is that it will become a breaking change...

roggervalf: About the second one, also I was thinking on setting a field in job object, I can try the second one

roggervalf:
@manast this pr is ready for another round xd please when you had a chance

manast:
This looks good to me for manually processing children. I can merge it if you do not have anything else to add. 
I wont be able to write about this in here: https://docs.bullmq.io/patterns/manually-fetching-jobs because this use case is still a bit foreign to me, so I will probably write something that is not reflecting the real use cases.

roggervalf:
> This looks good to me for manually processing children. I can merge it if you do not have anything else to add.
> I wont be able to write about this in here: https://docs.bullmq.io/patterns/manually-fetching-jobs because this use case is still a bit foreign to me, so I will probably write something that is not reflecting the real use cases.

Sure, it is ready

PR #484, opened 2021-04-19
Passing token in processor function

roggervalf:


src/interfaces/worker-options.ts
```diff
@@ -6,6 +6,7 @@ import { AdvancedOptions, QueueBaseOptions, RateLimiterOptions } from './';
  */
 export type Processor<T = any, R = any, N extends string = string> = (
   job: Job<T, R, N>,
+  token?: string,
```
manast, line 9: I think it cannot be optional, it is always there but you optionally use it or not :).

manast (changes requested):
looks good, just a small comment

src/interfaces/worker-options.ts
roggervalf: let me address it really quick :sonic:

Issue #64, opened 2019-11-07
Job#moveToCompleted on delayed job moves it to active


jamesholcomb:
When I add a job to a queue with a long delay (e.g. 60s), then subsequently call moveToCompleted before the delay has passed, the job goes active and is executed by the QueueScheduler.

My current workaround is to Job#remove but I would rather keep the history of all jobs.

Is this by design?

manast:
ok, so I think we overlook moveToCompleted, it is not taking into consideration that the job could be in the delayed status. So it moves the job to the completed set, but since it is also still in the delayed set it will execute when the delayed time is over.

roggervalf:
@jamesholcomb you can try last version, now moveToCompleted throw an error if you try to move to completed if the job is not in active state

jamesholcomb:
Can you summarize the changes that were made?  It's been a while since I looked at this particular issue.

roggervalf:
sure, now moveToCompleted verifies when deleting the job from active key, if that record was deleted or not, in case it was not deleted because there wasn't one, it throws an error that says that the actual job was not in active state, so job is not moved to completed, so now we should make sure that the job is in active state before moving it to completed state being aware of this

manast:
So I think we could close this issue now?

jamesholcomb:
Thanks @roggervalf. I reviewed my code and do not utilize moveToCompleted any longer.

PR #485, opened 2021-04-20
Throw an error if job is not in active state when moveToCompleted

roggervalf:
This PR updates the current test that were failing, the reason was that workers were missing, so the jobs were getting waiting state #64


manast:
great, very much appreciated!.

PR #486, opened 2021-04-21
Just publish dist files

roggervalf:
Currently, a lot of files are published, docs, dist, src, ... etc. We can just publish the ones that are in dist since these files are built. you can now use yarn dry-run to see which files are going to be published to verified this, also with this change we are going to decrease the size of this package a lot.


manast:
good point!. Thanks!.

PR #497, opened 2021-04-24
Fix moveToComplete, should not move to complete before children

roggervalf:
Currently we can manually move parent to complete, even if it has children. We should verify that this job does not have pending children first, if not throw an error


manast (approved):

manast:
thanks!.

PR #499, opened 2021-04-25
Feat: allow passing parent in job.create

roggervalf:
For now we should first call new Job and then Job.create to create the job with a parent. This PR allow us to call Job.create directly with the parent params


src/test/test_flow.ts
```diff
@@ -326,7 +328,7 @@ describe('flows', () => {
     expect(numJobs).to.be.equal(0);
 
     await flow.close();
-    await parentWorker.close();
```
roggervalf, line 329: this line was called in line 325, so it is a duplicate

manast:
Looks very good, thanks.

PR #505, opened 2021-04-27
Dev dependency chai as promised

roggervalf:
We could extend the error expectation functionality with this dev dependency instead of p-reflect


manast:
looks good, thanks!.

PR #507, opened 2021-04-28
Fix Redis cluster CROSSSLOT

arsonik:
When a job has no parent &/or children 
The code send a command with 2 empty keys to the Lua script
Resulting in a CROSSSLOT error.


manast:
Thanks for the PR. However I think a cleaner solution would be to just pass the parents keys as args. In fact, since BullMQ does not take advantage of redis clusters there is no really need for sending keys at all.

gregorg:
Many users of BullMQ uses an existing Redis Cluster, you should consider to support it.

manast:
> Many users of BullMQ uses an existing Redis Cluster, you should consider to support it.

I am just suggesting an alternative implementation that is simpler...

gregorg:
> > Many users of BullMQ uses an existing Redis Cluster, you should consider to support it.
> 
> I am just suggesting an alternative implementation that is simpler...

And I answered your 2nd sentence...

manast:
> > > Many users of BullMQ uses an existing Redis Cluster, you should consider to support it.
> > 
> > 
> > I am just suggesting an alternative implementation that is simpler...
> 
> And I replyed to your 2nd sentence...

my second sentence means that if no keys where defined in the lua script it would be cluster compatible by default...

gregorg:
> > > > Many users of BullMQ uses an existing Redis Cluster, you should consider to support it.
> > > 
> > > 
> > > I am just suggesting an alternative implementation that is simpler...
> > 
> > 
> > And I replyed to your 2nd sentence...
> 
> my second sentence means that if no keys where defined in the lua script it would be cluster compatible by default...

True. I am agree, your suggestion should be cleaner.

arsonik:
better like that !

even though I have no idea if there could be any side effect ?

src/commands/addJob-10.lua
```diff
@@ -23,8 +23,6 @@
       KEYS[6] 'priority'
       KEYS[7] events stream key
       KEYS[8] delay stream key
-      KEYS[9] waitChildrenKey key.
```
roggervalf, line 26: hi @arsonik, when keys number should match with the lua script name, in this case should be changed to addJob-8.lua, also you can try to run test locally to check this change 👀

arsonik: Ok 
Well that was my first time editing an Lua script ;)
```    1) "after each" hook


  176 passing (46s)
  5 pending
  1 failing

  1) workers
       "after each" hook:
     Uncaught Error: Connection is closed.
```

PR #511, opened 2021-04-28
Feat pass parent opts to addBulk

roggervalf:


src/test/test_bulk.ts
```diff
@@ -51,6 +51,50 @@ describe('bulk jobs', () => {
...
+    const childrenWorker = new Worker(queueName);      
+    await parentWorker.waitUntilReady();
+    await childrenWorker.waitUntilReady();
+
+    const parent = await parentQueue.add('parent', { some: 'data' });
+    const jobs = await queue.addBulk([
+      { name, data: { idx: 0, foo: 'bar' }, opts: {
+          parent: {
+            id: parent.id,
+            queue: 'bull:' + parentQueueName
```
manast, line 69: we can use string interpolation here:
```ts
`bull:${parentQueueName`
```

roggervalf: let me address it really quick

Issue #504, opened 2021-04-27
Flow task removed doesn't remove data relative


koiszzz:
```javascript

const worker1 = new Worker('test1', async job => {
       return 1;
});
const worker2 = new Worker('test2', async job => {
       return 2;
});
const flowPro = new FlowProducer();

(async() => {
    await flowPro.add({
            name: 'parent-task', data: {}, queueName: 'test1', opts: {removeOnComplete: true},
            children: [{name: 'children-task', data: {}, queueName: 'test2', opts: {removeOnComplete: true}}]
    });
})();
```

After running these codes, the redis still remain some data that named <pid:processed> and  <pid:event>.

manast:
not sure if it was a typo in your exampel code but you wrote: ```removeIOnComplete``` instead of ```removeOnComplete```. Let me know if still does not work and I will take a look into it.

koiszzz:
Please ignore that. I have spelled wrongly when I post this issue. In my code , it was spelled correctly. The parent job and child job were removed after running this code, but there still remains a hash type data named parent-job-id:process and a stream type data named parent-job-id:events.

roggervalf:
@manast one doubt, the parent events stream should also be removed?

manast:
the events are for the queue not for the job so they should not be removed.

manast:
but why are there event streams for "jobs", I think there is a bug that is making the events end in the wrong key actually.

PR #513, opened 2021-04-30
Fix delete processed hash

roggervalf:
This PR is trying to handle part of #504


src/classes/job.ts
```diff
@@ -559,13 +559,27 @@ export class Job<T = any, R = any, N extends string = string> {
...
+  }
+
   private async isInList(list: string) {
     return Scripts.isJobInList(this.queue, this.queue.toKey(list), this.id);
   }
 
   addJob(client: Redis, parentOpts?: ParentOpts): string {
     const queue = this.queue;
 
+    this.assignParentOpts(parentOpts);
```
roggervalf, line 582: this change is because we could use first new Job without passing parent opts and then job.addJob, here we pass the parent opts, as we are using job.opts.parent keys to moveToFinishedArgs, we need to set those missing options before we create the record in Redis

src/commands/moveToFinished-7.lua
```diff
@@ -80,7 +80,8 @@ if rcall("EXISTS",jobIdKey) == 1 then -- // Make sure job exists
         if result == 1 then 
             local processedSet = parentKey .. ":processed"
             rcall("HSET", processedSet, jobIdKey, ARGV[4])
-            if rcall("SCARD", dependenciesSet) == 0 then 
+            local activeParent = redis.call("ZSCORE", parentQueue .. ":waiting-children", parentId)
```
roggervalf, line 83: I found that if we don't have this check, we are going to have a possible scenario where the parent is active and then we add it to wait state, so when the parent is completed, the next job will be the same parent job, so we should check here that is only in waiting-children to move it to wait

manast: Not sure I understand the whole edge case, but this is the kind of scenario I would like to avoid at all cost, i.e. that a parent that has children is not in the expected waiting-children zset, because if it is not there then some other logic is probably not sound. This check maybe is only the aspirine to a bigger pain.

roggervalf: I think that this fix also will help for those manual jobs that are creating children and the job is still active.

manast: What I mean is that we should make sure this situation does not happen instead of being defensive about it. There are several reasons why, first because it makes the code more complex and difficult to understand, second because this implies there are some "unknown" situation whose effects we try to mitigate with defensive code. This often results in very difficult to maintain code without breaking stuff. I would like to better understand why this situation can happen and understand if we need to do design changes in order for it to not happen. We can discuss on slack if it is easier.

src/classes/job.ts
roggervalf: actually I think that user should aware that is better to use Job.create to avoid this issue, since the solution that I provided seems a little bit tricky

PR #514, opened 2021-04-30
Redis Cluster Support

arsonik:
Adding types, handling jobs so the lib can handle Redis Clusters servers


arsonik:
No idea why it's failing on th CI !

```
/V/W/bullmq ❯❯❯ nvm use 14                                                                                                                                                                                                                                                                                       ✘ 127 
Now using node v14.16.1 (npm v6.14.12)
/V/W/bullmq ❯❯❯ npm run test:cluster

> bullmq@1.22.2 test:cluster /Volumes/Work/bullmq
> ts-mocha --paths src/test/test_cluster.ts --exit



  Cluster
    .create
      ✓ saves the job in redis
      ✓ should use the custom jobId if one is provided
    JSON.stringify
      ✓ retains property types
      ✓ omits the queue property to avoid a circular json error on node 8
      ✓ should correctly handle zero passed as data (39ms)
    .update
      ✓ should allow updating job data
    .remove
      ✓ removes the job from redis
    .progress
      ✓ can set and get progress as number
      ✓ can set and get progress as object
    .log
      ✓ can log two rows with text
    .moveToCompleted
      ✓ marks the job as completed and returns new job
      ✓ should not fail if queue meta key is missing
      ✓ should not complete a parent job before its children (46ms)
    .moveToFailed
      ✓ marks the job as failed (54ms)
      ✓ moves the job to wait for retry if attempts are given
      ✓ marks the job as failed when attempts made equal to attempts given
      ✓ moves the job to delayed for retry if attempts are given and backoff is non zero
      ✓ applies stacktrace limit on failure
      ✓ saves error stacktrace
    .promote
      ✓ can promote a delayed job to be executed immediately
      ✓ should process a promoted job according to its priority (471ms)
      ✓ should not promote a job that is not delayed
      ✓ should promote delayed job to the right queue if queue is paused
    .getState
      when redisVersion is less than 6.0.6
        ✓ should get job actual state (40ms)
      when redisVersion is greater or equal than 6.0.6
        ✓ should get job actual state (42ms)
    .finished
      ✓ should resolve when the job has been completed
      ✓ should resolve when the job has been completed and return object
      ✓ should resolve when the job has been delayed and completed and return object (617ms)
      ✓ should resolve when the job has been completed and return string
      ✓ should reject when the job has been failed (531ms)
      ✓ should resolve directly if already processed (511ms)
      ✓ should reject directly if already processed (510ms)


  32 passing (4s)
```

manast:
Two tests are failing:
https://github.com/taskforcesh/bullmq/pull/514/checks?check_run_id=2474576109#step:7:361

src/commands/index.ts
```diff
@@ -12,7 +12,7 @@
  */
 'use strict';
 
-import { Redis } from 'ioredis';
+import { RedisClient } from '@src/classes';
```
manast, line 15: we should use just regular relative paths instead

src/test/test_job.ts
```diff
@@ -304,7 +304,7 @@ describe('Job', function() {
 
       await job.moveToFailed(new Error('test error'), '0', true);
 
-      await waiting;
+      // await waiting;
```
manast, line 307: why this change?

src/test/test_cluster.ts
```diff
@@ -0,0 +1,721 @@
+import { Queue, Job, getParentKey } from '@src/classes';
```
manast, line 1: I don't think is a good idea to copy paste other tests into this test_cluster, instead we should run the existing tests in a cluster setup.

manast (changes requested):
Thanks. It would be easier to merge if the RedisClient changes are isolated in its own PR since AFAICS it is unrelated to redis cluster, correct me if I am mistaken.

src/test/test_job.ts
arsonik: to see if it was passing the tests without

arsonik:
> Two tests are failing:
> https://github.com/taskforcesh/bullmq/pull/514/checks?check_run_id=2474576109#step:7:361

I know.
But I have no idea how to reproduce/debug that on GitHub end.

I did spent a lot of time though, figuring it out locally.
Because I was reusing the same redis cluster connection variable for the QueueEvents and Queue & Worker.
Wich lead into a blocking state (redis side) while waiting (XREAD BLOCK).

And it definitely look like what's happening during the CI !
 locally all the tests are ✅

```
  233 passing (1m)
  6 pending

✨  Done in 71.40s.
```

src/test/test_cluster.ts
arsonik: Yes definitely !
Any idea on how to proceed ?

src/test/test_job.ts
arsonik: wrong file actually

src/test/test_cluster.ts
manast: I think with github actions you can run the tests once with standard redis and then a second time with the cluster.

arsonik: well I guess we can set an env variable or something.
but that mean patching all the calls to these methods `new Queue|Worker|QueueEvents`
to provide a fresh newly made redis instance  every time !
unless ....
with a global bullmq/cluster config we patch `redis-connection.ts` to handle that env variable.
```if (!isRedisInstance(opts)) {
      this.opts = {
        port: 6379,
        host: '127.0.0.1',
        retryStrategy: function(times: number) {
          return Math.min(Math.exp(times), 20000);
        },
        ...opts,
      };
    } else {
      this._client = <RedisClient>opts;
    }
```

manast: what I mean is that you can create another job in node.yml with a cluster redis and run the tests again.

roggervalf:
hi @arsonik, this repo is using semantic-release, so when we make a commit we should follow these rules  https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format

manast:
> hi @arsonik, this repo is using semantic-release, so when we make a commit we should follow these rules https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format

If the PR gets fixed by the author so that it passes the tests I can squash the commits and convert to a conventional commit standard manually. :).

arsonik:
Ok I've striped the docker's cluster test image for now.

I'll probably do another PR with a default configuration setter
for the connection configuration & prefix.
Along with tests for redis cluster.

But for now it is working with cluster out of the box ! ✨

arsonik:
@manast I was able to put the PR in Draft . But I can't set it back to "ready"

arsonik:
Hello ! Any ETA on a possible merge ?

src/classes/flow-producer.ts
```diff
@@ -31,7 +31,7 @@ export class FlowProducer extends EventEmitter {
     super();
 
     this.opts = {
-      prefix: 'bull',
+      prefix: '{bull}',
```
manast, line 34: This would imply a breaking change so we do not really want to do this.

src/classes/compat.ts
```diff
@@ -537,7 +537,7 @@ export class Queue3<T = any> extends EventEmitter {
   }
 
   /**
-   * Returns Queue name with keyPrefix (default: 'bull')
+   * Returns Queue name with keyPrefix (default: '{bull}')
```
manast, line 540: should default to "bull" or it will be a breaking change.

src/classes/queue-base.ts
```diff
@@ -18,7 +17,7 @@ export class QueueBase extends EventEmitter {
     super();
 
     this.opts = {
-      prefix: 'bull',
+      prefix: '{bull}',
```
manast, line 20: better keep just bull as before.

src/classes/queue-keys.ts
```diff
@@ -1,7 +1,7 @@
 export type KeysMap = { [index in string]: string };
 
 export class QueueKeys {
-  constructor(public readonly prefix: string = 'bull') {}
+  constructor(public readonly prefix: string = '{bull}') {}
```
manast, line 4: keep bull as before.

manast (changes requested):
The problem with this PR is that it has a breaking change, not sure I want to bump to a major version just to change the default prefix. What other change is required in this PR for cluster to work according to you?

arsonik:
> The problem with this PR is that it has a breaking change, not sure I want to bump to a major version just to change the default prefix. What other change is required in this PR for cluster to work according to you?

As long as we don't have any tests against a redis cluster I guess we can remove the "cluster" prefix.
Will do it right away.

manast:
thanks!

manast (approved):

Issue #516, opened 2021-04-30
Feature request: non-blocking version of getNextJob


Janpot:
Looking for ways to compose workers in a more elaborate way, e.g. to add priority, or throttling, or round-robin balancing between workers, e.g.:
```ts
const prio1 = new Worker('my-queue-1')
const prio2 = new Worker('my-queue-2')
const prio3a = new Worker('my-queue-3a')
const prio3b = new Worker('my-queue-3b')

async function getNextJob(token) {
  // highest priority first
  let job = await prio1.getNextJob(token)
  if (job) return job
  
  // second highest priority, but maximum 1 per second
  const lockAcquired = await redis.set('prio2-throttle', 'lock', 'NX', 'PX', 1000);
  if (lockAcquired) {
    job = await prio2.getNextJob(token)
    if (job) return job
  }
  
  // toggle between two queues
  const index = await redis.incr('prio3-index');
  const queue = [prio3a, prio3b][index % 2]
  job = await queue.getNextJob(token)
  if (job) return job
  
  // we're empty, just wait for the next poll
  await sleep(1000)
}

async function run (token) {
  while (true) {
    const job = await getNextJob(token)
    try {
      const result = await process(job)
      await job.moveToCompleted(result, token);
    } catch (err) {
      await job.moveToFailed(err, token);
    }
  }
}
```
When trying this out, the blocking behavior of `worker.getNextJob` makes this use-case impossible. It would be helpful if there was a non-blocking version of `getNextJob` that just returns immediately. Perhaps when setting `drainDelay` to `0`?

manast:
drainDelay 0 means block for ever as per Redis semantics. It would be quite easy though to add an opts parameter with "block: boolean".

LBC000:
```js
// bullMQUtils.ts
import { delay, Job, Queue, Worker } from "bullmq";

/**
    // 调用手动队列处理函数
    const token = nanoid();
    const { queue: manuallyQueue, worker } = processManualQueue(
      "manuallyQueue",
      {
        connection: connection,
        concurrency: 5,
        delayTime: 3000,
        successCallback: async (data) => {
          if (data.video == 3) return false;
          return true;
        },
      },
      token
    ); 
 */
export function processManualQueue(queueName, options, token) {
  const { connection, concurrency, delayTime, successCallback } = options;
  const queue = new Queue(queueName, { connection });
  const worker = new Worker(queueName, null, {
    connection,
    concurrency: concurrency || 5,
  });

[...]

PR #517, opened 2021-04-30
feat: create option for non-blocking getNextJob

Janpot:
Closes https://github.com/taskforcesh/bullmq/issues/516

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -186,9 +186,9 @@ export class Worker<
       return;
     }
 
-    if (this.drained) {
+    if (this.drained && block) {
       try {
-        const jobId = await this.waitForJob(block);
+        const jobId = await this.waitForJob();
 
         if (jobId) {
           return this.moveToActive(token, jobId);
@@ -214,7 +214,7 @@ export class Worker<
     return this.nextJobFromJobData(jobData, id);
   }
 
-  private async waitForJob(block = true) {
+  private async waitForJob() {
     const client = await this.blockingConnection.client;
 
     let jobId;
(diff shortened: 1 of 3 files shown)


src/classes/worker.ts
```diff
@@ -189,7 +188,7 @@ export class Worker<
 
     if (this.drained) {
       try {
-        const jobId = await this.waitForJob();
+        const jobId = await this.waitForJob(block);
```
manast, line 191: This will break the semantics since waitForJob will not wait if block == false, instead what we should do is to directly call ```this.moveToActive(token)```,  i.e., same code path as when the queue is drained.

manast: In other words change the if sentence above to: 
```ts
if(this.drained && block){
```

Janpot: Ah, that's simpler, indeed 👍

src/test/test_worker.ts
```diff
@@ -1707,4 +1707,64 @@ describe('workers', function() {
...
+
+  describe('non-blocking', async () => {
+    it('should block by default', async () => {
+      const worker = new Worker(queueName);
+      const token = 'my-token';
+
+      // make sure worker is in drained state
+      await worker.getNextJob(token);
+
+      let [
```
manast, line 1719: use const

src/test/test_worker.ts
```diff
@@ -1707,4 +1707,64 @@ describe('workers', function() {
...
+    });
+
+    it('shouldn\'t block when disabled', async () => {
+      const worker = new Worker(queueName);
+      const token = 'my-token';
+
+      // make sure worker is in drained state
+      await worker.getNextJob(token, { block: false });
+
+      let [
```
manast, line 1737: use const

PR #518, opened 2021-05-03
Replace tslint with eslint

roggervalf:
tslint is deprecated as in https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration, this PR is trying to migrate to eslint and also set pretty-quick, so the code could be formatted before committing


manast:
Is it possible to configure eslint so that it is compatible with current formatting? that would help a lot in order to replace tslint smoothly.

roggervalf:
> Is it possible to configure eslint so that it is compatible with current formatting? that would help a lot in order to replace tslint smoothly.

sure, let me see

src/test/fixtures/delay.js
```diff
@@ -1,5 +1,7 @@
+'use strict';
+
 module.exports = function delay(ms) {
-  return new Promise(function(resolve) {
+  return new Promise(function (resolve) {
```
manast, line 4: there is a new space now, is this because of eslint or why this change?

roggervalf: yep, there was a missing rule in eslint

manast:
excelent job!

PR #520, opened 2021-05-03
fix: add return type for job.waitUntilFinished()

NicolasDuran:
Hi

Method `job.waitUntilFinished()` returns the job's returnvalue when resolving but its return type is not specified (resulting in `Promise<any>`)

This PR fixes this typing issue.


manast:
great, thank!.

PR #522, opened 2021-05-03
Fix queueevents active type

### LINKED ISSUE #519: Add types for active event to QueueEvents
Add missing type active to https://github.com/taskforcesh/bullmq/blob/3283af26c1fead3c3395bf2fc82790cb0772380f/src/classes/queue-events.ts#L7

roggervalf:
Trying to fix #519


manast:
looks good, thanks!

PR #527, opened 2021-05-06
Fix RedisClient type

roggervalf:
This PR tries to fix the release failure generated by this pr #514, what I did is change how RedisClient was constructed, previously it was transpiled with an import declaration, now it is build without it, also I run yarn docs locally and the error is gone


manast:
Fantastic, thanks!.

PR #535, opened 2021-05-08
docs: fix typo in flows.md

onwd:
Fixed a typo in the code snippet inside `flows.md` file.

Also, the same typo exists in the `parent-child-jobs.md`, but I didn't include it into this PR, as this file seems to be unused. Maybe we could simply remove that file.


manast:
thanks!.

PR #536, opened 2021-05-08
Feat add immediately opt for repeat

roggervalf:
This PR allows us to create a repeated job that will be executed immediately


manast:
There is a small consideration here, and is that when using immediately, the next iteration would occur most certainly too soon. So in order to make it work perfectly we would need to store some offset that we need to apply in every iteration. This offset would be the difference between the current timestamp and the timestamp where we actually would have executed the job if we did not use "immediately".

roggervalf:
> There is a small consideration here, and is that when using immediately, the next iteration would occur most certainly too soon. So in order to make it work perfectly we would need to store some offset that we need to apply in every iteration. This offset would be the difference between the current timestamp and the timestamp where we actually would have executed the job if we did not use "immediately".

Let me think a little bit how to address this

manast:
excellent!.

PR #538, opened 2021-05-10
Fix getJobLogs no reversed pagination

roggervalf:
Fixes OptimalBits/bull#2031


manast:
Thanks for the PR. I guess it comes from this issue https://github.com/OptimalBits/bull/issues/2031
As I wrote, I think the best would be to introduce an "asc" parameter as we have in the other getters where default would be "false" so that we keep backwards compatibility since there are cases were we want to paginate from newest logs to oldest and others when we want to provide streaming.

roggervalf:
> Thanks for the PR. I guess it comes from this issue [OptimalBits/bull#2031](https://github.com/OptimalBits/bull/issues/2031)
> As I wrote, I think the best would be to introduce an "asc" parameter as we have in the other getters where default would be "false" so that we keep backwards compatibility since there are cases were we want to paginate from newest logs to oldest and others when we want to provide streaming.

Sure, in last version I refactored it. Thanks

src/classes/queue-getters.ts
```diff
@@ -172,12 +172,16 @@ export class QueueGetters extends QueueBase {
...
   }
 
-  async getJobLogs(jobId: string, start = 0, end = -1) {
+  async getJobLogs(jobId: string, start = 0, end = -1, asc = false) {
     const client = await this.client;
     const multi = client.multi();
 
     const logsKey = this.toKey(jobId + ':logs');
-    multi.lrange(logsKey, -(end + 1), -(start + 1));
+    if (asc) {
```
tjhiggins, line 180: Shouldn't this be flipped for asc? 
```ts
job.log('1')
job.log('2')
job.log('3')

getJobLogs(job.id, 0, -1, asc=true) === ['1', '2', '3']
getJobLogs(job.id, 0, -1, asc=false) === ['3', '2', '1']
```

src/classes/queue-getters.ts
```diff
@@ -172,12 +172,16 @@ export class QueueGetters extends QueueBase {
...
+      multi.lrange(logsKey, -(end + 1), -(start + 1));
```
tjhiggins, line 181: I'm pretty sure this logic is incorrect for the opposite dir `multi.lrange(logsKey, -(end + 1), -(start + 1));`

```ts
job.log('1')
job.log('2')
job.log('3')

// actual
getJobLogs(job.id, 0, 1, asc=false) === ['2', '3']

// expected 
getJobLogs(job.id, 0, 1, asc=false) === ['3', '2']
```

Def add some test cases to confirm either way

tjhiggins (changes requested):

src/classes/queue-getters.ts
tjhiggins: Actually hold off on updating - asking Manuel for clarification on expected behavior
https://github.com/OptimalBits/bull/issues/2031#issuecomment-841917858

roggervalf: hi @tjhiggins, yeah, actually, I based on the actual asc and desc implementation for lrange function call from https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-getters.ts#L130-L136, I guess this logic is here because lrange does not have a asc param, so just the indexes were changed, but still we should wait for @manast

manast:
I am re-studying the code since this was written long time ago and I did not remember it anymore. So basically the "asc" parameter is being used  in ```getRanges``` because depending on how the elements are inserted in the list, we need to lrange them differently to get them in the order we want, in other words, ```getRanges``` do not really support ascending or descending order its just telling it how to return the array items. 
For logs we put them in the list using "RPUSH", so they enter from the right side of the list. 
I also think the previous implementation of ```getLogs``` was wrong since it was not reversing the results for the descending case. So basically I think that what needs to be done is:
1) use ```multi.lrange(key, start, end);``` for the ASCENDING case, i.e. older logs first. (We consider it ascending since the log TIMESTAMP is increasing with time).
2) use ```multi.lrange(logsKey, -(end + 1), -(start + 1))```  and REVERSE the array before returning it for the DESCENDING case. i.e. newest logs first.
3) use ASCENDING as default (since the previous implementation was broken, this is actually the one that better matches the previous behaviour.

@roggervalf @tjhiggins let me know if this matches your expectations.

tjhiggins:
Agreed on all 3 points. Thanks for taking the time to dig in. I'll update the bull pr to reflect.

Issue #539, opened 2021-05-10
Docs: wrong type used in sandboxed processors page


TheoHertz:
in [this guide](https://docs.bullmq.io/guide/workers/sandboxed-processors), the example given uses the type `Job` for a sandboxed processors but it should be `SandboxedJob` instead

javiercr:
I'm a bit confused, the guide references a blog post that uses `Job` not `SandboxesJob`. Is that article outdated?

> If you are looking for a tutorial with code examples on how to use sandboxed processors using typescript you can find one [here](https://blog.taskforce.sh/using-typescript-with-bullmq/).

PR #540, opened 2021-05-10
docs: fix job type in sandbox-processors guide

TheoHertz:
fixes https://github.com/taskforcesh/bullmq/issues/539


manast:
thanks for catching and fixing this!.

PR #541, opened 2021-05-11
Feat add sizeLimit option when creating

roggervalf:
bringing this functionality from https://github.com/OptimalBits/bull/pull/1378

### DIFF
--- a/src/interfaces/jobs-options.ts
+++ b/src/interfaces/jobs-options.ts
@@ -98,5 +98,8 @@ export interface JobsOptions {
    */
   prevMillis?: number;
 
+  /**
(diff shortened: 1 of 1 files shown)


src/interfaces/jobs-options.ts
```diff
@@ -97,4 +97,6 @@ export interface JobsOptions {
    * Internal property used by repeatable jobs.
    */
   prevMillis?: number;
+
```
manast, line 100: Here we could add some small documentation like:
```js
/**
 * Limits the size in bytes of the job's data payload (as a JSON serialized string).
 */
```

manast (changes requested):
great addition!. Just a small comment.

PR #542, opened 2021-05-11
Feat Allow parent opts for the root FlowJob and AddBulk FlowJob

Stormtv:
The docs / changelog was giving me some issues so feel free to edit that or rework my changes here as you see fit.

It didn't make sense to me why we couldn't allow parent opts for the root FlowJob so I modified the code to allow that to happen. The use case for this is adding additional jobs to an already existing parent queue.

Adding bulk FlowJobs is a nice to have when adding multiple Flows at once. I attempted to change the least amount of code possible to make this happen.

Let me know if you want me to split this into two PRs I made the fork for myself but thought I would make the PR and see if this was any interest to the main repo.


src/classes/flow-producer.ts
```diff
@@ -87,6 +87,57 @@ export class FlowProducer extends EventEmitter {
...
+   *
+   * A flow is a tree-like structure of jobs that depend on each other.
+   * Whenever the children of a given parent are completed, the parent
+   * will be processed, being able to access the children's result data.
+   *
+   * All Jobs can be in different queues, either children or parents,
+   * however this call would be atomic, either it fails and no jobs will
+   * be added to the queues, or it succeeds and all jobs will be added.
+   *
+   * @param flows An object with a tree-like structure where children jobs
```
Stormtv, line 101:
```suggestion
   * @param flows An array of objects with a tree-like structure where children jobs
```

manast:
Thanks for the PR, appreciated. !

Yes the changelog can create some problems, the best is to always discard changes on it and try again, the reason is due to a symlink needed so that it appears correctly in gitbook...

Anyway, regarding the PR. addBulk is a fine and a nice addition. Regarding allowing specifying a parent id for the root of the flow. This is in my mind a bit controversial and I would like some solid arguments why it is important. The reason is that I am concerned people will shoot themselves in the foot, since from the moment we allow this we loose atomicity when creating the flow. Also I will need to document very clearly what are the risks of using this approach instead of defining the whole flow in one call...

Finally, the documentation is autogenerated when the PR is merged, so no need to commit it by hand. Before we merge this we will need to remove all those changes, I can do that my self too, no problem.

manast:
On idea I have thought a bit but that I am not 100% happy with is that we can have provide methods in the processor that allows the processor schedule stuff that will be atomically executed as soon as the processor completes. So for example, in my worker processor it would look like this:
```ts
processor(job: Job, postActions: PostActions){
  ...
  postActions.addFlow(children))
}
```
when the processor ends, BullMQ check if there are any postActions to do before completing the job, and if they are it will execute them atomically in Redis as a transaction. In your example, adding children will imply:
1) Move this job to wait-children instead of "complete"
2) add all the children jobs as its children.

However, the problem is that this same parent job will be executed again when the new children completes... but maybe there is some clever way to re-solve it too. We can get fancy by adding a new processor function that we call pre-processor that is executed the first time (adding the children), and when the children are completed executes the actual processor function, etc...
In any case this needs to be well thought so that the final solution is robust, consistent and easy to use.

Stormtv:
I've been a little busy this week so far but I'm still intending on making those edits.

Thanks for looking into and thinking around my use case I'll continue to think about it also and see if I can come up with anything.

manast:
I am thinking about a different approach, that would solve the "long file parsing" use case we have been requested recently, and maybe also the "dynamic graph/tree" use case that you explained above.

Namely to have another state for the parents. Currently we have waiting-children, but we could introduce a new one, something like "waiting-add-children". When all the children jobs have been added we can move the parent to waiting-children and then all the logic we have today applies. Of course when we change the status of the parent we would also need to check if all of its children have already been processed (this would be an edge case when children executes very fast).

For the use case of parsing a very long file, we can have one job that process the file by creating a new flow, and we add the children as the file is parsed. There will not be the same parent used for two purposes, but a whole new parent that is just waiting for adding children. When all children have been added we can now change the status of the parent to "waiting-children" and out current logic kicks in.

If the job that is parsing the file and adding the children fails in the middle of its operation, it should be possible to retry it, however for this to work and not create duplicates, both the flow parent id and the children must get always the same ids based on a given file, so that duplicate jobs will not be re-added, and also we need to make some improvements in ```removeOnCompete``` that should only remove children jobs for parents that have been fully completed.

The tree case is tricker specially if you want the tree root to be processed at the end when all of its children have been processed. What makes the tree case difficult is that the tree includes two operations: traverse and process. If it is possible to first traverse the tree and create the flow and then process the flow it is fine, but traverse and process simultaneously will need more thinking :/.

manast:
@Stormtv the package.json version is incremented automatically by semantic released based on the commit history so it should not be modified by hand :).

Stormtv:
> @Stormtv the package.json version is incremented automatically by semantic released based on the commit history so it should not be modified by hand :).

@manast Sorry was messing with some stuff on my end I'm going to create a clean fork with just the AddBulk options / tests.

Stormtv:
Closing this PR going to rebase and only resubmit the add bulk

Issue #543, opened 2021-05-11
Dynamically change delay of the job


wenq1:
Is it possible to advance the delay of the job while the job is in the delayed state? As far as I can see it involves resetting the delay and change the zscore of the job. Anything else needs to be done?

At the moment in such case I keep track of the the delay using a custom variable and promote the job instead. In the job I compare the current time with the custom and decide wither to proceed with the job or retry with the new delay. Although working, it seems a bit awkward. Is there a builtin way of doing that?

manast:
There is no built in way to do it, but as you mentioned is quite straightforward to implement. It could be a lua script that just updates the delay and zscore and sends an event to the delay stream (so that BullMQ knows when to pick it up), and some small bookkeeping to check that the job actually is delayed (to eliminate a possible race condition where the job has already started processing when we try to change the delay time). You can find how delay jobs are added in ```addJob-8.lua. ```.

roggervalf:
@wenq1 could you please try the last version, it should have the changeDelay method, so this issue could be deleted and #545 as well.

hoangbn:
This feature does not work for me. Opened a new issue for it: #787

PR #545, opened 2021-05-13
feat: add changedelay

wenq1:
#543


wenq1:
@manast  can you please review?

src/commands/changeDelay-2.lua
```diff
@@ -0,0 +1,34 @@
...
+    0 - OK
+   -1 - Missing job.
+
+  Events:
+    - delayed key.
+]]
+local rcall = redis.call
+
+if rcall("EXISTS", KEYS[1]) == 1 then
+
```
roggervalf, line 23: you also can add this validation https://github.com/taskforcesh/bullmq/blob/master/src/commands/getStateV2-7.lua#L31 so we are making sure that the job is in delayed state, if not maybe return -3 like in https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveToDelayed-5.lua#L18 Job not in delayed state

wenq1: OK

wenq1: fixed below. also changed filename

src/commands/changeDelay-4.lua
```diff
@@ -0,0 +1,40 @@
...
+    KEYS[4] delayed stream
+
+    ARGV[1] delayedTimestamp
+    ARGV[2] the id of the job
+    ARGV[3] queue token
+
+  Output:
+    0 - OK
+   -1 - Missing job.
+   -1 - Job not in delayed set.
```
manast, line 17: Maybe this should be -2

src/commands/changeDelay-4.lua
```diff
@@ -0,0 +1,40 @@
...
+local rcall = redis.call
+
+if rcall("EXISTS", KEYS[2]) == 1 then
+
+  local jobId = ARGV[2]
+  local score = tonumber(ARGV[1])
+  local delayedTimestamp = (score / 0x1000)
+
+  if redis.call("ZSCORE", KEYS[1], jobId) ~= false then
+    return -3
```
manast, line 31: return -2 ?

src/classes/job.ts
```diff
@@ -401,6 +401,27 @@ export class Job<T = any, R = any, N extends string = string> {
...
+    const client = await this.queue.client;
+
+    const queue = this.queue;
+
+    let command: string;
+    const multi = client.multi();
+
+    const args = Scripts.changeDelayArgs(queue, this.id, Date.now() + delay);
+
+    (<any>multi).moveToDelayed(args);
```
manast, line 414: Shouldn't we call the changeDelay script instead?

src/classes/job.ts
```diff
@@ -401,6 +401,27 @@ export class Job<T = any, R = any, N extends string = string> {
...
+
+    if (!this.queue.closing) {
+      const results = await multi.exec();
+      const code = results[results.length - 1][1];
+      if (code < 0) {
+        throw Scripts.finishedErrors(code, this.id, command);
```
manast, line 420: Since we have a new code expressing "missing job in delay set", finishedErrors will not be compatible.

src/classes/scripts.ts
```diff
@@ -344,6 +344,35 @@ export class Scripts {
...
+    timestamp: number,
+  ) {
+    //
+    // Bake in the job id first 12 bits into the timestamp
+    // to guarantee correct execution order of delayed jobs
+    // (up to 4096 jobs per given timestamp or 4096 jobs apart per timestamp)
+    //
+    // WARNING: Jobs that are so far apart that they wrap around will cause FIFO to fail
+    //
+    timestamp = typeof timestamp === 'undefined' ? 0 : timestamp;
```
manast, line 360: The timestamp calculation is the same code we have in moveToDelayedArgs, it would be better to refactor that code, maybe a function called "calcDelayTimestamp" and use it in both places. Also note that the code for calculating the timestamp that you are using is old, you are probably basing your PR on an old version of the codebase.

manast (changes requested):
Thanks for the PR. Please see my comments. Also it would be important to have at least one test that demonstrates it works.

manast:
Closing since functionality is already merged.

Issue #160, opened 2020-03-05
whenCurrentJobsFinished reconnect arg


wenq1:
Reconnect = true is never true.

If worker is paused, upon resuming, the connection is gone.

https://github.com/taskforcesh/bullmq/blob/b96f5ba3e9c36182174300af629b34fe43c62ecd/src/classes/worker.ts#L354-L365

Relates to #155  but is not the root cause, so I'm creating a separate issue.

thegedge:
I think we have the same issue, which was causing our CI to have some non-deterministic failures. I have a fix on the way.

thegedge:
Trying to figure out how to write a test to capture this, but here's the fix: https://github.com/taskforcesh/bullmq/compare/master...gadget-inc:fix-160-reconnect-bug

wenq1:
let needsReconnect = false;
 if (this.waiting) {
       await this.blockingConnection.disconnect()); 
       needsReconnect = true;
  }
  
   if (this.processing) { 
     await Promise.all(this.processing.keys()); 
   } 
  
   reconnect && needsReconnect && (await this.blockingConnection.reconnect()); 


Better fix should be like this.

This is exactly the fix we used to fix #155.

However, wrong fix in #391 is commited by @manast, who thinks the fix should be to "avoid the infinite loop" rather than "stop the cause to the infinite loop"

PR #548, opened 2021-05-14
Ensure connection reconnects when pausing.

thegedge:
Fixes https://github.com/taskforcesh/bullmq/issues/160

Currently, if we pause a worker while waiting to pop off the next job, we disconnect. The issue is that `this.waiting` can change to `false` before we try to reconnect, preventing us from reconnecting, even if it was requested (which is it, when pausing). Once we resume, the connection is closed, causing the `run` loop to throw `Client is disconnected` (since we're neither paused nor closing).

We fix this by _only_ checking the `reconnect` argument. To handle the case where we didn't disconnect, we just set reconnect to `false` so that we don't try to reconnect.

**For reviewers**
I believe it's clear that this can happen, and I've observed it in a flaky test suite in my company's CI, but I'm struggling to write a test for it. Maybe you all would have ideas on how I could test this? 🙇

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -218,6 +218,10 @@ export class Worker<
   private async waitForJob() {
     const client = await this.blockingConnection.client;
 
+    if (this.paused) {
+      return;
+    }
+
     let jobId;
     const opts: WorkerOptions = <WorkerOptions>this.opts;
 
@@ -394,13 +398,17 @@ export class Worker<
     //
     // Force reconnection of blocking connection to abort blocking redis call immediately.
     //
-    this.waiting && (await this.blockingConnection.disconnect());
+    if (this.waiting) {
+      await this.blockingConnection.disconnect();
+    } else {
+      reconnect = false;
+    }
 
     if (this.processing) {
       await Promise.all(this.processing.keys());
     }
 
-    this.waiting && reconnect && (await this.blockingConnection.reconnect());
+    reconnect && (await this.blockingConnection.reconnect());
   }
 
   close(force = false) {
--- a/src/test/test_pause.ts
+++ b/src/test/test_pause.ts
@@ -311,4 +311,21 @@ describe('Pause', function() {
     const isResumedQueuePaused = await queue.isPaused();
     expect(isResumedQueuePaused).to.be.false;
   });
+
+  it('should pause and resume worker without error', async function() {
+    const worker = new Worker(queueName, async job => {
+      await delay(100);
+    });
+
+    await worker.waitUntilReady();
+    await delay(10);
+    await worker.pause();
+    await delay(10);
+    worker.resume();
+    await delay(10);
+    await worker.pause();
+    await delay(10);
+
+    return worker.close();
+  });
 });


manast:
LGTM. It is pretty difficult to write a test for this unless you mock blockingConnection.disconnect and blockingConnection.reconnect, so that this.waiting is updated in the mock and thus recreating the edge case.

thegedge:
> LGTM. It is pretty difficult to write a test for this unless you mock `blockingConnection.disconnect` and `blockingConnection.reconnect`, so that `this.waiting` is updated in the mock and thus recreating the edge case.

I haven't had an opportunity to try this out. My thought is that mocking `blockingConnection` isn't going to make for a great test given it's a private implementation detail. I definitely found it difficult to write a test here to orchestrate a realistic sequence of events. As much as I'd like to write a regression test, it feels to me like it would be low ROI here.

I'll leave it up to you, but if you're okay without a test for this specific issue, feel free to merge. If not, I'll try to find some time this week to work on it (but I unfortunately don't have a lot of free time 😢).

roggervalf:
hi @thegedge, I tested your changes with this test:
```
it.only('should pause and resume worker without error', async function() {
    const worker = new Worker(queueName, async job => {
      console.log(`Doing some WORK, yo!`);
    });

    const workers = await queue.getWorkers();

    if (worker.isPaused()) {
      worker.resume();
      console.log("Worker resumed");
    } else {
      console.log("Pausing Worker...");
      await worker.pause();
      console.log("Worker paused");
    }

    if (worker.isPaused()) {
      worker.resume();
      console.log("Worker resumed");
    } else {
      console.log("Pausing Worker...");
      await worker.pause();
      console.log("Worker paused");
    }

    expect(workers).to.have.length(1);
    return worker.close();
  });
```
I tried this test in local, without your change it fails, @manast what do you think?

manast:
@roggervalf what error do you get?

roggervalf:
> @roggervalf what error do you get?

now I was playing with it and got:
Selection_025
Even with this change, but when I added:
```ts
    if(!this.paused){
        jobId = await client.brpoplpush(
          this.keys.wait,
          this.keys.active,
          opts.drainDelay,
        );  
      }
```
in https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L226-L230, it passes
Selection_028

manast:
I see, it seems that we need to check for pausing again to avoid an edge case.

thegedge:
@manast would you like me to incorporate that change into this PR, or will we do that separately?

manast:
@thegedge sure, if you have time you are welcome to do it.

thegedge:
Okay, I've incorporated the other check for paused, and also a slight tweak to the test that @roggervalf recommended.

The tweak is to remove the expectation, since that wasn't testing anything we wanted to test, and sprinkle in some delays (less about the delay, more about the `await`). What's fascinating is that this test can reliably reproduce the `Connection is closed` problem that we originally set out to fix. I wish I knew exactly how, but I'm going to call that a win 🙂🎉

PR #551, opened 2021-05-16
Fix flowProducer process parent with children as empty array

### LINKED ISSUE #547: Flows not working if there are no children
Flows don't appear to run the parent job if there are no children.

When I run the code:

```
import { FlowProducer } from 'bullmq';

const flowProducer = new FlowProducer();

const flow = await flowProducer.add({
	queueName: 'myQueueName',
	name: 'myJobName',
	data: { myData: 'my data' },
	children: []
});
```

My parent job doesn't run. However, when I add a child job, it does run (after the child job):

```diff
import { FlowProducer } from 'bullmq';

const flowProducer = new FlowProducer();

const flow = await flowProducer.add({
	queueName: 'myQueueName',
	name: 'myJobName',
	data: { myData: 'my data' },
-	children: []
+	children: [{
+		queueName: 'myQueueName',
+		name: 'myChildJobName',
+		data: { myData: 'my child\'s data' }
+	}]
});
```

I know that there is not much point in using flows if there are no children, but I don't know that until runtime (my `children` is actually a mapped list, which may or may not be empty).

roggervalf:
Fixes #547

### DIFF
--- a/src/test/test_flow.ts
+++ b/src/test/test_flow.ts
@@ -104,6 +104,48 @@ describe('flows', () => {
     await removeAllQueueData(new IORedis(), parentQueueName);
   });
 
+  it('should process parent when children is an empty array', async () => {
+    const parentQueueName = 'parent-queue';
+
+    let parentProcessor;
+
+    const processingParent = new Promise<void>((resolve, reject) => [
+      (parentProcessor = async () => {
+        try {
+          resolve();
+        } catch (err) {
+          console.error(err);
+          reject(err);
+        }
+      }),
+    ]);
+
+    const parentWorker = new Worker(parentQueueName, parentProcessor);
+
+    const flow = new FlowProducer();
+    const tree = await flow.add({
+      name: 'parent-job',
+      queueName: parentQueueName,
+      data: {},
+      children: [],
+    });
+
+    expect(tree).to.have.property('job');
+    expect(tree).to.not.have.property('children');
+
+    const { job } = tree;
+    const parentState = await job.getState();
+
+    expect(parentState).to.be.eql('active');
+
+    await processingParent;
+    await parentWorker.close();
+
+    await flow.close();
+
+    await removeAllQueueData(new IORedis(), parentQueueName);
+  });
+
   it('should process a chain of jobs', async () => {
     const name = 'child-job';
     const values = [
--- a/src/test/test_flow.ts
+++ b/src/test/test_flow.ts
@@ -27,7 +27,7 @@ describe('flows', () => {
       { qux: 'something' },
     ];
 
-    const parentQueueName = 'parent-queue';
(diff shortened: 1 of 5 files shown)


src/test/test_obliterate.ts
```diff
@@ -10,14 +10,14 @@ describe('Obliterate', () => {
   let queueEvents: QueueEvents;
   let queueName: string;
 
-  beforeEach(async () => {
+  beforeEach(async function() {
```
manast, line 13: why replacing the arrow function by a function?

src/test/test_flow.ts
```diff
@@ -104,6 +104,43 @@ describe('flows', () => {
...
   });
 
+  it('should process parent when children is an empty array', async () => {
+    const parentQueueName = `parent-queue-${v4()}`;
+
+    let parentProcessor;
+
+    const processingParent = new Promise<void>((resolve, reject) => [
+      (parentProcessor = async () => {
+        try {
```
manast, line 114: why the try & catch here? calling resolve cannot throw any exception.

src/test/test_flow.ts
```diff
@@ -104,6 +104,43 @@ describe('flows', () => {
     await removeAllQueueData(new IORedis(), parentQueueName);
   });
 
+  it('should process parent when children is an empty array', async () => {
+    const parentQueueName = `parent-queue-${v4()}`;
+
+    let parentProcessor;
+
+    const processingParent = new Promise<void>((resolve, reject) => [
```
manast, line 112: I noticed we are using an array as return value for the promise callback, I see we also use it in other places, do you know why? maybe I used it by mistake myself and now it is spreading around :).

manast (commented):
Thanks for the PR.

src/test/test_obliterate.ts
roggervalf: yeah I was asking me why we are using functions in beforeEach, I was just changed for trying it, in the beginning it used function, so I returned it to that maybe because it was a special reason xd, I can changed again to arrow function.

src/test/test_flow.ts
roggervalf: you are right, I forgot to delete this part of the test

src/test/test_flow.ts
roggervalf: yes, those shouldn't be arrays, let me change it

src/test/test_obliterate.ts
manast: The reason "function" is needed in mocha is when we need to access the "this" object, which does not happen very often, usually when we want to change the "timeout" of a test. However it is possible to specify the timeout on the returning value for example:
```js
it('should test something', () => { ... }).timeout(50000);
```

PR #553, opened 2021-05-16
Fix flowProducer: use custom jobId as parentId for children

### LINKED ISSUE #552: FlowProducer: parent job does not contain dependencies if I set jobId
If I set jobId:
```javascript
const parentQueue = new Queue('parent-queue');
const flowProducer = new FlowProducer();

const flow = await flowProducer.add({
  name: 'p1',
  queueName: 'parent-queue',
  opts: { jobId: 'my-parent-job-id' },
  children: [
    { name: 'c1', queueName: 'child-queue' },
    { name: 'c2', queueName: 'child-queue' },
    { name: 'c3', queueName: 'child-queue' },
  ],
});

console.log('create flow job', flow.job.id);
const parentJob = await parentQueue.getJob(flow.job.id);
// or const parentJob = await parentQueue.getJob('my-parent-job-id');
console.log('deps', await parentJob.getDependencies());
```

Output:
```
create flow job my-parent-job-id         
deps { processed: {}, unprocessed: [] }  
```

Everything is fine without jobId:
```javascript
const flow = await flowProducer.add({
  name: 'p1',
  queueName: 'parent-queue',
  children: [
    { name: 'c1', queueName: 'child-queue' },
    { name: 'c2', queueName: 'child-queue' },
    { name: 'c3', queueName: 'child-queue' },
  ],
});

console.log('create flow job', flow.job.id);
const parentJob = await parentQueue.getJob(flow.job.id);
console.log('deps', await parentJob.getDependencies());
```

Output: 
```
create flow job 389cc496-336f-4748-a978-6cbaf95335a2                               
deps {                                                                             
  processed: {},                                                                   
  unprocessed: [ 'bull:child-queue:1', 'bull:child-queue:3', 'bull:child-queue:2' ]
}
```

I want to set the jobId to the parent in order to avoid re-adding the same parent task if it is already in the queue and track its progress.
The parent task is formed on the basis of 2 parameters that come from the user: `p1` and `p2`. These are short text strings, I planned to use them as ```const jobId = `${p1}:${p2}`;```

Workers of both queues process tasks, but for the given `jobId` of the parent I cannot get neither children nor dependencies.

roggervalf:
Fixes #552

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -110,8 +110,8 @@ export class FlowProducer extends EventEmitter {
     },
   ): JobNode {
     const queue = this.queueFromNode(node, new QueueKeys(node.prefix));
-    const parentId = node.children && uuid.v4();
 
+    const jobId = node.opts?.jobId || uuid.v4();
     const job = new Job(
       queue,
       node.name,
@@ -120,8 +120,9 @@ export class FlowProducer extends EventEmitter {
         ...node.opts,
         parent: parent?.parentOpts,
       },
-      node.opts?.jobId || parentId,
+      jobId,
     );
+    const parentId = node.children && jobId;
 
     const parentKey = getParentKey(parent?.parentOpts);
 
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -122,12 +122,12 @@ export class FlowProducer extends EventEmitter {
       },
       jobId,
     );
-    const parentId = node.children && jobId;
 
     const parentKey = getParentKey(parent?.parentOpts);
 
     if (node.children && node.children.length > 0) {
       // Create parent job, will be a job in status "waiting-children".
+      const parentId = jobId;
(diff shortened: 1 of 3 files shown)


src/classes/flow-producer.ts
```diff
@@ -120,8 +120,9 @@ export class FlowProducer extends EventEmitter {
         ...node.opts,
         parent: parent?.parentOpts,
       },
-      node.opts?.jobId || parentId,
+      jobId,
     );
+    const parentId = node.children && jobId;
```
manast, line 125: seems like in current code parentId is only used if ```node.children``` is truthy, so we can just move this assignment into the ```if (node.children)```  block and keep it like ```const parentId = jobId```

manast (commented):
Looks good. 
I did a small thinking about the case the issue reporter mentioned, i.e. that he wants the jobs with the same id to be ignored. This will work for the parent, but if the children get different job Ids it is going to add the same jobs to an existing parent. I will write a comment in the issue to make it clear to avoid unintended results.

PR #558, opened 2021-05-18
Missing version messages in changelog

roggervalf:
This PR adds those version messages that are missing

### DIFF
--- a/docs/gitbook/CHANGELOG.md
+++ b/docs/gitbook/CHANGELOG.md
@@ -278,68 +278,68 @@
 
 - job finish queue events race condition ([355bca5](https://github.com/taskforcesh/bullmq/commit/355bca5ee128bf4ff37608746f9c6f7cca580eb0))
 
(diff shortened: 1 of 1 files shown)


manast:
great thanks. However is better to no change the old formatting since semantic-release will automatically format the next logs in the same way as the older ones, so it will look inconsistent.

roggervalf:
Yes, I agree with you, but in this case, but after 1.14.1 version in https://github.com/taskforcesh/bullmq/pull/558/files#diff-4bef80b39a2e0ea36f37a8d0cccfd69b7ad556c929755a6afab32eccc957017aR281, semantic-release was removing one # from feat and fix versions, I noticed this from the ui https://docs.bullmq.io/changelog
Selection_029
so since 1.14.1 semantic release was using ### and #### that in our ui is taking those in the same level, to keep consistency with the new version of how semantic release is formatting the new changes I deleted one # following the same rules as we are having for the last versions

PR #559, opened 2021-05-21
Fix event parent stream

### LINKED ISSUE #504: Flow task removed doesn't remove data relative

roggervalf:
Fix #504

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -63,7 +63,7 @@ export class Job<T = any, R = any, N extends string = string> {
   stacktrace: string[] = null;
 
   /**
-   * Timestamp when the job was created (unless overrided with job options).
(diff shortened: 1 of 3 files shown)


manast:
well done!

PR #568, opened 2021-05-29
Fix getWorkers

### LINKED ISSUE #564: getWorkers() includes workers for other queues with the same name prefix
Getting workers for the queue "queue-1" seems to include workers for the queue named "queue-10" - that is a queue who's name starts with the requested queue's name. The items in the returned array point to "queue-1" tough (so no way to filter them out), but their count increases when I add a worker for "queue-10". Using bullmq 1.28.0

roggervalf:
Fixes #564

### DIFF
--- a/docs/gitbook/CHANGELOG.md
+++ b/docs/gitbook/CHANGELOG.md
@@ -40,8 +40,6 @@
 
 * **getjoblogs:** no reversed pagination ([fb0c3a5](https://github.com/taskforcesh/bullmq/commit/fb0c3a50f0d37851a8f35cb4c478259a63d93461))
 
-# Changelog
-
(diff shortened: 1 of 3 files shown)


docs/gitbook/CHANGELOG.md
```diff
@@ -40,8 +40,6 @@
 
 * **getjoblogs:** no reversed pagination ([fb0c3a5](https://github.com/taskforcesh/bullmq/commit/fb0c3a50f0d37851a8f35cb4c478259a63d93461))
 
-# Changelog
```
roggervalf, line 43: this was added when changelog was failing

manast:
interesting, I wonder where that startsWith comes from...

PR #569, opened 2021-05-30
Fix moveStalledJobsToWait: send failedReason to queueEvents

roggervalf:


src/commands/moveStalledJobsToWait-8.lua
```diff
@@ -59,11 +60,13 @@ if (#stalling > 0) then
...
                                            1)
                 if (stalledCount > MAX_STALLED_JOB_COUNT) then
+                    local failedReason = "job stalled more than allowable limit" 
                     rcall("ZADD", KEYS[4], ARGV[3], jobId)
                     rcall("HSET", jobKey, "failedReason",
-                          "job stalled more than allowable limit")
+                          failedReason)
                     rcall("XADD", KEYS[8], "*", "event", "failed", "jobId",
-                          jobId, 'prev', 'active')
+                          jobId, 'prev', 'active', 'failedReason',
```
roggervalf, line 68: queueEvents has failed event and it expects a failedReason, here we are not sending any, so I add same failedReason

src/test/test_stalled_jobs.ts
```diff
@@ -126,10 +129,28 @@ describe('stalled jobs', function() {
     await worker.close(true);
```
manast, line 130: you could also assign the error message to a const here since you are using it in two places :).

manast (commented):
good catch!

src/test/test_stalled_jobs.ts
roggervalf: sure, I missed that part xd

PR #571, opened 2021-05-30
Feat adding healthCheck

roggervalf:
Exploring a feature to verify if a queue is still connected


src/classes/redis-connection.ts
```diff
@@ -138,6 +138,16 @@ export class RedisConnection extends EventEmitter {
     }
   }
 
+  async healthCheck(): Promise<boolean> {
+    try {
+      await this._client.info();
+      return true;
+    } catch (error) {
+      this.emit('error', error);
```
manast, line 146: emit here is a side-effect, we should not create side-effects in a function that is supposed to just return true/false

manast (commented):
Thanks for the PR. I think that it is better to keep a utility function like this outside of the library since it is easy to implement and should not really be part of the scope of BullMQ. It we were to use that same function internally it would a different thing of course.

roggervalf:
> Thanks for the PR. I think that it is better to keep a utility function like this outside of the library since it is easy to implement and should not really be part of the scope of BullMQ. It we were to use that same function internally it would a different thing of course.

Let me see, I guess I can try something

roggervalf:
I tried to use it to fix this test https://github.com/taskforcesh/bullmq/blob/master/src/test/test_connection.ts#L152 but sadly it doesn't work, I'll close this pr for now

PR #574, opened 2021-06-01
Fix QueueScheduler: adding initial event typing

roggervalf:
For now I added only 2 events, at least those that I could check from tests


manast:
these are nice little improvements, thanks!.

manast:
I am going to merge is as a "feat" though, since it could actually break a CI if they have using a wrong signature.

PR #575, opened 2021-06-02
Feat getDependencies with pagination

roggervalf:
One job could have many dependencies that are unprocessed, in order to not get all at once, could be nice to have a wait to get those using sscan, this could be helpful to use it in a ui. Also I added getProcessedDependencies to have pagination on those values using hscan, same purpose. Note: this pr is still in progress


manast:
Functionality wise this looks good. But I think it would be better to put these two methods inside current ```getDependencies``` in order to keep the API surface more compact.
Maybe you can add more params to getDependencies so that you can control if you want to read only processed, unprocessed and optional pagination?

roggervalf:
> Maybe you can add more params to getDependencies so that you can control if you want to read only processed, unprocessed and optional pagination?

Let me think how to address this, also we have a separate method to only return the processed values (getChildrenValues), so this method should also have same feature

roggervalf:
> also we have a separate method to only return the processed values (getChildrenValues), so this method should also have same feature

Nevermind, to add this feature using hscan I would need to return the cursor, this should be a breaking change in this method :/

manast:
I think it is possible to do it without a breaking change. If no opts are provided return all data as in current implementation, if opts is provided then use the new cursor based code.

roggervalf:
> I think it is possible to do it without a breaking change. If no opts are provided return all data as in current implementation, if opts is provided then use the new cursor based code.

Sure no problem, I updated it base on your suggestion

manast:
Nice work!

PR #576, opened 2021-06-02
Fix getDependencies to parse processed values

roggervalf:
I'm curious if this was intended or not, since the values are not parsed and from getChildrenValues are parsed, so this pr tries also to parse processed values from getDependencies method


src/test/test_flow.ts
```diff
@@ -285,10 +285,8 @@ describe('flows', () => {
           const { processed } = await job.getDependencies();
           expect(Object.keys(processed)).to.have.length(1);
 
-          const childrenValues = await job.getChildrenValues();
```
manast, line 288: why remove this? still good to check if getChildrenValues() returns the correct data.

manast (commented):
Thanks, this was overlooked from my part.

src/test/test_flow.ts
roggervalf: sure I can restore it, I was removing it since there are other test cases that also check same children values, but I can restore it

PR #581, opened 2021-06-05
Feat job: adding getDependenciesCount method

roggervalf:
Would be nice to have a method to get processed and unprocessed counts without bringing all the dependencies values


src/test/test_worker.ts
```diff
@@ -8,7 +8,7 @@ import { after, times, once } from 'lodash';
 import { RetryErrors } from '../enums';
 import * as sinon from 'sinon';
 
-describe.only('workers', function() {
```
roggervalf, line 11: sorry, my bad

src/classes/job.ts
```diff
@@ -569,8 +571,8 @@ export class Job<T = any, R = any, N extends string = string> {
     const client = await this.queue.client;
     const multi = client.multi();
     if (!opts.processed && !opts.unprocessed) {
-      await multi.hgetall(this.toKey(`${this.id}:processed`));
```
roggervalf, line 572: these awaits are not necessary

ccollie:
Hey @roggervalf, while you're at it, could you export `DependenciesOpts` ?

https://github.com/taskforcesh/bullmq/blob/87b360323ed3ab4d43ccd3e278413137479cdf3b/src/classes/job.ts#L57

roggervalf:
> Hey @roggervalf, while you're at it, could you export `DependenciesOpts` ?
> 
> https://github.com/taskforcesh/bullmq/blob/87b360323ed3ab4d43ccd3e278413137479cdf3b/src/classes/job.ts#L57

@ccollie, sure no problem

roggervalf:
@manast please when you get a change, some :eyes:

manast:
LGTM. Is this ready for merge?

roggervalf:
@manast yes it is ready

manast:
Great. One little comment though, I noticed that you sometimes use "gerund" verb form in the commits. According to  the semantical commits standard it should be the imperative form of the verb. So instead of "adding" just "add". Keep up with the good work!

PR #582, opened 2021-06-05
Feat getFlow method

roggervalf:
This method tries to help in getting and entire flow from a parent reference. Also as we can have a lot of children for a job I think is a good idea to limit this method by a max children and depth level, so we don't too much information

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -23,9 +23,12 @@ export interface JobNode {
 }
 
 /**
- * This class allows to add jobs into one or several queues
- * with dependencies between them in such a way that it is possible
- * to build complex flows.
+ * This class allows to add jobs with dependencies between them in such
+ * a way that it is possible to build complex flows.
+ * Note: A flow is a tree-like structure of jobs that depend on each other.
+ * Whenever the children of a given parent are completed, the parent
+ * will be processed, being able to access the children's result data.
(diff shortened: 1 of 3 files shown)


src/classes/flow-producer.ts
```diff
@@ -87,6 +96,36 @@ export class FlowProducer extends EventEmitter {
     return jobsTree;
   }
 
+  /**
+   * @method getFlow
+   * Get a flow.
+   *
+   * A flow is a tree-like structure of jobs that depend on each other.
```
manast, line 103: maybe you could refactor this comment to the Flow class, since it becomes redundant to have it both for add and get.

src/classes/flow-producer.ts
```diff
@@ -182,6 +255,31 @@ export class FlowProducer extends EventEmitter {
     return nodes.map(node => this.addNode(multi, node, parent));
   }
 
+  private getChildren(
+    client: RedisClient,
+    processed: string[],
```
manast, line 260: There is no reason for this method to know about "processed" and "unprocessed", so you can just have one input array with job keys instead. It should be the caller of this method the one that needs to concat the arrays if they come from more than one array.

manast (commented):
It looks great even without pagination support. Thanks!.

src/classes/flow-producer.ts
roggervalf: Sure I can move it

PR #583, opened 2021-06-05
Fix completed: consider removing processed key when removeOnComplete

roggervalf:


manast:
Another great PR, thanks a lot!

PR #589, opened 2021-06-09
Fix flaky test related to getDependencies

roggervalf:
There is one comment to be aware when using sscan and hscan, here is explained in the documentation https://redis.io/commands/scan#why-scan-may-return-all-the-items-of-an-aggregate-data-type-in-a-single-call

### DIFF
--- a/src/test/test_worker.ts
+++ b/src/test/test_worker.ts
@@ -1902,30 +1902,33 @@ describe('workers', function() {
 
       expect(currentState).to.be.equal('active');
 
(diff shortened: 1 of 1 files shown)


manast:
yeah, this test seems to be a bit undeterministic, would be great to ensure they are deterministic if it is possible.

roggervalf:
> yeah, this test seems to be a bit undeterministic, would be great to ensure they are deterministic if it is possible.

I tried to refactor it, at least now it looks green
Selection_036

Issue #167, opened 2020-03-29
Possible memory leak in workers.


Crazy-kun:
Steps to reproduce:

1. Create worker.
`    const worker = new Worker("Test", resolve(__dirname, "./Test.js"), {
        connection: redisConfig
    });`
2. Create jobs.
`const queue = new Queue("Test", { connection: redisConfig });
 const promises = [];
        for (let i = 0; i < 10000; i++) {
            promises.push(
                queue.add(
                    "Test",
                    {
                        test: "test"
                    }
                )
            );
        }
        await Promise.all(promises);
        await queue.close();`

Test.js

`import { Job } from "bullmq";
import { redisConfig } from "../config";

export default (job: Job, done: any) => {
        let { test } = job.data;
        console.log(test);
        done();
};`

Application used memroy before worker starts processing jobs: ~100MB.
Application used memory during processing: ~600MB.

3. Removing worker after all jobs done.
`await worker.close();
`
Application used memory after worker removing: ~600MB.

Memory, used inside workers, do not disposes.

wenq1:
Memory not being freed not necessarily mean there's a leak. Could be that GC has not yet kicked in.

manast:
as @wenq1 is pointing out, you cannot prove there is a memory leak by a code like that, you need to make sure GC has run before you compare the memory. 
https://marmelab.com/blog/2018/04/03/how-to-track-and-fix-memory-leak-with-nodejs.html

Crazy-kun:
Well, if i repeat this worker several times, memory doesn't allocates over ~600MB. Ok, maybe it's really GC feature. Thanks for explanation!

PR #590, opened 2021-06-10
fix(redis-connection): remove event listeners from shared client

Derbdale:
Fixed an issue where event listeners on shared redis connections meant that Workers which had been "closed" were holding onto references to redis-connection in memory.

I believe this will also fix the issue described in #167

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -14,6 +14,7 @@ export class RedisConnection extends EventEmitter {
   private initializing: Promise<RedisClient>;
   private closing: boolean;
   private version: string;
+  private handleClientError: (e: Error) => void;
 
   constructor(private readonly opts?: ConnectionOptions) {
     super();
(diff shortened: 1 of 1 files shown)


manast:
Good catch, but I think the code is wrong. The "this" pointer will not be the same when the error callback is called by the redis client. If you write a test you will notice that it raises an exception. If you define the error handler as an arrow function in a class property it will be fine.

manast:
Btw, maybe it will not throw an exception since "redis client" also is a subclass of the EventEmitter class, but it will be emitting the error on the wrong instance.

Derbdale:
What is your preferred solution:
* Bind the handler to "this" in the constructor (this.handleClientError = this.handleClientError.bind(this);)
* Replace the function with an arrow function in the class.

manast:
@Derbdale I think it is more elegant to define a private class property and initialize it with an arrow function in the constructor.

Derbdale:
@manast That should all be sorted now.

src/classes/redis-connection.ts
```diff
@@ -34,10 +34,14 @@ export class RedisConnection extends EventEmitter {
     this.initializing = this.init();
 
     this.initializing
-      .then(client => client.on('error', err => this.emit('error', err)))
+      .then(client => client.on('error', this.handleClientError))
       .catch(err => this.emit('error', err));
   }
 
+  handleClientError = (err: Error): void => {
```
manast, line 41: I am not sure this works still, I think you need to initialize handleClientError in the constructor:
```ts
this.handleClientError = (err: Error): void => {
   this.emit('error', err);
};
```

Derbdale:
Ah, I misunderstood before!

Should be correct now :)

PR #592, opened 2021-06-11
Feat expose queueName in job

roggervalf:
As we have getFlow method, we can get a tree for an specific parent node, but in order to know from which queue the children and grand children belong, it would be nice to expose the name of the queue from the job instance.


roggervalf:
Please @manast when you get a chance ti get some 👀 on this pr

manast:
Looks good!

Issue #523, opened 2021-05-04
missing types for Worker event listeners


m-ronchi:
`Worker` class does not have signatures for its event listeners, therefore it always defaults to `EventEmitter`'s `on(event: string | symbol, listener: (...args: any[]) => void): this;`.

for comparison, both bull 3 queue (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bull/index.d.ts#L797 / https://github.com/taskforcesh/bullmq/blob/master/src/classes/compat.ts#L430) and `QueueEvents` (https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-events.ts#L6) have them

roggervalf:
@m-ronchi please try the last version, it should have event listener typing

PR #605, opened 2021-06-19
Feat add drained event typing on worker

roggervalf:
Same as in title. Ref #523


src/classes/worker.ts
```diff
@@ -18,6 +18,7 @@ import { isRedisInstance, delay } from '../utils';
 export const clientCommandMessageReg = /ERR unknown command ['`]\s*client\s*['`]/;
 
 export declare interface Worker {
+  on(event: 'drained', listener: () => void): this;
```
roggervalf, line 21: I saw here https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveToActive-8.lua#L99 that any value is provided to the listener

manast:
thanks!

PR #606, opened 2021-06-20
Fix missing event typing

roggervalf:
The actual drained event does not contain other args apart of event name, the typing was expecting that the event id is passing so I updated the logic here, also the active typing was wrong so it's updated now. I noticed that there are not keywords, so I added some what do you think https://docs.npmjs.com/cli/v7/configuring-npm/package-json#keywords?

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -6,7 +6,7 @@ import { StreamReadRaw } from '../interfaces/redis-streams';
 export declare interface QueueEvents {
   on(
     event: 'active',
-    listener: (args: { jobId: string }, prev?: string) => void,
+    listener: (args: { jobId: string; prev?: string }, id: string) => void,
   ): this;
   on(
     event: 'waiting',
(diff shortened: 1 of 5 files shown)


src/classes/queue-events.ts
```diff
@@ -99,8 +99,14 @@ export class QueueEvents extends QueueBase {
                 break;
             }
 
-            this.emit(args.event, args, id);
-            this.emit(`${args.event}:${args.jobId}`, args, id);
+            const { event, ...restArgs } = args;
+
+            if (event === 'drained') {
+              this.emit(event, id);
```
roggervalf, line 105: I'm not emitting same value as in line 108 because there is none jobId in restArgs for drained event

src/classes/worker.ts
```diff
@@ -18,9 +18,17 @@ import { isRedisInstance, delay } from '../utils';
 export const clientCommandMessageReg = /ERR unknown command ['`]\s*client\s*['`]/;
 
 export declare interface Worker {
-  on(event: 'drained', listener: () => void): this;
+  on(
+    event: 'active',
+    listener: (job: Job, result: null, prev: string) => void,
```
roggervalf, line 23: curious why we are having the second param as null always, https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L345

manast: I do not remember but maybe it was used to pass a promise that could be cancelled to cancel the job that became active, however it is not a good design soo we should not port it over to BullMQ.

manast:
Looks good.

src/classes/worker.ts
roggervalf: @manast, could we delete the result param in this case, also I just put result name variable because I didn't know which name to use xd?

manast: Yes, we could delete it.

Issue #414, opened 2021-03-04
Sandboxed updateProgress()


codetheweb:
I think that either the types or the function parameters are messed up for sandboxed processes in regards to `Job.updateProgress()`. The function is available when using non-sandboxed workers, but it's not part of the job object passed to sandboxed processors (however, `Job.progress()` is a thing).

manast:
yes I see what you mean, thanks for reporting!

JosephScript:
It also doesn't have `job.isActive` or any of the other functions:

```js
{
  id: '1',
  name: '1',
  data: {
    id: '1',
  },
  opts: {
    attempts: 10,
    delay: 5000,
    jobId: '1',
    backoff: { type: 'custom' }
  },
  progress: [Function: progress],
  attemptsMade: 0,
  processedOn: 1618000638460,
  timestamp: 1618000573242,
  stacktrace: '[]',
  returnvalue: 'null',
  returnValue: null,
  log: [Function: log]
}
```

JosephScript:
I'm not sure if I should open a new issue or not since it seems related, but the event listeners also do not work on Sandboxed processors:

```js
const processorFile = path.join(__dirname, 'processor.js')

const worker = new Worker(qName, processorFile, {
  limiter: {
    max: 10,
    duration: 1000,
  },
  connection: redis,
  settings: {
    backoffStrategies: {
      //...
    },
  },
})

// None of these are firing:
worker.on('progress', (job, progress) => {
  console.info(`Job Progress: ${progress}`)
})
worker.on('completed', (job) => {
  console.info(`Job Complete: ${job.id} has completed!`)
})

worker.on('stalled', (job) => {
  console.error(`Job Error: ${job.id} has stalled.`)
})

worker.on('failed', (job, err) => {
  console.error(`Job Error: ${job.id} has failed with: ${err.message}\n${err}`)
})
```

roggervalf:
@JosephScript, I'm seeing test cases for completed, failed, progress, and active events for example on https://github.com/taskforcesh/bullmq/blob/master/src/test/test_sandboxed_process.ts#L39, maybe you could check that test file

PR #607, opened 2021-06-20
Fix sandbox: use updateProgress method name

roggervalf:
Ref #414


manast:
Thanks. Yes, this is a good fix. I am a bit worried though that if we release this fix it may break existing production systems. We can release as a breaking change or add an "alias" function so that it still works for users using the wrongly named method in sandboxed processors, what do you think?

roggervalf:
> Thanks. Yes, this is a good fix. I am a bit worried though that if we release this fix it may break existing production systems. We can release as a breaking change or add an "alias" function so that it still works for users using the wrongly named method in sandboxed processors, what do you think?

I can add an alias for it, but as the original was using an external variable to be modified I had to copy the same code in the alias :/

roggervalf:
I found a better way to handle that reference, now should be fine

src/test/fixtures/fixture_processor_progress.js
```diff
@@ -9,8 +9,8 @@ const delay = require('./delay');
 module.exports = function(job) {
   return delay(50)
     .then(() => {
-      job.progress(10);
-      job.log(job.progress());
+      job.updateProgress(10);
+      job.log(job.updateProgress());
```
manast, line 13: here we should use just progress() since updateProgress() should always update the progress not return the current progress.

src/interfaces/sandboxed-job.ts
```diff
@@ -5,12 +5,13 @@ import { JobsOptions } from './jobs-options';
  * @see {@link https://docs.bullmq.io/guide/workers/sandboxed-processors}
  */
 export interface SandboxedJob<T = any, R = any>
-  extends Omit<JobJson, 'data' | 'opts' | 'progress' | 'log' | 'returnValue'> {
```
roggervalf, line 8: JobJson does not have log attribute, so we are safe deleting it here

PR #608, opened 2021-06-20
Add commitizen for semantic-release format helper

roggervalf:
We have a reference of commitizen on package.json but the package is missing, this package could help to new users that don't know about semantic release commit format, with a new script: yarn cm. Also a new contributing file was added in order to explain how to submit a pr.


contributing.md
```diff
@@ -0,0 +1,34 @@
+# Contributing
+
+## Commit messages
+
+This package is using semantic-release to automate the release process, and this depends on a specific format for commit messages. In case you are new using semantic-release or you just want a helper to make your commits, please run `yarn cm` to use `commitizen` to properly format your commit messages so they can be automatically processed and included in release notes.
```
manast, line 5: I think we should link to the format specification for the commit messages, since we are using default settings it should be the angular commit format: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format

However the scopes are of course not the same, it is possible to actually configure which scopes are allowed but I  do not think that gives us too much and sometimes it is limiting.

contributing.md
```diff
@@ -0,0 +1,34 @@
...
+```sh
+yarn
+yarn test
+```
+
+- Provide feedback on the PR about your changes and results.
+
+## Doing a release
+
+We are using semantic-release instead of this:
```
manast, line 23: I think we could rephrase to something like this:
Releases are automatically performed by semantic-release and consists on the following:

manast (commented):
Great stuff man. Actually I added "husky" with a hook for linting the commits, however for some users this hook is not working, I could not determine why, it works for me.

contributing.md
roggervalf: sure, I added in last commit.

contributing.md
roggervalf: same as above.

PR #611, opened 2021-06-23
fix: fixed circular dependency

prisis:


src/classes/flow-producer.ts
```diff
@@ -1,11 +1,12 @@
-import uuid = require('uuid');
+import { v4 } from 'uuid';
 import { Redis, Pipeline } from 'ioredis';
 import { EventEmitter } from 'events';
 import { QueueBaseOptions } from '../interfaces';
 import { RedisClient, RedisConnection } from './redis-connection';
 import { KeysMap, QueueKeys } from './queue-keys';
 import { FlowJob } from '../interfaces/flow-job';
 import { Job } from './job';
+import { getParentKey } from '@src/utils';
```
manast, line 9: Using @src usually leads to problems compiling the library, it is better to use relative paths.

manast (commented):
Thanks. I wonder, which circular dependency is this PR addressing?

prisis:
Found it when i was using rollupjs   
  
Circular dependency: src/classes/job.ts -> src/classes/flow-producer.ts -> src/classes/job.ts  
Circular dependency: src/classes/index.ts -> src/classes/queue-getters.ts -> src/classes/worker.ts -> src/classes/index.ts

roggervalf:
@prisis could you please try the last versión to verify if there is any other circular reference

PR #619, opened 2021-06-28
Feat save worker info

roggervalf:
This PR is just an idea following the comment in worker file, I would like to receive some feedback about it. Note: It's still in progress


src/classes/worker.ts
```diff
@@ -226,6 +239,16 @@ export class Worker<
     return this.nextJobFromJobData(jobData, id);
   }
 
+  private async saveWorkerInfo(now: number) {
+    const workerKey = `${this.workerName}:${this.id}`;
```
roggervalf, line 243: should we consider a prefix in this key?

src/classes/queue-scheduler.ts
```diff
@@ -168,6 +169,18 @@ export class QueueScheduler extends QueueBase {
     return [0, 0];
   }
 
+  private async workerCleaner(client: RedisClient) {
```
manast, line 172: I think we can use one commando to achieve the same, namely https://redis.io/commands/zremrangebyscore so it can remove all the workers that are older than a given timestamp, simpler and more efficiently. Also use a const instead of 30000, ideally a configurable option.

src/classes/queue-scheduler.ts
```diff
@@ -168,6 +169,18 @@ export class QueueScheduler extends QueueBase {
...
+    const [workerKey, timestamp] = await client.zrange(
+      this.keys.workers,
+      0,
+      0,
+      'WITHSCORES',
+    );
+    if (Number(timestamp) + 30000 < Date.now()) {
```
manast, line 179: I think instead of 30000 it should be double of the time used when updating the worker.

src/classes/worker.ts
```diff
@@ -149,8 +154,16 @@ export class Worker<
     const tokens: string[] = Array.from({ length: opts.concurrency }, () =>
       v4(),
     );
+    let now = Date.now();
```
manast, line 157: maybe a better name instead of now, "lastWorkerInfoUpdate" or something better.

src/classes/worker.ts
```diff
@@ -149,8 +154,16 @@ export class Worker<
...
 
+    await this.saveWorkerInfo(now);
     while (!this.closing) {
+      const currentTime = Date.now();
+      if (Date.now() - now > 30000) {
```
manast, line 162: reuse currentTime instead of calling Date.now() again.

manast: parametrize 30000 as a worker option

src/classes/worker.ts
```diff
@@ -226,6 +239,16 @@ export class Worker<
     return this.nextJobFromJobData(jobData, id);
   }
 
+  private async saveWorkerInfo(now: number) {
+    const workerKey = `${this.workerName}:${this.id}`;
+    const client = await this.blockingConnection.client;
+    const multi = client.multi();
+    multi.zadd(this.keys.workers, now, workerKey);
```
manast, line 246: what if instead of using a new key we just store the info as a json object in the zset? that would also simplify deletion logic.

manast (changes requested):

src/classes/worker.ts
```diff
@@ -149,8 +154,16 @@ export class Worker<
     const tokens: string[] = Array.from({ length: opts.concurrency }, () =>
       v4(),
     );
+    let lastWorkerInfoUpdate = Date.now();
 
+    await this.saveWorkerInfo(lastWorkerInfoUpdate);
     while (!this.closing) {
+      const currentTime = Date.now();
+      if (currentTime - lastWorkerInfoUpdate > 30000) {
```
manast, line 162: This 30000 must be somehow "syched" with workerCleanerTimeout in queue scheduler for this mechanism to work properly.

src/classes/worker.ts
manast: I was probably tired when I wrote this, of course my suggestion above is not possible at all.
Just trying to figure out how we could solve this so that QueueScheduler and Worker do not need to work in sync for this feature to work.

src/classes/worker.ts
```diff
@@ -231,6 +245,16 @@ export class Worker<
     return this.nextJobFromJobData(jobData, id);
   }
 
+  private async saveWorkerInfo(now: number) {
+    const workerKey = `${this.workerName}:${this.id}`;
+    const client = await this.blockingConnection.client;
+    const multi = client.multi();
+    multi.zadd(this.keys.workers, now, workerKey);
+    multi.hmset(workerKey, 'name', this.workerName, 'queueName', this.name);
+    multi.expire(workerKey, 30);
```
manast, line 254: this expire is not so useful since we still need to delete the worker key from the workers set or the two data structures will be off-sync

roggervalf:
I'm still thinking how to address some comments

manast:
> I'm still thinking how to address some comments

will try to give you more feedback as soon as I get some extra time...

roggervalf:
> will try to give you more feedback as soon as I get some extra time...

no worries, I was notified that it is still in progress status

roggervalf:
This feature was moved to another updated branch

manast:
So basically we are back to square one. But the major advantage of using client setname/list is that we do not need to manually keep updating which worker is currently alive or not... sorry for leading you in the wrong direction when we discussed this issue the other day.

PR #620, opened 2021-06-28
Fix reprocess: do not store job.id in added list

roggervalf:
In order to be updated with this fix https://github.com/OptimalBits/bull/pull/2075


roggervalf:
@manast please when you get a chance 👀. This is ready for a review

manast:
Thanks!

PR #622, opened 2021-06-30
Feat queue and worker: add missing event typings

roggervalf:
Same as the title


src/classes/queue.ts
```diff
@@ -209,7 +214,7 @@ export class Queue<
       | 'paused'
       | 'delayed'
       | 'failed' = 'completed',
-  ) {
+  ): Promise<string[]> {
```
manast, line 217: This change is breaking a couple of tests.

roggervalf: Thank you, I updated it, now should be fine for another :eyes:

src/test/test_worker.ts
```diff
@@ -240,11 +241,11 @@ describe('workers', function() {
       expect(job.data.foo).to.be.eql('bar');
 
       return new Promise(resolve => {
-        worker.on('failed', async jobId => {
+        worker.on('failed', async (job, error) => {
```
roggervalf, line 244: I changed these test cases because when digging into those, I found that job were instances and not id strings

manast: True, local events on workers returns instances, this is the same as in Bull v3.

manast (approved):

Issue #599, opened 2021-06-17
Dynamically add childJobs while the main job is running


Haruki-Mortis:
Maybe I missed it, but it would be awesome if it was possible to add childJobs to a job while it is running.

Example:
```typescript
const mainProcessor = async (job: Job) => {
  try {
    // do something
  } catch (error) {
    await job.addChild('errorProcessing', {error: error, parentId: job.id});
    throw error; 
  }
}
```

roggervalf:
@Haruki-Mortis you can do it by using https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/api/bullmq.job.create.md that has JobOptions.parent https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/api/bullmq.jobsoptions.md, as I understand by your example, when you create that child, your parent job will not move to failed since you will have a pending child, at some point you would need to move your parent job to waiting children state https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L801, there is a method to do this, but it is used on manual processing, you could use it but you should be aware that it will throw an error when the worker will try to move the parent job to failed or completed, then the parent job will be re-processed when the job is returned to active state. For now there is not an atomic way to manage this, but it is still in discussion about how to design it.

Haruki-Mortis:
@roggervalf  Thank you very much for the answer.
It helps me a lot.
In my current case, I don't have to wait for the child to be processed.
In the first case it is important to see all child/parent jobs in the job information.

PR #624, opened 2021-07-01
Feat job: add option to create independent children

roggervalf:
There is a case where we only want to keep the relationship between parent-child and not waiting for children, so it is possible that parent completes first
I have also this case, I would like your opinion about it.
re #599


src/interfaces/flow-job.ts
```diff
@@ -5,6 +5,6 @@ export interface FlowJob {
   queueName: string;
   data?: any;
   prefix?: string;
-  opts?: Omit<JobsOptions, 'parent'>;
+  opts?: Omit<JobsOptions, 'parent' | 'independence'>;
```
ccollie, line 8: "independence" should probably be "independent"

roggervalf: makes sense to me, thank you

manast:
Trying to wrap my head around this. So what is the reason for having the "independents" jobs? can't this be solved on top of BullMQ in user code?

roggervalf:
> Trying to wrap my head around this. So what is the reason for having the "independents" jobs? can't this be solved on top of BullMQ in user code?

Sure I can explain this idea, in my case I also have some children jobs that I don't need to wait to wait for them, but for easy tracking would be nice to have the relationship as those jobs are still children of one parent, I was thinking on extending the children get methods to support this case, for now I extended the getDependenciesCount

Visot:
what about orphans instead of independents?

roggervalf:
> what about orphans instead of independents?

I tempted to use this word now, ty

manast:
> > what about orphans instead of independents?
> 
> I tempted to use this word now, ty

Would "orphan" (lack of parent) really represent the intent? (considering your explanation above where you wrote that the case is to keep track children of a parent but without needing to wait for them).

roggervalf:
> > > what about orphans instead of independents?
> > 
> > 
> > I tempted to use this word now, ty
> 
> Would "orphan" (lack of parent) really represent the intent? (considering your explanation above where you wrote that the case is to keep track the children of a parent but without needing to wait for them).

Yes, that's the intention, would you prefer independents in this case?

manast:
Sorry for the late response.  I am a bit reluctant for this kind of functionality that is for bookkeeping purposes if it can build on top of BullMQ, the reason is that this will otherwise lead to feature creep and I really want to avoid that at all costs.

roggervalf:
> Sorry for the late response. I am a bit reluctant for this kind of functionality that is for bookkeeping purposes if it can build on top of BullMQ, the reason is that this will otherwise lead to feature creep and I really want to avoid that at all costs.

No worries, I can keep this pr opened to show the idea about how it could be implemented

PR #628, opened 2021-07-03
fix(options): add typing to backoffStrategies

clouedoc:
## Commit description

```
Before, `AdvancedOptions.backoffStrategies` was a plain object (`{}`).
This induced no intellisense in code editors and also weaker typing.
With this commit, we get autocomplete when creating custom backoff strategies.

BREAKING CHANGE: Invalid backoff strategies will not be compiled anymore.
```

## Why ? 

This allow for better discoverability. I had to read BullMQ's code to understand that backoff can be computed from the error message and Job data.

### Before

image

(no autocompletion)

### After

image

(nice autocompletion and type checking)

### DIFF
--- a/src/interfaces/advanced-options.ts
+++ b/src/interfaces/advanced-options.ts
@@ -1,8 +1,10 @@
+import { BackoffFunction } from '../classes/backoffs';
+
 export interface AdvancedOptions {
   /**
    * A set of custom backoff strategies keyed by name.
    */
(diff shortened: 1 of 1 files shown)


manast:
Thanks for the fix. I guess this will be the next major release of BullMQ then :).

clouedoc:
Happy to be able to contribute something :)

roggervalf:
In version 3, typing changed

PR #630, opened 2021-07-06
Feat parent opts in root job for a flow

roggervalf:
This PR allows to add parent opts for a root job in a flow, it makes sense only allow this for the first parent job since children are concatenated into their parents so children should not have this option, this option is generated by add flow method.

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -7,7 +7,6 @@ import {
   FlowQueuesOpts,
   FlowOpts,
   QueueBaseOptions,
-  QueueOptions,
   RedisClient,
 } from '../interfaces';
 import { getParentKey, jobIdForGroup } from '../utils';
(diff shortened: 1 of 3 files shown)


src/interfaces/flow-job.ts
```diff
@@ -1,10 +1,19 @@
...
 
-export interface FlowJob {
+export interface FlowChildJob {
   name: string;
   queueName: string;
   data?: any;
   prefix?: string;
   opts?: Omit<JobsOptions, 'parent'>;
-  children?: FlowJob[];
+  children?: FlowChildJob[];
```
manast, line 9: We could refactor these two almost identical interfaces. Two approaches comes to mind, either define a base interface with the common properties and have to especialized interfaces extending from it, or create a generic, where the opts type is passed when defining the concrete interfaces, I think this can be done:
```ts
interface FlowJobBase<T> {
  name: string;
  queueName: string;
  data?: any;
  prefix?: string;
  opts?: T;
  children?: FlowChildJob[];
}

export type FlowChildJob = FlowJobBase<Omit<JobsOptions, 'parent'>>;
export type FlowJob = FlowJobBase<JobsOptions>>;
```

src/classes/flow-producer.ts
```diff
@@ -67,7 +68,16 @@ export class FlowProducer extends EventEmitter {
     const client = await this.connection.client;
     const multi = client.multi();
 
-    const jobsTree = this.addNode(multi, flow);
+    const parentOpts = flow?.opts?.parent;
```
manast, line 71: what about just adding an extra optional parameter to "add" instead of passing the parent options in the FlowJob interface?

manast (changes requested):

src/classes/flow-producer.ts
roggervalf: I'm thinking on the addBulk method https://github.com/taskforcesh/bullmq/pull/542, maybe would be expected to create flows with different parents too, what do you think?

src/test/test_flow.ts
```diff
@@ -111,6 +111,103 @@ describe('flows', () => {
     await removeAllQueueData(new IORedis(), parentQueueName);
   });
 
+  it('should allow parent opts on the root job', async () => {
```
manast, line 114: The part that still intrigues me is how this is handle in a real life scenario where I assume there would be a grandparent worker which currently would not wait for the parentJob to be created so that the grandparent will wait for it to complete... I apologize if I miss something obvious🙏

roggervalf: I was thinking on this scenario and now that that you mention this, I think it would have more sense to use it in manual processing since the grandparent should not have a good way to wait for the parentJob in the test case that I did. Maybe this pr could wait until we get the feature of waiting children from function processors.

PR #637, opened 2021-07-13
feat(worker+scheduler): add a "running" attribute for healthchecking

sven-codeculture:
Hey everyone,

we sometimes run into the problem, that the worker stops doing its job. I've added this code for a possibility to check, if the "run()" loop is still up and healthy so that i can restart my process in case it is not.

Would be great to have this as soon as possible since we had some bigger issue in some of our projects the last day regarding died workers.

Thanks,
Sven


sven-codeculture:
Just added the same functionality to the scheduler since this can have the same problem.

PR #642, opened 2021-07-16
Hotfix for connection hanging when `close()`

agent-ly:
Summary: When supplying your own IORedis instance and `lazyConnect` is true, the RedisConnection instance will not connect on its own because it does not check if it's `status` is "wait". While this may be intended behavior, there is no direct way to call `connect` on the duplicated Redis instance causing it to never connect and indefinitely wait until ready.

Context: Came across this issue when trying to create unit tests for my application. I didn't look at any guidelines for PRs or issues because its such a simple fix so if you don't merge mine at least you know the problem and how to fix it yourself.

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -66,6 +66,10 @@ export class RedisConnection extends EventEmitter {
 
         client.once('ready', handleReady);
         client.once('error', handleError);
+        
+        if (client.status === 'wait') {
+          client.connect(); 
+        }
       }
     });
   }


manast:
Looks good, thanks.

agent-ly:
> 
> 
> Looks good, thanks.

I'll say that there is probably more to it (race conditions) but this solves the issue I had. An alternate solution would be allowing us to start the duplicated client ourselves instead of it being hard coded.

arbourd:
Wouldn't it make sense to set lazyConnect to false for unit tests, in this case? BullMQ is able to operate just fine with `lazyConnect: true` otherwise.

I don't entirely understand the need for this change.

manast:
> I don't entirely understand the need for this change.

Basically it makes possible to use redis instances that have "lazyConnect: true" as option. The question is, is there are any issue due to this change?

arbourd:
`lazyConnect :true` did work before, and it would connect lazily (as is when something tries to interact with Redis via ioredis). This change effectively forces a connection as soon as BullMQ is initialized.

The only issue is it effectively disables the lazyConnect functionality.

PR #650, opened 2021-07-20
Fix moveToFinished: remove stalled jobs when finishing

roggervalf:
Same as the title


manast:
interesting that it was out-commented :). Thanks!.

PR #657, opened 2021-07-22
fix: connection hangs with failed connection

### LINKED ISSUE #656: RedisClient.waitUntilReady never resolves
In the event that `RedisClient` fails to connect due to `ECONNREFUSED`, it remains stuck in an undetectable unresolved state. All other methods (including `close`) that internally depend on `_client` become inoperable. This makes it impossible to gracefully shut down an app that has failed to connect to redis.

This issue is related to this change https://github.com/taskforcesh/bullmq/commit/fb3377280b3bda04a15a62d2901bdd78b869e08c, where `waitUntilReady` was updated to simply ignore `ECONNREFUSED` errors.

10xjs:
Update the `RedisConnection.waitUntilReady` static method to better reflect the internal ioredis status.

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -6,6 +6,11 @@ import { load } from '../commands';
 import { ConnectionOptions, RedisOptions } from '../interfaces';
 import { isRedisInstance } from '../utils';
 
+/**
+ * See https://github.com/luin/ioredis/blob/d65f8b2/lib/utils/index.ts#L338
+ */
+const CONNECTION_CLOSED_ERROR_MSG = 'Connection is closed.';
+
 export type RedisClient = Redis | Cluster;
 
 export class RedisConnection extends EventEmitter {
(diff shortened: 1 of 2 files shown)


src/classes/redis-connection.ts
```diff
@@ -6,6 +6,11 @@ import { load } from '../commands';
 import { ConnectionOptions, RedisOptions } from '../interfaces';
 import { isRedisInstance } from '../utils';
 
+/**
+ * See https://github.com/luin/ioredis/blob/d65f8b2/lib/utils/index.ts#L338
+ */
+const CONNECTION_CLOSED_ERROR_MSG = 'Connection is closed.';
```
manast, line 12: I see the const ```CONNECTION_CLOSED_ERROR_MSG``` is exported by ioredis, can't we import it so that we eliminate the duplication?

src/test/test_connection.ts
```diff
@@ -148,20 +152,41 @@ describe('connection', () => {
...
-      try {
-        await queueFail.waitUntilReady();
-        reject(new Error('Did not fail connecting to invalid redis instance'));
-      } catch (err) {
-        expect(err.code).to.be.eql('ECONNREFUSED');
-        await queueFail.close();
-        resolve();
-      }
+    await expect(queueFail.waitUntilReady()).to.be.eventually.rejectedWith(
+      'Connection is closed.',
```
manast, line 161: use ioredis' CONNECTION_CLOSED_ERROR_MSG instead of string.

manast (commented):
thanks for this nice PR. Just a small comment.

src/classes/redis-connection.ts
10xjs: I must have totally missed that. I'll try to import it here.

src/classes/redis-connection.ts
```diff
@@ -48,29 +51,31 @@ export class RedisConnection extends EventEmitter {
...
   static async waitUntilReady(client: RedisClient) {
-    return new Promise<void>(function(resolve, reject) {
-      if (client.status === 'ready') {
+    if (client.status === 'ready') {
+      return;
+    }
+
+    if (client.status === 'wait') {
+      return client.connect();
+    }
```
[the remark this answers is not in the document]
manast: @airhorns maybe it is possible to check if lazyConnect is true and in that case not connect if we are in the wait state?
Also I wonder if it is possible to end in the "wait" state *after* having been connected to redis?, if so we should not only check if lazyConnect is true but also if we have been connected already... happy to merge a patch if you work this out :).

arbourd:
This was the suggestion from your comment though, yes? 

Are you saying you'd prefer that no lazy ioredis connections is preferred?

PR #659, opened 2021-07-23
Fix/output error before trying

manast:


manast:
closing as this is not relevant anymore

PR #660, opened 2021-07-25
Fix docs changelog style

roggervalf:
The list of version are not being showed properly, it is about # in each version after 1.39.4

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -11,13 +11,11 @@ import {
   QueueSchedulerOptions,
   WorkerOptions,
 } from '../interfaces';
-import { array2obj } from '../utils';
-import { MoveToChildrenOpts } from './job';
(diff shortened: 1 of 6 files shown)


manast:
Thanks. For some reason there are other changes in this PR unrelated to the documentation fixes.

roggervalf:
> Thanks. For some reason there are other changes in this PR unrelated to the documentation fixes.

Sorry, I didn't put comments related to those chages, I was moving getParentKey to utils file since it is used in job, job-producer and script files

src/classes/scripts.ts
```diff
@@ -11,13 +11,11 @@ import {
   QueueSchedulerOptions,
   WorkerOptions,
 } from '../interfaces';
-import { array2obj } from '../utils';
-import { MoveToChildrenOpts } from './job';
```
roggervalf, line 15: job is imported twice

PR #663, opened 2021-07-26
Fix rateLimiter: check groupKey is not undefined

roggervalf:
when processing a job that does not have a groupKey, the id is generated with undefined at the end, so every job with this content also ends with undefined too

### DIFF
--- a/src/test/test_worker.ts
+++ b/src/test/test_worker.ts
@@ -82,20 +82,23 @@ describe('workers', function() {
       expect(job.id).to.be.ok;
       expect(job.data.foo).to.be.eql('bar');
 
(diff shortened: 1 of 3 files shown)


src/test/test_worker.ts
```diff
@@ -82,20 +82,23 @@ describe('workers', function() {
...
             resolve();
           } catch (err) {
             reject(err);
           }
         });
       });
+
+      await completed;
+
+      await worker.close();
```
roggervalf, line 101: for some reason this test failed in one of my build so I just refactored to put at least the close worker method out

manast:
Looks good!

PR #674, opened 2021-08-04
Feat add bulk flow jobs

Stormtv:
Adding bulk FlowJobs is a nice to have when adding multiple Flows at once. I attempted to change the least amount of code possible to make this happen.

Sorry for the delay in doing this PR but this should be cleaner than the last one #542. Let me know if there are any changes you would suggest.


src/test/test_flow.ts
```diff
@@ -791,6 +791,120 @@ describe('flows', () => {
     await removeAllQueueData(new IORedis(), parentQueueName);
   });
 
+  it('should process job when using addBulk', async () => {
```
manast, line 794: typo: job -> jobs

src/classes/flow-producer.ts
```diff
@@ -121,6 +121,52 @@ export class FlowProducer extends EventEmitter {
...
+    }
+    const client = await this.connection.client;
+    const multi = client.multi();
+
+    const jobsTrees = this.addNodes(multi, flows);
+
+    const result = await multi.exec();
+
+    let index = 0;
+    const updateJobIds = (jobTree: JobNode) => {
```
manast, line 150: I think you can refactor the ```updateJobIds``` function and change it so that it returns the index and so you can call it for every "jobTree" and avoid the duplication of code. So the loop will become something like this:
```ts
   let index = 0;
   for (const jobTree of jobsTrees) {
      index = updateJobIds(jobTree, result, index);
    }
```

manast (changes requested):
Looks great, please check my comments.

src/classes/flow-producer.ts
Stormtv: Pushed up a refactor for this good call

src/classes/flow-producer.ts
```diff
@@ -51,6 +51,32 @@ export class FlowProducer extends EventEmitter {
...
+    result: [Error, string][],
+    index: number,
+  ): number {
+    // TODO: Can we safely ignore result errors? how could they happen in the
+    // first place?
+    jobsTree.job.id = result[index][1];
+    index++;
+    const children = jobsTree.children;
+    if (children) {
+      for (let i = 0; i < children.length; i++) {
```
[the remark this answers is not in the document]
manast: After writing this comment I realized that updateIds is not really necessary since the jobs have already the ids before being added to the queue as they are generated with a uuid, so I removed this code and made a PR: https://github.com/taskforcesh/bullmq/pull/684
So unless I have missed something you can also do the same in your PR.

Stormtv:
@manast I merged the changes and removed the update job ids but the precommit script seemed to have linted some of the other files.

manast:
I removed the changes in changelog and merged it in this PR instead, thanks for the contribution!
https://github.com/taskforcesh/bullmq/pull/690

Issue #666, opened 2021-07-28
Redis connection error event handler is set ONLY after connection is established


YimingIsCOLD:
Bull Version: 1.40.1

```ts
const queue = new Queue('test', {
    connection: {
        host: 'localhost',
        port: 6379,
    },
});

queue.on('error', () => console.log('Something went wrong!'));
```

An unhandled error event is logged by `IORedis` when creating a new queue with no redis server available.

```
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1142:16)
```

If there is a redis server available during the queue creation and somehow lost connection to the redis server, the queue will emit an error with `Something went wrong!`.

After some investigation, I found out that [the error event hook is set after a `initializing` has been resolved](https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts#L45). However, in the case where there is no redis server available during queue creation, it will be stuck in [`waitUntilReady` promise](https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts#L66-L79) and thus not setting the error event hook.

I will suggest to place the hook at [here](https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts#L91) instead of after the `initializing` has been resolved. What you do think?

manast:
It sounds like a reasonable change.

YimingIsCOLD:
Alright! I will create a PR to address this.

YimingIsCOLD:
@manast How do I run the test locally? I got a fresh copy of the repo, did `yarn` and spin up a redis server with docker. When I tried to run the test with `yarn test`, it always failed with the following logs. I tried digging around but can't seems to find the root cause.

```
  connection
Error: ECONNRESET
    at Context.<anonymous> (/Users/yiming/workspace/bullmq/src/test/test_connection.ts:56:32)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    ✓ should recover from a connection loss (39ms)
Error: ECONNRESET
    at Worker.<anonymous> (/Users/yiming/workspace/bullmq/src/test/test_connection.ts:105:36)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    ✓ should handle jobs added before and after a redis disconnect
[...]

PR #675, opened 2021-08-05
Fix redis-connection: add error event in waitUntilReady

roggervalf:
Following #666 I tried to address this issue, so I don't see this log
```sh
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1142:16)
```
anymore when running the test


manast (commented):
Looks good. I think in the future we should improve this class so that it is actually possible to instantiate it even if Redis is temporarily unavailable.

PR #678, opened 2021-08-08
Add redis docker-compose to support windows development

roggervalf:
This PR allows windows users to run tests locally. Sometimes I'm using Windows to work in this package, so I use redis docker-compose in order to set redis locally, also https://github.com/microsoftarchive/redis/releases this is pretty old. In order to test this you could run the scripts yarn dc:up to start redis then you could use yarn test, test in Windows should run without problem.
Note: the docker-compose file is not included in releasable files, to check this, you can run yarn dry-run

### DIFF
--- a/src/test/test_rate_limiter.ts
+++ b/src/test/test_rate_limiter.ts
@@ -258,7 +258,7 @@ describe('Rate Limiter', function() {
           const timeDiff = Date.now() - startTime;
           // In some test envs, these timestamps can drift.
           expect(timeDiff).to.be.gte(30);
-          expect(timeDiff).to.be.below(100);
+          expect(timeDiff).to.be.below(325);
 
           let count = 0;
           let prevTime;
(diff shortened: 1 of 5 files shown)


src/test/test_rate_limiter.ts
```diff
@@ -258,15 +258,15 @@ describe('Rate Limiter', function() {
           const timeDiff = Date.now() - startTime;
           // In some test envs, these timestamps can drift.
           expect(timeDiff).to.be.gte(30);
-          expect(timeDiff).to.be.below(100);
+          expect(timeDiff).to.be.below(325);
```
roggervalf, line 261: I had to update this test since in Windows was failing

src/test/test_worker.ts
```diff
@@ -82,7 +82,7 @@ describe('workers', function() {
       expect(job.id).to.be.ok;
       expect(job.data.foo).to.be.eql('bar');
 
-      const completed = new Promise((resolve, reject) => {
+      const completed = new Promise<void>((resolve, reject) => {
```
roggervalf, line 85: this changes are related to ts errors about missing return type in these Promises

Issue #621, opened 2021-06-29
Flows - Ratelimiting issues


germanoeich:
Hello! 

I have observed some wierd behaviour when using the Worker limiter and a flow. No matter what I try, I just can't achieve a groupKey like ratelimiter with flows, in fact, when groupKey is set, the whole ratelimit for the flow jobs is broken. I noticed FlowJob has a `rateLimiterKey` property, I am unsure what the functionality of this field is, but I've also tried a bunch of combinations (job data field name, random values for each parent, setting it on the children, etc) and it doesn't seem to affect the ratelimiter at all.

Here's some repro code:
```typescript
import { FlowProducer, Job, Queue, QueueScheduler, Worker } from 'bullmq';

export default async function process(job: Job, token?: string) {
    console.log(`Received job for ${job.name} ${job.id}`);
}

const queueName = 'test';

const worker = new Worker(queueName, process, {
    concurrency: 100,
    limiter: {
        max: 1,
        duration: 5000,
        groupKey: 'guild',
    },
});

const qp = new QueueScheduler(queueName);
const f = new FlowProducer();
f.add({
    name: 'test',
    queueName: queueName,
    data: { guild: 1 },
    children: [
        {
            name: 'testd1-1',
            data: { guild: 1 },
            queueName,
            children: [
                {
                    name: 'testd2-1',
                    queueName,
                    data: { guild: 1 },
                    children: [{
                        name: 'testd3-1',
                        queueName
                    }],
                },
            ],
        },
    ],
});

f.add({
    name: 'test2',
    queueName: queueName,
    data: { guild: 2 },
    children: [
        {
            name: 'testd1-2',
            data: { guild: 2 },
            queueName,
            children: [
                {
                    name: 'testd2-2',
                    queueName,
                    data: { guild: 2 },
                    children: [{
                        name: 'testd3-2',
                        queueName
                    }],
                },
            ],
        },
    ],
});
```

**Version**: 1.36.1

**What I expect to happen**: Job testd3-2 and job testd3-1 are fired simultaneously, after 5s, testd2-1 and testd2-2 are fired simultaneously, so on and so forth
**What is happening**: Jobs are not being ratelimited at all, only respecting the children -> parent order.

If I remove the groupKey from the worker, the jobs _are_ ratelimited correctly (In this case, only 1 job is ran every 5s), however that's not the behaviour I need.

Am I approaching this issue in the wrong way / is there any other way of achieving this?

manast:
[...]

PR #692, opened 2021-08-20
Feat(flows): add rate limit

roggervalf:
re #621 pass queuesOpts to allow rate limit

### DIFF
--- a/src/test/test_connection.ts
+++ b/src/test/test_connection.ts
@@ -28,7 +28,7 @@ describe('connection', () => {
   it('should recover from a connection loss', async () => {
     let processor;
 
-    const processing = new Promise(resolve => {
+    const processing = new Promise<void>(resolve => {
       processor = async (job: Job) => {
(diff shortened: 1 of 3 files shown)


src/test/test_events.ts
```diff
@@ -94,7 +94,7 @@ describe('events', function() {
 
     await queue.add('test', {});
 
-    const completed = new Promise(resolve => {
+    const completed = new Promise<void>(resolve => {
```
roggervalf, line 97: missing return type

src/test/test_connection.ts
```diff
@@ -66,7 +66,7 @@ describe('connection', () => {
     let count = 0;
     let processor;
 
-    const processing = new Promise((resolve, reject) => {
+    const processing = new Promise<void>((resolve, reject) => {
```
roggervalf, line 69: same case of missing return type

manast (approved):
LGTM

PR #695, opened 2021-08-21
Docs: apply jsdoc

roggervalf:
This pr adds some jsdoc sentences and fix some others

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ yarn-error.log*
 *.sw*
 temp
 docs/gitbook/api
+package-lock.json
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -20,3 +20,4 @@ yarn-error.log*
 *.sw*
 temp
 docs/gitbook/api
+package-lock.json
```
manast, line 23: I think we need to keep package-lock.json for users that do not use yarn.

manast (changes requested):

.gitignore
roggervalf: Selection_042
I used npm in windows but it seems that when using yarn, then appears this warning, also our scripts are using yarn, I can still delete line 23, but what do you think?

manast: In that case we must recommend yarn as the tool to install BullMQ... in any case this is not so important since package-lock and yarn.lock are only used when developing BullMQ but not when using it as a dependency.

PR #697, opened 2021-08-22
fix: protect emit calls with throw/catch

### LINKED ISSUE #596: [exiting] Fatal exception: Missing lock for job err
I have a  worker server that was knocked down  due to an unhanded exception.  Upon restart it started to pull jobs from the queue and It again began to shut down due to Fatal exception: Error: Missing lock for job 4189 failed

Please advise

manast:
fixes #596


roggervalf (approved):
LGTM

PR #698, opened 2021-08-23
Fix/protect emit calls

roggervalf:


src/test/test_events.ts
```diff
@@ -89,6 +89,29 @@ describe('events', function() {
     await worker.close();
   });
 
+  it('emits error event when there is an error on other events', async function() {
+    const worker = new Worker(queueName, async job => {}, {
+      drainDelay: 1,
+    });
+
+    worker.once('completed', (job: any) => {
+      console.log(job.bar.id);
```
manast, line 98: We should not console.log in test cases.

src/test/test_events.ts
```diff
@@ -89,6 +89,29 @@ describe('events', function() {
...
+    });
+
+    const error = new Promise<void>(resolve => {
+      worker.once('error', resolve);
+    });
+
+    await queue.add('test', { foo: 'bar' });
+
+    await error;
```
manast, line 107: What is causing the error event? I am re-reading the test several times but cannot find the part that triggers the error 😂

manast: ah ok, it is trying to access id on bar on the job instance (since bar is undefined),

roggervalf: yes, jejeje 😆

PR #699, opened 2021-08-24
Feat(events): add added event when job is created

roggervalf:
Add added event when a job is created by addJob script

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -9,20 +9,11 @@ export declare interface QueueEvents {
     listener: (args: { jobId: string; prev?: string }, id: string) => void,
   ): this;
   on(
-    event: 'waiting',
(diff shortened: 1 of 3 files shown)


manast:
Thanks. Wondering, what is the purpose of this new event? I kind of have an old memory about this event in some context but I do not remember it anymore. As you know when adding a new job we emit either the "waiting" or the "delayed" events.

src/classes/queue-events.ts
```diff
@@ -9,20 +9,11 @@ export declare interface QueueEvents {
     listener: (args: { jobId: string; prev?: string }, id: string) => void,
   ): this;
   on(
-    event: 'waiting',
```
roggervalf, line 12: moved in alphabetical order

manast (approved):

PR #701, opened 2021-08-24
Style(tsdoc): add eslint-plugin-tsdoc

roggervalf:
Eslint support tsdoc, so we can know when there are unsupported tags or errors in our docs. Also the docs generator is using tsdoc as standard, no jsdoc

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -156,13 +156,12 @@ export class Job<T = any, R = any, N extends string = string> {
   }
 
   /**
-   * @method create
    * Creates a new job and adds it to the queue.
    *
-   * @param queue the queue where to add the job.
-   * @param name  the name of the job.
-   * @param data  the payload of the job.
-   * @param opts the options bag for this job.
+   * @param queue - the queue where to add the job.
+   * @param name - the name of the job.
+   * @param data - the payload of the job.
+   * @param opts - the options bag for this job.
    * @returns
    */
   static async create<T = any, R = any, N extends string = string>(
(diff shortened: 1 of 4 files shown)


src/classes/job.ts
```diff
@@ -156,13 +156,12 @@ export class Job<T = any, R = any, N extends string = string> {
   }
 
   /**
-   * @method create
```
roggervalf, line 159: method directive is not supported, just in jsdoc

src/classes/job.ts
```diff
@@ -156,13 +156,12 @@ export class Job<T = any, R = any, N extends string = string> {
...
    * Creates a new job and adds it to the queue.
    *
-   * @param queue the queue where to add the job.
-   * @param name  the name of the job.
-   * @param data  the payload of the job.
-   * @param opts the options bag for this job.
+   * @param queue - the queue where to add the job.
```
roggervalf, line 161: a hyphen should be added, even when there are not description

src/classes/job.ts
```diff
@@ -273,11 +270,10 @@ export class Job<T = any, R = any, N extends string = string> {
   }
 
   /**
-   * @method fromId
    * Fetches a Job from the queue given the passed job id.
    *
-   * @param queue the queue where the job belongs to.
-   * @param {string} jobId the job id.
```
roggervalf, line 280: just the name of the param, not type

roggervalf:
we will continue using jsdoc

roggervalf:
Reopened to be allowed to see tsdoc warnings

PR #716, opened 2021-08-28
refactor(error-codes): add error-codes enum

roggervalf:


manast (changes requested):
Looks good. However, it would be better to keep the same error codes returned by the .lua scripts so that they are in sync to what we return in the Bull v3 .lua scripts.

manast (approved):

PR #717, opened 2021-08-28
fix(retry): throw error when retry non failed job

roggervalf:


src/test/test_worker.ts
```diff
@@ -1717,11 +1712,7 @@ describe('workers', function() {
 
       await activating;
 
-      try {
-        await job.retry();
-      } catch (err) {
-        expect(err.message).to.equal('Retried job not failed');
-      }
+      await expect(job.retry()).to.be.rejectedWith('Retried job not failed');
```
roggervalf, line 1715: I recommend expect errors in this way, if not we could get silent failures in our tests as in this case

Issue #436, opened 2021-03-22
[Feature] Call Worker and QueueScheduler methods without launching the process


bogdan:
Manually processing jobs like this: https://docs.bullmq.io/patterns/manually-fetching-jobs can cause problems because `new Worker` launches the process instantly. This can cause a lot of unpredictable effects in support scripts if I just want to manually resolve issues in a queue without processing any single job from there.

Is it possible to add a constructor option to prevent the process from being launched instantly. Like:

```
const worker = new Worker(name, () => {}, {autorun: false})
```

Overall, I think calling `run()` in constructor is an anti-pattern: you don't make all the methods of `Worker` class easily accessible this way.

manast:
> Manually processing jobs like this: https://docs.bullmq.io/patterns/manually-fetching-jobs can cause problems because `new Worker` launches the process instantly.

No, it does not. If you do not define a process function the worker will not do anything.


> Overall, I think calling `run()` in constructor is an anti-pattern: you don't make all the methods of `Worker` class easily accessible this way.

I can agree on this. However not the comment above, it does not work as you assumed.

bogdan:
Good to know about missing processor, however this is something that is not very obvious. It probably requires at least documentation.

It is interesting that redis connection is always cloned. Maybe this is required for "launch" mode with processor, but is it required for "support" mode without processor?

manast:
> It is interesting that redis connection is always cloned. Maybe this is required for "launch" mode with processor, but is it required for "support" mode without processor?

yes, that must be an oversight from my part, the connection does not need to be clones in this case at all.

bogdan:
@roggervalf looks good! Can you also add methods I mentioned in https://github.com/taskforcesh/bullmq/issues/436#issuecomment-804891421 or can I open another issue to add them?

manast:
@bogdan regarding my comment above. Although technically you do not need to duplicate the connection, in practice whatever connection you use will be block as soon as you call ```getNextJob```. So I think the duplication makes sense to be performed always to avoid weirs issues where the user expects something to happen but nothing happens because the connection is blocked.

PR #724, opened 2021-08-31
feat(worker): launch without launching process

roggervalf:
re #436

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -103,27 +103,10 @@ export class Worker<
       }
       this.timerManager = new TimerManager();
 
-      if (this.opts.autorun)
-        this.execute().catch(error => {
-          this.running = false;
-          console.error(error);
-        });
+      if (this.opts.autorun) this.run();
     }
 
     this.on('error', err => console.error(err));
-  }
-
-  run(): void {
-    if (this.processFn) {
-      if (!this.running) {
-        this.execute().catch(error => {
-          this.running = false;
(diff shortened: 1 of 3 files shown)


src/classes/worker.ts
```diff
@@ -69,6 +69,7 @@ export class Worker<
       concurrency: 1,
       lockDuration: 30000,
       runRetryDelay: 15000,
+      process: true,
```
manast, line 72: I think "autorun" is a good name here actually.

src/classes/worker.ts
```diff
@@ -102,13 +103,23 @@ export class Worker<
...
+        this.run().catch(error => {
+          this.running = false;
+          console.error(error);
+        });
+    }
+
+    this.on('error', err => console.error(err));
+  }
+
+  process(): void {
```
manast, line 116: I think we can just make "run" public, so the error handling will be taken care by the caller. And inside the run() function we could just throw an exception if processFn is not defined (as it does not make sense to start the run loop if no process function is defined.

manast (changes requested):
Good that you fixed this issue! Please check the comments.

src/classes/worker.ts
roggervalf: sure, let me refactor it a little bit

manast (commented):
Looks good. I think you should squeeze this PR and use "feat" since you are adding a new feature.

PR #725, opened 2021-08-31
test: add missing close

roggervalf:


src/test/test_compat.ts
```diff
@@ -578,7 +578,8 @@ describe('Compat', function() {
       expect(count.waiting).to.be.eql(2);
       expect(count.completed).to.be.eql(2);
 
-      return Promise.all([worker1.close(), worker2.close()]);
+      await worker1.close();
```
manast, line 581: In theory these two should be the same as the Promise.all() except they run in parallel so it is a bit faster, did you have an issue with it?

manast (commented):
Looks good.

src/test/test_compat.ts
roggervalf: No issues, I was trying to be consistent with the other test cases where we are closing without returning a Promise.all

PR #727, opened 2021-08-31
fix(queuescheduler): handle shared connections fixes #721

### LINKED ISSUE #721: QueueScheduler & QueueEvents make the test suite suspend
This is the simplified version of our code base to reproduce the bug.
```
const Redis = require('ioredis')
const {QueueScheduler} = require('bullmq')

const connection1 = new Redis(6379, 'localhost')

let queueScheduler;
describe('test', ()=>{
    beforeAll(()=>{
        queueScheduler = new QueueScheduler('q1', {connection: connection1})
    })

    afterAll(async()=>{
        await queueScheduler.close()
        // await connection1.quit()
    }, 10000)
    
    it('nothin', ()=>{
        expect(true).toBe(true)
    })
})
```
The expected behavior is that teardown phase should successfully close the connection and test should end smoothly, but when running the test, it's suspended and throws this warnings: 
```
Jest did not exit one second after the test run has completed.
This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
When run the test with `--detectOpenHandles` option it complains about the Redis connection and Queuescheduler.

This condition makes our CI pipeline fail randomly because of timeout limit and I'm looking for a way to close the connections smoothly.

Environment:
`bullmq`: 1.44.2
`ioredis`: 4.27.8
`redis`: 6.0.9

manast:


src/classes/redis-connection.ts
```diff
@@ -142,7 +145,7 @@ export class RedisConnection extends EventEmitter {
   async close() {
     if (!this.closing) {
       this.closing = true;
-      if (this.opts != this._client) {
+      if (!this.shared) {
```
roggervalf, line 148: In queue-scheduler, this should be always true, but we could pass connection params only and this._client would not be set I think

manast: But in other cases we pass the connection and it still shared, for example if you pass a redis connection when using the Queue instance. If I haven't made a mistake I think we cannot avoid to explicitly tell RedisConnection if the connection is shared or not.

roggervalf (approved):
Lgtm

PR #729, opened 2021-09-01
feat(queue-scheduler): launch without launching process

roggervalf:
re #436

### DIFF
--- a/src/classes/queue-scheduler.ts
+++ b/src/classes/queue-scheduler.ts
@@ -37,7 +37,7 @@ export class QueueScheduler extends QueueBase {
 
   constructor(
     name: string,
-    { connection, ...opts }: QueueSchedulerOptions = {},
+    { connection, autorun = true, ...opts }: QueueSchedulerOptions = {},
   ) {
     super(name, {
       maxStalledCount: 1,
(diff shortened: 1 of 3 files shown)


src/classes/queue-scheduler.ts
```diff
@@ -53,97 +53,94 @@ export class QueueScheduler extends QueueBase {
...
+              streamLastId = id;
+            } else {
+              this.nextTimestamp = Number.MAX_VALUE;
+            }
+          }
         }
+        this.running = false;
+      } catch (error) {
+        this.running = false;
+        console.error(error);
```
manast, line 136: Instead of console.error here we should just throw the exception so the user calling this method should handle it

src/classes/queue-scheduler.ts
```diff
@@ -53,97 +53,94 @@ export class QueueScheduler extends QueueBase {
     }
 
     if (autorun) {
-      this.execute().catch(error => {
-        this.running = false;
-        console.error(error);
-      });
+      this.run();
```
manast, line 56: For this case we should keep the catch and use console.error.

manast (changes requested):
Almost there, please see my two comments.

manast (approved):

PR #730, opened 2021-09-01
Test/close instances

roggervalf:


src/test/test_worker.ts
```diff
@@ -1033,18 +1034,14 @@ describe('workers', function() {
...
-      done(new Error('Should throw an exception'));
-    } catch (err) {
-      done();
-    }
+    expect(
+      () =>
+        new QueueScheduler(queueName, {
+          stalledInterval: 0,
+        }),
+    ).to.throw('Stalled interval cannot be zero or undefined');
```
roggervalf, line 1044: in this way we could test errors

PR #742, opened 2021-09-09
fix(add-job): throw error when parent is completed

roggervalf:


src/classes/job.ts
```diff
@@ -116,6 +116,10 @@ export class Job<T = any, R = any, N extends string = string> {
    */
   parentKey?: string;
 
+  parentId?: string;
```
manast, line 119: Maybe it is time to put this parent related fields into a new interface to avoid poluting the job class.

roggervalf: sure, I refactor it to use an interface, when you get a chance please give some :eyes:

src/interfaces/jobs-options.ts
```diff
@@ -1,6 +1,11 @@
 import { RepeatOptions } from './repeat-options';
 import { BackoffOptions } from './backoff-options';
 
+export interface ParentKeys {
```
manast, line 4: Maybe a better name could be: "ParentJobOpts"

PR #746, opened 2021-09-12
fix(is-finished): reject when missing job key

### LINKED ISSUE #85: waitUntilFinished hang if removeOnComplete is set without TTL
When using the `removeOnComplete` option on a job, calling the `waitUntilFinished` will have an undefined behavior (unless ttl has been set).

If it's called before the job is cleared from the queue, it will work as intended but if the job is already removed, it will just hang indefinitely.

I think the `waitUntilFinished` function should warn the user if `removeOn...` is set as it may never resolve.

roggervalf:
fixes #85

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -1,3 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { CONNECTION_CLOSED_ERROR_MSG } from 'ioredis/built/utils';
 import { QueueEventsOptions } from '../interfaces';
 import { array2obj, delay, isRedisInstance } from '../utils';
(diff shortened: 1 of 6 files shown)


src/classes/queue-events.ts
```diff
@@ -244,7 +244,7 @@ export class QueueEvents extends QueueBase {
           }
         }
       } catch (err) {
-        if (err.message !== 'Connection is closed.') {
+        if ((err as Error).message !== 'Connection is closed.') {
```
manast, line 247: Here we can use ```CONNECTION_CLOSED_ERROR_MSG``` from:
```ts
import { CONNECTION_CLOSED_ERROR_MSG } from 'ioredis/built/utils';

```

manast (changes requested):
LGTM, just a small comment.

manast (approved):

PR #749, opened 2021-09-14
fix(wait-for-job): add catch block and emit error

### LINKED ISSUE #623: Promise rejection error in waitForJob
We are running redis on AWS in cluster mode, which is returning this redis error when it swaps master/replica:
```
UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)
```
That is then causing an error to throw in our bullmq services, full error:
```
ReplyError: UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
    at parseError (/code/node_modules/redis-parser/lib/parser.js:179:12)
    at parseType (/code/node_modules/redis-parser/lib/parser.js:302:14) {
  command: {
    name: 'brpoplpush',
    args: [
      'bull:task:jobs:wait',
      'bull:task:jobs:active',
      '5'
    ]
  }
}
```

It seems that this try/finally is missing a catch causing this error to bubble up, would that be correct?
https://github.com/taskforcesh/bullmq/blob/220b4f619b30a8f04979e9abd0139e46d89b424d/src/classes/worker.ts#L241

roggervalf:
fixes #623

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -7,7 +7,7 @@ import { CONNECTION_CLOSED_ERROR_MSG } from 'ioredis/built/utils';
 import * as semver from 'semver';
 import { load } from '../commands';
 import { ConnectionOptions, RedisOptions } from '../interfaces';
-import { isRedisInstance, handleError } from '../utils';
+import { isRedisInstance, isNotConnectionError } from '../utils';
 
 export type RedisClient = Redis | Cluster;
 
@@ -150,7 +150,9 @@ export class RedisConnection extends EventEmitter {
           await this._client.quit();
         }
       } catch (error) {
-        handleError(error);
+        if (isNotConnectionError(error)) {
+          throw error;
+        }
       } finally {
         this._client.off('error', this.handleClientError);
       }
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -7,7 +7,7 @@ import { CONNECTION_CLOSED_ERROR_MSG } from 'ioredis/built/utils';
 import * as semver from 'semver';
 import { load } from '../commands';
 import { ConnectionOptions, RedisOptions } from '../interfaces';
-import { isRedisInstance } from '../utils';
+import { isRedisInstance, handleError } from '../utils';
 
 export type RedisClient = Redis | Cluster;
 
@@ -64,6 +64,11 @@ export class RedisConnection extends EventEmitter {
     }
 
     return new Promise<void>((resolve, reject) => {
+      let lastError: Error;
+      const errorHandler = (err: Error) => {
+        lastError = err;
+      };
+
       const handleReady = () => {
         client.removeListener('end', endHandler);
         client.removeListener('error', errorHandler);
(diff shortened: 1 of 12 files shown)


src/classes/worker.ts
```diff
@@ -289,6 +292,11 @@ export class Worker<
         this.keys.active,
         opts.drainDelay,
       );
+    } catch (error) {
```
roggervalf, line 295: don't really know how to make a test for this :thinking:

manast: I think the only way to do it is by monkey patching ```client``` so that ```brpoplpush``` always throws an exception. So in the test overwrite this.blockingConnection.client with a stub. Sinon is the best for this, as we are already using it in other tests.

src/classes/worker.ts
```diff
@@ -289,6 +292,11 @@ export class Worker<
...
+      if ((error as Error).message !== CONNECTION_CLOSED_ERROR_MSG) {
+        await delay(2000);
```
manast, line 297: Just wondering, shouldn't we always delay? because otherwise if the error persists for a long time we will produce an endless loop here that may consume all CPU.
Also it would be great to refactor this magic number into a constant.

src/classes/worker.ts
roggervalf: so after adding the delay after if statement seems like in our test we are getting into that block, so I had to stub the delay method

src/test/test_repeat.ts
```diff
@@ -323,7 +325,7 @@ describe('repeat', function() {
   });
 
   it('should remove repeated job when using removeOnComplete', async function() {
-    this.timeout(200000);
+    this.timeout(20000);
```
roggervalf, line 328: these changes are because we are adding a higher number but we could choose a lower one in those cases, even when these changes are going to fail because of a change, test should not get stopped for too much time

src/classes/worker.ts
```diff
@@ -289,12 +297,22 @@ export class Worker<
...
+        this.emit('error', error);
+      }
+      await this.delay();
     } finally {
       this.waiting = false;
     }
     return jobId;
   }
 
+  async delay(): Promise<void> {
```
roggervalf, line 312: I need to do this in order to stub this function

manast: Ok, but we can keep it private.

roggervalf: when it is prívate I cannot stub it :/

manast: aha, thats right. Then just a comment to explain why it is needed is enought.

manast (approved):

src/classes/redis-connection.ts
```diff
@@ -150,7 +150,9 @@ export class RedisConnection extends EventEmitter {
           await this._client.quit();
         }
       } catch (error) {
-        handleError(error);
+        if (isNotConnectionError(error)) {
```
manast, line 153: Nice change. It is always better to have explicit functions instead of "handle" which can be anything. 👍

PR #750, opened 2021-09-14
feat(queue-events): launch without launching process

roggervalf:


src/classes/queue-events.ts
```diff
@@ -194,64 +199,77 @@ export class QueueEvents extends QueueBase {
...
           }
         }
-      } catch (err) {
-        if ((err as Error).message !== CONNECTION_CLOSED_ERROR_MSG) {
-          throw err;
-        }
-        await delay(5000);
+      } catch (error) {
+        this.running = false;
+        this.emit('error', error);
```
manast, line 269: Instead of emitting an error I think we just should throw the error so that the caller can handle it.

src/classes/queue-events.ts
```diff
@@ -194,64 +199,77 @@ export class QueueEvents extends QueueBase {
...
-        const data: StreamReadRaw = (await client.xread(
-          'BLOCK',
-          opts.blockingTimeout,
-          'STREAMS',
-          key,
-          id,
-        )) as any;
+        while (!this.closing) {
+          this.running = true;
+          try {
```
manast, line 220: I think it would be more elegant to refactor the while loop into a private method, maybe we can reuse the name "consumeEvents" for this. It will also make it nicer since the inner try/catch will be hidden inside the method.

src/classes/queue-events.ts
```diff
@@ -194,12 +199,29 @@ export class QueueEvents extends QueueBase {
       this.opts,
     );
 
-    this.consumeEvents().catch(err => this.emit('error', err));
+    if (autorun)
+      this.run().catch(error => {
```
manast, line 203: I think in this case we can emit the error, and actually we could also remove the console.log also, so the user needs to listen to "errors" to actually know about this error.

manast (commented):
Looks beautiful, just a small comment.

src/classes/queue-events.ts
```diff
@@ -194,12 +199,26 @@ export class QueueEvents extends QueueBase {
       this.opts,
     );
 
-    this.consumeEvents().catch(err => this.emit('error', err));
+    if (autorun) this.run().catch(error => this.emit('error', error));
```
manast, line 202: Always use { } for it statements. Don't we have a rule for this in the linter? I thought it was standard.

src/test/test_events.ts
```diff
@@ -24,6 +24,50 @@ describe('events', function() {
...
+      const queueName2 = `test-${v4()}`;
+      const queue2 = new Queue(queueName2);
+      const queueEvents2 = new QueueEvents(queueName2, { autorun: false });
+      await queueEvents2.waitUntilReady();
+
+      const waiting = new Promise(resolve => {
+        queue2.on('waiting', resolve);
+      });
+
+      queueEvents2.run();
```
manast, line 38: We could store a promise here and check that the call to run actually resolves when the queue is closed.

src/classes/queue-events.ts
```diff
@@ -250,7 +271,7 @@ export class QueueEvents extends QueueBase {
           }
         }
       } catch (err) {
-        if (isNotConnectionError(err)) {
+        if (isNotConnectionError(err) || this.connection.closed) {
```
roggervalf, line 274: as we were checking that this error is not a connection error, the promise never resolves, I had to add a boolean at connection level when end events is called, so here we could get into that condition and resolves this promise

src/test/test_events.ts
```diff
@@ -24,6 +24,53 @@ describe('events', function() {
...
+      const queueName2 = `test-${v4()}`;
+      const queue2 = new Queue(queueName2);
+      const queueEvents2 = new QueueEvents(queueName2, { autorun: false });
+      await queueEvents2.waitUntilReady();
+
+      const waiting = new Promise(resolve => {
+        queue2.on('waiting', resolve);
+      });
+
+      const run = new Promise(resolve => {
```
manast, line 38: This is the same as: 
```ts
const running = queueEvents2.run();
```

src/test/test_events.ts
```diff
@@ -24,6 +24,53 @@ describe('events', function() {
...
+    });
+
+    describe('when run method is called when queueEvent is running', function() {
+      it('throws error', async () => {
+        const queueName2 = `test-${v4()}`;
+        const queue2 = new Queue(queueName2);
+        const queueEvents2 = new QueueEvents(queueName2, { autorun: false });
+        await queueEvents2.waitUntilReady();
+
+        queueEvents2.run();
```
manast, line 59: Store running promise and check it is fulfilled after closing.

manast (approved):

PR #751, opened 2021-09-14
fix(connection): fail only if redis connection does not recover

roggervalf:
ref https://github.com/OptimalBits/bull/pull/2152

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -64,6 +64,8 @@ export class RedisConnection extends EventEmitter {
     }
 
     return new Promise<void>((resolve, reject) => {
+      const errorHandler = (err: Error) => {};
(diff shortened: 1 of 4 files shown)


src/classes/redis-connection.ts
```diff
@@ -76,14 +78,8 @@ export class RedisConnection extends EventEmitter {
         reject(new Error(CONNECTION_CLOSED_ERROR_MSG));
```
manast, line 78: Why not save "lastError" from the error handler ? We could initialize ```lastError``` with ```new Error(CONNECTION_CLOSED_ERROR_MSG)``` and then update it with the errors listened in the error handler, as in the bull case, don't you think?

manast (approved):

PR #752, opened 2021-09-15
refactor(utils): add handleError

roggervalf:
This PR is for have error handler in utils to be reused


src/classes/queue-events.ts
```diff
@@ -247,9 +244,7 @@ export class QueueEvents extends QueueBase {
           }
         }
       } catch (err) {
-        if ((err as Error).message !== CONNECTION_CLOSED_ERROR_MSG) {
```
roggervalf, line 250: we were missing to check here with and operator

manast (approved):

PR #756, opened 2021-09-17
fix(running): move running attribute before first async call

roggervalf:


src/classes/worker.ts
```diff
@@ -110,10 +110,11 @@ export class Worker<
       }
       this.timerManager = new TimerManager();
 
-      if (this.opts.autorun)
+      if (this.opts.autorun) {
         this.run().catch(error => {
           console.error(error);
```
manast, line 115: For consistency we should also replace this console.error by an emit('error')

manast (approved):

PR #764, opened 2021-09-22
feat(queue): add drain lua script

roggervalf:


src/test/test_worker.ts
```diff
@@ -1016,22 +1016,6 @@ describe('workers', function() {
     await worker.close();
   });
 
-  it('count added, unprocessed jobs', async () => {
```
roggervalf, line 1019: this test was moved to queue tests, makes more sense, since here there is no worker

manast (commented):
Looks great. I think we just need one test to cover a paused queue as well.

manast (approved):

PR #766, opened 2021-09-24
perf(obliterate): do not pass unused variables

roggervalf:


src/commands/obliterate-2.lua
```diff
@@ -32,25 +32,26 @@ local function getSetItems(keyName, max)
     return rcall('SMEMBERS', keyName, 0, max)
 end
 
-local function removeJobs(parentKey, keys)
+local function removeJobs(keys)
```
roggervalf, line 35: parentKey is passed here, but it's not used

manast (approved):

PR #772, opened 2021-09-29
Feat/extensions support

manast:


src/classes/job.ts
```diff
@@ -426,9 +430,10 @@ export class Job<T = any, R = any, N extends string = string> {
     fetchNext = false,
   ): Promise<void> {
     const client = await this.queue.client;
+    const message = err?.message;
```
roggervalf, line 433: is there a posible case where moveToFailed get something different than an error?

manast: In theory no, but there was some issue reported where it somehow err was not defined. This is not a solution but a painkiller.

roggervalf (approved):
LGTM

PR #794, opened 2021-10-06
feat(child-pool): Helpful Error on Interpolation

kody-abe:
If you have an error in your sandboxed code that a causes NodeJS to crash (ie. doing a require to a file that does not exist) you get an un-helpful error `Error initializing child: Internal Exception Handler Run-Time Failure and signal null`.

This change propagates the actual error up as a job failure.

### DIFF
--- a/src/classes/master.ts
+++ b/src/classes/master.ts
@@ -53,7 +53,13 @@ process.on('SIGINT', waitForCurrentJobAndExit);
 process.on('message', msg => {
   switch (msg.cmd) {
     case 'init':
-      processor = require(msg.value);
+      try {
+        processor = require(msg.value);
+      } catch (err) {
+        processor = function() {
+          return Promise.reject(err); 
+        }
+      }
       if (processor.default) {
         // support es2015 module.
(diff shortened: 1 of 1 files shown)


roggervalf:
hi @kody-abe, thanks for this pr, could you also try to add a test case to cover this change, you could add it in test_sandboxed_process.ts

src/classes/master.ts
```diff
@@ -53,7 +53,13 @@ process.on('SIGINT', waitForCurrentJobAndExit);
 process.on('message', msg => {
   switch (msg.cmd) {
     case 'init':
-      processor = require(msg.value);
+      try {
+        processor = require(msg.value);
+      } catch (err) {
+        processor = function() {
```
manast, line 59: Seems like what will happen here is that we will get a processor function that always fails with said error, so it will start processing the jobs in the queue and fail them all (If I am not too sleepy to miss something here).
Instead we should catch this error and do a process.send with a new "cmd" maybe like "init-failed", and then handle the error higher up.

PR #797, opened 2021-10-07
feat(sandboxed-process): handle init-failed error

roggervalf:


src/classes/child-pool.ts
```diff
@@ -53,6 +53,9 @@ async function initChild(child: ChildProcess, processFile: string) {
       if (msg.cmd === 'init-complete') {
         resolve();
         child.off('message', onMessageHandler);
+      } else if (msg.cmd === 'init-failed') {
+        reject(new Error(msg.err));
+        child.off('message', onMessageHandler);
```
manast, line 58: we could refactor out this line from the if clauses
```
child.off('message', onMessageHandler);
```

src/classes/master.ts
```diff
@@ -53,7 +53,16 @@ process.on('SIGINT', waitForCurrentJobAndExit);
 process.on('message', msg => {
   switch (msg.cmd) {
     case 'init':
-      processor = require(msg.value);
+      try {
+        processor = require(msg.value);
+      } catch (err) {
+        status = 'IDLE';
```
manast, line 59: We probably need a new status here too, like "ERRORED"

PR #805, opened 2021-10-11
fix: if -> id

zhennann:


docs/gitbook/api/bullmq.queue.remove.md
```diff
@@ -16,7 +16,7 @@ remove(jobId: string): Promise<any>;
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  jobId | string | The if of the job to remove |
+|  jobId | string | The id of the job to remove |
```
roggervalf, line 19: this file is auto-generated, with the another change should be fine.

PR #809, opened 2021-10-12
fix(queue-getters): only getting the first 2 jobs

ATheCoder:
I believe the functions that are supposed to get the jobs that are in a queue had the wrong default argument value set causing the functions to get only the first two jobs in a given state.

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -102,27 +102,27 @@ export class QueueGetters extends QueueBase {
     return this.getJobCountByTypes('waiting-children', 'paused');
   }
 
-  getWaiting(start = 0, end = 1) {
+  getWaiting(start = 0, end = -1) {
     return this.getJobs(['waiting'], start, end, true);
   }
 
-  getWaitingChildren(start = 0, end = 1) {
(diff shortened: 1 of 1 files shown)


ATheCoder:
Seems like the coverage is not enough, should I add tests for these changes?

manast:
Looks good. What can be done is to check that the data of every returned job actually matches the data of the added jobs, they should be returned in the correct order as well.

src/classes/queue-getters.ts
```diff
@@ -102,27 +102,27 @@ export class QueueGetters extends QueueBase {
...
   }
 
-  getDelayed(start = 0, end = 1) {
+  getDelayed(start = 0, end = -1) {
     return this.getJobs(['delayed'], start, end, true);
   }
 
-  getCompleted(start = 0, end = 1) {
+  getCompleted(start = 0, end = -1) {
     return this.getJobs(['completed'], start, end, false);
```
ATheCoder, line 122: For completed and failed jobs the order is not ascending because the `asc` parameter is `false`.

Changing this can be a major change, so I didn't do it.

Issue #808, opened 2021-10-12
bullmq can't  install  in Node 16.x


sharper:
msgpack@1.0.3  module can't  complie in Node 16.x

PR #810, opened 2021-10-12
fix(msgpack): switch msgpack dependency to msgpackr

Niek:
Fixes: https://github.com/taskforcesh/bullmq/issues/808


package.json
```diff
@@ -86,6 +85,7 @@
     "eslint-plugin-promise": "4.2.1",
     "eslint-plugin-tsdoc": "^0.2.14",
     "husky": "^3.1.0",
+    "is-extglob": "^2.1.1",
```
manast, line 88: why is this dependency required?

manast:
Thanks. The problem is that msgpackr will not work as direct replacement due to how to handles undefined properties.

manast:
As soon as you use the Packr constructor it generates incompatible msgpack data:
```ts
import { Packr } from 'msgpackr';
import { unpack } from 'msgpack';

const packer = new Packr();
const b = packer.pack({});

console.log(packer.unpack(b), unpack(b));
```
Will yield 
```
/Users/manuelastudillo/Dev/taskforce/bullmq/src/classes/scripts.ts:14
console.log(packer.unpack(b), unpack(b));
                              ^
TypeError: Encountered unknown MesssagePack object type
    at Object.<anonymous> (/Users/manuelastudillo/Dev/taskforce/bullmq/src/classes/scripts.ts:14:31)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Module.m._compile (/Users/manuelastudillo/Dev/taskforce/bullmq/node_modules/ts-mocha/node_modules/ts-node/src/index.ts:439:23)
```

So the only way to use msgpackr today to generate compatible msgpack binary data is by making sure you do not have any undefined fields in the objects you try to encode.

manast:
Closing since I managed to get it working with msgpackr.

Issue #851, opened 2021-11-01
Repeatable Flows - only parent job get's repeated


humont:
# Problem

I seem to be having an issue with repeatable flows where only the parent job will repeat, but any dependencies get lost and never run again.

# Ideal solution

Ideally, I think we should be able to set the entire flow to repeat as a complete unit, all dependencies included.

# Code Example

I have tried the code below with the repeat set on only the child, only the parent, or both child and parent jobs, and in all cases, on the parent job repeats.

```typescript
const flowProducer = new FlowProducer({connection: redisConn);

flowProducer.add({
  name: 'parent job',
  queueName: '...somename',
  data: 'someData',
  opts: {repeat: {cron: "10 * * * * *"}},
  children: [
    {name: 'child job',
    queueName: '...somename',
    data: 'someData',
    opts: {repeat: {cron: "10 * * * * *"}}}
  ]
});
```

manast:
This not supported att the moment, and it is going to be quite complex to implement actually so probable will not be supported in the foreseeable future.

manast:
@roggervalf the best would be to remove "repeat" option at least from the children and maybe throw an exception if using both repeat opts in the parent and the children option at the same time.

manast:
Even better maybe it is possible to use this https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html instead of throwing an exception in runtime.

humont:
Thanks for the quick answers.

My solution has been to create a Job which queues up the flow. The job itself has a cron schedule. So far this seems to be working fine in my testing.

I would agree that `repeat` options should be removed from flows (parent *and* children) - since their entire purpose is to have conditional task completion, having the ability to re-run _just_ the parent seems confusing!

manast:
> My solution has been to create a Job which queues up the flow. The job itself has a cron schedule. So far this seems to be working fine in my testing.

That is a perfectly fine solution and I have used it myself aswell.

PR #853, opened 2021-11-02
fix(flow): remove repeat option from children

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/851

### DIFF
--- a/src/interfaces/flow-job.ts
+++ b/src/interfaces/flow-job.ts
@@ -6,7 +6,7 @@ export interface FlowJob {
   queueName: string;
   data?: any;
   prefix?: string;
-  opts?: Omit<JobsOptions, 'parent'>;
+  opts?: Omit<JobsOptions, 'parent' | 'repeat'>;
   children?: FlowJob[];
 }
 


src/interfaces/flow-job.ts
```diff
@@ -1,13 +1,25 @@
...
+  opts?: Omit<JobsOptions, 'parent' | 'repeat'>;
+  children?: FlowChildJob[];
+}
+
+export interface FlowJob extends BaseFlowNode {
   name: string;
   queueName: string;
   data?: any;
   prefix?: string;
   opts?: Omit<JobsOptions, 'parent'>;
```
manast, line 21: hmm, here we still allow repeat, is it correct?

roggervalf: I was understanding from this comment https://github.com/taskforcesh/bullmq/issues/851#issuecomment-957025297 that parent could have repeat option, but children not :thinking:

manast: No, what I meant is that as soon as you have parent/child properties in options we should not allow also having repeat options, as they need to be mutually exclusive, either you use repeat or you use flows but not both.

roggervalf: Got it, I'll update this fix

manast (approved):

PR #855, opened 2021-11-03
refactor(job): save parent data as object

roggervalf:
parentKey should be a get property

BREAKING CHANGE: save parent info as object (id, queueKey)

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -47,14 +47,14 @@ export interface JobJsonRaw {
   failedReason: string;
   stacktrace: string[];
   returnvalue: string;
-  parentKey?: string;
+  parent?: string;
 }
 
 export interface MoveToChildrenOpts {
   timestamp?: number;
   child?: {
     id: string;
-    queue: string;
+    queueKey: string;
   };
 }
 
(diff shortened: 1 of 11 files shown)


src/classes/job.ts
```diff
@@ -47,14 +47,14 @@ export interface JobJsonRaw {
   failedReason: string;
   stacktrace: string[];
   returnvalue: string;
-  parentKey?: string;
+  parent?: string;
```
manast, line 50: how is parent represented here?

manast (commented):
I guess we also need to update the examples in the documentation to reflect these changes.

roggervalf:
> I guess we also need to update the examples in the documentation to reflect these changes.

Yes, I was forgetting that, thank you

src/classes/job.ts
roggervalf: It's a string that was encoded by cjson, this string should represent an object containing id and queueKey attributes

manast: Ok, I am trying to think how could we make it more explicit then, so that we can know what data it holds just by looking at the interface.

manast: So, ok, it is a json encoded string. And it represents both parentId and queueKey? what does queue key mean here?

roggervalf: Yep, it represents parentId and parentQueueKey, as it is part of parent property I think we are good just having id and queueKey as names

manast: would it be possible to just store the parentKey since parentQueueKey is actually a substring of the parentKey?

roggervalf: we were saving that value but sometimes we should separate those values in order to check the parent queue, so probably would be more helpful (no more have to separate those values any time we need to) for us to have it just separated

PR #857, opened 2021-11-04
perf(added): remove data and opts from event

roggervalf:
BREAKING CHANGE: remove data and opts from added event


manast:
Something is not right with this PR, there are a lot of changes from autogenerated doc files.

roggervalf:
> Something is not right with this PR, there are a lot of changes from autogenerated doc files.

yeah, that was because next is not updated with the last master version

manast (commented):
Just remember that the text: 

BREAKING CHANGE: remove data and opts from added event


Must be on the commit itself in a new line in order to be picked up by semantic release https://github.com/semantic-release/semantic-release#commit-message-format

manast (approved):

PR #870, opened 2021-11-10
fix: allow lazy redis connections

arbourd:
#642 introduced a change that force connected lazy connections on boot of BullMQ. In this [comment](https://github.com/taskforcesh/bullmq/pull/657#discussion_r728602683), the suggestion is to essentially undo that by checking to see if `lazyConnect: true`.

In the case of #642, which appears to be introduced to satisfy unit tests, perhaps the solution would be not lazily connecting for unit tests?

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -85,7 +85,7 @@ export class RedisConnection extends EventEmitter {
       return;
     }
 
-    if (client.status === 'wait') {
+    if (client.status === 'wait' && !client.options.lazyConnect) {
       return client.connect();
     }
 


manast:
I am a bit reluctant to this kind of changes because messing with the connection settings can lead to queues that stop processing jobs and currently with the latest changes it is working quite well.

arbourd:
This was the suggestion from your comment though, yes? 

Are you saying you'd prefer that no lazy ioredis connections is preferred?

manast:
> This was the suggestion from your comment though, yes?
> 
> Are you saying you'd prefer that no lazy ioredis connections is preferred?

Ah ok. I forgot that already. Since we had so many problems regarding connections lately I was scared this could introduce a new problem.

manast:
But I also added this in my comment: "Also I wonder if it is possible to end in the "wait" state after having been connected to redis?, if so we should not only check if lazyConnect is true but also if we have been connected already..."

arbourd:
Okay I can look into that and see if it's possible.

PR #881, opened 2021-11-18
fix(compat): remove Queue3 class

roggervalf:
do not support compatibility class

BREAKING CHANGE: do not support compatibility class Queue3


manast (commented):
Do you know why we are getting all the changes on the apidocumentation files?

manast (approved):

PR #884, opened 2021-11-20
fix(job): use this when use new statements

roggervalf:


src/classes/job.ts
```diff
@@ -175,7 +175,7 @@ export class Job<
   ) {
     const client = await queue.client;
 
-    const job = new Job<T, R, N>(queue, name, data, opts, opts && opts.jobId);
+    const job = new this<T, R, N>(queue, name, data, opts, opts && opts.jobId);
```
manast, line 178: This was a really interesting solution...

Issue #883, opened 2021-11-19
RFC: Script Loader


ccollie:
## RFC: scriptLoader
A replacement for the script loading mechanism in `bullmq`

See WIP branch [here](https://github.com/ccollie/bullmq/tree/script-loader) and tests [here](https://github.com/ccollie/bullmq/blob/script-loader/src/test/test_script-loader.ts)

### Why ?
To allow modularity and composability of `lua` scripts, with recursive dependency handling and path mapping.

#### Example

`isNil.lua`
```lua
local function isNil(x)
  return x == nil or x == cjson.null
end
```

`string.lua`
```lua
--- @include "isNil"
local function safeLen(x)
  return not isNil(x) and #x or 0
end
```

`math.lua`
```lua
--- @include "isNil"
local function sign(x)
  if isNil(x) then return nil end
  if (x == 0) then return 0 end
  return x < 0 and -1 or 1
end
```

`utils.lua`
```lua
--- @include "string"
--- @include "math"
```

`app.js`
```javascript
const script = await loadScript('path/to/utils.lua')
```

`result`
```lua
local function isNil(x)
  return x == nil or x == cjson.null
end
local function safeLen(x)
  return not isNil(x) and #x or 0
end
local function sign(x)
  if isNil(x) then return nil end
  if (x == 0) then return 0 end
  return x < 0 and -1 or 1
end
```

All children are loaded transitively and in dependency order to any depth.

Path mapping is supported, so we can do the following:

`utils.lua`
```lua
--- @include "<includes>/math"
--- @include "<includes>/string"
--- @include "<includes>/jobs"
```
`app.ts`
```javascript
addScriptPathMapping('includes', 'relative/path/to/includes');
const script = await loadScript('utils.lua');
```
The `~` character at the beginning of an include maps to the project root (location of the project's `package.json`)

For convenience, we can do `glob` includes

`utils.lua`
```lua
--- @include "<includes>/util-*.lua"
```

## loadScript() function

<b>Signature:</b>

```typescript
export declare function loadScript(filePath: string, cache?: Map<string, ScriptInfo>): string;
```

## Parameters

|  Parameter | Type | Description |
|  --- | --- | --- |
|  filePath | string | path to the top-level script we want to load |
|  cache | Map<string, ScriptInfo> | an optional map to return the metadata or all files gathered to fulfil the request |

<b>Returns:</b>

string

The interpolated script

## addScriptPathMapping() function

[...]

PR #893, opened 2021-11-25
feat: new lua script loader with enhanced include support

ccollie:
See https://github.com/taskforcesh/bullmq/issues/883


ccollie:
There's a bit of git funkiness happening. I'll fix and create an updated PR.

PR #897, opened 2021-11-26
feat(script-loader): lua script loader with include support

ccollie:


src/commands/index.ts
```diff
@@ -1,58 +1,444 @@
...
+  name: string;
+  numberOfKeys?: number;
+} {
+  const longName = path.basename(filePath, '.lua');
+  const [name, num] = longName.split('-');
+  const numberOfKeys = num && parseInt(num, 10);
+  return { name, numberOfKeys };
+}
+
+async function handleGlob(pattern: string): Promise<string[]> {
```
manast, line 228: Maybe a more descriptive name would make it easier to understand the intent of this function.

src/commands/index.ts
```diff
@@ -1,58 +1,444 @@
...
+}
+
+/**
+ * Recursively collect all scripts included in a file
+ * @param file - the parent file
+ * @param cache - a cache for file metadata to increase efficiency. Since a file can be included
+ * multiple times, we make sure to load it only once.
+ * @param stack - internal stack to prevent circular references
+ */
+async function resolveDependencies(
```
manast, line 243: I think it would be easier to understand if this function had well-defined inputs and outputs. I have not managed to understand the whole function yet so I am not sure if I am missing other side effects than ones on the cache input.

manast (commented):
This is a great PR with functionality that will enable us to vastly refactor and improve all the lua scripts. I wrote some comments, looks very good overall, but I think it would be better if we could get rid of the module globals and instead of having functions that affect this globals as side-effects, put them in a class and have well defined methods on it.

src/commands/index.ts
```diff
@@ -1,58 +1,444 @@
...
+import { RedisClient } from 'src/interfaces';
+import { promisify } from 'util';
+
+const readFile = promisify(fs.readFile);
+const readdir = promisify(fs.readdir);
 
-const readdir = util.promisify(fs.readdir);
-const readFile = util.promisify(fs.readFile);
-const exists = util.promisify(fs.exists);
+const GLOB_OPTS = { dot: true, silent: false };
```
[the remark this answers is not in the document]
ccollie: Old time habits from `c` :-)

src/commands/index.ts
ccollie: The `cache` and `stack` are set initially in `parseScript`. resolveDependencies then passes them to itself.
`file` is the metadata of the Lua script we’re currently parsing. resolveDependencies simply parses its list of includes recursively and adds them to file.includes. It does not modify any global state.

ccollie:
So I refactored the code into a class: `ScriptLoader` with the appropriate methods exposed. The question is how to expose it to the outside. Currently in `commands/index` there's the `load` function, which we can replace easily with something like,

```javascript
async function load(client) {
  return new ScriptLoader().load(client);
}
``` 

but if we want to support path mapping we'd probably want a singleton instance of `ScriptLoader` and expose the path related methods as functions.

Thoughts ?

manast:
Yes, I think exporting an instance of the ScripLoader class is the way to go here. 

As a related note, It may be of interest that after reading your comment it struck me that currently, we are loading all the lua scripts per every connection to Redis, even if it is to the same queue and in the same Queue or Worker class instance, so we are not reusing them properly, which maybe it is not a big deal but I never thought that this is not very optimal. The question is if there is any way to avoid reloading the commands at least if we want to keep using ioredis mechanism for loading lua scripts... anyway, nothing that affects this PR at all, I just wanted to share this insight with you.

ccollie:
Sounds good.

r.e. efficiency, thats one of the reasons I allowed using a `cache` parameter to minimize file system access on subsequent requests. If we move to a singleton, a `Command` cache is trivial to add - the issue would be the key (i imagine it would be directory name).

docs/gitbook/changelog.md
```diff
@@ -1,3 +1,10 @@
+## [1.54.5](https://github.com/taskforcesh/bullmq/compare/v1.54.4...v1.54.5) (2021-11-26)
```
manast, line 1: hmm, something must have gone wrong because this change should not appear here...

manast:
Actually I am not sure, but I guess if we load the same script into the same redis instance it would just replace the previous one as long as the sha1 key is the same, however the question is how do we actually clean old scripts that are not in use anymore? (which should happen as soon as you make a change in a script). There is this "script flush" command, but it does not seem safe at all to use without introducing hazards.

docs/gitbook/changelog.md
ccollie: Not sure what caused that  ... Seems like something triggered by `CI`

ccollie:
> but it does not seem safe at all to use without introducing hazards.

I don't know of any way of cleaning old scripts safely without downtime. What do you do with already connected clients, for instance ? They will be pointing to old `sha1` keys on any script change so flushing would cause requests to fail.

manast:
> I don't know of any way of cleaning old scripts safely without downtime. What do you do with already connected clients, for instance ? They will be pointing to old `sha1` keys on any script change so flushing would cause requests to fail.

Exactly. It is just imposible to do in practice... there is a new functionality in Redis roadmap, called FUNCTIONS that maybe improves this aspect of scripting.

ccollie:
I already see improvements for version 2.0. In particular there are a few places where code depends on the `redis` version, and I'd like to avoid what's done r.e. `getState`. So future state would look like:

```lua
--[[
      Check if an id exists in a redis list.
      Input:
        KEYS[1]   key
        ARGV[1]   job id
      Output:
        1 if id is in any of the lists specified by KEYS
        0 otherwise
]]
local function isInList (key, item)
-- not exact, but you get the idea
--- @if VERSION > 6.2.0.  
 return rcall("LPOS", k, v) ~= false         
--- @else 
  local list = rcall("LRANGE", key, 0, -1)
  for _, v in pairs(list) do
     if v == item then
         return true
     end
  end
  return false
--- @endif 
end
```

then the client's `redis` version (and other context) is passed as context to `scriptLoader.load`

ccollie:
Looks like some funkiness with an extraneous `gitbook` commit again

roggervalf:
@ccollie is your pr ready or there will be more changes? it LGTM

ccollie:
@roggervalf No more changes for this version, so its a go from my end.

roggervalf (approved):

lukepolo:
this seemed to cause an issue 

```
(node:935) UnhandledPromiseRejectionWarning: TypeError: n.paths is not iterable
    at getPkgJsonDir (/opt/qumulex/dist/webpack:/node_modules/bullmq/dist/esm/commands/script-loader.js:346:34)
    at new ScriptLoader (/opt/qumulex/dist/webpack:/node_modules/bullmq/dist/esm/commands/script-loader.js:38:25)
    at Module.call (/opt/qumulex/dist/webpack:/node_modules/bullmq/dist/esm/commands/index.js:3:22)
    at __webpack_require__ (/opt/qumulex/dist/webpack:/webpack/bootstrap:19:32)
    ```

**Comment by manast** [comment]
@lukepolo can you provide more context, please?

**Comment by lukepolo** [comment]
I’ll get more details here soon , it’s lost likely just issue and not bulls

**Comment by lukepolo** [comment]
```
// Determine the project root
// https://stackoverflow.com/a/18721515
function getPkgJsonDir() {
    console.info("module", module)
    for (const modPath of module.paths) {
```


module is `undefined`

manast:
Can we close this now or is it still an issue?

lukepolo:
its closed~

lukepolo:
fixed via #897

PR #904, opened 2021-12-01
style: fix styled github warnings

roggervalf:


src/classes/child-pool.ts
```diff
@@ -2,14 +2,10 @@ import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
 import { flatten } from 'lodash';
 import * as getPort from 'get-port';
-import * as fs from 'fs';
-import { promisify } from 'util';
 import { killAsync } from './process-utils';
 import { ParentCommand, ChildCommand } from '../interfaces';
 import { parentSend } from '../utils';
 
-const stat = promisify(fs.stat);
```
roggervalf, line 11: this variable is not used

manast: should not be needed, and also there is ```fs.promises``` so no need for promisification anymore.

PR #916, opened 2021-12-05
feat(worker): make moveToActive protected

manast:


tests/fixtures/delay.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
roggervalf, line 1: If the submodules is deleted, how we are manage the tests sync with bullmq-pro? 🤔

manast: I am not sure exactly how, I will experiment having bullmq as a submodule instead.

PR #919, opened 2021-12-08
feat(queue-events): add better event typing

roggervalf:
ref https://rjzaworski.com/2019/10/event-emitters-in-typescript

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -780,8 +780,8 @@ export class Job<
       const completedEvent = `completed:${jobId}`;
       const failedEvent = `failed:${jobId}`;
 
-      queueEvents.on(completedEvent, onCompleted);
-      queueEvents.on(failedEvent, onFailed);
+      queueEvents.on(completedEvent as any, onCompleted);
+      queueEvents.on(failedEvent as any, onFailed);
       this.queue.on('closing', onFailed);
 
       const removeListeners = () => {
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -9,7 +9,7 @@ import { DELAY_TIME_5 } from '../utils';
 import { QueueBase } from './queue-base';
 import { RedisConnection } from './redis-connection';
 
-export interface QueueEventsList {
+export interface QueueEventsListener {
   /**
    * Listen to 'active' event.
    *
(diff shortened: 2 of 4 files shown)


tests/test_script_loader.ts
```diff
@@ -38,11 +38,6 @@ describe('scriptLoader', () => {
   });
 
   describe('when using path mappings', () => {
-    it('correctly determines the path to the project root', () => {
```
roggervalf, line 41: this test is deleted because rootPath is a private property

src/classes/queue-events.ts
```diff
@@ -9,88 +9,80 @@ import { DELAY_TIME_5 } from '../utils';
 import { QueueBase } from './queue-base';
 import { RedisConnection } from './redis-connection';
 
-export interface QueueEventsDeclaration {
+export interface QueueEventsList {
```
manast, line 12: I would call this interface: "QueueEventsListener" which in my opinion represents better what this interface is.

src/classes/queue-events.ts
```diff
@@ -99,90 +91,61 @@ export interface QueueEventsDeclaration {
...
-   * @param {'waiting-children'} event
-   * @callback listener
    */
-  on(
-    event: 'waiting-children',
-    listener: (args: { jobId: string }, id: string) => void,
+  'waiting-children': (args: { jobId: string }, id: string) => void;
+}
+
+export interface QueueEventsDeclaration {
```
manast, line 139: Don't we also need a "off" or "removeEventListener" declaration?

src/classes/job.ts
```diff
@@ -780,8 +780,8 @@ export class Job<
       const completedEvent = `completed:${jobId}`;
       const failedEvent = `failed:${jobId}`;
 
-      queueEvents.on(completedEvent, onCompleted);
-      queueEvents.on(failedEvent, onFailed);
+      queueEvents.on(completedEvent as any, onCompleted);
```
manast, line 783: This is not super nice but I guess there are not so many better options...

src/classes/queue-events.ts
```diff
@@ -195,6 +158,29 @@ export interface QueueEventsDeclaration {
 export class QueueEvents extends QueueBase implements QueueEventsDeclaration {
   private running = false;
 
+  emit<U extends keyof QueueEventsList>(
```
manast, line 161: If we add these methods, why do we need this declaration?

```QueueEventsDeclaration```

roggervalf: yeah we don't need QueueEventsDeclaration in this case

src/classes/job.ts
roggervalf: yep, as now we are using static events, there is no glob version for add these cases, I tried with [key: string] but it does not work either

manast (approved):

PR #929, opened 2021-12-13
fix(redis-connection): allow undef client.options

boredland:
In some scenarios (ioredis-mock for example) the `client.options` will be null. Due to https://github.com/taskforcesh/bullmq/pull/927, bullmq isn't checking `opts.maxRetriesPerRequest` anymore. While this should be solved in ioredis-mock eventually (currently "client" is missing [here](https://github.com/stipsan/ioredis-mock/blob/master/compat.md#supported-commands-)), making this optional here could be fine enough.

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -52,8 +52,8 @@ export class RedisConnection extends EventEmitter {
       this._client = <RedisClient>opts;
       this.checkOptions(deprecationMessage, this._client.options);
       if (
-        (<RedisOptions>this._client.options).maxRetriesPerRequest ||
-        this._client.options.enableReadyCheck
+        (<RedisOptions>this._client.options)?.maxRetriesPerRequest ||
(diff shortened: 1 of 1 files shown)


manast:
Thanks for the PR. I am a bit reluctant to merge this since I am not a huge fan of defensive programming. The typescript typings should guarantee that options are defined and we should not need to rely on checking ourselves if they are or not...

boredland:
> Thanks for the PR. I am a bit reluctant to merge this since I am not a huge fan of defensive programming. The typescript typings should guarantee that options are defined and we should not need to rely on checking ourselves if they are or not...

All good points. I guess then ioredis-mock isn't realistic any time soon. I'll find sth else.

manast:
@boredland wouldn't be quite easy for ioredis-mock to also mock those two options?

boredland:
probably. there even is a related PR open. but ya, found a workaround (mocking bullmq instead of redis 🤷🏼)

PR #934, opened 2021-12-14
fix(redis-connection): consider cluster redisOptions config

roggervalf:


src/classes/redis-connection.ts
```diff
@@ -68,7 +70,10 @@ export class RedisConnection extends EventEmitter {
   }
 
   private checkOptions(msg: string, options?: RedisOptions) {
-    if (options && (options.maxRetriesPerRequest || options.enableReadyCheck)) {
+    if (
+      options &&
+      ((<IORedis.RedisOptions>options)?.maxRetriesPerRequest || options.enableReadyCheck)
```
manast, line 75: why is needed this type coercion here if options is already of the RedisOptions type?

roggervalf: it was because in line 72 I was using RedisOptions, it should be ioredis.redisOptions to not have complains

manast (approved):

PR #935, opened 2021-12-14
Feat/allow sandbox extensions

manast:


manast:
Closing as we have much better code for handling child processes now.

PR #954, opened 2021-12-17
fix(package): add exports prop

roggervalf:


manast:
I do not think this is needed anymore now that we are exporting both CJS and ESM. The only downside seems to be if the end-user mix requires and imports to BullMQ in the same project, but I would argue it is not really our fault.

PR #958, opened 2021-12-21
Feat transform commands

roggervalf:


commands/addJob-9.lua
```diff
@@ -56,7 +56,7 @@ local parentQueueKey
 local parentData
 
 -- Includes
---- @include "includes/destructureJobKey"
+--- @include "destructureJobKey"
```
manast, line 59: I think it is good to keep the underlying structure, it is a pity to only support a global flat directory structure.

ccollie: From my experience, a flat structure can get unruly. My `includes` folder currently has 50 files, and I'd imagine that over time "includes" could greatly outnumber main scripts in `bullmq`.

ccollie: Another thought: I was planning on refactoring some of the scripts to be more reusable - for example `removeJob-1.lua` (to implement `removeJobsByPattern`). I think it would be confusing to have a main script and its utility file named similarly in the same folder.

roggervalf: Sounds good to me, I'll try to transform our scripts keeping the folder structure in some way

package.json
```diff
@@ -42,14 +42,19 @@
...
+    "transform:test": "npm-run-all transform:test:commands transform:test:includes",
     "tsc:all": "tsc && tsc -p tsconfig-cjs.json"
   },
   "dependencies": {
     "cron-parser": "^2.18.0",
     "get-port": "^5.1.1",
-    "glob": "^7.2.0",
     "ioredis": "^4.28.2",
     "lodash": "^4.17.21",
+    "minimatch": "^3.0.4",
```
manast, line 57: What does minimatch provide that cannot be achieved with glob?
The thing with glob vs minimatch is that glob seems to be much better mantained and used

roggervalf: minimatch is used in glob https://github.com/isaacs/node-glob/blob/master/package.json#L23 and both libraries belongs to the same maintainer too

PR #959, opened 2021-12-21
fix(worker): fix undefined moveToActive

### LINKED ISSUE #946: v1.57.4 client.moveToActive is not a function
Series issue with change client.moveToActive is not a function on recent upgrade.

TypeError: client.moveToActive is not a function
16/12/2021 13:32:26    at Function.moveToActive (/lokl/node_modules/bullmq/src/classes/scripts.ts:684:40)
16/12/2021 13:32:26    at processTicksAndRejections (internal/process/task_queues.js:95:5)

We have had to rollback our servers.

manast:
Fixes #946

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -309,6 +309,7 @@ export class Worker<
           while (!this.closing) {
             if (processing.size < opts.concurrency) {
               const token = tokens.pop();
+
               processing.set(
                 this.retryIfFailed<Job<any, any, string>>(
                   () => this.getNextJob(token),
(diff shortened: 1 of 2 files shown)


src/classes/worker.ts
```diff
@@ -694,6 +689,7 @@ export class Worker<
       try {
         return await fn();
       } catch (err) {
+        this.emit('error', <Error>err);
```
roggervalf, line 692: Should we check if this is a connection error here?

roggervalf (approved):
LGTM just one question

PR #965, opened 2021-12-24
feat(worker): save worker name into client

### LINKED ISSUE #456: [Feature request] Assign meaningful names to workers
To keep a consistent overview of my current workers, I would like to assign each of them a meaningful name, the hostname of the computer they are running on for example.
That way, when calling `queue.getWorkers()` I can tell which machine the worker is running on.
Trying to set the `worker.name` property changes the entire queue that the worker is listening to, unfortunately.
And checking `queue.getWorkers()...name` returns the name of the queue, which I already know.

roggervalf:
separate queueName from worker name

BREAKING CHANGE: name property is renamed to queueName and name is passed into opts

fix #456

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -247,8 +247,10 @@ export class QueueGetters extends QueueBase {
         client[key] = value;
       });
       const name = client['name'];
-      if (name && name === this.clientName()) {
-        client['name'] = this.name;
+      const clientName = this.clientName();
+      if (name && name.startsWith(clientName)) {
+        const workerName = name.substring(clientName.length + 1);
+        client['name'] = workerName ? workerName : this.name;
         clients.push(client);
       }
     });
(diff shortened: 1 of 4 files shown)


src/classes/queue-getters.ts
```diff
@@ -247,8 +247,10 @@ export class QueueGetters extends QueueBase {
         client[key] = value;
       });
       const name = client['name'];
-      if (name && name === this.clientName()) {
-        client['name'] = this.name;
+      const clientName = this.clientName();
+      if (name && name.startsWith(clientName)) {
+        const workerName = name.substring(clientName.length + 1);
```
manast, line 252: wouldn't workerName be the same as clientName at this point? (i.e. no need to do a substring of name unless I am missing something).

roggervalf: no really, clientName would be bull:{base64QueueName} and substring of name when passing just a number would bring all the characters after that index, so in this case only the workerName in case exists

manast (approved):

PR #976, opened 2021-12-31
docs(flow): add getFlow into guide structure

roggervalf:


docs/gitbook/guide/flows/get-flow-tree.md
```diff
@@ -0,0 +1,65 @@
...
+const tree = await flow.getFlow({
+  id: topJob.id,
+  queueName: 'topQueueName',
+});
+
+const { children, job } = tree;
+```
+
+{% hint style="info" %}
+Each children may have a job property and in case they have children as well, they would have children property
```
ccollie, line 51: Should be `Each _child_` instead of `Each children`

roggervalf: Nice @ccollie for catching it

PR #990, opened 2022-01-14
fix(remove-job): consider removing parent dependency key in lua scripts

roggervalf:


ccollie (commented):
@roggervalf make sure to include the `destructureJobKey` script in `removeParentDependencyKey` to ensure that it is self-contained.

roggervalf:
> @roggervalf make sure to include the `destructureJobKey` script in `removeParentDependencyKey` to ensure that it is self-contained.

thanks @ccollie I had that fix in local, I was waiting to commit it with my last commit xd

PR #996, opened 2022-01-18
Added TTL description on .waitUntilFinished()

sublime93:


docs/gitbook/api/bullmq.job.waituntilfinished.md
```diff
@@ -17,7 +17,7 @@ waitUntilFinished(queueEvents: QueueEvents, ttl?: number): Promise<ReturnType>;
 |  Parameter | Type | Description |
 |  --- | --- | --- |
 |  queueEvents | [QueueEvents](./bullmq.queueevents.md) |  |
-|  ttl | number |  |
+|  ttl | number | Time in milliseconds to wait for job to finish before timing out. |
```
roggervalf, line 20: hi @sublime93 thanks for the pr, we should not change docs files directly, these files are generated, the description should be placed in https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L732 following ts-docs format i.e: https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L803

roggervalf:
this change is on version 1.64.4

PR #1005, opened 2022-01-19
Fix reduce initial value

majidsajadi:
if you don't provide `type` to `getJobCountByTypes` get an error would be thrown.

```bash
TypeError: Reduce of empty array with no initial value
```

I guess we can validate the parameters. but in `getJobCount` if you don't provide `type` an empty object is returned. so I guess this patch is acceptable.


roggervalf (approved):
LGTM, thank you @majidsajadi

Issue #1002, opened 2022-01-19
Improve the logic of "attemptsMade"


manast:
# Introduction

Currently "attemptsMade" is coupled to the retry mechanism, as it is only incremented when a retry has been issued. This is semantically awkward as in that this number does not really represent the number of times a job has been attempted to be processed, for example, a worker may die causing its current job to stall. This job will be picked up later by a worker and effectively-being processed a second time, but "attemptsMade" will still be 0.

# Requirements

The "attemptsMade" field should be incremented when and only a job is moved to active. This will better reflect its semantical intent and also be more efficient with functions like "retry all failed jobs", which otherwise would require to update the "attemptsMade" field for every job, making it much more slower.

Since attemptsMade will now be 1 after the first time a job is processed, the retry logic must also take this change into account so that jobs are retried automatically at the max number specified in the "attempts" settings. Note that this change is not 100% backwards compatible, since an older job processed with a precious version of BullMQ will be retried 1 more time with the new code, but I do not think this slight change motivates a major version release, a minor would be OK IMHO.

PR #1009, opened 2022-01-20
fix(job): increase attemptsMade when moving job to active

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1002


manast (approved):
looks great

tests/test_worker.ts
```diff
@@ -1560,7 +1560,7 @@ describe('workers', function () {
       await new Promise<void>(resolve => {
         worker.on('completed', () => {
           const elapse = Date.now() - start;
-          expect(elapse).to.be.greaterThan(2000);
+          expect(elapse).to.be.greaterThan(1000);
```
manast, line 1563: Why is this reduction in time needed?

roggervalf: because, the first time is moved to active, attemptsMade is set as 1, before it was 0 the first time

manast (approved):

PR #1010, opened 2022-01-20
feat(queue-events): add retries-exhausted event

roggervalf:


src/commands/moveToFinished-8.lua
```diff
@@ -132,7 +132,7 @@ if rcall("EXISTS",jobIdKey) == 1 then -- // Make sure job exists
     rcall("XADD", KEYS[6], "*", "event", ARGV[5], "jobId", jobId, ARGV[3],
           ARGV[4])
 
-    if ARGV[15] then
+    if ARGV[15] > 0 then
```
manast, line 135: So this will not happen if the job completed instead of failing right?

manast:
LGTM, although I guess we also need to update queue-events with the typings and documentation for this event.

src/classes/scripts.ts
```diff
@@ -300,6 +300,9 @@ export class Scripts {
       target,
       token,
       fetchNext,
+      job.opts.attempts && job.attemptsMade >= job.opts.attempts
```
manast, line 303: this will work mostly, but there would be more robust to do this check in the .lua script itself if possible. If it is awkward or hard we could postpone it, but good to have it documented.

src/commands/moveToFinished-8.lua
roggervalf: I'll verify it

roggervalf:
> LGTM, although I guess we also need to update QueueEvents with the typings and documentation for this event.

I added the event typing

manast (approved):
Lots of changes, but I guess they are not so many in practice so it will look better if doing a rebase.

roggervalf:
> Lots of changes, but I guess they are not so many in practice so it will look better if doing a rebase.

Oh, in reality because I did a rebase from master some of those commits are appearing here, we should be good with a squash, for some reason some of the changes that are already on master are appearing like new changes but they are not

roggervalf:
https://github.com/taskforcesh/bullmq/commit/e476f35f5c3f9b1baf2bbc3d46712b8ba597f73c as reference the real changes

Issue #1006, opened 2022-01-19
Provide state  enum


majidsajadi:
Can we refactor the code to use `JobState` enum (or union type) instead of `string[]` and export it?

Issue #744, opened 2021-09-10
TypeScript type for job status


fmoessle:
The "old" **bull** package included a type for the status of a job:

```typescript
import { JobStatus } from 'bull'
```

As I see no such type exists in **bullmq**.

Is this intentional?

roggervalf:
we currently have JobState union in 1.65.0. Thanks for reporting this

PR #1011, opened 2022-01-20
feat(queue): add JobTypes for better typing

roggervalf:
fix https://github.com/taskforcesh/bullmq/issues/1006 https://github.com/taskforcesh/bullmq/issues/744

### DIFF
--- a/src/classes/compat.ts
+++ b/src/classes/compat.ts
@@ -20,7 +20,7 @@
 // TypeScript Version: 2.8
 
 import { EventEmitter } from 'events';
-import { JobTypes } from '../types';
+import { JobType } from '../types';
 import { Job } from './job';
 import { QueueScheduler } from './queue-scheduler';
 import { Queue } from './queue';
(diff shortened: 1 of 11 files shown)


src/classes/compat.ts
```diff
@@ -312,7 +313,7 @@ export class Queue3<T = any> extends EventEmitter {
    * Optional parameters for range and ordering are provided.
    */
   getJobs(
-    types: string[] | string,
+    types: JobTypes[] | JobTypes,
```
manast, line 316: Just a small thing, normally a class, enum, or interface should be in the singular form.

manast (approved):

majidsajadi (commented):
i guess we can use `JobType` in `getState` in `queue-base.ts` too? also `Scripts.getState`. 
care that there is an `unknown`  type too.

roggervalf:
> i guess we can use `JobType` in `getState` in `queue-base.ts` too? also `Scripts.getState`. care that there is an `unknown` type too.

Good idea, I updated it

PR #1013, opened 2022-01-20
dont loop through empty modules path

### LINKED ISSUE #1012: Webpack Production Build: paths is not iterable
During booting of the app we get an error ` n.paths is not iterable` , which points to this function below 

```
// Determine the project root
// https://stackoverflow.com/a/18721515
function getPkgJsonDir() {
    console.info("module", module)
    for (const modPath of module.paths) {    
```

module is {` children: [], exports: [Setter] } ` and seems to be missing the  `paths key`

lukepolo:
if the modules paths is empty don't loop through them fixes #1012

### DIFF
--- a/src/commands/script-loader.ts
+++ b/src/commands/script-loader.ts
@@ -484,10 +484,7 @@ async function getFilenamesByPattern(pattern: string): Promise<string[]> {
 // Determine the project root
 // https://stackoverflow.com/a/18721515
 function getPkgJsonDir(): string {
-  if(!module.paths) {
-    return;
-  }
-  for (const modPath of module.paths) {
+  for (const modPath of (module.paths || [])) {
     try {
       const prospectivePkgJsonDir = path.dirname(modPath);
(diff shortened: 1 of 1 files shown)


lukepolo:
PR : https://github.com/taskforcesh/bullmq/pull/1013


before: 

image


after; 

image

src/commands/script-loader.ts
```diff
@@ -484,6 +484,9 @@ async function getFilenamesByPattern(pattern: string): Promise<string[]> {
 // Determine the project root
 // https://stackoverflow.com/a/18721515
 function getPkgJsonDir(): string {
+  if(!module.paths) {
```
roggervalf, line 487: what about:
`for (const modPath of (module.paths||[])) {`

lukepolo: Sure that works too

lukepolo: making sure this still works as expected

roggervalf:
Thanks @lukepolo, it's released

PR #1016, opened 2022-01-25
perf(retry): delete props in retryJob lua script

roggervalf:


src/commands/reprocessJob-4.lua
```diff
@@ -20,6 +21,7 @@ if (rcall("EXISTS", KEYS[1]) == 1) then
   local jobId = ARGV[1]
   if (rcall("ZREM", KEYS[3], jobId) == 1) then
     rcall(ARGV[2], KEYS[4], jobId)
+    rcall("HDEL", KEYS[1], "finishedOn", "processedOn", ARGV[3])
```
roggervalf, line 24: probably we should set attempstMade as 0 too, what do you think @manast ?

manast: Not sure, why would we like to reset the number of attempts here?

roggervalf: if the job failed after the last attempt, if we retry the job, it would keep the last maximum attempt number

roggervalf: as we are not going to reset the attempstMade, this pr should be ready

src/classes/job.ts
```diff
@@ -844,23 +844,16 @@ export class Job<
   /**
    * Attempts to retry the job. Only a job that has failed can be retried.
```
manast, line 845: Since we can actually retry completed jobs, this comment is incorrect right now.

manast (approved):

Issue #1001, opened 2022-01-19
Support for an efficient "retry all" failed jobs


manast:
# Introduction

In many situations, for example, let's say that due to a bug in your worker you end with thousands of failed jobs, the bug has been fixed and now you want to retry all these jobs as fast as possible, it is important then to be able to just retry all the jobs that have failed in an efficient way.

# Requirements

- An atomic method to move all the failed jobs from the failed set to the wait list.
- Preserve the order so that the jobs end in the wait list in the same order as they were failed.
- The retried jobs must be processed before any existing jobs are already in the wait list.

PR #1024, opened 2022-01-26
feat(queue): add retryJobs method

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1001


src/classes/queue.ts
```diff
@@ -361,6 +361,13 @@ export class Queue<
     } while (cursor);
   }
 
+  async retryAllFailedJobs(count?: number): Promise<void> {
```
roggervalf, line 364: probable we could rename this as retryFailedJobs, to be shorter :thinking:

src/commands/retryAllFailedJobs-4.lua
```diff
@@ -0,0 +1,44 @@
...
+
+local function getZSetItems(keyName, max)
+  return rcall('ZRANGE', keyName, 0, max - 1)
+end
+
+local jobs = getZSetItems(KEYS[3], maxCount)
+
+for i, key in ipairs(jobs) do
+  local jobKey = baseKey .. key
+  if (rcall("ZREM", KEYS[3], key) == 1) then
```
manast, line 29: It should not be needed to check that we in fact removed the value from the zset since .lua scripts are atomic and we got the value from a previous call to ZRANGE. By skipping this test we can remove the values much faster in one call (up to 7000 values since there is an issue with more than 8k values in Redis), check how we did it here: https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveStalledJobsToWait-8.lua#L103-L107
Same concept can be applied to LPUSH, however not possible with HDEL and HSET. I think the reset of finishOn, etc could be made when moving the job to active so that it is done lazily and therefore this call will become much faster, what do you think?
Also, regarding attemptsMade, I see there is a point in resetting it, but not sure we must, that would also make this call faster as we do not need to iterate through all the keys one by one.

manast: Sadly we cannot optimize the call to XADD, so that will still need to be iterated...

src/classes/queue.ts
manast: the minimal name would be just "retryFailed", or even just "retry" if in the future we would allow retry also "completed" jobs, which is not completely unlikely, the case being, a lot of jobs completed but they did not do the proper job, so we want to retry all of them.

manast: maybe "retryJobs" then.

tests/test_queue.ts
```diff
@@ -195,4 +196,48 @@ describe('queues', function () {
...
+
+      const failedCount = await queue.getJobCounts('failed');
+      expect(failedCount.failed).to.be.equal(4);
+
+      const completing = new Promise(resolve => {
+        worker.on('completed', after(4, resolve));
+      });
+
+      fail = false;
+      await queue.retryAllFailedJobs(2);
```
manast, line 233: It is also important to check that the order of the retried jobs is correct, as the oldest failed must be the first to be retried.

manast (changes requested):

manast:
Note: we need to backport this feature to Bull aswell.

src/classes/queue.ts
roggervalf: sounds good to me, I will change the count prop to be into an object, so when we introduce the case for completed jobs, a new param could be place into this object

src/commands/retryJobs-4.lua
```diff
@@ -0,0 +1,42 @@
...
+  -- Emit waiting event
+  rcall("XADD", KEYS[2], "*", "event", "waiting", "jobId", key);
+end
+maxCount = maxCount - #jobs
+
+if(maxCount <= 0) then
+  return 1
+end
+
+return 0
```
manast, line 42: Notice that there is no newline at the end.

src/classes/queue.ts
```diff
@@ -361,6 +361,13 @@ export class Queue<
     } while (cursor);
   }
 
+  async retryJobs(opts: { count?: number } = {}): Promise<void> {
```
manast, line 364: We should add a documentation snippet for this method so that it appears in the API documentation.

src/commands/includes/batches.lua
```diff
@@ -0,0 +1,12 @@
+local function batches(n, batchSize)
```
manast, line 1: Nice refactoring

src/commands/retryJobs-4.lua
```diff
@@ -0,0 +1,51 @@
...
+
+for i, key in ipairs(jobs) do
+  local jobKey = baseKey .. key
+  rcall("HDEL", jobKey, "finishedOn", "processedOn", "failedReason")
+
+  -- Emit waiting event
+  rcall("XADD", KEYS[2], "*", "event", "waiting", "jobId", key);
+end
+
+if (#jobs > 0) then
```
manast, line 38: This "if" can also be moved to include also the "ipairs" loop

src/commands/retryJobs-4.lua
```diff
@@ -0,0 +1,51 @@
...
+  for from, to in batches(#jobs, 7000) do
```
manast, line 39: Very nice

manast (commented):
Looks great, just a small comment regarding the documentation of the retryAll method.

manast (approved):

PR #1033, opened 2022-01-28
docs: update documentation of waitUntillFinished to match behavior

### LINKED ISSUE #1031: waitUntillFinished throws error for failed jobs
Hello,
I have the following code. Assume there is some processor that throws an error when processing job
```typescript
const queue = new Queue('Cars');
const events = new QueueEvents('Cars')
const job = await queue.add('paint', { colour: 'red' });
await job.waitUntillFinished(events) // error thrown here
const failed = await job.isFailed();
console.log(failed) // expect to print  "true" 
```

I would expect this code to print "true". However, the waitUntillFinished line instead throws the error, causing the process to end. Why is this? The description for the method states `Returns a promise the resolves when the job has finished. (completed or failed).` I interpreted this to mean that the promise will always resolve even if the job fails. Does the documentation need to be changed?

bilalshaikh42:
updates the documentation of the method to match the actual behavior, which returns a rejected promise when job failes. 

closes #1031

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -733,7 +733,8 @@ export class Job<
   }
 
   /**
-   * Returns a promise the resolves when the job has finished. (completed or failed).
+   * Returns a promise the resolves when the job has completed (containing the return value of the job),
+   * or rejects when the job has failed (containing the failedReason). 
    *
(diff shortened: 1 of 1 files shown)


roggervalf (approved):
LGTM, thanks @bilalshaikh42

PR #1044, opened 2022-02-02
docs(process-step-jobs): add new pattern

roggervalf:


docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -0,0 +1,44 @@
+# Process Step jobs
+
+Sometimes, people would like to break their processor function into small pieces that will be processed depending on the previous executed step, we could handle this kind of logic by using switch blocks:
```
manast, line 3: "people would like to" -> "it is useful to"

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -0,0 +1,44 @@
...
+  queueName,
+  async job => {
+    const initialStep = 'initialStep';
+    const secondStep = 'secondStep';
+    const finishStep = 'finishStep';
+    let step = job.data.step;
+    while (step !== finishStep) {
+      switch (step) {
+        case initialStep: {
+          await job.update({
```
manast, line 18: I think we should show that for every step we do some work specific for that step, like  a call to an imaginary function:
```
await doInitialStepStuff();
```

etc.

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -0,0 +1,44 @@
...
+            step: secondStep,
+          });
+          step = secondStep;
+          break;
+        }
+        case secondStep: {
+          if (job.attemptsMade < 3) {
+            throw new Error('Not yet!');
```
manast, line 26: Why do we do this check and exception ?

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -0,0 +1,44 @@
...
+          throw new Error('invalid step');
+        }
+      }
+    }
+  },
+  { connection },
+);
+```
+
+As you can see, we should save the step value, in this case we are saving it into data. So even in an error, it would be retried in the last step that was saved.
```
manast, line 44: Also, we could add a reference to a similar version of this example but based on observables, to highlight that it is not needed to update manually the step.

manast (approved):

docs/gitbook/patterns/process-step-jobs.md
roggervalf: as we talked, this would be kept in our tests, for this example we could keep it simple

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -0,0 +1,47 @@
...
+          throw new Error('invalid step');
+        }
+      }
+    }
+  },
+  { connection },
+);
+```
+
+As you can see, we should save the step value, in this case we are saving it into data. So even in an error, it would be retried in the last step that was saved (in case we use a backoff strategy).
```
manast, line 43: "in an error" -> "in the case of an error"

manast (approved):

Issue #1050, opened 2022-02-04
Job.queueName is undefined in sandboxed processor


jgillick:
I'm trying to use a single processor script that will automatically route jobs from all queues to the correct handlers, however, it doesn't seem possible to get the queue name from the job.

**Launch Worker**
```js
const processorFile = path.join(__dirname, `processor.js`);
const worker = new Worker('comms', processorFile);
```

**Processor File**
```ts
export default async function (job: Job) {
    console.log(job.queueName);  // always returns undefined
});
```

I'm using BullMQ version 1.68.1

Any ideas why this would be happening?

manast:
There is no queueName in the job class. Please consider using typescript to eliminate this issues in the future.

manast:
UPDATE.
Yes, there is actually. However, it is not exposed to the sandbox processor. I will mark it as an enhacement.

jgillick:
@manast Thanks. I've submitted a PR for this: https://github.com/taskforcesh/bullmq/pull/1051

PR #1051, opened 2022-02-04
feat(job): add queueName to JobJson

jgillick:
Added queueName to the JobJson type so that it will be available to sandboxed workers.

In response to [#1050](https://github.com/taskforcesh/bullmq/issues/1050)

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -314,6 +314,7 @@ export class Job<
       failedReason: JSON.stringify(this.failedReason),
       stacktrace: JSON.stringify(this.stacktrace),
       returnvalue: JSON.stringify(this.returnvalue),
+      queueName: this.queueName,
     };
   }
 
@@ -737,7 +738,7 @@ export class Job<
 
   /**
    * Returns a promise the resolves when the job has completed (containing the return value of the job),
-   * or rejects when the job has failed (containing the failedReason). 
+   * or rejects when the job has failed (containing the failedReason).
    *
    * @param queueEvents - Instance of QueueEvents.
    * @param ttl - Time in milliseconds to wait for job to finish before timing out.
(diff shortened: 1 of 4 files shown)


src/interfaces/job-json.ts
```diff
@@ -14,6 +14,7 @@ export interface JobJson {
   stacktrace: string;
   returnvalue: string;
   parentKey?: string;
+  queueName?: string;
```
roggervalf, line 17: I think ? is not needed, queueName should be always present

src/classes/job.ts
```diff
@@ -314,6 +314,7 @@ export class Job<
       failedReason: JSON.stringify(this.failedReason),
       stacktrace: JSON.stringify(this.stacktrace),
       returnvalue: JSON.stringify(this.returnvalue),
+      queueName: this.queueName,
```
roggervalf, line 317: you could modify one of the sandbox test cases to consider this change too

roggervalf (commented):
LGTM, just some minor comments

jgillick:
Thanks @roggervalf. I created an entirely new sandbox unit test for this.

manast (changes requested):
Thanks for the PR. 
Unfortunatelly, and I do not want to be a party pooper :), but we cannot store the queueName in the job structure, since this will take a lot of Redis memory and most importantly it should not be needed. Instead we have to implement it here: https://github.com/taskforcesh/bullmq/blob/master/src/classes/child-processor.ts#L127
Since the original Job class already has a getter that returns the queueName:
https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L544

roggervalf:
Thanks @manast, sorry @jgillick but currently asJSON method is used in https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L893, so it would be saved in redis, we should not save queueName as it is a getter as mentioned above. I can take this feature on my own for this changes.

jgillick:
All good. I'll refactor it this week or next. Thank you.

jgillick:
@roggervalf This works. Thank you!

Issue #1056, opened 2022-02-07
Removing a job does not remove it's children 


bilalshaikh42:
Based on the documentation at https://docs.bullmq.io/guide/flows#jobs-removal,
```
If a parent job is removed, all its children will also be removed.
```
However, this is not working for us. We have a flow defined as follows:
```typescript
const runId ="CommonId"
// JobQueue is a enum with queue names
const manifestJob = {
  name: 'Manifest',
  queueName: JobQueue.manifest,
  opts: {
    jobId: `${runId}`,
    },
  },
};

const extractJob = {
  name: 'Sedml',
  queueName: JobQueue.extract,
  opts: {
    jobId: `${runId}`,
  },
};

// Process the manifest and post the manifest to the API
const filesJob = {
  name: 'Files',
  queueName: JobQueue.files,
  opts: {
    jobId: `${runId}`,
  },
  // Needs extraction completed and manifest completed
  children: [extractJob, manifestJob],
};

const thumbnailsProcessJob = {
  name: 'Create thumbnails',
  queueName: JobQueue.thumbnailProcess,
  opts: {
    jobId: `${runId}`,
  },
  // Needs file processing completed
  children: [filesJob],
};

const postThumbnailsJob = {
  name: 'Post thumbnails',
  queueName: JobQueue.thumbnailPost,
  opts: {
    jobId: `${runId}`,
  },
  children: [thumbnailsProcessJob],
};

const sedmljob = {
  name: 'SED-ML',
  queueName: JobQueue.sedmlProcess,
  opts: {
    jobId: `${runId}`
    },
  },
};
const postSedmlJob = {
  name: 'Post SED-ML',
  queueName: JobQueue.sedmlPost,
  opts: {
    jobId: `${runId}`,
    attempts: 5,
    },
  },
  children: [sedmljob],
};

const outputsJob = {
  name: 'Output',
  queueName: JobQueue.output,
  opts: {
    jobId: `${runId}`,
    attempts: 5,
    },
  },
};

const logsJob = {
  name: 'Logs',
  queueName: JobQueue.logs,
  opts: {
    jobId: `${runId}`,
    },
  },
};
const postLogsJob = {
  name: 'Post logs',
  queueName: JobQueue.logsPost,
  opts: {
    jobId: `${runId}`,
    },
  },
  children: [logsJob],
};

const metadataJob = {
  name: 'Metadata',
  queueName: JobQueue.metadata,
};

const postMetadataJob = {
[...]

PR #1060, opened 2022-02-09
fix(flow): remove processed children

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1056


src/commands/includes/removeParentDependencyKey.lua
```diff
@@ -19,9 +19,10 @@ local function moveParentToWait(parentPrefix, parentId, emitEvent)
   end
 end
 
-local function removeParentDependencyKey(jobKey, hard, baseKey)
-  local parentKey = rcall("HGET", jobKey, "parentKey")
-  if( (type(parentKey) == "string") and parentKey ~= "" and (rcall("EXISTS", parentKey) == 1)) then
+local function removeParentDependencyKey(jobKey, hard, parentKey, baseKey)
```
roggervalf, line 22: sometimes we have parentKey before calling this function so, it would be great to pass it as a parameter and check if it is present or not, so we save 1 extra call to redis

src/commands/moveToFinished-8.lua
```diff
@@ -116,25 +116,18 @@ if rcall("EXISTS", jobIdKey) == 1 then -- // Make sure job exists
 
         -- Remove old jobs?
         local prefix = ARGV[9]
-        local function removeJob(jobId)
```
roggervalf, line 119: this code is the same as in our include removeJob

manast (approved):

PR #1065, opened 2022-02-10
feat(sandbox): pass parent property

roggervalf:


tests/test_sandboxed_process.ts
```diff
@@ -390,6 +391,48 @@ describe('sandboxed process', () => {
...
+      worker.on('completed', async (job: Job, value: any) => {
+        try {
+          expect(job.data).to.be.eql({ foo: 'bar' });
+          expect(value).to.be.eql({
+            id: 'job-id',
+            queueKey: `bull:${parentQueueName}`,
+          });
+          expect(Object.keys(worker['childPool'].retained)).to.have.lengthOf(0);
+          expect(worker['childPool'].free[processFile]).to.have.lengthOf(1);
+          await worker.close();
```
manast, line 413: use finally in the try/catch to avoid repetition of ```worker.close()```

PR #1069, opened 2022-02-11
docs(step-jobs): pattern to wait children

roggervalf:


docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -40,8 +38,82 @@ const worker = new Worker(
...
+            },
+          );
+          await job.update({
+            step: thirdStep,
+          });
+          step = thirdStep;
+          break;
+        }
+        case thirdStep: {
+          const shouldWait = await job.moveToWaitingChildren(token);
```
bilalshaikh42, line 97: I am trying to implement this pattern. However, since my worker is not fetching jobs manually, I don't have a token. Is this possible to do without fetching jobs manually?

bilalshaikh42: Ah, I see. Thanks. We have been using the nestjs wrapper for this, which is not fully updated to BullMQ from Bull, so I completely missed that processors receive the lock. This certainly makes a lot of things easier!

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -40,8 +38,82 @@ const worker = new Worker(
...
+      switch (step) {
+        case initialStep: {
+          await doInitialStepStuff();
+          await childrenQueue.add(
+            'child-1',
+            { foo: 'bar' },
+            {
+              parent: {
+                id: job.id,
+                queue: `bull:${parentQueueName}`,
```
bilalshaikh42, line 68: Another question here. Is this possible to retrieve from the parent job itself? If we have a prefix other than "bull" for example, can this be retrieved from either the job or queue?

roggervalf: hi @bilalshaikh42, probably you are meaning if flows supports custom prefix https://github.com/taskforcesh/bullmq/blob/master/tests/test_flow.ts#L317 we are covering that case as in that test

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -40,8 +41,85 @@ const worker = new Worker(
...
+enum Step {
+  Initial,
+  Second,
+  Third,
+  Finish,
+}
+
+const worker = new Worker(
+  parentQueueName,
+  async job => {
```
manast, line 62: job -> (job, token) =>

manast (commented):
Just a small comment

docs/gitbook/patterns/process-step-jobs.md
roggervalf: hey @bilalshaikh42 I think I got your point, currently you could get the queueName from the parent, the prefix is not available directly, but it makes sense to have a getter as in this pr https://github.com/taskforcesh/bullmq/pull/1077

PR #1074, opened 2022-02-12
feat(job): validate UnrecoverableError presence

roggervalf:


src/classes/unrecoverable-error.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * UnrecoverableError
+ *
+ * Error to pass a job to failed even if the attemptsMade
```
manast, line 4: pass -> move

src/classes/job.ts
```diff
@@ -442,8 +443,8 @@ export class Job<
    * @param fetchNext - true when wanting to fetch the next job
    * @returns void
    */
-  async moveToFailed(
-    err: Error,
+  async moveToFailed<E extends Error>(
```
manast, line 446: Will this work for "sandboxed" processors?

src/classes/unrecoverable-error.ts
```diff
@@ -0,0 +1,14 @@
...
+ * are lower than the expected limit.
+ *
+ */
+export class UnrecoverableError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = this.constructor.name;
+    Object.setPrototypeOf(this, new.target.prototype);
```
roggervalf, line 12: base on https://stackoverflow.com/questions/41102060/typescript-extending-error-class

src/classes/job.ts
roggervalf: yes, it will work, I also added a test case for this

manast (approved):
Good job!

Issue #1075, opened 2022-02-13
Add method to get existing Queue Schedulers on a particular Queue


bilalshaikh42:
Add a getQueueSchedulers() or similar method that allows for retrieving a list of the current QueueSchedulers that have been registered for a particular queue. This is motivated by the following usecase:

>I am using a single container for both the queues and queues schedulers, but this leads to many queue schedulers created if scaling horizontally. Is there a way to check if a queue has a queue scheduler attached? That way, my code would only create a queue scheduler if a certain number are not already present

_Originally posted by @bilalshaikh42 in https://github.com/taskforcesh/bullmq/discussions/1072_

roggervalf:
hi @bilalshaikh42, please try version 1.73.0, this feature should be possible now

bilalshaikh42:
Thank you!

PR #1078, opened 2022-02-16
feat(queue-getters): add getQueueSchedulers

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1075


manast (approved):
Looks great to me.

PR #1080, opened 2022-02-17
fix(flow): respect defaultJobOptions from queue opts

roggervalf:


manast (approved):
Great documentation!

yourtallness:
So queue options still need to be duplicated even with this change?

https://docs.bullmq.io/guide/flows#provide-options

PR #1082, opened 2022-02-18
feat(retry-jobs): allow to retry completed jobs

roggervalf:


src/classes/queue.ts
```diff
@@ -7,6 +7,7 @@ import { QueueGetters } from './queue-getters';
 import { Repeat } from './repeat';
 import { Scripts } from './scripts';
 import { RedisConnection } from './redis-connection';
+import { FinishedTarget } from '..';
```
manast, line 10: Maybe a better name for this type would be "FinishedStatus"

manast (commented):
Looks great, just a small comment.

manast (approved):

PR #1085, opened 2022-02-19
feat(queue-getters): add getQueueEvents

roggervalf:


src/classes/queue-base.ts
```diff
@@ -73,6 +74,16 @@ export class QueueBase extends EventEmitter {
     return this.opts.prefix + ':' + this.base64Name();
   }
 
+  protected async setClientName(client: RedisClient): Promise<void> {
```
manast, line 77: wouldn't it be more elegant to move this method to the RedisClient class itself?

manast: And refactor this.clientName as a input parameter of "setClientName"

src/classes/queue-events.ts
```diff
@@ -231,6 +233,14 @@ export class QueueEvents extends QueueBase {
     }
   }
 
+  /**
+   * @override
+   */
+  protected clientName(): string {
+    const queueNameBase64 = this.base64Name();
+    return `${this.opts.prefix}:${queueNameBase64}:qe`;
```
manast, line 241: 'qe' could be refactored to a constant like "postfix" or something, since it is used in several places.

src/classes/queue-getters.ts
```diff
@@ -306,6 +306,28 @@ export class QueueGetters<
     }
   }
 
+  /**
+   * Get queue events list related to the queue.
+   *
+   * @returns - Returns an array with queue events info.
+   */
+  async getQueueEvents(): Promise<
```
manast, line 314: This code could be refactored and put into the base class or maybe on the RedisClient class I am not sure, seems like the only difference is the "postfix" constant.

manast (commented):
Looks good, some comments.

Issue #1094, opened 2022-02-22
Jobs are processing more than the max attempts


roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1062#issuecomment-1044111610
this is happening because in moveToFinished we are moving jobs to active but we are missing to increment the attemptsMade

PR #1095, opened 2022-02-22
fix(move-to-finished): increment attemptsMade when moving job to active

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1094


coder-pm:
I believe that the fix introduced a regression because attemptsMade now starts with 1 on the first try. However, according to the official BullMQ documentation, attemptsMade should indicate the number of attempts made after the job has failed, not including the first attempt. If you're using felixmosh/bull-board, you can clearly see the regression, as every job is marked as attempt 2.

PR #1096, opened 2022-02-22
docs: named processor

roggervalf:


docs/gitbook/patterns/named-processor.md
```diff
@@ -0,0 +1,30 @@
...
+    }
+  },
+  { connection },
+);
+```
+
+You could use a simple switch case to differentiate your logic, in this example we are using the job name.
+
+{% hint style="warning" %}
+This was a feature in Bull package, but it creates a lot of confusion, so in order to provide an alternative, you can use this pattern. See [#297](https://github.com/taskforcesh/bullmq/issues/297) and [#69](https://github.com/taskforcesh/bullmq/issues/69) as reference
```
manast, line 29: "in Bull package" -> "in the Bull package"

manast (commented):
Looks good, just a small comment.

Issue #1087, opened 2022-02-19
Warn users that BullMQ is not compatible with Upstash


manast:
Due to the recent popularity of https://upstash.com as a Redis hosting service, we are starting to get reports of people trying Bull and getting all kinds of weird behaviors. Currently, the way Upstash is implemented is not compatible with BullMQ. Hopefully, in the future, it will be, but for now, we should throw an error when using a host using the upstash domain so that users do not use the library as it will not work.

I created this issue in the Upstash repo to give visibility to this problem and so that everybody can track its resolution: https://github.com/upstash/issues/issues/18

PR #1098, opened 2022-02-23
fix(connection): advertise that Upstash is not compatible with Bullmq

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1087

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -20,6 +20,10 @@ const deprecationMessage = [
   'On the next versions having this settings will throw an exception',
 ].join(' ');
 
+const upstashMessage = ['BullMQ: Upstash is not compatible with BullMQ.'].join(
+  ' ',
+);
+
 export class RedisConnection extends EventEmitter {
   static minimumVersion = '5.0.0';
   protected _client: RedisClient;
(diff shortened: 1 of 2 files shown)


src/classes/redis-connection.ts
```diff
@@ -20,6 +20,10 @@ const deprecationMessage = [
   'On the next versions having this settings will throw an exception',
 ].join(' ');
 
+const upstashMessage = ['BullMQ: Upstash is not compatible with BullMQ.'].join(
```
manast, line 23: No need to use "join" in this case as the string is quite short. Also using ` it is possible to split in lines.

src/classes/redis-connection.ts
```diff
@@ -48,14 +52,16 @@ export class RedisConnection extends EventEmitter {
...
     } else {
       this._client = <RedisClient>opts;
-      let options = this._client.options;
+      let options = <IORedis.RedisOptions>this._client.options;
       if ((<IORedis.ClusterOptions>options)?.redisOptions) {
         options = (<IORedis.ClusterOptions>options).redisOptions;
       }
 
       this.checkOptions(deprecationMessage, options);
+      this.checkUpstashHost(options.host);
```
manast, line 64: I think we should actually throw an exception in this case, since it is not going to work.

manast (approved):

Issue #800, opened 2021-10-08
Parent job does not execute when child job fails


m4nuC:
I have a FlowProducer that runs a parent job after a set of children jobs have been executed. There are about 10k children that are being run concurrently. I've set the option `removeOnFail` to `true` on both the children and the parents but it seems that if a child fail the execution of the parent just hangs. 

Could this be a bug or is it a configuration issue on my side?

Here is what the code that create the job looks like:

```
export const initializeCollectionStats = async () => {
  const flowProducer = new FlowProducer({connection: redisClient, sharedConnection: true});
  const totalSupply = 10000
  const tokenJobs = [];
  for (let i = firstTokenID; i <= totalSupply; i++) {
    tokenJobs.push({
      name: 'child',
      data: {
        hello: 'world'
      },
      queueName: "queueName",
      opts: {
        removeOnFail: true,
        attempts: 2,
        timeout: 3000
      }
    })
  }

  const initJob = await flowProducer.add({
    name: 'initProject',
    data: {
      contractAddress
    },
    queueName: PROJECT_INIT_QUEUE,
    children: tokenJobs,
    opts: {
      attempts: 1
    }
  })

  return initJob
}
```

and the worker that run the children job looks like this:

```
const worker = new Worker(QUEUE_NAME, async (job) => {
    const { hello } = job.data
    try {
        const metadata = await doSomethingAsync(hello)
        return "done"
    } catch (e) {
        logger.error(e)
        return null
    }
}, { connection: redisClient, concurrency: 300, sharedConnection: true});

```

manast:
Currently by design, until all child jobs have been completed the parent job will not be processed.

m4nuC:
@manast Gotcha thanks, so then then only would be to catch errors and manually mark child jobs as completed or is there an option for this ? I assumed that `removeOnFail` was doing something like this.

m4nuC:
@manast to clarify what I am trying to achieve is:
- Process child job before parent (that's a given)
- Retry child job on fail just like any other jobs 
- If max amount of retries is reach on children, then move job to fail
- Once all child jobs are either failed or completed, then run parent.

Slind14:
This affects us as well. Is there a possibility to see a configurable outcome in the near future?

manast:
@Slind14 what is your usecase?

Slind14:
> @Slind14 what is your use-case?

[...]

PR #1100, opened 2022-02-24
feat(job): add removeDependencyOnFail option

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/800


src/types/job-options.ts
```diff
@@ -0,0 +1,15 @@
+import { JobOptionsBase } from '../interfaces';
+
+export type JobsOptions = JobOptionsBase & {
+  /**
+   * If true, removes the job from its parent dependencies when it fails after all attempts.
+   */
+  removeDependencyOnFail?: boolean;
+};
+
```
manast, line 9: Maybe a comment here clarifying that these fields are the ones stored in Redis and thus with smaller keys for compactness.

src/commands/includes/updateParentIfNeeded.lua
```diff
@@ -0,0 +1,17 @@
+--[[
+  Validate and move parent to active if needed.
+]]
+
+local function updateParentIfNeeded(parentQueueKey, parentDependenciesKey, parentId )
```
manast, line 5: could this function be renamed to "moveParentToWaitifNeeded" ? seems like that is what it does, "update" it is a big vague and makes it more difficult to understand.

roggervalf: sounds good to me

src/classes/job.ts
```diff
@@ -276,6 +281,28 @@ export class Job<
...
+  static optsFromJSON(rawOpts?: string): JobsOptions {
+    const opts = JSON.parse(rawOpts || '{}');
+
+    const optionEntries = Object.entries(opts) as Array<
+      [keyof RedisJobOptions, any]
+    >;
+    const options = optionEntries.reduce<Partial<Record<string, any>>>(
+      (acc, item) => {
+        const [attributeName, value] = item;
+        if (attributeName === 'rdof') {
```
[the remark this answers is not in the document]
roggervalf: yes, this is the beginning of it

M3talen:
Dear @manast ,

I wanted to reach out and express my appreciation for the work you have done on the this project. I have been using the project for a few projects, and I think it is a fantastic resource.

I was wondering if you would be open to considering the addition of this feature that I believe would greatly benefit the project.  As right now there is no other way to complete a flow without removing the child manually. But the bigger problem for me is that removing child's from the array doesn't update the flow job.

Thank you for your time, and I look forward to hearing from you.

Sincerely,
@M3talen

roggervalf:
hi @M3talen,
could you please elaborate a little bit your use case, now I'm thinking that if a job is not required to be completed for an specific flow, this should mean that the parent-child relationship is not needed in that case. Also as you mentioned, when a child is deleted, the parent should be moved to wait if this is the last pending child for that job as in https://github.com/taskforcesh/bullmq/blob/master/tests/test_flow.ts#L2480-L2525.

M3talen:
hi @roggervalf
It might be a bit difficult to explain but ill try to show it off

1) in EngineQueue i create a new task (daily-flow) 
2) that task creates a new flow with a parent called 'flow-job' in a new queue (FlowQueue) 
3) as children i put tasks that need to be done daily in other queues 
3.1) Flow starts executing first children.
3.2) job in EngineQueue gets put to completed. (here start the problems - if i wait for this job and a child fails it will stop procesing the queue)
Now i don't care about what tasks fail but when one fails i would like to add a stacktrace/log/update data on the 'flow-job' so i know what failed but i want the 'flow-job' to be executed when all children either fail or complete. 

Currently when a task fails (a child) the 'flow-job' is not executed until i delete it
even if the failed task fails all attempts and gets moved to failed it doesn't execute the 'flow-job'

I hope this explains a bit of my problem.

roggervalf:
hi @M3talen, in your case you can use this function https://api.docs.bullmq.io/classes/FlowProducer.html#addBulk, for example:
```ts
const trees = await flow.addBulk([
        {
          name: 'flow-job',
          queueName: 'Flow-Queue',
          data: {},
          children: [
            {
              name,
              data: { idx: 0, foo: 'bar' },
              queueName: 'Flow-Queue',
            },
          ],
        },
        {
          name: 'non-dependent-child',
          queueName: 'Flow-Queue',
          data: {},
          children: [
            {
              name,
              data: { idx: 1, foo: 'baz' },
              queueName: 'Flow-Queue',
            },
          ],
        },
      ]);

```
so when you have children where you don't care if they complete or fail, those could be created apart of the main flow, with addBulk you can create all those flows together

roggervalf:
in favor of #1953

PR #1107, opened 2022-02-28
Feat/add support for metrics

### LINKED ISSUE #1104: Add metrics/statistics support
Currently, the only way to calculate metrics from a queue is by analyzing the completed (and/or failed) jobs. This has many drawbacks, too many to list here.

Instead, we would like to have some simple but powerful mechanism that can be used to calculate metrics. The goals for this feature are:
- Able to provide real-time statistics of performance over periods of time.
- Minimum impact on a queue performance.
- Minimum usage of RAM.

The initial proposal I have is to store the number of jobs processed per minute during the last 2 weeks. 
If we assume that a queue is processing 1M jobs per minute, this will require less than 120Kb of space: 6 bytes per data point (as maximum value would be 999999, ints are stored as strings in Redis), 60 data points per hour * 24 hours * 14 days = 120960 bytes. This is a negligible amount of data for any queue, of course the maximum time window should be configurable.

To accomplish this, we can have a  "completed/failed" counter, this is one key such as ```bull:myqueue:finish-count``` that is increased every time a job is completed or failed, as well as the count for the latest datapoint (kind of the prev count).
Every time a job is processed, we will check if we have crossed a minute boundary, if so we compute a delta between current and last data point count and store that delta in a LIST (```bull:myqueue:metrics``` maybe), then we update prev count with current count and continue like this for ever. We probably also need to store the latest timestamp used for the latest data point in the LIST, this will help us know where in time the metrics list ends, and also if we need to insert zeroes in the case the queue has not processed any jobs during several minutes.

A similar mechanism could also be used for calculating data points when adding jobs to the queue, having two data point arrays, one for completed jobs and one for jobs added to the queue would allow for even more interesting metrics calculation, such as average time jobs spend in the queue, throughput and much more. This is super useful to implement up/down scale strategies as well as monitoring capabilities.

manast:
fixes #1104


docs/gitbook/guide/metrics/metrics.md
```diff
@@ -0,0 +1,49 @@
...
+  },
+});
+```
+
+{% hint style="warning" %}
+You need to use the same setting on all your workers to get consistent metrics.
+{% endhint %}
+
+In order to get the metrics you just use the `getMetrics` method on the Queue class. You can choose to get the metrics
+for the completed or failed jobs:
```
roggervalf, line 25: there is an extra new line here

docs/gitbook/guide/metrics/metrics.md
```diff
@@ -0,0 +1,49 @@
+# Metrics
+
+BullMQ provides a simple metrics gathering functionality that allows you to track the performance of your queues.
+The workers can count the number of jobs they have processed per minute and store this data in a list to be consumed later.
+
+You enable it on the worker settings by specifying how many data points you want to keep, we recommend 2 weeks of metrics data which
+should take a very small amount of space, just around 120Kb of RAM per queue.
```
roggervalf, line 7: there is an extra new line before this line

roggervalf (approved):
Awesome, lgtm

roggervalf (approved):

Issue #1097, opened 2022-02-22
Why not create a FlowJob with parent id?


swherden:
Context: Inside a main job, more complex sub jobs have to be created. The complex jobs would fit perfectly into a FlowJob Structure. 
The definite structure is unknown inside the main job. Hence, the FlowJob structure has to be created at runtime and the main job depends on it.
Problem Why is it not possible neither to put the new FlowJob structure as a child into the main job nor to create a FlowJob with the main job as parent?

`const flow:FlowJob = new FlowJob(..., parent: { id: string, queue: string})`

or 

```
const flow:FlowJob = new FlowJob(...);
job.addChild(flow);
```

roggervalf:
hi @swherden, it seems like there is a pr that could handle this case https://github.com/taskforcesh/bullmq/pull/630, could you also elaborate your case a little more. Also it is probably that you should check this pattern https://docs.bullmq.io/patterns/process-step-jobs too

swherden:
Thanks for the PR #630. It looks like that it is solving my demand. 

For the details. I have the requirements to divide a main into x child jobs dynamically at runtime. Furthermore, I would like to track the completion state of all children as well as of the parent job. Therefore, it would be great to add further FlowJobs dynamically as childs to a parent Job.

swherden:
I try @roggervalf 's comment and it is a solution.

PR #1110, opened 2022-03-02
feat(flow-producer): allow parent opts in root job when adding a flow

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1097


fabio-filho:
@manast, can this PR be merged? This will solve a problem of mine.

fabio-filho:
@roggervalf, is there any reason to keep this PR open?

roggervalf:
hi @fabio-filho , this a pretty old pr, @manast what do you think about this?

fabio-filho:
Me too, @pavelbely, I'm doing a workaround to handle it.

fabio-filho:
@manast, 😬

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -124,3 +124,86 @@ const worker = new Worker(
...
+              {
+                name,
+                data: { idx: 1, foo: 'baz' },
+                queueName: 'grandchildrenQueueName',
+              },
+            ],
+            opts: {
+              parent: {
+                id: job.id,
+                queue: `${job.prefix}:${job.queueName}`,
```
manast, line 170: I think It would be nicer if we had a ```job.qualifiedQueueName()``` so that the user does not need to know how the key is constructed in Redis.

manast: Btw, why is the property called "queue" in the parent options, but queueName in the children? if they were the same then it would work with just: ```queueName: job.queueName```

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -124,3 +124,86 @@ const worker = new Worker(
 {% hint style="info" %}
 Bullmq-Pro: this pattern could be handled by using observables; in that case, we do not need to save next step.
 {% endhint %}
+
+# Chaining Flows
+
+Another use case is to add flows at runtime and then wait for the children to complete.
+
```
manast, line 131: Add this: "For example, we can add children dynamically in the processor function of a worker."

manast (commented):
Just some small comments on the documentation.

docs/gitbook/patterns/process-step-jobs.md
roggervalf: when flows were created, it used queue parameter name, probably in v3 we can rename it as queuePrefix because it contains prefix as well

manast: I think queuePrefix is not the best name, as we already have queue.prefix which is not the same thing so it may be confusing. I am aware we use queuePrefix in other places, including scripts, maybe the name should be queueQualifiedName which is a more standard naming for this kind of thing, for example: https://www.gnu.org/software/gawk/manual/html_node/Qualified-Names.html

roggervalf: good catch, I'll add it as a reference in the new property docs

manast (approved):

Issue #1112, opened 2022-03-02
Queue.getJobs may return duplicates


Toilal:
I wrote this test to reproduce an issue I hit in my app and really confused me.

```typescript
  it('repro', async () => {
    const queue = new Queue('test')
    await queue.add('test', {})
    const jobs = await queue.getJobs(['completed', 'failed', 'paused', 'repeat', 'wait', 'active', 'delayed', 'waiting', 'waiting-children'])
    expect(jobs).toHaveLength(1)
   // Error:
   // Expected length: 1
   // Received length: 2
  })
```

It returns 2 items instead of expected 1, with the job instance duplicated. It is due to 'wait' and 'waiting' status beeing a duplicate ...

Problem is the same if you pass duplicates state in first parameter.

```typescript
  it('repro', async () => {
    const queue = new Queue('test')
    await queue.add('test', {})
    const jobs = await queue.getJobs(['waiting', 'waiting', 'waiting', 'waiting', 'waiting'])
    expect(jobs).toHaveLength(1)
   // Error:
   // Expected length: 1
   // Received length: 5
  })
```

To go further about this issue, I expect `getJobs()` to work without giving any state to return all jobs, as it's the behavior for `getJobCounts()`. Maybe there's a technical limitation due to Redis, but it would be easier to use.

manast:
Yeah, we are not checking if you send duplicate statuses in the call... not sure we should do that kind of defensive programming. Normally you only call "getJobs" for a given status or maybe a couple of them, not sure either it gives so much value to implement a new short hand but sure it could be added if you want to submit a PR for it.

Toilal:
When implementing a dashboard or an API that monitors execution of jobs, you may have to grab all jobs regardless the status.

As a new user of bullmq, and it's not easy to find out the list of all possible statuses either.

I'll open a pull request.

PR #1113, opened 2022-03-02
Get jobs enhancements

Toilal:
See https://github.com/taskforcesh/bullmq/issues/1112

The PR is splitted in 3 incremental commits. Each commit brings more changes, so you can choose what to merge or not.

First two commits should be safe to merge. 3rd commit improves consistency and should ensure that `getJobs` always returns the same count as `getJobsCount`, but it's more intrusive.

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -49,7 +49,13 @@ export class QueueGetters<
     const currentTypes = typeof types === 'string' ? [types] : types;
 
     if (Array.isArray(currentTypes) && currentTypes.length > 0) {
-      return [...new Set(currentTypes)];
+      const sanitizedTypes = [...currentTypes];
+
+      if (sanitizedTypes.indexOf('waiting') !== -1) {
+        sanitizedTypes.push('paused');
+      }
+
+      return [...new Set(sanitizedTypes)];
     }
 
     return [
@@ -145,7 +151,7 @@ export class QueueGetters<
    * Returns the number of jobs in waiting or paused statuses.
    */
   getWaitingCount(): Promise<number> {
-    return this.getJobCountByTypes('waiting', 'paused');
+    return this.getJobCountByTypes('waiting');
   }
 
   /**
@@ -252,9 +258,6 @@ export class QueueGetters<
   ): Promise<Job<DataType, ResultType, NameType>[]> {
     types = this.sanitizeJobTypes(types);
 
-    if (types.indexOf('waiting') !== -1) {
-      types = types.concat(['paused']);
-    }
     const jobIds = await this.getRanges(types, start, end, asc);
 
     return Promise.all(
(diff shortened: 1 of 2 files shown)


ATheCoder:
@manast

manast (approved):

src/classes/queue-getters.ts
```diff
@@ -45,16 +45,40 @@ export class QueueGetters<
...
+    if (typeof types === 'string') {
+      types = [types];
+    }
+
+    if (Array.isArray(types) && types.length > 0) {
+      const sanitizedTypes = [...types];
+
+      if (
+        sanitizedTypes.indexOf('waiting') !== -1 ||
+        sanitizedTypes.indexOf('waiting-children')
```
roggervalf, line 58: we are not considering paused when waiting-children is provided in the current master branch, you could update your branch with master

src/classes/queue-getters.ts
```diff
@@ -45,16 +45,40 @@ export class QueueGetters<
     });
   }
 
+  private sanitizeJobTypes(types: JobType[] | JobType | undefined): JobType[] {
+    if (typeof types === 'string') {
```
roggervalf, line 49: what about:
`const currentTypes = typeof types === 'string' ? [types] : types;`

roggervalf (changes requested):

src/classes/queue-getters.ts
Toilal: Il would create `[undefined]` array if parameter is undefined

src/classes/queue-getters.ts
Toilal: I have removed `sanitizedTypes.indexOf('waiting-children')` from this condition

src/classes/queue-getters.ts
```diff
@@ -69,12 +75,7 @@ export class QueueGetters<
     Returns the number of jobs waiting to be processed.
   */
   count(): Promise<number> {
-    return this.getJobCountByTypes(
-      'waiting',
-      'paused',
-      'delayed',
-      'waiting-children',
-    );
+    return this.getJobCountByTypes('waiting', 'delayed', 'waiting-children');
```
Toilal, line 78: I'm not sure if 'paused' should be removed here

manast: It should not be removed. When the queue is paused what happens is that the wait key gets renamed to "paused", but other than that the status of a job is still waiting.

Toilal: Thanks, I have rollbacked this method.

src/classes/queue-getters.ts
roggervalf: hey @Toilal, but if types is `undefined`, `typeof types === 'string'` would be `false`, in that case currentTypes will be `undefined`, not `[undefined]` :thinking:

Toilal: Ho yes my mistake, I have updated the PR :)

roggervalf (approved):
LGTM

Issue #982, opened 2022-01-05
Recommended Express integration


namoscato:
We have a relatively straightforward integration of this library on Heroku (that generally follows [this guide](https://devcenter.heroku.com/articles/node-redis-workers)):

1. `worker` process – small Node.js process that instantiates a `Worker` instance, leveraging [throng](https://github.com/hunterloftis/throng) / clustering
1. `web` process – Express server instantiates a singleton ioredis connection and `Queue` instance; in practice, some of our Express routes add a job to the queue

We're instantiating the ioredis connection like so:

```ts
new IORedis(redisUrl, {
  maxRetriesPerRequest: null,
  enableReadyCheck: false,
});
```

per the recommendation added in https://github.com/taskforcesh/bullmq/pull/836.

**Is this the recommended Express setup?** We're specifically concerned that disabling `maxRetriesPerRequest` will pose issues in our web context where ideally the number of retries would be finite (per [ioredis' default](https://github.com/luin/ioredis#auto-reconnect)) and commands would "fail fast" to deliver a timely HTTP response.

manast:
This is a very interesting point and I am afraid I may have overlooked it. Trying to add to a queue that has lost a connection currently does not fail, which in most use cases is not what you want. So it seems like the optimal settings for a worker are not the same as to a client adding jobs to the queue... I will try to find a solution for this.

manast:
Still I am not sure what would be the proper way to solve this. So we want the call to add to fail fast, but we also want the queue connection to reconnect to the Redis server as soon as that server is online/reachable again, so we need infinite retries but also failing the command immediately if not currently connected, do you know which ioredis settings would accomplish this?

manast:
Ok, so after digging a bit more into it, it is possible we could just enforce the null on maxRetriesPerRequest only on connections that use "blocking" commands, that way we will fail fast when calling BullMQs api and still keep the blocking commands running stable during disconnection episodes-

namoscato:
Thanks for the quick reply, @manast!

> we need infinite retries but also failing the command immediately if not currently connected, do you know which ioredis settings would accomplish this?

Maybe this is what you are getting at, but I don't think you can accomplish this with a single ioredis instance / connection. Certainly their [default `maxRetriesPerRequest: 20`](https://github.com/luin/ioredis/blob/v4.28.2/lib/redis/RedisOptions.ts#L84) would accomplish the latter.

[...]

PR #1119, opened 2022-03-06
fix(connection): remove Queue reconnect overrides

namoscato:
Fixes https://github.com/taskforcesh/bullmq/issues/982

* Add optional `QueueBaseOptions.persistentConnection`
* Add optional `persistent` argument to `RedisConnection`, defaulting to `true`
* Tighten up some `RedisConnection` type checking


src/classes/redis-connection.ts
```diff
@@ -47,21 +53,23 @@ export class RedisConnection extends EventEmitter {
           return Math.min(Math.exp(times), 20000);
         },
         ...opts,
-        maxRetriesPerRequest: null,
-        enableReadyCheck: false,
```
namoscato, line 51: I'm not too familiar with `IORedis.Cluster`, but I'm not sure these were properly overriding `IORedis.ClusterOptions.redisOptions` in that scenario. With that said, I just left this [as is](https://github.com/taskforcesh/bullmq/pull/1119/files#diff-51c66308ccf996681d88c66c21de3203d273ff3f48071d8a12f696eb6d194f64R59-R60).

src/classes/redis-connection.ts
```diff
@@ -47,21 +53,23 @@ export class RedisConnection extends EventEmitter {
...
-      this.checkUpstashHost(this.opts.host);
-    } else {
-      this._client = <RedisClient>opts;
-      let options = <IORedis.RedisOptions>this._client.options;
-      if ((<IORedis.ClusterOptions>options)?.redisOptions) {
-        options = (<IORedis.ClusterOptions>options).redisOptions;
+
+      if (this.persistent) {
+        this.opts.maxRetriesPerRequest = null;
+        this.opts.enableReadyCheck = false;
```
namoscato, line 60: I don't fully understand how this `enableReadyCheck` override affects things and am unsure if it should be kept "coupled" with `maxRetriesPerRequest`.

src/interfaces/queue-options.ts
```diff
@@ -20,6 +20,11 @@ export interface QueueBaseOptions {
    */
   sharedConnection?: boolean;
 
+  /**
+   * Denotes commands should retry indefinitely.
+   */
+  persistentConnection?: boolean;
```
namoscato, line 26: Open to a better name here.

src/classes/redis-connection.ts
manast: In fact I think enableReadyCheck is not needed at all, so we can remove it.

src/interfaces/queue-options.ts
manast: I think it is ok.

src/utils.ts
```diff
@@ -60,14 +60,18 @@ export function delay(ms: number): Promise<void> {
   });
 }
 
-export function isRedisInstance(obj: any): boolean {
+export function isRedisInstance(obj: any): obj is Redis | Cluster {
```
manast, line 63: I wasn't aware of this syntax (obj is Redis | Cluster) do you have a link that explains it?

manast (commented):
Thanks for this fine PR. I have been ill the latest days so it has taking me a long time to process all issues and prs. 
I would say that the part that is missing is to tell the connection it is persistent in the Worker, QueueScheduler and  QueueEvents classes, as these are the ones that require persistency, whereas Queue does not. I solved it in Bull a couple of weeks ago: 
https://github.com/OptimalBits/bull/commit/d3b91386e30d7205efdc19bcd18fe1e5fefa3542

namoscato (commented):
> I would say that the part that is missing is to tell the connection it is persistent in the Worker, QueueScheduler and QueueEvents classes, as these are the ones that require persistency, whereas Queue does not.

Yeah, I'm happy to make this explicit, but it's handled via the `RedisConnection` class [argument default](https://github.com/namoscato/bullmq/blob/23e858351f010d70598d6c45c8a6f2813443ea1e/src/classes/redis-connection.ts#L42).

manast:
> > I would say that the part that is missing is to tell the connection it is persistent in the Worker, QueueScheduler and QueueEvents classes, as these are the ones that require persistency, whereas Queue does not.
> 
> Yeah, I'm happy to make this explicit, but it's handled via the `RedisConnection` class [argument default](https://github.com/namoscato/bullmq/blob/23e858351f010d70598d6c45c8a6f2813443ea1e/src/classes/redis-connection.ts#L42).

What I mean is that when the Worker is instantiated it should enable the persistent connection setting, as we do with the shared connection today: https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L148

src/classes/redis-connection.ts
namoscato: Removed in 2eb7c0204974d317ee64da70f04daf71df67ff03.

namoscato (commented):
> > > I would say that the part that is missing is to tell the connection it is persistent in the Worker, QueueScheduler and QueueEvents classes, as these are the ones that require persistency, whereas Queue does not.
> > 
> > 
> > Yeah, I'm happy to make this explicit, but it's handled via the `RedisConnection` class [argument default](https://github.com/namoscato/bullmq/blob/23e858351f010d70598d6c45c8a6f2813443ea1e/src/classes/redis-connection.ts#L42).
> 
> What I mean is that when the Worker is instantiated it should enable the persistent connection setting, as we do with the shared connection today

Sure, I explicitly set this option in 2ee0f54e7f6a5f767f48dee93275c4dcfc4920e7.

namoscato:
@manast, I just rebased this off `master`, but do you have any other feedback? Happy to help push this over the finish line.

manast:
@roggervalf do you have any comments?

tests/test_connection.ts
```diff
@@ -20,26 +19,43 @@ describe('connection', () => {
...
-        (<IORedis.RedisOptions>client.options).maxRetriesPerRequest,
-      ).to.be.equal(null);
-      expect(
-        (<IORedis.RedisOptions>client.options).enableReadyCheck,
-      ).to.be.equal(false);
-    }
-
-    const queue = new QueueBase(queueName, opts);
-    checkOptions(await queue.client);
+    beforeEach(async () => {
```
roggervalf, line 25: if there is only one test case per describe scope, you can move this logic  before line 37

tests/test_connection.ts
```diff
@@ -20,26 +19,43 @@ describe('connection', () => {
...
+    it('should override maxRetriesPerRequest: null as redis options', () => {
+      expect(options.maxRetriesPerRequest).to.be.equal(null);
+    });
+  });
+
+  describe('non-persistent', () => {
+    let options: IORedis.RedisOptions;
+
+    beforeEach(async () => {
+      const queue = new QueueBase(queueName, {
```
roggervalf, line 45: same as above, but moved before line 57

roggervalf (commented):
LGTM, just 2 styled comments

tests/test_connection.ts
namoscato: Yeah, sometimes I find it clearer to separate setup from assertion (i.e. the conventional ["Arrange, Act, Assert" pattern](https://xp123.com/articles/3a-arrange-act-assert/)), but I updated this in dec8986057c3be5ba53459bacfceb759ac4c146c.

tests/test_connection.ts
namoscato: Updated in dec8986057c3be5ba53459bacfceb759ac4c146c.

manast:
I think this PR is ready, I just have a small request, changing the name from "persistent" to "blocking", because in fact, all connections are normally persistent, is just that when we need a blocking connection we must retry forever instead of failing directly. After this small change, we can merge it.

src/classes/redis-connection.ts
```diff
@@ -26,19 +30,21 @@ export class RedisConnection extends EventEmitter {
...
   private initializing: Promise<RedisClient>;
   private closing: boolean;
   private version: string;
   private handleClientError: (e: Error) => void;
 
   constructor(
-    private readonly opts?: ConnectionOptions,
+    opts?: ConnectionOptions,
     private readonly shared: boolean = false,
+    private readonly persistent = true,
```
roggervalf, line 42: hi @namoscato, could you please also change **persistent** to **blocking** when have a chance, because all the connections are persistent. The difference is that blocking is when you want infinite retries and non-blocking when you want that the call fails directly.

namoscato:
> changing the name from "persistent" to "blocking"

Makes sense! Updated in 65b938a071895d68387e4327b0c4f0e8acdaa98b.

PR #1124, opened 2022-03-09
[Docs] Clarify jobs priority

slavivanov:
Note that jobs without priority are processed last. This is not what I had assumed, so it's better to clarify it.


docs/gitbook/guide/jobs/prioritized.md
```diff
@@ -8,6 +8,8 @@ Adding prioritized jobs is a slower operation than the other types of jobs, with
 
 Priorities goes from 1 to MAX_INT, whereas lower number is always higher priority than higher numbers.
 
+In a queue where only some jobs have a priority assigned, jobs without priority will be processed last.
```
manast, line 11: I think it would be better like this:
"Jobs without a priority assigned will get the least priority."

slavivanov: Fixed, thanks

PR #1131, opened 2022-03-11
fix(move-to-finished): consider rate limit

roggervalf:


src/classes/scripts.ts
```diff
@@ -757,6 +760,17 @@ export class Scripts {
       opts.limiter.groupKey && args.push(true);
     }
 
+    /* const args: (string | number | boolean | Buffer)[] = [
```
manast: out commented code

manast (commented):
Looks good. Small comment only.

src/classes/scripts.ts
roggervalf: yes, because it's in progress :sweat_smile:, I also want to refactor the moveJobToActive to consider the rate limit logic

manast (approved):

PR #1133, opened 2022-03-13
feat(./src/index.ts): add worker and job infering types

Everspace:
Add the QueueWorker<T> and QueueJob<T> types to allow one to infer the types on a corrisponding
worker easily.

This allows for nice typing such as the following

`queues.ts`
```ts
import { Queue } from "bullmq"

export const exampleQueue = new Queue<
  { buildId: number },
  boolean,
  "pass" | "fail"
>("exampleQueue")
```

`workers.ts`
```ts
import { QueueWorker, Worker } from "bullmq"
import type { exampleQueue } from "./queues"

const worker: QueueWorker<typeof exampleQueue> = new Worker("exampleQueue", async (job) => {
  switch(job.name) { // Strongly typed
    case "pass": 
      return true // also checked
    case "fail":
      return false
  }
})
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,3 +4,23 @@ export * from './enums';
 export * from './interfaces';
 export * from './types';
 export * from './utils';
+
+import type { Queue } from './classes/queue';
+import type { Worker } from './classes/worker';
+import type { Job } from './classes/job';
+
+export type QueueWorker<Q> = Q extends Queue<
+  infer DataType,
+  infer ResultType,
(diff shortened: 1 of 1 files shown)


SneakySensei:
I understand the usecase but I'm wondering if this offers the best type inference on the queue and the worker.
I've suggested an alternative method here, https://github.com/taskforcesh/bullmq/issues/3140#issuecomment-3708454237

Please let me know if that solves your usecase as well.

Issue #1147, opened 2022-03-21
Add UTC option for repeatable job RepeatOptions


pauliusg:
As I see, under the hood CRON rule next invocation is calculated using cron-parser npm package. This package supports UTC time:  https://github.com/harrisiirak/cron-parser (look at utc option).

In the code, RepeatOptions are passed as cron-parser options:
https://github.com/taskforcesh/bullmq/blob/da587c2844763a5e7035f799541ce0d6e09aa296/src/classes/repeat.ts#L225-L228

So it is enough to add `utc?: boolean` property to RepeatOptions. Currently I am ignoring TypeScript type and everything works as expected:
```javascript
...
const schedulerQueue = new Queue(TasksSchedulerQueueName);
...
await schedulerQueue.add(data.taskName, data, {
jobId: data.id.toString(),
// @ts-ignore Officially utc option is not supported.
// But as BullMQ uses cron-parser under the hood
// and passes RepeatOptions to that package, it works.
repeat: { cron: taskSchedule.cronRule, utc: true },
});
```

PR #1149, opened 2022-03-22
feat(cron-parser): upgrades version to 4.2.1

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1147

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "tsc:all": "tsc && tsc -p tsconfig-cjs.json"
   },
   "dependencies": {
-    "cron-parser": "^2.18.0",
+    "cron-parser": "^4.2.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
     "lint-staged": "10.5.4",
     "mocha": "^6.2.3",
     "mocha-lcov-reporter": "^1.3.0",
+    "moment": "^2.29.1",
```
manast, line 100: Do we really need moment? I could not find references to it in this PR.

roggervalf: Yes, moment is used in our tests, in the old version of cron-parser, moment was included, but in the new one it is not

manast (approved):

Issue #1151, opened 2022-03-23
When adding repeatable job, not all redis-connection.ts error event listeners are removed


pauliusg:
This is happening for me with shared Redis client. My code:
```javascript
const queue = new Queue('scheduler, { connection: sharedRedisClient });

await queue.add(data.taskName, data, {
	jobId: data.id.toString(),
	repeat: { cron: taskSchedule.cronRule },
});


for (const listener of sharedRedisClient.listeners('error')) {
	console.log({ listenerBeforeClose: listener.toString() });
}

await queue.close();

for (const listener of sharedRedisClient.listeners('error')) {
	console.log({ listenerAfterClose: listener.toString() });
}
```
And it still returns listener after close. This an issue for me because this code is executed from API many times and eventually I get MaxListenersExceededWarning.

Temporary I added this code below queue close to resolve my issue:
```javascript
await queue.close();

const repeat = await schedulerQueue.repeat; // Added code
await repeat.close(); //Added code
```

Then all listeners are removed. 

Looks like there should be a close method added in https://github.com/taskforcesh/bullmq/blob/e17b3f21606757a16630988a69c9607e8c843bd2/src/classes/queue.ts
```javascript
close(): Promise<void> {
  if (!this.closing) {
    this.closing = async () => {
      if (this._repeat) {
        const repeat = this._repeat();
        await repeat.close();
      }
      await super.close(); // Call close in queue-base.ts
    };
  }
  return this.closing;
}
```

PR #1154, opened 2022-03-24
fix(queue): close repeat connection when calling close

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1151


src/classes/repeat.ts
```diff
@@ -3,8 +3,7 @@ import { JobsOptions, RepeatOptions } from '../interfaces';
 import { QueueBase } from './queue-base';
 import { Job } from './job';
 import { Scripts } from './scripts';
-import { CronDate, parseExpression } from 'cron-parser';
```
roggervalf, line 6: there were some unused variables here

manast (approved):
Looks good.

PR #1158, opened 2022-03-25
fix(clean): consider processedOn and finishedOn attributes

roggervalf:


src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
 --- @include "includes/removeJob"
 
-local jobs = rcall(command, KEYS[1], rangeStart, rangeEnd)
-local deleted = {}
-local deletedCount = 0
-local jobTS
-local deletionMarker = ''
-local jobIdsLen = #jobs
-if ARGV[4] == "active" then
+local function cleanList(listKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp)
```
roggervalf, line 33: so I tried to separate this script into 4 functions because, some logic are only for specific states and in order to keep this script faster I decided to split it

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+  local deletionMarker = ''
+  local jobIdsLen = #jobs
   for i, job in ipairs(jobs) do
     if limit > 0 and deletedCount >= limit then
       break
     end
   
-    local jobKey = ARGV[1] .. job
+    local jobKey = jobKeyPrefix .. job
     if (rcall("EXISTS", jobKey .. ":lock") == 0) then
```
roggervalf, line 85: the difference between cleanList and cleanActive is this check that is made per each job

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
-    if (not jobTS or jobTS < ARGV[2]) then
-      if isList then
-        -- replace the entry with a deletion marker; the actual deletion will
-        -- occur at the end of the script
-        rcall("LSET", KEYS[1], rangeEnd - jobIdsLen + i, deletionMarker)
-      end
-      removeJob(job, true, ARGV[1])
+    local jobKey = jobKeyPrefix .. job
+    -- * finishedOn says when the job was completed, but it isn't set unless the job has actually completed
+    jobTS = rcall("HGET", jobKey, "finishedOn")
```
roggervalf, line 127: for completed and failed statuses, finishedOn should be the last updated timestamp attribute to be updated, so we don't need to get the other 2 params

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+    if limit > 0 and deletedCount >= limit then
+      break
+    end
+  
+    local jobKey = jobKeyPrefix .. job
+    -- Find the right timestamp of the job to compare to maxTimestamp:
+    -- * processedOn represents when the job was last attempted, but it doesn't get populated until the job is first tried
+    -- * timestamp is the original job submission time
+    -- Fetch all 2 of these (in that order) and use the first one that is set so that we'll leave jobs that have been active within the grace period:
+    for _, ts in ipairs(rcall("HMGET", jobKey, "processedOn", "timestamp")) do
```
roggervalf, line 161: in delayed status, this is a special state because a job could not pass from completed or failed to delayed, so finishedOn should not be considered

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+      break
+    end
+  
+    local jobKey = jobKeyPrefix .. job
+    -- Find the right timestamp of the job to compare to maxTimestamp:
+    -- * finishedOn says when the job was completed, but it isn't set unless the job has actually completed
+    -- * processedOn represents when the job was last attempted, but it doesn't get populated until the job is first tried
+    -- * timestamp is the original job submission time
+    -- Fetch all three of these (in that order) and use the first one that is set so that we'll leave jobs that have been active within the grace period:
+    for _, ts in ipairs(rcall("HMGET", jobKey, "finishedOn", "processedOn", "timestamp")) do
```
roggervalf, line 51: finishedOn should be considered here because we can retried a job, that pass to wait and then active

src/commands/cleanJobsInSet-2.lua
manast: Can we just pass a parameter like "checkLock" and in that case do the check, otherwise not do it and thus keep only one copy of the function?

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+
+  return {deleted, deletedCount}
+end
+
+local function cleanSet(setKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp)
+  local jobs = rcall("ZRANGE", setKey, rangeStart, rangeEnd)
+  local deleted = {}
+  local deletedCount = 0
+  local jobTS
+  local deletionMarker = ''
```
manast, line 118: deletionMarker is not used here

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+  local jobIdsLen = #jobs
```
manast, line 119: jobIdsLen is not used here

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+  
+  return {deleted, deletedCount}
+end
+
+local function cleanDelayed(setKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp)
+  local jobs = rcall("ZRANGE", setKey, rangeStart, rangeEnd)
+  local deleted = {}
+  local deletedCount = 0
+  local jobTS
+  local deletionMarker = ''
```
manast, line 149: deletionMarker unused

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+  local jobIdsLen = #jobs
```
manast, line 150: jobIdsLen unused

src/commands/cleanJobsInSet-2.lua
```diff
@@ -38,64 +30,167 @@ end
...
+  if(#deleted > 0) then
+    for from, to in batches(#deleted, 7000) do
+      rcall("ZREM", setKey, unpack(deleted, from, to))
+    end
+  end
+  
+  return {deleted, deletedCount}
+end
+
+local function cleanDelayed(setKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp)
```
manast, line 144: It would be great if cleanSet could also be used for cleanDelayed to avoid duplication of code.

roggervalf: yes I think a found a way

manast (approved):

PR #1164, opened 2022-03-30
refactor(queue): pack options

roggervalf:


src/classes/queue.ts
```diff
@@ -115,10 +123,11 @@ export class Queue<
     this.waitUntilReady()
       .then(client => {
         if (!this.closing) {
-          client.hset(
+          (<any>client).addQueueOpts(
             this.keys.meta,
-            'opts.maxLenEvents',
-            get(opts, 'streams.events.maxLen', 10000),
+            pack({
+              maxLenEvents: get(opts, 'streams.events.maxLen', 10000),
```
roggervalf, line 129: the limiter attribute should be considered as well after this refactor

manast:
I am a bit confused about this PR, what is it improving? I am worried that the json.decode call will affect performance since it is done very often.

roggervalf:
> I am a bit confused about this PR, what is it improving? I am worried that the json.decode call will affect performance since it is done very often.

Yeah, I was doing that because I was seeing that opts.maxLenEvents looks as a parameter into an object, but probably is better to keep it just as maxLenEvents attribute, I refactored to decode this key, also keeping the old version when getting the value to be backward compatible

src/interfaces/queue-options.ts
```diff
@@ -36,6 +36,18 @@ export interface QueueOptions extends QueueBaseOptions {
    * Options for the rate limiter.
    */
   limiter?: {
+    /**
+     * Max number of jobs to process in the time period
+     * specified in `duration`.
+     */
+    max?: number;
```
roggervalf, line 43: these parameters are optional in order to not be a breaking change

PR #1165, opened 2022-03-30
Misspelled rate limited

shahabbasi:


manast (approved):
Thanks for the correction!

PR #1168, opened 2022-03-31
refactor(backoffoptions): improved IntelliSense for backoff types

CodingMeSwiftly:
Currently, the `type` field is typed to `string` which makes it hard for users to know which values
work out-of-the-box without consulting the online documentation. This change fixes that by improving
the discoverability of the framework-defined backoff types while typing. Moreover, users will better
understand what values are supported by looking at the type definition file without having to leave
the IDE.

### DIFF
--- a/src/interfaces/backoff-options.ts
+++ b/src/interfaces/backoff-options.ts
@@ -7,7 +7,7 @@ export interface BackoffOptions {
   /**
    * Name of the backoff strategy.
    */
-  type: string;
+  type: 'fixed' | 'exponential' | Omit<string, 'fixed' | 'exponential'>;
   /**
    * Delay in milliseconds.
    */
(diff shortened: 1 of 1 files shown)


src/interfaces/backoff-options.ts
```diff
@@ -7,7 +7,7 @@ export interface BackoffOptions {
   /**
    * Name of the backoff strategy.
    */
-  type: string;
+  type: 'fixed' | 'exponential' | Omit<string, 'fixed' | 'exponential'>;
```
roggervalf, line 10: if I run `yarn build`, it throws an error

roggervalf: what about `'fixed' | 'exponential' | (string & {});`

CodingMeSwiftly: Wow. I thought I had run `yarn build` prior to committing.. sorry. `'fixed' | 'exponential' | (string & {})` has the same effect and doesn't break the build, so.. kudos :) I've updated the PR with your proposed change.

PR #1173, opened 2022-04-04
Fix imports for ESM modules.

vorillaz:
Using ESM imports along with [Esbuild](https://esbuild.github.io/content-types/#es-module-interop) or Tsup fails on runtime due to default importing the `get-port` and `ioredis` modules. This pull request changes the `tsconfig` file and the failing imports across the BullMQ package in order to play nicely with these bundlers.

_Reference: trying to bundle BullMQ failed Esbuild error_
Screenshot 2022-04-04 at 1 35 03 PM

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,8 +12,8 @@
     "jsx": "preserve",
     "importHelpers": true,
     "moduleResolution": "node",
-    "esModuleInterop": false,
-    "allowSyntheticDefaultImports": false,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
     "strictNullChecks": false,
     "baseUrl": ".",
     "paths": {
--- a/src/classes/child-pool.ts
+++ b/src/classes/child-pool.ts
@@ -1,7 +1,7 @@
 import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
 import { flatten } from 'lodash';
-import * as getPort from 'get-port';
+import getPort from 'get-port';
 import { killAsync } from './process-utils';
 import { ParentCommand, ChildCommand } from '../interfaces';
 import { parentSend } from '../utils';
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -12,8 +12,8 @@
     "jsx": "preserve",
     "importHelpers": true,
     "moduleResolution": "node",
-    "esModuleInterop": false,
-    "allowSyntheticDefaultImports": false,
+    "esModuleInterop": true,
```
manast, line 15: This setting could have negative effects in other scenarios, there is a risk this could break production build chains if changed.

vorillaz: Hi @manast 
I was a bit concerned regarding the building process but the TSC checks are passing, esbuild and babel are unable to bundle BullMQ tho.

manast: But maybe there are no drawbacks with this setting, I am not sure.  I remember battling this setting in the past, but I do not remember the details.

manast: Here is some old discussion, since this flag was actually true in the initial version: https://github.com/taskforcesh/bullmq/issues/129
Maybe they lured me into changing it  :). But the question is, will BullMQ still work for those dependants that have the flag set to false?

vorillaz: Huh, you are right but basically, that's more an issue with `ioredis`. It seems that one way or another the build will fail for some of us based on the way we are consuming `bullmq`. Honestly, I think I could live by using `bullmq` as a common-js module in order to avoid the hustle.

manast: Is this something that ioredis can solve, in that case you should create a issue in that repo, @luin is usually quite responsive.

boredland:
I took the time in https://github.com/taskforcesh/bullmq/pull/1346 to perform the needed upgrades. Would be great to get some feedback from @vorillaz, @manast and @luin over there!

PR #1197, opened 2022-04-15
fix: remove Error from Promise return in moveToWaitingChildren

### LINKED ISSUE #1196: job.moveToWaitingChildren never returns an Error
The return type for `job.moveToWaitingChildren` is `Promise<boolean | Error>` but the code appears to only be able to return a `boolean`. This causes an unnecessary type guard to make Typescript happy. The function can throw, but it doesn't directly return an `Error`.

https://github.com/taskforcesh/bullmq/blob/7ff86b23525392937914c2336f22c7d2ae89f6e9/src/classes/job.ts#L855-L860

in Scripts:
https://github.com/taskforcesh/bullmq/blob/7ff86b23525392937914c2336f22c7d2ae89f6e9/src/classes/scripts.ts#L601-L613

coreyvan:
Fixes #1196

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -855,7 +855,7 @@ export class Job<
   moveToWaitingChildren(
     token: string,
     opts: MoveToChildrenOpts = {},
-  ): Promise<boolean | Error> {
+  ): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


roggervalf:
Thank you @coreyvan

PR #1203, opened 2022-04-20
fix(queue-base): emit close error when no closing

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1205

### DIFF
--- a/src/classes/queue-base.ts
+++ b/src/classes/queue-base.ts
@@ -38,7 +38,9 @@ export class QueueBase extends EventEmitter {
       opts.sharedConnection,
       opts.blockingConnection,
     );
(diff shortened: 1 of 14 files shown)


src/classes/queue-base.ts
```diff
@@ -59,9 +59,9 @@ export class QueueBase extends EventEmitter {
     } catch (err) {
       try {
         return super.emit('error', err);
-      } catch (err) {
+      } catch (error) {
```
manast, line 62: is this renaming really needed?

roggervalf: no needed, it was changed as part of my testing

manast (approved):
LGTM

PR #1213, opened 2022-04-26
Fix delete lock move waiting children

roggervalf:


src/commands/moveToWaitingChildren-4.lua
```diff
@@ -21,7 +21,15 @@
 ]]
 local rcall = redis.call
 
-local function move_to_waiting_children (activeKey, waitingChildrenKey, jobId, timestamp)
+local function move_to_waiting_children (activeKey, waitingChildrenKey, jobId, timestamp, lockKey, token)
+  if token ~= "0" then
```
roggervalf, line 25: moved here as we are checking the jobKey existence first in other scripts

src/commands/moveToWaitingChildren-4.lua
```diff
@@ -21,7 +21,15 @@
 ]]
 local rcall = redis.call
 
-local function move_to_waiting_children (activeKey, waitingChildrenKey, jobId, timestamp)
+local function move_to_waiting_children (activeKey, waitingChildrenKey, jobId, timestamp, lockKey, token)
```
manast, line 24: This function is not following the naming convention we used for other functions, i.e. camel case.

roggervalf: you're right, changed in last commit

manast (approved):

PR #1214, opened 2022-04-26
feat(repeat): save repeatJobKey reference

roggervalf:


src/classes/job.ts
```diff
@@ -5,12 +5,19 @@ import {
...
   JobJsonRaw,
-  JobsOptions,
   ParentKeys,
   RedisClient,
   WorkerOptions,
 } from '../interfaces';
-import { FinishedStatus, JobState, JobJsonSandbox } from '../types';
+import {
+  FinishedStatus,
+  JobsOptions,
```
manast, line 14: shouldn't JobOptions come from "interfaces" ?

src/classes/job.ts
```diff
@@ -5,12 +5,19 @@ import {
...
+  JobState,
+  JobJsonSandbox,
```
manast, line 16: All these seems to be interfaces not types.

src/classes/job.ts
```diff
@@ -276,6 +283,32 @@ export class Job<
     return job;
   }
 
+  private static optsFromJSON(rawOpts?: string): JobsOptions {
```
manast, line 286: use for loops instead of reduce since this function will be used a lot and it will affect performance.

src/classes/job.ts
```diff
@@ -323,6 +356,28 @@ export class Job<
     };
   }
 
+  private optsAsJSON(opts: JobsOptions = {}): RedisJobOptions {
```
manast, line 359: use for loops

src/classes/job.ts
manast: Ok, now I see that there are new types for some of the interfaces. We should not use the same names then, because it is going to be very confusing. What is the difference, functionally/semantically of these new types vs the interface counterparts? that difference may help us in finding new names that are more appropiate.

roggervalf: these types are only extensions for the new attributes, there is for example a JobOptionsBase interface that will be extended in JobsOptions and RedisJobOptions

manast (commented):
It looks good. There is only one thing that I do not like, and that is that we modify the opts object by adding the repeatJobKey there (job.opts.repeat.repeatJobKey). I think it would be better if this property would be an optional public property of the job class itself.

manast (approved):
LGTM

PR #1223, opened 2022-04-29
fix(add-bulk): use for loop and throw if error is present

roggervalf:


manast (commented):
Looks good, but I think it should be a "fix" rather than "perf".

manast (approved):

Issue #1171, opened 2022-03-31
Stalled/failed jobs stay in queue


CodingMeSwiftly:
We're seeing something very similar to [this issue](https://github.com/taskforcesh/bullmq/issues/977), where stalled/failed jobs remain in the queue, even when `removeOnComplete` and `removeOnFail` are `true` in the job's options.

#1042 claims to fix it by not allowing progress updates after a job was removed from the queue (missing job key). The PR was released in [v1.68.1](https://github.com/taskforcesh/bullmq/releases/tag/v1.68.1), we're on [v1.78.2](https://github.com/taskforcesh/bullmq/releases/tag/v1.78.2).

We can work around the issue as mentioned in #977 by periodically clearing the queue ourselves, but it definitely feels like a bug with job scheduler.

In either case, setting the `removeOnComplete` and `removeOnFail` flags in job options should make sure the job is removed, even if stalled. Effectively leaving an empty queue if no job is scheduled for a certain amount of time. We see failed jobs in the queue even after having a job scheduler connected and running for >5h:

![Image 31 03 22 at 19 38](https://user-images.githubusercontent.com/6830248/161130486-aacf2cb8-ac1d-4ce9-ae21-a80d8c6d2860.jpg)

I've noticed that the jobs in question have in common that they perform very long running http requests (>1min), but they're async so stalling should not be caused by that, right?

CodingMeSwiftly:
I can confirm `moveStalledJobsToWait` is called in the scheduler, but that doesn't remove the stalled jobs either.

manast:
Is there any code that reproduces this issue?

CodingMeSwiftly:
Well our codebase :) - I'll try to put together an MVP over the weekend and will share it here.

manast:
We can reproduce this. But I do not remember if this is actually a feature. It is possible I reasoned that we will never want to automatically remove stalled jobs, please try to convince me why is it a good thing to support?

CodingMeSwiftly:
I can just speak for myself here:
Stalled jobs or the 'stalled' state in general seem not to live on the API surface (there is no `getStalled` on the `Queue` class).
The [QueueScheduler documentation](https://docs.bullmq.io/guide/queuescheduler) states that stalled jobs, 'are moved back or failed depending on the settings selected when instantiating the class.' That suggests that jobs cannot stay `stalled` indefinitely if we use a `QueueScheduler`, they have to end up either completed or failed.

[...]

PR #1225, opened 2022-04-30
Fix remove on fail stalled

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1171

### DIFF
--- a/src/commands/includes/removeJobsByMaxCount.lua
+++ b/src/commands/includes/removeJobsByMaxCount.lua
@@ -1,5 +1,5 @@
 --[[
-  Functions to remove jobs by max age.
+  Functions to remove jobs by max count.
 ]]
 
 -- Includes
--- a/src/commands/includes/removeJobsByMaxCount.lua
+++ b/src/commands/includes/removeJobsByMaxCount.lua
@@ -1,0 +1,15 @@
+--[[
+  Functions to remove jobs by max age.
+]]
+
+-- Includes
+--- @include "removeJob"
+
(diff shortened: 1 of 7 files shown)


src/commands/includes/removeJobsByMaxCount.lua
```diff
@@ -0,0 +1,15 @@
+--[[
+  Functions to remove jobs by max age.
```
manast, line 2: wrong comment

src/commands/moveStalledJobsToWait-8.lua
```diff
@@ -67,14 +70,38 @@ if (#stalling > 0) then
...
+                  local failedReason = "job stalled more than allowable limit" 
+                  rcall("HMSET", jobKey, "failedReason",
+                        failedReason, "finishedOn", ARGV[3])
+                  rcall("XADD", KEYS[8], "*", "event", "failed", "jobId",
+                        jobId, 'prev', 'active', 'failedReason',
+                        failedReason)
+
+                  if removeOnFailType == "number" then
+                    removeJobsByMaxCount(ARGV[3], opts["removeOnFail"], KEYS[4], ARGV[2])
+                  elseif removeOnFailType == "boolean" or removeOnFailType == "nil" then
```
manast, line 86: Why is ```removeOnFailType == "nil"``` needed?

manast: Maybe what is needed is to test if ```removeOnFailType ~= "nil"``` on the next else, the one that uses the age and count opts?

manast (commented):
Very nice refactorization. I had just a couple of comments.

PR #1237, opened 2022-05-10
fix(remove-job): pass right prev param in removed event

roggervalf:


manast (commented):
Ok, this works, but since you are checking that the REM operation returned 1, you can shortcircuit as soon as the job is removed and not keep removing on the other states.
One way to do this:
```lua
while(true) {
  if rcall("LREM", prefix .. "wait", 0, jobId) == 1 then
    prev = "wait"
    break;
  end   
  if rcall("ZREM", prefix .. "waiting-children", jobId) == 1 then
    prev = "waiting-children"
    break;
  end
   ...
break;
}
```

manast (commented):
Ok, this works, but since you are checking that the REM operation returned 1, you can shortcircuit as soon as the job is removed and not keep removing on the other states.
One way to do this:
```lua
while(true) {
  if rcall("LREM", prefix .. "wait", 0, jobId) == 1 then
    prev = "wait"
    break;
  end   
  if rcall("ZREM", prefix .. "waiting-children", jobId) == 1 then
    prev = "waiting-children"
    break;
  end
   ...
break;
}
```

src/commands/removeJob-1.lua
```diff
@@ -42,6 +42,26 @@ local function isLocked( prefix, jobId)
     return true
 end
 
+local function getPrevState( prefix, jobId)
```
manast, line 45: hmmm, the function is named "getPrevState" but it is actually removing the job AND returning the state. I think the name could just be "removeJobFromAnyState" or something similar.

Issue #1235, opened 2022-05-09
Acknowledge removal of job with removeRepeatable


kraten:
I found two issues while using the queue removeRepeatable method:
1. It should throw some error when repeatable job doesn't exist with the provided arguments. Instead it just returns a promise that resolves to null value.
2. Even when the repeatable job options are correct and job is removed, the promise gets resolved to null. It would be better to have some return value to know that the job got removed.

To summarize, I couldn't find a way to distinguish if the repeatable job is actually removed or the options are incorrect, as they both resolve to null value.

roggervalf:
hi @kraten, this is related to this issue https://github.com/taskforcesh/bullmq/issues/1211, for now we are only emiting the removed event, I marked this one as an enhancement

manast:
Fixed here https://github.com/taskforcesh/bullmq/pull/1239

PR #1239, opened 2022-05-11
feat(remove-repeatable): return boolean depending on job existence

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1235


manast (approved):
Looks good.

PR #1240, opened 2022-05-12
Refactor scripts class

roggervalf:


src/classes/worker.ts
```diff
@@ -430,7 +430,7 @@ export class Worker<
     token: string,
     jobId?: string,
   ): Promise<Job<DataType, ResultType, NameType>> {
-    const [jobData, id] = await Scripts.moveToActive(this, token, jobId);
+    const [jobData, id] = await this.scripts.moveToActive(this, token, jobId);
```
manast, line 433: Why is needed to pass "this" in this case?

manast (commented):
Looks good to me, just a small question.

src/classes/worker.ts
roggervalf: yes, here it was missing that change, in last commit should be addressed

src/classes/job.ts
```diff
@@ -297,12 +300,16 @@ export class Job<
...
+            queue,
+            (<unknown>jobData) as JobJsonRaw,
+            jobId,
+          );
     }
   }
 
   toJSON() {
-    const { queue, ...withoutQueue } = this;
+    const { queue, scripts, ...withoutQueue } = this;
```
manast, line 312: now it should be called: withoutQueueAndScripts. :).
In general, I think this pattern can lead to bugs because it is easy to forget to remove what you do not want and end the serialization of unneeded things. Even if more tedious, it is probably better to manually select the properties you want to serialize.

manast (approved):
Looks good, a smal comment.

Issue #320, opened 2020-11-27
RRULE - RFC5545


rogueturnip:
I want to ask has anyone done a conversion from an RRULE to creating scheduled reoccurring tasks in bullmq?
https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html

acro5piano:
@rogueturnip 
Do you think cron syntax is not enough?

rogueturnip:
Cron would be fine if there was a conversion from RRULE.  I'm finding that this is common across calendering and task management and it would be very handy to leverage that syntax.

acro5piano:
@rogueturnip 
Oh you want RRULE to cron converter. 
As far as I know, there are no such things at least written in JS... You seems have to write your own converter.

- https://github.com/jakubroztocil/rrule/issues/51
- https://github.com/simshaun/recurr/issues/169

rogueturnip:
I may have found something.  I have to test it to be sure but if anyone is curious.
```
import * as cronstrue from "cronstrue";

function rtoc(r) {
  r = r.replace("RRULE:", "");
  const C_DAYS_OF_WEEK_RRULE = ["MO", "TU", "WE", "TH", "FR", "SA", "SU"];
  const C_DAYS_WEEKDAYS_RRULE = ["MO", "TU", "WE", "TH", "FR"];
  const C_DAYS_OF_WEEK_CRONE = ["2", "3", "4", "5", "6", "7", "1"];
  const C_DAYS_OF_WEEK_CRONE_NAMED = [
    "MON",
    "TUE",
    "WED",
    "THU",
    "FRI",
    "SAT",
    "SUN",
  ];
  const C_MONTHS = [
    "JAN",
    "FEB",
    "MAR",
    "APR",
    "MAY",
    "JUN",
    "JUL",
    "AUG",
    "SEP",
    "OCT",
    "NOV",
    "DEC",
  ];
  let result = "";
  let dayTime = "0 0 0";
  let dayOfMonth = "?";
  let month = "*";
  let dayOfWeek = "?";
  let year = "*";

  let FREQ = "";
  let INTERVAL = -1;
  let BYMONTHDAY = -1;
  let BYMONTH = -1;
  let BYDAY = "";
  let BYSETPOS = 0;

  let rarr = r.split(";");
  for (let i = 0; i < rarr.length; i++) {
    let param = rarr[i].split("=")[0];
    let value = rarr[i].split("=")[1];
[...]

PR #1245, opened 2022-05-17
Feat/support RRULE syntax for repeated jobs

emanuelet:
added support for [RRULE](https://datatracker.ietf.org/doc/html/rfc5545) notation to set-up repeated jobs schedule. 

RRULE is the de-facto standard when it comes to repeated calendar events, used by iCalendar and Google Calendar, and allows a much more granular and complex set of rules compared to cron.

The addition uses https://github.com/jakubroztocil/rrule to parse the string passed in the payload and generate the next job. 

ref https://github.com/taskforcesh/bullmq/issues/320


manast:
Thank you for this fine PR. The implementation is fine, but I am trying to figure out if we could refactor a functionality like this as a separate plugin. There are basically two reasons for this 1) to avoid adding a new dependency, 2) to free us from more future maintenance burdens. I think a plugin mechanism for the Repeat functionality would not be very difficult to achieve considering it is mostly a matter of implementing the ```getNextMillis``` method in the repeat class. In fact, it may already be possible by subclassing Repeat. If we do this then we can link to the RRULE plugin on BullMQs main page.

emanuelet:
so just so I understand what you are referring to, are you talking about doing something like this https://stackoverflow.com/a/26309279/3998747 ? or using this https://www.npmjs.com/package/patch-package ?

manast:
@emanuelet no, I mean subclassing BullMQ classes. In this case, you probably need to subclass Queue and Repeat. You could have a QueueRRule and a RepeatRRule class, which just accept the RRULE options and implement a custom getNextMillis based on the RRULE. So similar to what you did but in a separate repo that you own.

manast:
ah, and you also will need to subclass the Worker class, so that it uses the getNextMilllis method from the RepeatRRule class...
Maybe @roggervalf has more ideas on how to implement this smoothly.

roggervalf:
I was thinking that maybe we can handle it like custom cronStrategies like in https://github.com/taskforcesh/bullmq/pull/1248

emanuelet:
At the end of the day, I just wish to be able to pass RRULE strings to `bullmq` so I am happy either way

roggervalf:
In favor of #1248

emanuelet:
@roggervalf whats the status of #1248 ?

roggervalf:
hey @emanuelet sorry for the delay, the feature is merged

emanuelet:
thank you, much appreciated

Issue #1243, opened 2022-05-16
TypeError: Invalid Version: undefined


arsonik:
Occurred after updating to any version greater than 1.79.0 using redis cluster
Trying to investigate ...

Calling:
```ts
await job.getState()
```
```
TypeError: Invalid Version: undefined
    at new SemVer (node_modules/bullmq/node_modules/semver/semver.js:314:11)
    at compare (node_modules/bullmq/node_modules/semver/semver.js:647:10)
    at Function.lt (node_modules/bullmq/node_modules/semver/semver.js:688:10)
    at Scripts.getState (node_modules/bullmq/src/classes/scripts.ts:443:16)
```

PR #1246, opened 2022-05-17
Fix missing cluster init options, update dependencies

arsonik:
Fix #1243

PR #1247, opened 2022-05-18
fix(redis-connection): save cluster opts

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/1246
fixes https://github.com/taskforcesh/bullmq/issues/1243


manast (approved):
LGTM

PR #1248, opened 2022-05-19
Feat repeat cron strategies

roggervalf:


emanuelet:
Is it possible to merge this PR?

emanuelet:
is there anything holding up this PR? I am waiting for this functionality for some crucial updates in our  application

roggervalf:
hi @emanuelet, this pr is not ready yet

manast (commented):
A couple of general comments:
- The name CronStrategy refers to Cron but in this case it is more general, so maybe RepeatStrategy is a better name.
- The cron behaviour that we have today should also be implemented using a RepeatStrategy, so we use internally the same mechanism that we expose externally.
- The way to pass arguments to the strategy uses the "cron" field, but for the same reasons as above, this would not express correctly that it is a generic expression that depends on the current RepeatStrategy used by the worker. The question is how to add a new generic argument in an elegant way 🤔
- What happens if some worker due to a bug is not using the correct strategy? (the user forgot to add it or something). I like the concept of using "name" in this case the worker could throw an exception or fail the job if no correct strategy has been provided.
- I think for a feature like this we need a documentation entry in the guide.

src/classes/crons.ts
```diff
@@ -0,0 +1,70 @@
+import { RepeatOptions } from '../interfaces/repeat-options';
+import { CronStrategy } from '../types/cron-strategy';
+import { parseExpression } from 'cron-parser';
+
+export interface CronStrategies {
+  [index: string]: CronStrategy;
+}
+
+export class Crons {
```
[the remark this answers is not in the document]
roggervalf: I think this file is from an old version of this pr, this class does not exist in the current state

docs/gitbook/guide/jobs/repeatable.md
```diff
@@ -108,3 +108,72 @@ For instance, let's say that you have a job that is repeated every second, but t
...
+  'Paint',
+  async () => {
+    doSomething();
+  },
+  { settings },
+);
+```
+
+{% hint style="warning" %}
+As you may notice, repeat strategy setting should be provided in queue and worker classes, in order to work as intended.
```
manast, line 174: Hmm, the strategy itself is only provided in the worker, right?

tests/test_repeat.ts
```diff
@@ -162,6 +164,7 @@ describe('repeat', function () {
         endDate: 12345,
         tz: null,
         cron: '10 * * * * *',
+        pattern: '10 * * * * *',
```
manast, line 167: should we allow both cron and pattern at the same time?

docs/gitbook/guide/jobs/repeatable.md
manast: Ok. So the reason we need in both places is because the first time we add the job to the queue we need to calculate when is the next iteration, but after that the Worker takes over and we use the worker settings. Maybe adding a clarification in the "hint" above would be good so that it does not leave people scratching their heads.

roggervalf: I used this comment as clarification xd

manast (approved):
LGTM

Issue #1231, opened 2022-05-05
Queue error: Ccnnection is closed.


Philluxx:
Hey there,

we we are adding jobs to a queue within an expressjs application. Therefore we are initializing the queue at startup.

Now we see almost regularly console log messages `Connection is closed.` If adding an Error-Listener this error is catched (stating the same error message). Nevertheless new jobs can be added to the queue successfully.

The same behaviour can be seen in a plain NodeJS application. This error does not appear for workers.

My question is: how to deal with this error. Adding an error listener with queue.on("error",...) and supressing log messages feels quite bad. Can anyone give me a hint to search for the cause? Is this a bug in BullMQ or an issue in IORedis?

Thank you so much and best regards!

```typescript
import { Queue } from 'bullmq'

(async () => {
     const queue = new Queue("test-queue",
        {
            connection: {
                host: "host",
                port: 6380,
                password: "topsecret",
                tls: { servername: "host" },
            },
        });

    queue.on("error", (error: Error) => {
        // Suppressing it feels bad...
        // console.log(error.message);
    });
})();

```

manast:
Are you closing the queue somewhere? The error seems to indicate that you tried to perform some operation on a closed connection.

Philluxx:
Hey, no the queue is not closed. I tested it now with different Redis servers. With a local redis in docker there are no error log messages. With an instance in Azure this will show the described behaviour.

What confuses me is that it still works - although the error message is thrown, I can still add new jobs to the queue.

If I don't want the log to be flooded with these messages, I have to suppress all error log messages, but then "real", relevant messages may no longer be displayed.

bicrypt:
same issue for me, local redis works, managed redis on digitalocean doesn't.

the console is spammed with these error messages since 1.80.4 as far as I can tell, since reverting to 1.80.3 resulted in 0 of these error messages thereafter and bumping the version up to 1.80.4 instantly caused these errors to reappear.

full stack trace for anything after 1.80.3:

```
Error: Connection is closed.
    at EventEmitter.RedisConnection.handleClientClose (/some/masked/path/node_modules/.pnpm/bullmq@1.80.4/node_modules/bullmq/dist/cjs/classes/redis-connection.js:54:32)
    at EventEmitter.emit (node:events:527:28)
    at EventEmitter.emit (node:domain:475:12)
    at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

[...]

PR #1251, opened 2022-05-20
fix(close): emit ioredis:close event instead of error

roggervalf:


bicrypt:
@roggervalf thank you for addressing issue #1231 🔥

PR #1252, opened 2022-05-21
perf(remove-job): send prefix key instead of jobKey

roggervalf:


src/commands/removeJob-1.lua
```diff
@@ -107,7 +63,7 @@ local function removeJob( prefix, jobId, parentKey)
     rcall("XADD", prefix .. "events", "*", "event", "removed", "jobId", jobId, "prev", prev);
 end
 
-local prefix = getJobKeyPrefix(KEYS[1], ARGV[1])
```
roggervalf, line 110: at least for the main job we could only take the queue prefix as key so we do not need to destructure the jobKey

manast (approved):
LGTM

Issue #1254, opened 2022-05-23
Getter test 'gets only workers related only to one queue' inaccurate if connection = new IORedis()


EarthlingDavey:
Hi first off, I want to thank the developers, maintainers and contributors for such a useful project.

2nd up I would like to report the only bug that I've seen so far in a couple of years use.

The error occured on my projct, but I can reproduce the result on the test suite.

I found that if a connection passed into bullmq is an IORedis instance.
Then, queue.getWorkers is incorrect. 

Here is a link to a fork with reproducion in the test suite: https://github.com/usedandloved/bullmq/commit/859b593c6a6f275ceeda85d0dd8ce10771b7a7f0

And this is the terminal log.

```
$ ts-mocha -p tsconfig-cjs.json --config ./.mocharc.js tests/test_getters.ts


  Jobs getters
    .getWorkers
      1) gets all workers for this queue only
      2) gets only workers related only to one queue


  0 passing (145ms)
  2 failing

  1) Jobs getters
       .getWorkers
         gets all workers for this queue only:

      AssertionError: expected [ { id: '849', …(26) } ] to have a length of 2 but got 1
      + expected - actual

      -1
      +2
      
      at Context.<anonymous> (tests/test_getters.ts:96:35)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)

  2) Jobs getters
       .getWorkers
         gets only workers related only to one queue:

      AssertionError: expected [] to have a length of 1 but got +0
      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (tests/test_getters.ts:117:31)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

manast:
Thanks for reporting this issue. Which version are you using?

EarthlingDavey:
No problem. My project is 1.83.0, also the linked repo is forked from master branch just after v1.83.0 tag.

EarthlingDavey:
Wow...that was fast. @manast @roggervalf thank you! 😍

PR #1255, opened 2022-05-24
fix(get-workers): use blockingConnection to set clientName

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1254


manast (approved):
LGTM

Issue #22, opened 2019-09-03
Feature-Request/Change the number of concurrent processes


manast:
The way concurrent works in BullMQ4.x makes it very trivial to implement:

https://github.com/OptimalBits/bull/issues/1439

PR #1256, opened 2022-05-24
feat(worker): change the number of concurrent processes

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/22

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -328,15 +328,15 @@ export class Worker<
 
           const processing = (this.processing = new Map());
 
-          while (!this.closing) {
-            const opts: WorkerOptions = <WorkerOptions>this.opts;
-            const tokens: string[] = Array.from(
-              { length: opts.concurrency },
-              () => v4(),
-            );
+          let concurrency = this.opts.concurrency;
+          const tokens: string[] = Array.from({ length: concurrency }, () =>
+            v4(),
+          );
 
-            if (processing.size < opts.concurrency) {
-              const token = tokens.pop();
+          while (!this.closing) {
+            const getNextJob = processing.size < concurrency;
+            if (getNextJob) {
+              const token = tokens.pop() || v4();
 
               processing.set(
                 this.retryIfFailed<Job<DataType, ResultType, NameType>>(
@@ -365,14 +365,19 @@ export class Worker<
               // reuse same token if next job is available to process
               processing.set(
                 this.retryIfFailed<void | Job<DataType, ResultType, NameType>>(
-                  () => this.processJob(job, token),
+                  () =>
+                    this.processJob(job, token, processing.size <= concurrency),
                   this.opts.runRetryDelay,
                 ),
                 token,
               );
             } else {
-              tokens.push(token);
+              if (getNextJob) {
+                tokens.push(token);
+              }
             }
+
+            concurrency = this.opts.concurrency;
           }
           this.running = false;
           return Promise.all([...processing.keys()]);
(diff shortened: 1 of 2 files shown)


src/classes/worker.ts
```diff
@@ -283,6 +283,10 @@ export class Worker<
     return this.blockingConnection.client;
   }
 
+  set concurrency(concurrency: number) {
+    this.opts = { ...this.opts, concurrency };
```
manast, line 287: why not simply ```this.opts.concurrency = concurrency```

src/classes/worker.ts
```diff
@@ -362,14 +365,19 @@ export class Worker<
...
-                  () => this.processJob(job, token),
+                  () =>
+                    this.processJob(job, token, processing.size <= concurrency),
                   this.opts.runRetryDelay,
                 ),
                 token,
               );
             } else {
-              tokens.push(token);
+              if (getNextJob) {
```
manast, line 375: I do not think this is correct. We only push the token back to the tokens array when there are no new jobs to process right now so that this same token can be reused in the future. It is not related to if we called getNextJob or not.

src/classes/worker.ts
```diff
@@ -362,14 +365,19 @@ export class Worker<
               // reuse same token if next job is available to process
               processing.set(
                 this.retryIfFailed<void | Job<DataType, ResultType, NameType>>(
-                  () => this.processJob(job, token),
+                  () =>
+                    this.processJob(job, token, processing.size <= concurrency),
```
manast, line 369: The problem here is that ```processing.size <= concurrency``` could be true or false at the time we call ```processJob```, however at the time we use the boolean ```fetchNext``` inside ```processJob``` the assertion may not be true or false anymore.

manast (changes requested):

src/classes/worker.ts
roggervalf: yes, but in this case I'm considerong when we decrease the concurrencia, so tokens should decrease as well

manast: Yes but you have asynchronous calls in between (line 363), so the "getNextJob" flag may not represent the same thing anymore at this point in the code, since the concurrency could have changed.

src/classes/worker.ts
roggervalf: this is because opts is a public attribute, in order to not affect external references is better to use this spread operator

src/classes/worker.ts
roggervalf: you are right, now I'm considering using a callback function

src/classes/worker.ts
manast: If that is the case, then we should copy opts already in the constructor, because otherwise if the user modifies "opts" outside of bull it would affect the worker's behaviour as a side effect.

src/classes/worker.ts
```diff
@@ -362,14 +361,18 @@ export class Worker<
               // reuse same token if next job is available to process
               processing.set(
                 this.retryIfFailed<void | Job<DataType, ResultType, NameType>>(
-                  () => this.processJob(job, token),
+                  () =>
+                    this.processJob(
+                      job,
+                      token,
+                      () => processing.size <= concurrency,
```
manast, line 368: here we need to use this.opts.concurrency, since the concurrency level could have been changed since it was assigned to the variable concurrency in line 375. I also think it is not needed to assing to a local concurrency variable, we can just use this.opts.concurrency.

manast (approved):

PR #1268, opened 2022-06-02
Fix the Queue.addBulk return signature

danew:
Currently the Queue.addBulk method's return type is incorrectly typed as
```ts
Promise<Job<DataType, DataType, NameType>[]>
```
Where it should be typed as 
```ts
Promise<Job<DataType, ResultType, NameType>[]>
```


roggervalf (approved):
LGTM, thanks

PR #1284, opened 2022-06-12
fix(queue): get rid of repeat options from default

### LINKED ISSUE #1282: repeat option is ignored when using "defaultJobOptions"
Despite being allowed to use the "repeat" option when instantiating a Queue with "defaultJobOptions", the repeat options are ignored when adding the job. Since having "repeat" as default option is quite unusual I think the solution. for this bug is just to remove the repeat option from the allowed defaultJobOptions.

roggervalf:
fixes #1282

### DIFF
--- a/src/interfaces/jobs-options.ts
+++ b/src/interfaces/jobs-options.ts
@@ -1,6 +1,6 @@
 import { RepeatOptions, KeepJobs, BackoffOptions } from './';
 
-export interface JobsOptions {
+export interface BaseJobOptions {
   /**
    * Timestamp when the job was created.
    * @defaultValue Date.now()
@@ -29,16 +29,6 @@ export interface JobsOptions {
   attempts?: number;
 
   /**
-   * Repeat this job, for example based on a `cron` schedule.
-   */
-  repeat?: RepeatOptions;
-
-  /**
-   * Base repeat job key.
-   */
-  repeatJobKey?: string;
-
-  /**
    * Rate limiter key to use if rate limiter enabled.
    *
    * @see {@link https://docs.bullmq.io/guide/rate-limiting}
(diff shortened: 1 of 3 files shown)


src/interfaces/jobs-options.ts
```diff
@@ -94,6 +69,38 @@ export interface JobsOptions {
...
+  /**
+   * Base repeat job key.
+   */
+  repeatJobKey?: string;
+
+  /**
+   * The number of milliseconds after which the job should be
+   * fail with a timeout error.
+   */
+  timeout?: number;
```
roggervalf, line 93: timeout, I think is not used, could I delete it?

manast: Not used. But I thinkn it is used in the Pro version. So we need to add it to Pro then.

src/interfaces/jobs-options.ts
```diff
@@ -94,6 +69,38 @@ export interface JobsOptions {
...
+export interface JobsOptions extends BaseJobOptions {
+  /**
+   * Repeat this job, for example based on a `cron` schedule.
+   */
+  repeat?: RepeatOptions;
+
+  /**
+   * Base repeat job key.
+   */
+  repeatJobKey?: string;
```
manast, line 87: Is this option really needed? I am thinking the user is not able to configure this key, it is computed automatically from the "repeat" option.

manast (commented):
just one comment

src/interfaces/jobs-options.ts
roggervalf: yes, in this case is like prevMillis option, I'll change the description as prevMillis saying that this is an internal param

manast (approved):

Issue #1294, opened 2022-06-24
Can we make queue.close() quick?


snawaz:
Currently, the `queue.close()` takes almost `0.5s`  to close, because `close()` eventually calls `waitForJob` which `delay()` for `100ms` currently (and that's being multiplied, eventually `.close()` becoming `0.5s` slow). 

Anyway, if `close()` has been called, can we detect this scenario and then avoid calling `delay()` in that case? Or maybe some better solution.. 🤔 

https://github.com/taskforcesh/bullmq/blob/5b1a71a6200b4b5c8e42f9dd618441c223ffe834/src/classes/worker.ts#L467-L472

In the `catch` block, the error is: `Error: Connection is closed` (I just added the log to see it).

Two weeks back I faced a similar issue with the `bull` library. However, now I've switched to `bullmq`. Lots of the implementation I see is very similar. So maybe the other thread would be helpful as well, so here it is:  https://github.com/OptimalBits/bull/issues/2372

snawaz:
PR created: https://github.com/taskforcesh/bullmq/pull/1295

snawaz:
Closed via https://github.com/taskforcesh/bullmq/pull/1295

PR #1295, opened 2022-06-25
fix: avoid calling delay() if queue is being closed

snawaz:
One easy way to fix this: https://github.com/taskforcesh/bullmq/issues/1294

Please advise if there is any better way.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -139,6 +139,7 @@ export class Worker<
   private waiting = false;
   private running = false;
   private blockTimeout = 0;
+  private beingClosed = false;
 
   protected processFn: Processor<DataType, ResultType, NameType>;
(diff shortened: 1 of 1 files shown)


src/classes/worker.ts
```diff
@@ -468,7 +469,9 @@ export class Worker<
       if (isNotConnectionError(<Error>error)) {
         this.emit('error', <Error>error);
       }
-      await this.delay();
+      if (!this.beingClosed) {
```
roggervalf, line 472: hi @snawaz, did you try to use !this.closing instead of this new attribute?

snawaz: @roggervalf : ah, that makes sense. I just removed the new variable and used `this.closing` itself.

Issue #1333, opened 2022-07-21
Last Job Failed In Flow Results in Flow Parent not being called


lukepolo:
I was doing some testing and found that if you create a flow 

```
import { FlowProducer } from 'bullmq';

// A FlowProducer constructor takes an optional "connection"
// object otherwise it connects to a local redis instance.
const flowProducer = new FlowProducer();

const flow = await flowProducer.add({
  name: 'renovate-interior',
  queueName: 'renovate',
  children: [
    { name: 'paint', data: { place: 'ceiling' }, queueName: 'steps' },
  ],
});
```

And the last job `ceiling`  (in this case the only job), fails the parent will not be called.

I would at least expect the parent to called `failed`

lukepolo:
as a temp fix for me i just added

```
.on("failed", async (job: Job, error) => {

        if(job.parent) {
          await job.remove();
        }
 
 ```

manast:
The problem is, sometimes a child fails but you want to retry it manually after fixing the reason for failure, so you would probably want to wait before failing the parent.

lukepolo:
I could see that , but not when you have millions of jobs , seems more like a `dev` type of thing rather than production thing.

manast:
It could be a new option "failParentOnChild" or something like that, and in that case the parent would fail as soon as any of the children fails, however it would be not possible to skip processing the rest of the children even if the parent has failed. What do you think @roggervalf ?

rosslavery:
This is a similar use-case to what we'd like.

Imagine a parent with 2 million children. There may be an acceptable number of failures that we'd like to continue running the parent, or we may want it to fail and also fail the parent.

Although @manast note about not being able to stop processing the remaining children isn't ideal for our use case 🤔

lukepolo:
@rosslavery Yah same here , failures are `ok`  (internet goes out for 10 minutes or something odd)

roggervalf:
> It could be a new option "failParentOnChild" or something like that, and in that case the parent would fail as soon as any of the children fails, however it would be not possible to skip processing the rest of the children even if the parent has failed. What do you think @roggervalf ?

it sounds good to me, this is still in my list

PR #1339, opened 2022-07-28
feat(job-options): add failParentOnFail

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1333


src/classes/job.ts
```diff
@@ -357,6 +384,26 @@ export class Job<
...
   }
 
+  private optsAsJSON(opts: JobsOptions = {}): RedisJobOptions {
+    const optionEntries = Object.entries(opts) as Array<
+      [keyof JobsOptions, any]
+    >;
+    const options = optionEntries.reduce<Partial<Record<string, any>>>(
+      (acc, item) => {
+        const [attributeName, value] = item;
+        if (attributeName === 'failParentOnFail') {
```
roggervalf, line 394: probably this new option should be placed into parent object option

src/commands/includes/moveParentFromWaitingChildrenToFailed.lua
```diff
@@ -0,0 +1,25 @@
+--[[
+  Function to recursively moves from waitingChildren to failed.
```
manast, line 2: "to recursively move"

src/commands/addJob-9.lua
```diff
@@ -72,6 +72,11 @@ if parentKey ~= nil then
   local parent = {}
   parent['id'] = parentId
   parent['queueKey'] = parentQueueKey
+
+  if opts['fpof'] ~= nil then
```
manast, line 76: Is it possible to move this code out of Lua and put it in Js instead?

src/commands/includes/moveParentFromWaitingChildrenToFailed.lua
```diff
@@ -0,0 +1,25 @@
...
+
+local function moveParentFromWaitingChildrenToFailed( parentQueueKey, parentKey, parentId, jobId, timestamp)
+  if rcall("ZREM", parentQueueKey .. ":waiting-children", parentId) == 1 then
+    rcall("ZADD", parentQueueKey .. ":failed", timestamp, parentId)
+    rcall("HMSET", parentKey, "failedReason", "child " .. jobId .. " failed", "finishedOn", timestamp)
+
+    local rawParentData = rcall("HGET", parentKey, "parent")
+
+    if rawParentData ~= false then
+      local parentData = cjson.decode(rawParentData)
```
manast, line 13: Now it is probably too late to change since it would imply a breaking change, but it would have been nice to use msgpack instead of json for more compact data.

src/commands/includes/moveParentFromWaitingChildrenToFailed.lua
```diff
@@ -0,0 +1,25 @@
+--[[
+  Function to recursively move from waitingChildren to failed.
+]]
+
+local function moveParentFromWaitingChildrenToFailed( parentQueueKey, parentKey, parentId, jobId, timestamp)
+  if rcall("ZREM", parentQueueKey .. ":waiting-children", parentId) == 1 then
```
manast, line 6: If the parent is failed, should we also fail the children, or will they just continue processing normally?

manast (commented):
Looks good, I wrote some comments. Also we would need a new section in the flows documentation about this case.

src/commands/includes/moveParentFromWaitingChildrenToFailed.lua
roggervalf: I will handle children failure in a separated pr

src/classes/job.ts
```diff
@@ -314,6 +319,26 @@ export class Job<
     return job;
   }
 
+  private static optsFromJSON(rawOpts?: string): JobsOptions {
```
manast, line 322: I think we could refactor optsFromJson and optsAsJson by using a mapping object, it is also more generic, something along this lines:
```ts
const optsDecodeMap = {
  fpof: "failParentOnFailure"
}

const optsEncodeMap = _.invertBy(optsDecodeMap);

private static translateJSONOptions(translateMap: optsDecodeMap | optsDecodeMap, rawOpts?: string): JobsOptions {
    const opts = JSON.parse(rawOpts || '{}');

    const optionEntries = Object.entries(opts) as Array<
      [keyof RedisJobOptions, any]
    >;

    const options: Partial<Record<string, any>> = {};
    for (const item of optionEntries) {
      const [attributeName, value] = item;
      if(optsDecodeMap[attributeName]) {
        options[optsDecodeMap[attributeName]] = value;
      } else {
        options[attributeName] = value;
      }
    }

    return options as JobsOptions;
  }

```

manast (approved):

Issue #201, opened 2020-05-18
Provide ESM


arlac77:
As node now supports ESM natively. Can You please provide ESM directly.

By compiling ts into es2015

```
tsc --module es2015 --target es2015
```

```sh
  import { Queue, Worker } from "bullmq";
           ^^^^^
  SyntaxError: The requested module 'bullmq' does not provide an export named 'Queue'
```

manast:
How can we do that without breaking older versions of node?

arlac77:
[Conditional Exports](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_conditional_exports)

PR #1346, opened 2022-08-02
chore: allow esm usage

boredland:
By creating this PR I hope we can talk about remaining issues and brainstorm ideas to maybe get rid of some typecasts.

this heavily relies on dropping node 12 support. which has been eol for nearly two years (20 Oct 2020).

superseeds https://github.com/taskforcesh/bullmq/pull/1173
contributes to https://github.com/taskforcesh/bullmq/discussions/1305
closes https://github.com/taskforcesh/bullmq/issues/201

also included:
- package: bumped all dependencies to current versions
- package: use cpy-cli instead of copyfiles for nicer syntax
- package: conditional exports
- package: removed duplicate declarations
- package: added npmignore to reduce packaged bloat
- package: moved auto yarn build to prepublishOnly
- ci: added more useful node versions for ci tests (current lts, previous lts, current version)
- ci: added more useful redis versions for ci tests (5,6,7)
- ci: use global node cache instead of actions/cache
- ci: collect coverage for current lts and current stable redis
- ci: pinned most action versions to sha
- ci: merged coverage and node-tests into one workflow
- ci: auto cancel concurrent test runs
- ci: removed recursing into submodules as afaik there are none
- ci: [set up renovate](https://github.com/taskforcesh/bullmq/discussions/1348?notification_referrer_id=NT_kwDOAEcl3LI0MTE4Njg0OTc3OjQ2NjI3NDg#discussioncomment-3315218) and pinned github actions


src/classes/redis-connection.ts
```diff
@@ -69,7 +69,7 @@ export class RedisConnection extends EventEmitter {
         const hosts = (<any>this._client).startupNodes.map(
           (node: { host: string }) => node.host,
         );
-        this.checkUpstashHost(this._client.options.redisOptions?.host || hosts);
```
boredland, line 72: this removal provokes an error in the test suite:
```
1) connection
       when host belongs to Upstash
         when using redisOptions
           throws an error:
     AssertionError: expected [Function] to throw an error
      at Context.<anonymous> (tests/test_connection.ts:96:72)
      at processImmediate (node:internal/timers:466:21)
```

I removed this, because "host" does not exist on the type of redisOptions.

package.json
```diff
@@ -68,8 +68,9 @@
     "@types/chai": "^4.2.22",
     "@types/chai-as-promised": "^7.1.4",
     "@types/glob": "^7.2.0",
-    "@types/ioredis": "^4.28.2",
     "@types/lodash": "^4.14.173",
+    "@types/lodash.defaults": "^4.2.7",
```
boredland, line 72: without this I got:

```

node_modules/ioredis/built/utils/lodash.d.ts:1:27 - error TS7016: Could not find a declaration file for module 'lodash.defaults'. '/workspaces/bullmq/node_modules/lodash.defaults/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/lodash.defaults` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash.defaults';`

1 import defaults = require("lodash.defaults");
```

package.json
```diff
@@ -68,8 +68,9 @@
...
+    "@types/lodash.isarguments": "^3.1.7",
```
boredland, line 73: without this I got:
```
node_modules/ioredis/built/utils/lodash.d.ts:2:30 - error TS7016: Could not find a declaration file for module 'lodash.isarguments'. '/workspaces/bullmq/node_modules/lodash.isarguments/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/lodash.isarguments` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash.isarguments';`

2 import isArguments = require("lodash.isarguments");
```

src/classes/redis-connection.ts
boredland: ah. looked into it and its just that one test not working due to upstream api changes. Will remove it.

boredland:
sooo this escalated a bit. I could well split things out. Feedback very much welcome!

boredland:
my changes CI setup brought the execution time for tests down to 3m 36s in the [last run](https://github.com/boredland/bullmq/actions/runs/2788854945), coming from 6m 17s with the [current setup](https://github.com/taskforcesh/bullmq/actions/runs/2788526200).

boredland:
@manast could you have a look here? if its too big, I am happy to split it up into multiple PRs.

src/classes/child-pool.ts
```diff
@@ -22,7 +21,7 @@ const convertExecArgv = async (execArgv: string[]): Promise<string[]> => {
       standard.push(arg);
     } else {
       const argName = arg.split('=')[0];
-      const port = await getPort();
+      const port = (await import('get-port')).default();
```
manast, line 24: why this change?

package.json
```diff
@@ -2,11 +2,16 @@
...
   "description": "Queue for messages and jobs based on Redis",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "jsnext:main": "dist/esm/index.js",
-  "types": "dist/esm/index.d.ts",
-  "source": "src/index.ts",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/esm/index.d.ts",
+  "exports": {
```
manast, line 8: There is some risk that this will affect existing applications depending on BullMQ, as it may affect how to import the library.

package.json
```diff
@@ -2,11 +2,16 @@
   "name": "bullmq",
   "version": "1.87.0",
   "description": "Queue for messages and jobs based on Redis",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "jsnext:main": "dist/esm/index.js",
-  "types": "dist/esm/index.d.ts",
-  "source": "src/index.ts",
```
manast, line 9: Why remove "source" ?

src/classes/redis-connection.ts
manast: @roggervalf do you remember why the need for checking host on redisOptions?

manast (commented):
Thanks for all these nice improvements and additions. As you mentioned, it would be much easier to merge these changes if divided in smaller units, and even easier to merge the units that do not require a breaking change such as upgrading the node version to >= 14. Even though node 12 is EOL, there are users running production systems with it and if we require an upgrade we may break their deployments, therefore we cannot upgrade in a minor release, but a major and with a BREAKING CHANGE notice. 
I wrote some comments in the PR, and here I list the units I identified:
- Upgrade dependencies (that do not need code changes)
- add "exports" to package.json (may be a breaking change not sure).
- Redis upgrade.
- Add renovate.
- Upgrade to node 14.
- Use native uuid instead of external dependency (requires node 14).
- upgrade docker-compose
- New test.yml github workflow (not sure why it is better to have coverage.yml and node.js.yml merged into one file though).
- Changes in husky configuration.
- Uppercase Redis commands.

I may have missed something but this is what I found so far.

src/classes/child-pool.ts
boredland: get-port is esm only and this is the recommended workaround to use it in commonJs.

package.json
boredland: ah, probably just a mistake. will readd it.

package.json
boredland: relatively sure :tm: that this wont be the case, as explained [here](https://github.com/taskforcesh/bullmq/pull/1361)

src/classes/child-pool.ts
manast: But I thought using import * was the recommended approach, has this changed recently?

package.json
manast: If it is merged as a separate PR it could be rolled-back if we notice it start breaking things.

src/classes/child-pool.ts
boredland: na, the problem is afaik that we're building a cjs module too and are not using a cool build tool like tsup/esbuild. That way to be able to build a cjs module, we have to stick to this [commonjs-compatible workaround](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) using a dynamic import.

boredland: I can gladly try this when we rebased this PR though, as I'd prefer it without too.

package.json
boredland: d'accord https://github.com/taskforcesh/bullmq/pull/1361

src/classes/redis-connection.ts
roggervalf: hey, because host is an available param in redisOptions, it can contain Upstash host that is not working well with bullmq

src/classes/child-pool.ts
boredland: trying this out here: https://github.com/taskforcesh/bullmq/pull/1374

boredland:
closing in favor of https://github.com/taskforcesh/bullmq/pull/1346

PR #1357, opened 2022-08-13
fix(clean): consider priority when cleaning waiting jobs

roggervalf:


src/commands/cleanJobsInSet-2.lua
```diff
@@ -32,11 +32,11 @@ end
 
 local result
 if ARGV[4] == "active" then
-  result = cleanList(KEYS[1], ARGV[1], rangeStart, rangeEnd, ARGV[2], false)
+  result = cleanList(KEYS[1], ARGV[1], rangeStart, rangeEnd, ARGV[2], false, false)
```
roggervalf, line 35: as both last attributes are false and in wait/paused case, both are true, probably we could combine both attributes in one, like isNotActive, wdyt @manast

manast: Or maybe just "isWaiting" (being paused actually irrelevant, as it is still the wait list, just in a different state).

manast (approved):

PR #1358, opened 2022-08-14
chore: bump non-breaking dependency versions

boredland:


boredland:
@manast updated it to update typescript-eslint too, as that was having problems with current eslint.

PR #1360, opened 2022-08-14
chore: use some unused types properly

boredland:
while browsing the files I saw those generics were not used at all and hence pretty useless. I suppose this is what they were supposed to do.


manast:
The reason of the lacking of types is because the compat.ts layer is deprecated since it did not gain any adoption. But lets merge this just for completeness.

PR #1362, opened 2022-08-14
chore: de-duplicate types

boredland:
When looking at https://github.com/taskforcesh/bullmq/pull/1361, I realized

1. that the types are emitted twice and used once and 
2. that the package was including lots of unrelated things.

Solves 1. by only emitting types for esm which is linked to by the "types" field.

Solved 2. by adding a .npmignore that omits everything not needed to use the package.


package.json
```diff
@@ -20,10 +20,11 @@
     "dist"
   ],
   "scripts": {
-    "build": "npm-run-all tsc:all copy:lua copy:includes:lua clean:temp:files",
-    "clean:temp:files": "rimraf dist/cjs/bullmq.d.ts dist/esm/bullmq.d.ts dist/tsdoc-metadata.json",
-    "copy:lua": "copyfiles -f ./src/commands/*.lua ./dist/cjs/commands && copyfiles -f ./src/commands/*.lua ./dist/esm/commands",
-    "copy:includes:lua": "copyfiles -f ./src/commands/includes/*.lua ./dist/cjs/commands/includes && copyfiles -f ./src/commands/includes/*.lua ./dist/esm/commands/includes",
+    "build": "run-p build:*",
```
boredland, line 23: this speeds up the build, because both tsc commands can be run in parallel

manast, line 23: This looks cleaner, however, shouldn't we also have a clean script?

manast: For CI it wont matter but when building locally it does.

boredland: hey! I discovered that tsc now has a [built-in cleaning](https://raghsonline.com/posts/typescript/typescript-clean-build/) tooling. I don't like their ergonomics (would have thought `--build --clean` cleans and builds, but in fact it just cleans.

boredland:
hi @manast is there anything left stopping this from being merged?

manast:
So there is no risk for incompatibilities? the types files are identical?

boredland:
> So there is no risk for incompatibilities? the types files are identical?

if they hadn't been identical, we would have seen issues before, wouldn't we? :wink:

boredland:
> So there is no risk for incompatibilities? the types files are identical?

perhaps I should be more clear about this: it is still the same esm configuration generating the types. I just changed it so the declaration is emitted only once. Before it has been emittet for esm and cjs, but used (read: referenced in the package.json) only once.

roggervalf:
this was merged in another pr having this pr as reference, thanks @boredland

PR #1359, opened 2022-08-14
ci: add renovate configuration

boredland:
as discussed in https://github.com/taskforcesh/bullmq/discussions/1348?notification_referrer_id=NT_kwDOAEcl3LI0MTE4Njg0OTc3OjQ2NjI3NDg#discussioncomment-3315218

PR #1363, opened 2022-08-14
chore: streamline ci setup

boredland:
1. pinned down actions to their commit hashes. this is a basic security practice, as relying on tags may introduce real risks. [renovate](https://github.com/taskforcesh/bullmq/pull/1359) will be able to update them.
2. use yarn global cache via `actions/setup-node` instead of `actions/cache`, as yarn is very efficient when using the global cache and the whole "download, extract, use" workflow of `actions/cache` takes up significantly more time
3. merged the "coverage" and the "nodejs" workflow into a "test" workflow, as both run tests and using multiple jobs we can gain concurrency anyway. I just did not see any benifit in having that in two different workflows.
4. moved the automated `yarn build` from the `prepare` to the `prepublishOnly` script, as prepare runs in a lot of occasions while I suppose this is meant to explicitly be done before publishing.
5. added some different redis versions to run tests against
6. instead of explicit versions we now use the lts/* notation of `actions/node`. range is from lts/-2 to current version, the current lts is tested against all configured versions of redis. lts/-2 (node 12 right now) IMO should not be actively supported, but removing that we can discuss when we discuss other upgrades.
7. switched to alpine redis containers, as they're smaller and thus a bit faster to download
8. made redis 7 the default container to test against, as I don't see a reason to stay at 6, as 7 is [downwards compatible](https://redis.com/blog/redis-7-generally-available/).

you can see it in action [here](https://github.com/boredland/bullmq/pull/10).


.github/workflows/api.yml
```diff
@@ -9,14 +9,15 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [lts/*]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3
```
manast, line 15: Why the reference to a git sha instead of a given version?

manast: aha, ok, I saw the comment on the PR now.

manast:
It is possible that this PR has  broken the release script, it hangs forever on the sematic-release/git: 
https://github.com/taskforcesh/bullmq/runs/7862461250?check_suite_focus=true

PR #1365, opened 2022-08-14
ci: set up nyc coverage limits and coveralls action

boredland:
- uses coveralls github action to submit coverage
- sets nyc limits to realistic values
- adds a check how much a PR lowered the coverage

you can see it in action [here](https://github.com/boredland/bullmq/pull/14).

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,7 @@
     "clean:temp:files": "rimraf dist/cjs/bullmq.d.ts dist/esm/bullmq.d.ts dist/tsdoc-metadata.json",
     "copy:lua": "copyfiles -f ./src/commands/*.lua ./dist/cjs/commands && copyfiles -f ./src/commands/*.lua ./dist/esm/commands",
     "copy:includes:lua": "copyfiles -f ./src/commands/includes/*.lua ./dist/cjs/commands/includes && copyfiles -f ./src/commands/includes/*.lua ./dist/esm/commands/includes",
-    "coverage": "nyc --reporter=text npm run test",
-    "coveralls": "nyc report --reporter=text-lcov | coveralls",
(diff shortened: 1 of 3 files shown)


boredland:
well well well. it seems coveralls isn't even set up? :joy:

boredland:
in coveralls I'd recommend these settings:

image

in github I'd recommend these settings:

image

package.json
```diff
@@ -120,7 +120,29 @@
     ],
     "exclude": [
       "bullmq-tests/test_*.ts"
-    ]
+    ],
+    "lines": 80,
```
manast, line 124: Are these settings related to coveralls?, I could not find documentation about them, are they supposed to be in the root of the package.json?

.github/workflows/coveralls.yml
```diff
@@ -0,0 +1,48 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests and submit coverage to coveralls
+# Since submitting the coverage needs the COVERALLS_REPO_TOKEN secret, this action runs in a dedicated pipeline
```
manast, line 2: This comment does not seem to be up-to-date

boredland: hi! I am currently working out a solution that allows to have a coverage feedback for guests like me. will keep you posted ;)

package.json
boredland: No, this has nothing to do with coveralls, but with nyc (and isn't in the root, but inside the nyc field). For some time now, the bullmq repo has not been reporting to coveralls, because nyc has failed before any coveralls submission happened, because the default nyc threshold of 80% has been hit with the branch coverage being near 70%.

Issue #1368, opened 2022-08-15
Error [ERR_REQUIRE_ESM]: require() of ES Module


AlexOwl:
Using ES module (type: module in package.json) as Worker throws the error

```
Error [ERR_REQUIRE_ESM]: require() of ES Module <project>/worker.js from <project>/node_modules/bullmq/dist/cjs/classes/child-processor.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in <project>/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

AlexOwl:
Current solution

Add script to `package.json`:
```
…
"scripts": {
    "FIX_ERR_REQUIRE_ESM": "node -e \"['cjs','esm'].map(a=>'./node_modules/bullmq/dist/'+a+'/classes/child-processor.js').map(a=>require('fs').writeFileSync(a,require('fs').readFileSync(a,{encoding:'utf8'}).replace('require(processorFile)','await import(processorFile)')))\""
},
…
```

Run it with `npm run FIX_ERR_REQUIRE_ESM`

redtogin:
Does anyone know the origin of the issue? Is it from recent bullmq version or some interaction with new version of nodejs/babel/etc?

```
17 Aug 2022 18:10:53.596202 <190>1 2022-08-17T16:10:53.216421+00:00 app web.1 - - (node:23) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
17 Aug 2022 18:10:53.596165 <190>1 2022-08-17T16:10:53.216446+00:00 app web.1 - - (Use `node --trace-warnings ...` to show where the warning was created)
17 Aug 2022 18:10:53.596138 <190>1 2022-08-17T16:10:53.225145+00:00 app web.1 - - /app/node_modules/bullmq/dist/esm/index.js:1
17 Aug 2022 18:10:53.596120 <190>1 2022-08-17T16:10:53.225146+00:00 app web.1 - - export * from './classes';
17 Aug 2022 18:10:53.596100 <190>1 2022-08-17T16:10:53.225147+00:00 app web.1 - - ^^^^^^
[...]

Issue #917, opened 2021-12-06
bullmq@1.56.0: SyntaxError: Named export 'Queue' not found


Niek:
Upgrading BullMQ from 1.55.1 to 1.16.0 results in this error:

```js
import { Queue } from 'bullmq'
         ^^^^^
SyntaxError: Named export 'Queue' not found. The requested module 'bullmq' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'bullmq';
const { Queue } = pkg;
```

Changing my code to this worked indeed, but it's not ideal:
```js
import bull from 'bullmq'
const { Queue } = bull;
```

I can't find which commit caused it, but it would be nice to have this fixed.

Niek:
It might be related to the TS 3->4 bump in https://github.com/taskforcesh/bullmq/commit/d2897ee7bbf4aee5251ac4fb28705f2bebbe7bfe

manast:
Interesting... did you test by upgrading your local typescript version too? Could it be that typescript introduced an incompatibility between version 3 and 4?

Niek:
> Interesting... did you test by upgrading your local typescript version too? Could it be that typescript introduced an incompatibility between version 3 and 4?

This is in a project not using TS at all, but ES imports.

manast:
@roggervalf maybe it is related to this line in tsconfig.json:
```
    "esModuleInterop": false,
```

daeschlerdavid:
Same for me.
Mixing commonJS and ESM is so much pain...
Lot of my dependency already did the switch to ESM, why not bullmq ?

manast:
@daeschlerdavid not sure what you mean. In BullMQ we only use ESM when importing modules.

manast:
Or put in another way. I do not know why your compiler is complaining that BullMQ is not an ESM module.

manast:
You can ignore my two last comments. It is properly done now, hope it works for you now.

daeschlerdavid:
still getting this error after TS compilation

```
import { QueueEvents, Job } from 'bullmq';
                      ^^^
SyntaxError: Named export 'Job' not found. The requested module 'bullmq' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'bullmq';
const { QueueEvents, Job } = pkg;
```

Niek:
[...]

PR #1369, opened 2022-08-15
Fix Error [ERR_REQUIRE_ESM]

AlexOwl:
Fix of the issue #1368

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -24,14 +24,20 @@ export class ChildProcessor {
   public async init(processorFile: string): Promise<void> {
     let processor;
     try {
-      processor = require(processorFile);
-    } catch (err) {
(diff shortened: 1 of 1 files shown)


manast:
Looks un-harmful but I have been biten several times with ESM support so I am not 100% this wont break somebody's deploy, need to be prepared to a revert if that happens.

manast:
Reverted since as I suspected this is already creating errors on current users. This is my recommended approach that just works: https://blog.taskforce.sh/using-typescript-with-bullmq/

AlexOwl:
I got the problem, better to check if current environment is ESM compatible then use import

roggervalf:
hi @AlexOwl, it should be related to https://github.com/taskforcesh/bullmq/issues/917, please try version 1.89.1

Issue #1160, opened 2022-03-27
changeDelay does not update job properties to reflect new expected execution time


hoangbn:
After fixing https://github.com/taskforcesh/bullmq/issues/787,  job executes correctly after changing delay now.
However,  I don't see the properties of the job updated to reflect the new execution time

For example, if I create a job:
```
  id: '123',
  name: 'someJob',
  data: { ... },
  opts: {
    delay: 20000,
    jobId: '123',
  },
  timestamp: 1648409394575
```
I calculate its earliest execution time to be: `1648409394575 + 20000`

After changing the job delay using changeDelay to say `40000`, I expect these properties to change so I can calculate the new execution time to be `now() + 40000` (currently nothing changes).

wenq1:
Upon the time you call now () + 40000, now () has already deviated from the calling time of changeDelay ().

I think delay is just an "suggestion". Maybe it is better for something like job._expected_start_date which is always fixed to be exposed. That doesn't prevent the 'delay' opts in this question to be changed thou.

Issue #1332, opened 2022-07-21
Support for ioredis v5


kaname-png:
Are there plans for bullmq to support ioredis v5?

kaname-png:
The ability to pass a username in the connection URI, also not rely on @types/ioredis types and compatibility with redis v7.

soulchild:
@manast I just happened to notice that the latest bullmq version still contains `@types/redis@4` in the [devDependencies](https://github.com/taskforcesh/bullmq/blob/master/package.json#L71), but ioredis@5 provides its own types now. I had some problems with that combination in the past and was just wondering whether it'd make sense to just remove `@types/redis` altogether? Or is there a specific reason why the types are still included?

PR #1389, opened 2022-08-20
fix(job): update delay when changeDelay

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1332
fixes https://github.com/taskforcesh/bullmq/issues/1160

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -457,15 +457,14 @@ export class Scripts {
   async changeDelay(jobId: string, delay: number): Promise<void> {
     const client = await this.queue.client;
 
-    const delayTimestamp = Date.now() + delay;
-    const args = this.changeDelayArgs(jobId, delayTimestamp);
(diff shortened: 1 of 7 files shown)


src/classes/scripts.ts
```diff
@@ -457,34 +457,33 @@ export class Scripts {
...
-  private changeDelayArgs(jobId: string, timestamp: number): string[] {
+  private changeDelayArgs(jobId: string, delay: number): (string | number)[] {
     //
     // Bake in the job id first 12 bits into the timestamp
     // to guarantee correct execution order of delayed jobs
     // (up to 4096 jobs per given timestamp or 4096 jobs apart per timestamp)
     //
     // WARNING: Jobs that are so far apart that they wrap around will cause FIFO to fail
     //
-    timestamp = Math.max(0, timestamp);
```
roggervalf, line 476: this logic is not needed as timestamp at least will have Date.now() value that should be >0

Issue #248, opened 2020-08-02
Undefined/empty jobs


jaschaio:
I am having a lot of failed jobs, but when I try to get them via:

```js
const jobs = await queue.getJobs( [ 'failed' ] );
```

They are returned as `undefined`. e.G.:

```js
[ undefined, undefined, undefined, ... ]
```

As I can't get the job ID this way, I double checked which keys it is trying to get by setting `process.env.DEBUG = "ioredis:*";`. When I manually check one of the keys that returns an `undefined` job via the `redis-cli` I get an empty result:

```
HGETALL bull:short:825F243AD7000000CF2B022C0100296E5A100417BA75E26AFC434291F746A7608C065246645F696400645F243AD7DE5C4A0015CB2CDD0004
(empty array)
```

Any reason how this could happen or how I could investigate further?

Anushil98:
same here please help !!!!!

manast:
Can you provide some code that reproduces the issue?

Anushil98:
Our team used Bull to solve the issue we had used a similar code snippet as mentioned in the first comment.
In our use case we tried to use the getJobs and getWaiting  for fetching the last n jobs in the queue inside a cron job for a certain purpose. The getWaitingCount gives us the correct job count but when we use the getJobs we get an array of undefined jobs. The main issue here is that this exact use case works in bull but not in bullmq.

Code couldn't be shown due to confidential information.
Thanks for responding.

manast:
you do not need to provide code that is confidential, just a test case that demonstrates the issue. Thing is, we have tests that cover this case and it works, so it must be something else that affects the results:
https://github.com/taskforcesh/bullmq/blob/master/src/test/test_getters.ts

Mikkal24:
I ran into a similar issue with the 'repeat' job type. I expected getJobs to return an array of jobs but, instead I got an array of 'undefined' which is weird. 

I wrote a test inside `test_getters.ts` like manast suggested to make sure I'm not crazy.

```  it('should return repeat jobs', async function(done) {
    try {
      await queue.add(
        'test',
        { foo: 1 },
        {
          repeat: {
            every: 15000,
          },
        },
      );

      const jobs = await queue.getJobs('repeat');
      expect(jobs[0]?.data?.foo).to.equal(1);
    } catch (e) {
      return done(e);
    }
  });
  ```

Anyway,  I hope this issue gets resolved soon since I'd like more data than what the `getRepeatableJobs()` call gives me.

[...]

PR #1395, opened 2022-08-29
fix(queue-getters): getting undefined when getting repeat jobs

ATheCoder:
This will fix https://github.com/taskforcesh/bullmq/issues/248

This happened because `repeat` jobs where actually considered to be delayed jobs when using the queue's `getJobs` function.

I also added a test for it.

### DIFF
--- a/tests/test_repeat.ts
+++ b/tests/test_repeat.ts
@@ -1174,4 +1174,16 @@ describe('repeat', function () {
     await worker.close();
     await queueScheduler.close();
   });
+
+  it('should get repeatable jobs using the getJobs method of the queue', async function() {
+    const queueScheduler = new QueueScheduler(queueName, { connection });
+    await queueScheduler.waitUntilReady();
+
+    await queue.add('test', { foo: 'bar' }, { repeat: { every: 1000 } });
+
+    const queueJobs = await queue.getJobs(); // right: 'repeat:5b8589ce9156aede92c064e935c4edcf:1000'
+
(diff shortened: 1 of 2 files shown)


ATheCoder:
@manast [this post](https://github.com/taskforcesh/bullmq/issues/248#issuecomment-804612022) in #248  is exactly the reproduction of this issue.

tests/test_repeat.ts
```diff
@@ -1174,4 +1174,16 @@ describe('repeat', function () {
     await worker.close();
     await queueScheduler.close();
   });
+
+  it('should get repeatable jobs using the getJobs method of the queue', async function() {
```
manast, line 1178: Is this test not already passing before this PR?

manast:
@ATheCoder thanks for the PR. 

The thing is that "delayed" jobs are not necessary repeat jobs, only the next iteration of the repeat job will be a delayed job. So this fix is not really semantically correct. We have a getter that returns the "meta" part of the repeatable jobs though.
In my opinion what this PR solves could be actually be more correctly resolved by better documentation...

ATheCoder:
@manast I totally agree. However, using the `getRepeatableJobs` method will only give the *meta data* of repeatable jobs which is not enough for some use cases where it is important to have all the data.

Do you have any ideas as to what would be a good fix that I can implement?

ATheCoder:
I also tried using `getRepeatableJobs` along with `getJob`. But, that did not work either. Because the `id` field was missing on the metadata provided by `getRepeatableJobs`.

PR #1413, opened 2022-09-14
feat: eliminate the need of having a QueueScheduler

manast:
…Scheduler

BREAKING CHANGE:
failed and stalled events are now produced by the Worker class instead of by the QueueScheduler.


src/classes/worker.ts
```diff
@@ -571,25 +580,29 @@ export class Worker<
...
     const handleCompleted = async (result: ResultType) => {
-      const completed = await job.moveToCompleted(
-        result,
-        token,
-        fetchNextCallback() && !(this.closing || this.paused),
-      );
-      this.emit('completed', job, result, 'active');
-      const [jobData, jobId] = completed || [];
-      return this.nextJobFromJobData(jobData, jobId);
+      if (!this.closing) {
```
roggervalf, line 583: there is a test that prevents this change https://github.com/taskforcesh/bullmq/blob/master/tests/test_worker.ts#L78-L110

src/classes/worker.ts
```diff
@@ -571,25 +580,29 @@ export class Worker<
...
     const handleFailed = async (err: Error) => {
-      try {
-        await job.moveToFailed(err, token);
-        this.emit('failed', job, err, 'active');
-      } catch (err) {
-        this.emit('error', <Error>err);
-        // It probably means that the job has lost the lock before completion
-        // The QueueScheduler will (or already has) moved the job back
-        // to the waiting list (as stalled)
+      if (!this.closing) {
```
roggervalf, line 596: there is a test that prevents this change https://github.com/taskforcesh/bullmq/blob/master/tests/test_worker.ts#L41-L76

manast: Interesting. But it feels more correct to not try to access Redis when it is highly likely it will fail since the connection is closing :/

src/classes/worker.ts
manast: Fixed by checking ```this.connection.closing``` instead

roggervalf (approved):
LGTM

PR #1421, opened 2022-09-20
fix(compat): remove Queue3 class

roggervalf:
BREAKING CHANGE: Worker and Queue classes should be used instead


manast:
I think the proper comment must have a carry return after "BREAKING CHANGE:", this can be fixed if we merge with Squash instead of rebase.

malsabbagh:
Hi @roggervalf, @manast do you guys recommend updating first to a package with Queue3 when migrating from Bull 3.x? Or should we move to the latest bullmq package and try to migrate with that?

manast:
If you are migrating from Bull to the latest BullMQ I recommend you use new queues instead of reusing the old ones as the code bases have diverged a bit, even though the data structures are very similar.
Here some advice in case you haven't read it yet: https://docs.bullmq.io/guide/migration-to-newer-versions

PR #1422, opened 2022-09-20
fix(sandbox): remove progress method

roggervalf:
BREAKING CHANGE: updateProgress should be used instead


manast (approved):
I think a carry return is required after "BREAKING CHANGE:"

Issue #921, opened 2021-12-09
Throw exception if trying to use a redis instance with a different eviction policy than "noeviction"


manast:
This can be achieved by issuing the "info" command just after connecting and checking the line [1]:
```
maxmemory_policy:allkeys-lru
```

[1] https://redis.io/commands/info
[2] https://redis.io/topics/lru-cache

kingston:
Is there a reason why volatile-*, e.g. volatile-lfu, policies should not be allowed? They behave like noeviction unless the key is set to expire (useful for caches where they can be evicted if memory pressure requires) and as far as I can tell, BullMQ does not use expiring keys. https://redis.io/docs/manual/eviction/

manast:
BullMQ does use expiring keys for several things actually, for example, locks on jobs. We have however relaxed this requirement on the library since some hosting providers do not allow us to gather the maxmemory policy, still, it is highly recommended to use noeviction to avoid strange behaviours down the road.

PR #1429, opened 2022-09-21
fix(connection): throw error when no noeviction policy

roggervalf:
BREAKING CHANGE: noeviction maxmemory-policy is the only option allowed
    
re #921


kevcodez:
Is there a way to ignore this? We're aware of the possible side effects here, but since other services write into the same Redis instance, we cannot simply use a different maxmemory-policy.

Issue #1437, opened 2022-09-22
Config Get maxmemory policy not possible on elasticache


autarchprinceps:
On Elasticache you cannot do a Config Get, so the check for noeviction vs volatile-lru always fails and breaks everything.
This is an issue going from 1.91.1 to 2.0.1. I have setup noeviction as policy, but in AWS ElastiCache Parameter Group settings.
Message: Eviction policy is volatile-lru. It should be "noeviction"
Could you provide a method to surpress this check, so we can use version 2?

manast:
Ok. I guess the best, in this case, is to handle the error and in such case just output a warning that the" maxmemory" policy could not be determined. @roggervalf can you look into this?

manast:
Error message could be something along the lines: "maxmemory policy could not be determined, please make sure you are using 'noeviction' to avoid issues"

patrickrbc:
It looks like this issue is preventing all ElastiCache users from updating to BullMQ's latest version. Is there a workaround or does it needs to be fixed here?

Is there anything blocking the fix from #1440?

roggervalf:
hi @patrickrbc, that pr should not address this issue, as it check the maxMemory policy presence and from the description of this issue, maxMemory policy is present but having volatile-lru. I'm wondering if elasticache is not reflecting this value and just showing volatile-lru as default, could someone clarify this behavior?

PR #1440, opened 2022-09-23
fix(connection): apply console.warn in noeviction message

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1437


hmvs:
Can we please stop writing to the console (breaks structured logging)? Or can we provide some flag that this message will not be logged?

manast:
The best would be able to provide the possibility of defining your own log functions, and by default using console.log/warn/error.

PR #1441, opened 2022-09-23
perf(scripts): pre-build scripts

roggervalf:


manast:
Instead of committing the built scripts is better to build them on the fly when running the tests and also build them when doing a new release.

commandTransform.js
```diff
@@ -0,0 +1,65 @@
+const path = require('path');
+const fs = require('fs');
+const { promisify } = require('util');
+const { argv } = require('process');
+
+const readFile = promisify(fs.readFile);
```
manast, line 6: It is not needed to promisify this functions as node also provides a version that returns a promise: https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_readfile_path_options

manast (commented):
Looks good, but it seems like for testing it will not work as before, since we would need to manually build before we run the tests, so I guess the test script needs to generate the scripts before running too.

sebmellen:
@roggervalf @lukepolo any luck on fixing this? Can we help get a PR in? We are having the same issue as @lukepolo right now, with this arg ID -> `4d7b090b56fe5384faa95e756e1375277faea973`

(CC @Pehesi97, @byennen, @SarjuHansaliya, @LucianBuzzo)

roggervalf:
hi @sebmellen there a pr that is ready, but we are having issues with our pipeline in pro versión, we are still fixing it, I'll let you know

sebmellen:
@roggervalf Thanks for the response! Do you know if there are any short-term workarounds we can try? This issue is really hurting our k8s cluster.

lukepolo:
@sebmellen , what we are doing for now is not sharing our connection params , it forces it to use a new connection and seems to make it work correctly

PR #1446, opened 2022-09-26
Get open port using built-in module instead of get-port

### LINKED ISSUE #1424: Error: Must use import to load ES module
Hello,

I'm seeing this error when a job fails while using `waitUntilFinished`:
```
Error: Must use import to load ES Module: /usr/src/app/node_modules/bullmq/node_modules/get-port/index.js
require() of ES modules is not supported.
require() of /usr/src/app/node_modules/bullmq/node_modules/get-port/index.js from /usr/src/app/node_modules/bullmq/dist/cjs/classes/child-pool.js is an ES module file as it is a .js file whose nearest parent package.json contains \"type\": \"module\" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove \"type\": \"module\" from /usr/src/app/node_modules/bullmq/node_modules/get-port/package.json.

    at QueueEvents.onFailed (/usr/src/app/node_modules/bullmq/src/classes/job.ts:851:16)
    at QueueEvents.emit (events.js:400:28)
    at QueueEvents.emit (domain.js:475:12)
    at QueueEvents.emit (/usr/src/app/node_modules/bullmq/src/classes/queue-base.ts:103:20)
    at QueueEvents.emit (/usr/src/app/node_modules/bullmq/src/classes/queue-events.ts:196:18)
    at QueueEvents.consumeEvents (/usr/src/app/node_modules/bullmq/src/classes/queue-events.ts:288:18)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async QueueEvents.run (/usr/src/app/node_modules/bullmq/src/classes/queue-events.ts:241:9)
```

Code to reproduce is something like:
```
const queue = new Queue(...)
const worker = new Worker(...)
const queueEvents = new QueueEvents(...)
const job = await queue.add('job-name', {})
await job.waitUntilFinished(queueEvents)
```

Then, when the job fails, this error is thrown. This is with BullMQ v1.91.1 and Node v14.20.0. Any advise would be greatly appreciated 🙂

theoilie:
Fixes #1424 by replacing the `get-port` module, which causes errors in some scenarios when imported, with a simple function that uses the built-in NodeJS `net` module to find an open port for child processes. This has been tested using [patch-package](https://www.npmjs.com/package/patch-package) in a project where sandboxed jobs were previously unable to be processed due to this issue.

### DIFF
--- a/src/classes/child-pool.ts
+++ b/src/classes/child-pool.ts
@@ -1,6 +1,7 @@
 import { ChildProcess, fork } from 'child_process';
 import * as path from 'path';
 import { flatten } from 'lodash';
+import net from 'net';
 import { killAsync } from './process-utils';
 import { ParentCommand, ChildCommand } from '../interfaces';
 import { parentSend } from '../utils';
@@ -10,6 +11,16 @@ const CHILD_KILL_TIMEOUT = 30_000;
 export interface ChildProcessExt extends ChildProcess {
   processFile?: string;
 }
+
+const getFreePort = async () => {
+  return new Promise((resolve) => {
+    const server = net.createServer();
+    server.listen(0, () => {
+      const port = server.address().port;
+      server.close((error: any) => resolve(port));
+    });
+  });
+};
 
 const convertExecArgv = async (execArgv: string[]): Promise<string[]> => {
   const standard: string[] = [];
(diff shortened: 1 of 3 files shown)


src/classes/child-pool.ts
```diff
@@ -11,6 +12,16 @@ export interface ChildProcessExt extends ChildProcess {
   processFile?: string;
 }
 
+const getFreePort = async () => {
```
manast, line 15: Is there any link that explains why this method works?

manast:
Thanks. I answered my own question so we are good to go.

PR #1455, opened 2022-10-01
fix(rate-limit): remove groups

roggervalf:
in favor of groups feature in pro version


src/commands/includes/moveJobFromWaitToActive.lua
```diff
@@ -24,36 +24,16 @@ local function moveJobFromWaitToActive(keys, keyPrefix, jobId, processedOn, opts
   if(maxJobs) then
```
manast, line 24: Just an observation; the name of this method "moveJobFromWaitToActive" is a bit off considering that it is not actually moving the job, it is called "after" moving the job. So we should find a better name for it. Maybe it can be break down in smaller functions, something like "applyRateLimit", etc.

src/commands/includes/moveJobFromWaitToActive.lua
```diff
@@ -24,36 +24,16 @@ local function moveJobFromWaitToActive(keys, keyPrefix, jobId, processedOn, opts
...
       local exceedingJobs = jobCounter - maxJobs
       local expireTime = tonumber(rcall("PTTL", rateLimiterKey))
-      local delay = expireTime + ((exceedingJobs - 1) * limiterDuration) / maxJobs;
-      local timestamp = delay + tonumber(processedOn)
 
-      -- put job into delayed queue
-      rcall("ZADD", keys[7], timestamp * 0x1000 + bit.band(jobCounter, 0xfff), jobId);
-      rcall("XADD", keys[4], "*", "event", "delayed", "jobId", jobId, "delay", timestamp);
+      -- put job into wait queue
+      rcall("LPUSH", keys[1], jobId)
```
manast, line 36: This call (and the LREM one) should only be required when we are calling the moveToActive script passing a jobId (since that means BRPOPLPUSH did actually move a job to the active list). When we do not pass the jobId, then we should not need to do this LPUSH (btw, I think it should be a RPUSH to move it back to the exact same position in the wait list as it was before).

manast:
I will write a blog post explaining in more detail why we want to do this change, but here is a TLDR; 
Rate limit by group key is currently quite complex and inefficient, it is based on using delayed jobs when jobs become rate limited, which is inaccurate and CPU intensive. Without proper group support, there is no other way we know of to make this code better, and the problem is that it affects the efficiency of the global rate limiter. In order to implement a simpler and more efficient global rate limiter, we decided to remove the rate limiter based on group keys. This change will imply a new release, 3.0, so if you really need the limiter by group keys and you do not want to use the Pro version, then you still have the option of either staying on the 2.x version or using the Bull library.

PR #1466, opened 2022-10-07
docs: fix Object.values usage in flows guide

janvennemann:
While implementing jobs using the "Flows" guide i noticed a `TypeError: Object(...).values is not a function` error when using the code from the docs. Changing it to `Object.values(childrenValues)` is the correct usage.


roggervalf (approved):
LGTM Thank you

PR #1474, opened 2022-10-13
fix(redis): replace throw exception by console.error

manast:


roggervalf (approved):
LGTM fixes https://github.com/taskforcesh/bullmq/issues/1437

PR #1486, opened 2022-10-20
docs: add missing breaking changes about noeviction

metzgegu:
### Description

Include missing v2 breaking changes about the Redis eviction policy


manast:
Thanks. The noeviction policy has always been necessary but we did a breaking change where we throw an exception if this is not setup correctly. However we have relaxed this now, so only an error is logged out.

PR #1492, opened 2022-10-24
feat(rate-limit): remove group key support and improve global rate limit

manast:
BREAKING CHANGE: limit by group keys has been removed in favor of a much simpler and efficent rate-limit implementation.


tests/test_flow.ts
```diff
@@ -1266,7 +1266,7 @@ describe('flows', () => {
     });
   });
 
-  it('should rate limit by grouping', async function () {
+  it.skip('should rate limit by grouping', async function () {
```
roggervalf, line 1269: could we remove this test?

tests/test_rate_limiter.ts
```diff
@@ -154,6 +154,66 @@ describe('Rate Limiter', function () {
...
+
+    const numJobs = 10;
+    const dynamicLimit = 250;
+    const duration = 100;
+
+    const worker = new Worker(
+      queueName,
+      async job => {
+        if (job.attemptsMade === 1) {
+          await worker.rateLimit(dynamicLimit);
```
roggervalf, line 168: could we do: await this.rateLimit(dynamicLimit);

roggervalf (approved):
LGTM

tests/test_rate_limiter.ts
manast: No, it won't work with this. But it is a good point, we need to support "this" in order to support sandboxed processors.

JCMais: just reading this, so, ratelimit is not supported for sandboxed processors?

Issue #1495, opened 2022-10-25
Job is `undefined` in worker `failed` listener


jstayton:
This doesn't happen frequently, but every so often, this error occurs in my worker's `failed` listener:

```
TypeError: Cannot read properties of undefined (reading 'attemptsMade')
  File "/srv/build/1666358956/src/jobs/base/worker_creator.js", line 30, col 33, in Worker.<anonymous>
    const isLastAttempt = job.attemptsMade === job.opts.attempts
  File "node:events", line 520, col 28, in Worker.emit
  File "node:domain", line 475, col 12, in Worker.emit
  File "/srv/build/1666358956/node_modules/bullmq/dist/cjs/classes/queue-base.js", line 77, col 26, in Worker.emit
    return super.emit(event, ...args);
  File "/srv/build/1666358956/node_modules/bullmq/dist/cjs/classes/worker.js", line 69, col 22, in Worker.emit
    return super.emit(event, ...args);
  File "/srv/build/1666358956/node_modules/bullmq/dist/cjs/classes/worker.js", line 499, col 42, in <anonymous>
    failedJobs.forEach((job) => this.emit('failed', job, new Error('job stalled more than allowable limit'), 'active'));
  File "/srv/build/1666358956/src/jobs/base/worker_creator.js", in Array.forEach
  File "/srv/build/1666358956/node_modules/bullmq/dist/cjs/classes/worker.js", line 499, col 20, in Worker.notifyFailedJobs
    failedJobs.forEach((job) => this.emit('failed', job, new Error('job stalled more than allowable limit'), 'active'));
  File "/srv/build/1666358956/node_modules/bullmq/dist/cjs/classes/worker.js", line 492, col 22, in Worker.moveStalledJobsToWait
    this.notifyFailedJobs(await Promise.all(jobPromises));
  File "node:internal/process/task_queues", line 96, col 5, in processTicksAndRejections
```

Here's a condensed version of the code in question:

```js
const worker = new Worker(...)

worker.on('failed', (job, error) => {
  // This is where the error occurs due to `job` being `undefined`:
  const isLastAttempt = job.attemptsMade === job.opts.attempts

  // More code...
})
```

[The docs](https://api.docs.bullmq.io/interfaces/WorkerListener.html#failed) seem to say that `job` should always be an instance of the `Job` class. Am I reading that right? Or should I guard against `job` being `undefined` in my code?

This is running BullMQ v2.1.1. Thanks for your help!

[...]

PR #1500, opened 2022-10-27
fix(worker): consider removed jobs in failed event

roggervalf:
there is a test case that covers this case https://github.com/taskforcesh/bullmq/blob/master/tests/test_stalled_jobs.ts#L311-L377
ref https://github.com/taskforcesh/bullmq/issues/1495

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -88,7 +88,7 @@ export interface WorkerListener<
    * This event is triggered when a job has thrown an exception.
    */
   failed: (
-    job: Job<DataType, ResultType, NameType>,
+    job: Job<DataType, ResultType, NameType> | undefined,
     error: Error,
     prev: string,
   ) => void;


manast:
I wonder, do we know the underlying reason for the job to be undefined in this particular case? is it possible to fix it so that it always is sent?

roggervalf:
it is undefined because it was removed when it was moved to failed as it has removeOnFailed as true

manast:
Yes, that makes sense, that it is removed from Redis, but the worker itself should still keep a reference of the job instance, no?

manast:
I mean, for the global failed event, yes it must be undefined, but is it really the case for the local worker event?

roggervalf:
hey, the escenario where this is happening is when we have stalled jobs that were removed https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L790-L797 in that case the only value that is available is the jobId that is returned by stalled script, but as we are creating an instance of a Job with the id of the removed job, we ended with an undefined value

manast (approved):

PR #1504, opened 2022-10-29
perf(promote): provide priority from job instance

roggervalf:


129lovely:
@roggervalf Is there a reason why this is just closed? :(

roggervalf:
hi @129lovely, you can do this by using changePriority method https://docs.bullmq.io/guide/jobs/prioritized#change-priority and then promote your job

PR #1515, opened 2022-11-03
fix(worker): restore dinamic concurrency change

roggervalf:


manast:
Is this really needed? I figured it was not needed anymore.

manast (approved):

PR #1519, opened 2022-11-03
fix(delayed): notify workers a delayed job is closer in time fixes #1505

### LINKED ISSUE #1505: Job is not retried when concurrency > 1 + repeatable job is scheduled
Reproduction steps

```
const {Queue, Worker} = require('bullmq');

const connection = {
  host: 'localhost',
};

const queue = new Queue('myQueue', {connection});
const worker = new Worker(
  'myQueue',
  (job) => {
    console.log('In my handler', job.id, job.attemptsMade);
    throw new Error('test failure');
  },
  {
    connection,
    concurrency: 2,
  },
);

// Adding this job messes up retrial of the job named `retriedJob`
queue.add('repeatedJob', 'data', {
  repeat: {
    pattern: '0 * * * *',
  },
});

queue.add('retriedJob', 'data', {
  attempts: 5,
  backoff: {
    type: 'fixed',
    delay: 1000,
  },
});
```

Expected: retriedJob is retried after 1 second and 'In my handler' is logged 5 times in a row.  
Actual: Job retriedJob never retried unless the worker is restarted

Other issue:
If concurrency = 2 and no repeatable job is scheduled, retrial works but instead of retrying after 1 second, it often retries after 5 seconds

manast:


src/classes/job.ts
```diff
@@ -605,6 +605,11 @@ export class Job<
     }
 
     const results = await multi.exec();
+    const anyError = results.find(result => result[0]);
+    if (anyError) {
+      throw new Error(`Error moving to failed: ${anyError}`);
```
roggervalf, line 610: not in all cases we are executing moveToFinished, we execute or moveToDelayed or moveToFinished or retryJob

manast: ok, I just took the name of the method where this error is produced, Job##moveToFailed. What would be a better error message?

roggervalf: I'm seeing that there is a command variable, probably we can use that one to  build the error message like `Error calling ${command} command`

src/commands/moveToDelayed-8.lua
```diff
@@ -29,35 +29,43 @@ local rcall = redis.call
...
-  rcall("ZADD", KEYS[4], score, jobId)
+    local numRemovedElements = rcall("LREM", KEYS[2], -1, jobId)
+    if (numRemovedElements < 1) then return -3 end
 
-  rcall("XADD", KEYS[6], "*", "event", "delayed", "jobId", jobId, "delay", delayedTimestamp);
+    -- Check if we need to push a marker job to wake up sleeping workers.
+    local target = getTargetQueueList(KEYS[8], KEYS[1], KEYS[7])
+    if rcall("LLEN", target) == 0 then
+        local nextTimestamp = getNextDelayedTimestamp(delayedKey)
+        rcall("SET", "DEBUG", type(nextTimestamp))
```
roggervalf, line 59: we could delete this line

PR #1525, opened 2022-11-05
Fix grammar

julian-hansen:
an number changed to a number


roggervalf:
Thanks @julian-hansen, but the actual change should be Made in https://github.com/taskforcesh/bullmq/blob/master/src/interfaces/base-job-options.ts#L45, the dile you changed is autogenerated

julian-hansen:
Thanks Rogger,

 

First time on the docs site – will make a note

 

Updated in the correct place.

PR #1526, opened 2022-11-06
Update base-job-options.ts

julian-hansen:
Change grammar "an number" should be "a number"


manast:
Thanks for the correction.

PR #1535, opened 2022-11-15
fix(rate-limit): check job is active before moving to wait

manast:


src/classes/scripts.ts
```diff
@@ -30,6 +30,7 @@ import { ErrorCode } from '../enums';
 import { array2obj, getParentKey, isRedisVersionLowerThan } from '../utils';
 import { QueueBase } from './queue-base';
 import { Job, MoveToWaitingChildrenOpts } from './job';
+import { ChainableCommander } from 'ioredis';
```
roggervalf, line 33: it seems that this value is not used

src/classes/worker.ts
```diff
@@ -834,10 +834,11 @@ export class Worker<
   ) {
     const multi = (await this.client).multi();
     multi.pttl(this.keys.limiter);
-    multi.lrem(this.keys.active, 1, job.id);
-    multi.rpush(this.keys.wait, job.id);
-    multi.del(`${this.toKey(job.id)}:lock`);
-    const [[err, limitUntil]] = await multi.exec();
-    return <number>limitUntil;
+    this.scripts.moveJobFromActiveToWait(multi, job.id);
```
roggervalf, line 837: this is more like a refactor

src/classes/scripts.ts
manast: Yes, it is here: https://github.com/taskforcesh/bullmq/pull/1535/files#diff-7858112f5b8e6ac9e61f03ef5e82b749bbc72c060061ba6b00451d98b97a5a0aR844

src/classes/worker.ts
manast: Not really, since now we check if the job was indeed in the active list before we try to move it to the wait list.

src/classes/scripts.ts
roggervalf: oh I see, you are right

roggervalf (approved):
LGTM

PR #1549, opened 2022-11-24
feat(queue-events): support duplicated event

roggervalf:


src/classes/queue-events.ts
```diff
@@ -61,6 +61,13 @@ export interface QueueEventsListener extends IoredisListener {
    */
   drained: (id: string) => void;
 
+  /**
+   * Listen to 'duplicated' event.
+   *
+   * This event is triggered when a job is not created because it already existed.
```
manast, line 67: already exited -> already exist in the queue

manast (approved):

PR #1556, opened 2022-11-30
fix(add-job): throw error when jobId integer

roggervalf:


lukashroch:
@roggervalf what is the rationale to have integers blocked? We've been using bullmq for quite a while to pass own unique sequences to bull jobs and this is a breaking change for us :-/

manast:
@lukashroch we had an edge case where a user did use custom job ids as integers but also sometimes it did not specify a custom Id, mixing custom and non-custom could end sometimes in jobs being ignored inadvertently.  But yes, I agree this should have been a breaking change.

manast:
@roggervalf maybe a good enough solution is to give a warning instead of throwing an error in this case? so the user must be aware of the potential edge case.

roggervalf:
Yeah the edge case that @manast is pointing is right, also now there is a duplicated event in last release so people could listen when a job is a duplicated one, but in this case I will console.warn this error and restore it the error in next breaking change

matjaeck:
> @lukashroch we had an edge case where a user did use custom job ids as integers but also sometimes it did not specify a custom Id, mixing custom and non-custom could end sometimes in jobs being ignored inadvertently. But yes, I agree this should have been a breaking change.

Hello, this unfortunately does not answer why I'm being warned about using numeric strings for job Ids. I know it's my responsibility to keep it consistent, but I don't get what the problem with Ids is or why I'm being warned about using integers when actually providing ids as strings (as the type enforces).

Please explain the actual problem with job Ids and what type of strings we are allowed to use. Thanks.

manast:
@matjaeck there is nothing more to it than what we already discussed here. Since mixing numeric job IDs with non-custom job IDs can result in the edge case of jobs inadvertently being ignored we have decided to warn for the use of numeric Ids. you can easily just attach a prefix or postfix to your job IDs to avoid any warnings.

matjaeck:
@manast Thanks, this clarifies it for me. I have switched to using uuids. Thanks for the reply.

PR #1564, opened 2022-12-03
fix(get-workers): set name when ready event in connection

roggervalf:


manast (approved):
Looks very good to me.

PR #1569, opened 2022-12-05
fix(job): console warn custom job ids when they represent integers

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/1556, in next breaking change we can restore throwing the error to avoid this edge case

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1058,7 +1058,8 @@ export class Job<
 
     if (`${parseInt(this.id, 10)}` === this.id) {
       //TODO: throw an error in next breaking change
-      console.warn('Custom Ids cannot be integers');
+      console.warn(`Custom Ids should not be integers to avoid potential edge cases.
+ref https://github.com/taskforcesh/bullmq/pull/1569`);
     }
 
     return this.scripts.addJob(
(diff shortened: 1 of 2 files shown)


matjaeck:
Hello, I'm sorry but I don't understand the issue that I'm being warned about at all and the linked ticket does not provide any more information.

> We had an edge case where a user did use custom job ids as integers but also sometimes it did not specify a custom Id, mixing custom and non-custom could end sometimes in jobs being ignored inadvertently

- What has this to do with the Id being an integer (it isn't)?

- `jobId` is of type `string` so how do I have to understand the warning "Custom Ids should not be integers" when the id clearly is a string!

What am I missing or what is this warning actually about? Numeric strings? What's wrong with them?

manast:
When you do not specify a custom ID, BullMQ assigns an increasing integer as the job ID. Since adding a job with an existing job ID results in the job being ignored, we want to avoid the potential edge case where you add an integer ID that collides with an automatically generated job ID. Since job IDs are strings we refer to the fact that the string contains a number.

eyalfishler:
so is it allowed to use jobID as a number? would it be blocked in future versions or this would remain as a warning?

manast:
@eyalfishler if you use custon ids stay safe by appending some prefix to your id and you will be fine.

eyalfishler:
seems like the code is taking a string and parse it to Int 
so if i pass "18736136782" as string for JobID it would fail, rather than passing 18736136782 as int which i understand is not allowed. why not allow for a number in string format as JobID ?

PR #1570, opened 2022-12-05
fix: floor PEXPIRE to integer

### LINKED ISSUE #1565: Rate limiter with `max: 1` having unexpected results
We are having issues when we set the rate limiter to:

```ts
limiter: {
    max: 1,
    duration: 100
},
```
It seems to be limiting the jobs more than it should.  It's taking much longer to move a job from `waiting` to `active` than 1 job every 100ms.

Also, when set it to:
```ts
limiter: {
    max: 1,
    duration: 1000/300 // 3.33333333333
},
```
it stopped processing jobs altogether.

Then in both cases, if we removed the rate limiter, it processed jobs immediately.

but after we use `duration: 1000/300`, if we add back the `duration: 100` limiter, it still didn't process jobs. The only way it would process jobs again was to remove the limiter completely.

We are running Redis version 6.2.6

Has anyone else been successful in setting the rate limiter to `max: 1`?

manast:
Fixes #1565

### DIFF
--- a/src/commands/includes/getRateLimitTTL.lua
+++ b/src/commands/includes/getRateLimitTTL.lua
@@ -3,7 +3,7 @@ local function getRateLimitTTL(opts, limiterKey)
   if maxJobs then
     local jobCounter = tonumber(rcall("GET", limiterKey))
     if jobCounter ~= nil and jobCounter >= maxJobs then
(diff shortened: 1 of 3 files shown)


roggervalf (approved):
LGTM

PR #1575, opened 2022-12-07
fix(worker): add token postfix

manast:
It is useful to have a different token per worker processing "task", as well as a token based on the worker id, so that we can have a better mapping between jobs and workers in the future.


roggervalf (approved):
LGTM

PR #1578, opened 2022-12-09
test(repeat): fix flaky test

roggervalf:


src/classes/worker.ts
```diff
@@ -267,6 +248,25 @@ export class Worker<
...
     }
+
+    this.blockingConnection = new RedisConnection(
+      isRedisInstance(opts.connection)
+        ? (<Redis>opts.connection).duplicate()
+        : opts.connection,
+    );
+    this.blockingConnection.on('error', error => this.emit('error', error));
+
+    this.blockingConnection.on('ready', async () => {
```
roggervalf, line 259: moved because ready event could be trigger so fast, before we can listen it

PR #1580, opened 2022-12-09
fix(job): fetch parent before job moves to complete

S3bb1:
When creating a flow with a long-running job in the beginning (e.g. takes multiple seconds) and during the processing of the job you want to add a new parent to this job (because it should run after the long-running job) it doesn't execute the job which was added after.

The root cause of this issue is, that the worker fetches the `Job` from redis before it moves it to active. When the `Job` then finishes, the worker passes the previous fetched instance of the job to `moveToCompleted`. When an additional Job is added as `parent` of the current processing job, it never gets executed because the instance doesn't have the correct parent key. 

This PR adds an additional "sync" of the parent key of the job before it gets moved to completed to have the current instance up to date when in the meantime a new parent was added. 

Added a test which runs the following flow:

- Create Job 1 without children
  - Flow moves Job to active
  - Worker processes job (wait 1000ms)
- Client waits 500ms
- Client adds new Job 2 with children [Job1]
- Worker finishes Job 1
- Worker starts Job 2 
- Worker finishes Job 2

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -516,11 +516,6 @@ export class Job<
       throw errorObject.value;
     }
 
-    const jobData = await Job.fromId(this.queue, this.id);
-    // overwrite parent with redis response
-    this.parent = jobData.parent;
-    this.parentKey = jobData.parentKey;
-
     return this.scripts.moveToCompleted(
       this,
       stringifiedReturnValue,
@@ -597,11 +592,6 @@ export class Job<
     }
 
     if (moveToFailed) {
-      const jobData = await Job.fromId(this.queue, this.id);
-      // overwrite parent with redis response
-      this.parent = jobData.parent;
-      this.parentKey = jobData.parentKey;
-
       const args = this.scripts.moveToFailedArgs(
         this,
         message,
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -597,6 +597,11 @@ export class Job<
     }
 
     if (moveToFailed) {
+      const jobData = await Job.fromId(this.queue, this.id);
+      // overwrite parent with redis response
+      this.parent = jobData.parent;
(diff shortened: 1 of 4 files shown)


src/classes/job.ts
```diff
@@ -516,6 +516,11 @@ export class Job<
       throw errorObject.value;
     }
 
+    const jobData = await Job.fromId(this.queue, this.id);
```
roggervalf, line 519: better to handle this logic atomically in moveToFinished script, so probably we may need to get parent values from there instead of providing it in args

S3bb1: Added the fetch of the parent and the parentKey into the lua scripts.

S3bb1:
Can we rerun the tests? On my local machine they went through

src/commands/moveToFinished-12.lua
```diff
@@ -72,9 +70,6 @@ if rcall("EXISTS", jobIdKey) == 1 then -- // Make sure job exists
     local opts = cmsgpack.unpack(ARGV[9])
 
     local token = opts['token']
-    local parentId = opts['parent'] and opts['parent']['id'] or ""
```
manast, line 75: Since we are not using opts['parent'] nor opts['parentKey'] anymore, we should also remove them from the opts object passed to the script to avoid passing unnecessary data in the Redis call.

manast (changes requested):

src/commands/moveToFinished-12.lua
S3bb1: Thanks! That makes sense :) removed the lines in the `moveToFinishedArgs` function

manast (approved):

S3bb1:
Any updates here @roggervalf?

roggervalf (approved):
Sorry for the delay, LGTM

PR #1586, opened 2022-12-15
perf(get-children-values): replace slow object destructuring with single object

roggervalf:


src/classes/job.ts
```diff
@@ -755,13 +751,10 @@ export class Job<
...
       ];
 
-      const transformedProcessed = Object.entries(processed).reduce(
-        (accumulator: Record<string, any>, [key, value]) => {
-          accumulator[key] = JSON.parse(value);
-          return accumulator;
-        },
-        {},
-      );
+      const transformedProcessed: Record<string, any> = {};
```
manast, line 754: As this is now practically the same code as above, can't we refactor it into one function?

Issue #1532, opened 2022-11-10
Add support for getJobState method in Queue class


LyzeOfKiel:
Feature request:
Now, the only way to get job's state is by calling `job.getState()`. And if I already know job id, but don't have the actual job instance in a function, it would be convenient to call `queue.getJobState(jobId)` without retrieving the job via `queue.getJob(jobId)` first.

manast:
Ok, it would be a bit faster as you do not need to fetch the job first. Maybe an enhancement for a rainy day :)

roggervalf:
hi @LyzeOfKiel, it should be available since version 3.5.0

PR #1593, opened 2022-12-20
feat(queue): add getJobState method

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1532


manast (approved):
Nice little one.

PR #1597, opened 2022-12-21
fix(worker): add max concurrency from the beginning

roggervalf:
ref https://github.com/taskforcesh/bullmq/discussions/1589


src/classes/worker.ts
```diff
@@ -368,14 +368,17 @@ export class Worker<
...
             ) {
-              const token = `${this.id}:${tokenPostfix++}`;
-              processing.set(
-                this.retryIfFailed<Job<DataType, ResultType, NameType>>(
-                  () => this.getNextJob(token),
-                  this.opts.runRetryDelay,
-                ),
-                token,
-              );
+              const restProcesses = this.opts.concurrency - processing.size;
```
manast, line 371: I think this code should be something like this: 
```ts
const restProcesses = this.opts.concurrency - processing.size;
              for (let i = 0; i < restProcesses; i++) {
                const gettingJob = this.retryIfFailed<
                  Job<DataType, ResultType, NameType>
                >(() => this.getNextJob(token), this.opts.runRetryDelay);
                processing.set(gettingJob, token);
                await gettingJob;
              }
```
This is because otherwise, we will be getting the same number of jobs as the concurrency factor in parallel without respecting the rate limiter.

roggervalf: sadly, with this change, the test is taking a lot of time. Also getNextJob calls moveJobToActive, so we will restore all the  rest of jobs as soon as the rate limit is reached

src/classes/worker.ts
```diff
@@ -488,7 +488,8 @@ export class Worker<
     // If we get the special delayed job ID, we pick the delay as the next
     // block timeout.
     if (jobId && jobId.startsWith('0:')) {
-      this.blockTimeout = parseInt(jobId.split(':')[1]);
+      const expectedBlockTimeout = parseInt(jobId.split(':')[1]) - Date.now();
```
manast, line 491: This is a good change, basing the delay on a timestamp will be more accurate most of the time.

src/commands/addJob-8.lua
```diff
@@ -144,9 +144,9 @@ elseif (delayedTimestamp ~= 0) then
   local target = getTargetQueueList(KEYS[3], KEYS[1], KEYS[2])
   if rcall("LLEN", target) == 0 then
     local nextTimestamp = getNextDelayedTimestamp(KEYS[5])
-    if not nextTimestamp or (delayedTimestamp < nextTimestamp) then
-      local delay = delayedTimestamp - tonumber(timestamp)
-      rcall("LPUSH", target, "0:" .. delay)
+    if nextTimestamp ~= nil then
+      --local delay = delayedTimestamp - tonumber(timestamp)
```
manast, line 148: Don't forget to remove the out-commented code.

src/commands/moveToDelayed-8.lua
```diff
@@ -54,17 +54,18 @@ if rcall("EXISTS", jobKey) == 1 then
       return -3
     end
 
+    rcall("ZADD", delayedKey, score, jobId)
```
manast, line 57: Hmm, I am confused, how did we miss to actually put the job in the delayed set?

src/classes/worker.ts
manast: But the question is, is it slower than before or slower than after your change? Because I think the reason it is slower is because it is respecting the rate limiter, whereas without doing the "await" on getting job, it will not respect the rate limit.

src/commands/moveToDelayed-8.lua
roggervalf: this logic was moved before checking the next delayed timestamp, we were not missing moving the job

src/classes/worker.ts
roggervalf: I make some tests using this test case:
before any change it took around 28500ms
with my change it took around 4500ms
with your change for I cannot see jobs getting processed, it just get stuck

manast: I am quite sure that if you do not wait for the ```getNextJob``` in this loop, it will not honor the rate limiter, for example, try with a high concurrency number, like 100 or more, and then a rate limiter of 1 job per second, I bet it will manage to process a couple of jobs before it actually gets rate limited.

roggervalf: sure, I added a test case for concurrency as 100 and max limiter as 1 per second

src/commands/includes/updateParentDepsIfNeeded.lua
```diff
@@ -27,8 +27,8 @@ local function updateParentDepsIfNeeded(parentKey, parentQueueKey, parentDepende
 
       if rcall("LLEN", parentTarget) == 0 then
         local nextTimestamp = getNextDelayedTimestamp(parentDelayedKey)
-        if not nextTimestamp or (delayedTimestamp <= nextTimestamp) then
-          rcall("LPUSH", parentTarget, "0:" .. delayedTimestamp - tonumber(timestamp))
+        if nextTimestamp ~= nil then
```
manast, line 30: This code is the same as in addJob, we could refactor into a "addDelayMarkerIfNeeded" function.

src/commands/moveToDelayed-8.lua
```diff
@@ -54,17 +54,18 @@ if rcall("EXISTS", jobKey) == 1 then
...
+    rcall("XADD", KEYS[6], "*", "event", "delayed", "jobId", jobId, "delay", delayedTimestamp)
+
     -- Check if we need to push a marker job to wake up sleeping workers.
     local target = getTargetQueueList(KEYS[8], KEYS[1], KEYS[7])
     if rcall("LLEN", target) == 0 then
         local nextTimestamp = getNextDelayedTimestamp(delayedKey)
-        if not nextTimestamp or (delayedTimestamp < nextTimestamp) then
-            rcall("LPUSH", target, "0:" .. delayedTimestamp - tonumber(ARGV[2]))
+        if nextTimestamp ~= nil then
```
manast, line 64: Here we have another instance of ```addDelayMarkerIfNeeded```

manast (approved):

Issue #1595, opened 2022-12-20
Reusing Cluster connection with DSN instead of host causes error


rysi3k:
Hello
IORedis Cluster connection allows to set node address not only by host and port but also string dsn like: `redis://host:port`.
When I try to reuse that connection, it causes error on startup.

Example script to reproduce:
```
import {Queue} from 'bullmq';
import IORedis from 'ioredis';

// throws error
const connection = new IORedis.Cluster(['redis://10.0.6.161:7379']);
// good:
// const connection = new IORedis.Cluster([{host:'10.0.6.161', port:7379}]);

const myQueue = new Queue('myqueue', {connection});
myQueue.add('wall', {color: 'pink'});
```

Error:
```
TypeError: Cannot read properties of undefined (reading 'endsWith')
    at /Users/user/projects/webdev/api-service/node_modules/.pnpm/bullmq@1.87.2/node_modules/bullmq/src/classes/redis-connection.ts:104:26
    at Array.some (<anonymous>)
    at RedisConnection.checkUpstashHost (/Users/user/projects/webdev/api-service/node_modules/.pnpm/bullmq@1.87.2/node_modules/bullmq/src/classes/redis-connection.ts:104:9)
    at new RedisConnection (/Users/user/projects/webdev/api-service/node_modules/.pnpm/bullmq@1.87.2/node_modules/bullmq/src/classes/redis-connection.ts:73:8)
```

It will be nice to fix this.

Regards

manast:
Thanks for reporting this.

manast:
I have tried this one but I cannot reproduce the error. I also followed the code path and it seems it should not throw that exception, at least with the latest version of BullMQ. Can you verify that you are running the latest version?

rysi3k:
@manast yes, you are right I have some old version. I see fix was implemented here: https://github.com/taskforcesh/bullmq/commit/8355182a372b68ec62e9c3953bacbd69e0abfc74

PR #1599, opened 2022-12-22
Fix/allow ioredis cluster connection by dsn string

manast:
Add a test to cover this case: https://github.com/taskforcesh/bullmq/issues/1595
Also use ```includes``` instead of ```endsWith``` which is a bit more reliable.


roggervalf (approved):
LGTM

PR #1607, opened 2022-12-27
fix #1603: Fix performance issues in `remove()`

### LINKED ISSUE #1603: Poor `remove()` performance
I noticed the performance of `job.remove()` (or `queue.remove(jobId)` for that matter) is very poor. Especially when the number of jobs gets very large (for example a few million), redis really starts to struggle at 100% CPU during a batch delete. Retrieving 10k jobs with e.g. `queue.getCompleted()` takes just a few seconds, but removing them takes more than 10 minutes (!).

I haven't benchmarked in detail, but I guess this could be due to the dependency checks in the remove lua script.

As I'm looking for a workaround, would it be possible to manually (batch-)remove completed jobs from redis with `DEL queueName:jobId` if the jobs are simple jobs (no dependencies) or will that mess up the queue state?

Niek:
This optimization of the `removeJobFromAnyState.lua` script makes removing jobs a **lot** faster.

Fixes #1603

### DIFF
--- a/src/commands/includes/removeJobFromAnyState.lua
+++ b/src/commands/includes/removeJobFromAnyState.lua
@@ -7,16 +7,16 @@
 
 local function removeJobFromAnyState( prefix, jobId)
   -- We start with the ZSCORE checks, since they have O(1) complexity
-  if rcall("ZSCORE", prefix .. "completed", jobId) ~= nil then
+  if rcall("ZSCORE", prefix .. "completed", jobId) then
     rcall("ZREM", prefix .. "completed", jobId)
     return "completed"
-  elseif rcall("ZSCORE", prefix .. "waiting-children", jobId) ~= nil then
+  elseif rcall("ZSCORE", prefix .. "waiting-children", jobId) then
     rcall("ZREM", prefix .. "waiting-children", jobId)
     return "waiting-children"
-  elseif rcall("ZSCORE", prefix .. "delayed", jobId) ~= nil then
+  elseif rcall("ZSCORE", prefix .. "delayed", jobId) then
     rcall("ZREM", prefix .. "delayed", jobId)
     return "delayed"
-  elseif rcall("ZSCORE", prefix .. "failed", jobId) ~= nil then
+  elseif rcall("ZSCORE", prefix .. "failed", jobId) then
     rcall("ZREM", prefix .. "failed", jobId)
     return "failed"
   -- We remove only 1 element from the list, since we assume they are not added multiple times
--- a/src/commands/includes/removeJobFromAnyState.lua
+++ b/src/commands/includes/removeJobFromAnyState.lua
@@ -6,20 +6,26 @@
 ]]
 
 local function removeJobFromAnyState( prefix, jobId)
-  if rcall("LREM", prefix .. "wait", 0, jobId) == 1 then
-    return "wait"
-  elseif rcall("LREM", prefix .. "paused", 0, jobId) == 1 then
-    return "paused"
-  elseif rcall("LREM", prefix .. "active", 0, jobId) == 1 then
-    return "active"
-  elseif rcall("ZREM", prefix .. "waiting-children", jobId) == 1 then
-    return "waiting-children"
-  elseif rcall("ZREM", prefix .. "delayed", jobId) == 1 then
-    return "delayed"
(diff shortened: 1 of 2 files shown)


src/commands/includes/removeJobFromAnyState.lua
```diff
@@ -6,20 +6,26 @@
...
 local function removeJobFromAnyState( prefix, jobId)
-  if rcall("LREM", prefix .. "wait", 0, jobId) == 1 then
-    return "wait"
-  elseif rcall("LREM", prefix .. "paused", 0, jobId) == 1 then
-    return "paused"
-  elseif rcall("LREM", prefix .. "active", 0, jobId) == 1 then
-    return "active"
-  elseif rcall("ZREM", prefix .. "waiting-children", jobId) == 1 then
+  -- We start with the ZSCORE checks, since they have O(1) complexity
+  if rcall("ZSCORE", prefix .. "completed", jobId) then
```
manast, line 10: It should not be needed to check if the jobId is in the ZSET, since ZREM will return an integer larger than 0 if it was able to remove it: https://redis.io/commands/zrem/

Niek: The whole reason to do this check is performance: `ZSCORE` has `O(1)` complexity while `ZREM` has `O(M*log(N))`. On a sorted set with a million entries, that makes a huge difference.

See: https://redis.io/commands/zscore/

roggervalf: it's better to see this with a test case, you can add this test:
```
    describe('when there are 4000 jobs to be removed', function () {
      it('removes the jobs in time rage of 4000ms', async function () {
        const numJobs = 4000;
        const jobsData = Array.from(Array(numJobs).keys()).map(index => ({
          name: 'test',
          data: { order: numJobs - index },
        }));
    
        const waitingJobs = await queue.addBulk(jobsData);
        const jobsDataWithDelay = Array.from(Array(numJobs).keys()).map(index => ({
          name: 'test',
          data: { order: numJobs - index },
          opts: {
            delay: 500 + (numJobs - index) * 150,
          },
        }));
    
        const delayedJobs = await queue.addBulk(jobsDataWithDelay);

        for( const job of delayedJobs){
          await job.remove();
        }

        for( const job of waitingJobs){
          await job.remove();
        }

        const countJobs = await queue.getJobCountByTypes('waiting', 'delayed');
        expect(countJobs).to.be.equal(0);
      });  
    });  

```js
Without your change, a timeout is thrown

Niek (commented):
Add test case, thanks @roggervalf

src/commands/includes/removeJobFromAnyState.lua
Niek: Added in https://github.com/taskforcesh/bullmq/pull/1607/commits/37b29d316c9afb8bad1eeda2a9432879c3489097, thanks!

roggervalf:
it looks good to me, what do you think @manast ?

Niek:
@manast Is there anything missing in this PR? The changes are minimal and the performance impact is enormous: a speedup of a million times when there are a few hundred k jobs in a queue.

src/commands/includes/removeJobFromAnyState.lua
manast: @Niek ok. That ZSCORE is much faster when there is no item to remove seems like a bug in Redis to me :) I mean, the first thing the ZREM implementation should do is find the item to delete, which should be O(1), then if the item exists proceed with the delete itself which will be O(m*logn) since it needs to rebalance the tree after removing the item... anyway if this is the case we do not have any other option than use ZSCORE.

manast: (I guess we should report this to the Redis team, it can have a big impact for users not aware of this behaviour)

manast (approved):

src/commands/includes/removeJobFromAnyState.lua
Niek: I do not consider this a bug in Redis - just strange behaviour, but clearly documented:

https://redis.io/commands/zrem/:
> Time complexity: O(M*log(N)) with N being the number of elements in the sorted set and M the number of elements to be removed.

https://redis.io/commands/zscore/:
> Time complexity: O(1)

Issue #1600, opened 2022-12-23
Ability to clear/limit job log


mariuszste:
It would be nice if we could limit the amount of lines saved in the job log. Essentially a first in, first out log with the ability to limit the maximum number of lines to some value.

This would be really useful when retrying completed jobs.

A simpler alternative like the ability to clear the log manually or automatically when a job is retried would also be nice.

CodyTseng:
I think it might be nice to add an option like `keepLogs` to `JobOptions` to limit the number of logs.

```
const job = queue.add('bar', {}, { keepLogs: 100 });
```

I would be happy to get familiar with the BullMQ code and implement this feature if needed.

mariuszste:
> I think it might be nice to add an option like keepLogs to JobOptions to limit the number of logs.
There is already some discussion about this on #1611. I'll try implementing this when i get the time for it but feel free to do it before me.

CodyTseng:
Looks great, looking forward to your update!

roggervalf:
This feature should be available on versión 3.6.0

PR #1611, opened 2023-01-03
feat(job): allow clearing job's log (#1600)

mariuszste:
ref #1600

First time contributing, let me know what what missed.


src/classes/job.ts
```diff
@@ -463,6 +463,15 @@ export class Job<
     return client.rpush(logsKey, logRow);
   }
 
+  /**
+   * Clears job's log
+   */
+  async clearLog(): Promise<void> {
+    const client = await this.queue.client;
+    const logsKey = this.toKey(this.id) + ':logs';
+    await client.del(logsKey);
```
roggervalf, line 472: I'm thinking on the proposal of the issue, in order to keep some logs in the list, you can a lua script using https://redis.io/commands/llen/ to see how many logs are in the list and then you can determine how many records should be extracted, so you can use https://redis.io/commands/lpop/. This new method could receive a number param. Also the name of this method should be in plural like cleanLogs

mariuszste: > This new method could receive a number param.

Should the number provided be the number of items that should be left in the list or the number to remove? Assuming you meant the first, couldn't we just use something like https://redis.io/commands/ltrim/?

mariuszste: Or do we also want to return the deleted records? Not sure if this is a good idea if someone will have a lot of logs. `getJobLogs` is paged and i assume that's for good reason.

roggervalf: ltrim usage sounds good to me in case we don't want to return the deleted records, clean method also returns the deleted records https://api.docs.bullmq.io/classes/Queue.html#clean, in this case we can provide the same start and stop parameters to this new method. In case we want to return the deleted records, using llen and lpop could be an option. What do you think @manast ?

manast: I think having an optional option object with a property like "keepLogs" would be nice. I do not think there is any need to return deleted logs in this case, most likely you wont need that, and if you need you can always get the logs first and then delete.

So the signature would be something like this:
```ts
async clearLogs( opts?: { keepLogs: number })
```

mariuszste:
added requested changes as well as a `keepLogs` job option that will automatically trim the logs as they come it

manast (approved):
LGTM

manast (commented):
Ok, I realized something, keepLogs will be stored in the jobs options. Currently we are trying to minimize the size of the options by using shorthands, so instead of keepLogs we should only store something like "kl". You can see how this is done in the job serialization code.

src/classes/job.ts
```diff
@@ -446,7 +446,37 @@ export class Job<
...
+
+    return this.opts.keepLogs
+      ? Math.min(this.opts.keepLogs, result[0][1])
+      : result[0][1];
+  }
+
+  /**
+   * Clears job's logs
+   */
+  async clearLogs(opts: { keepLogs?: number } = {}): Promise<void> {
```
roggervalf, line 471: as it only receives one attribute, could be just keepLogs?: number instead of an object

mariuszste:
> You can see how this is done in the job serialization code.

Yeah, but the `optsAsJSON` and `optsFromJSON` serialize `JobsOptions` as is. Renaming the field would also cause it to be renamed in the public interface. Should we rewrite these methods in the style of `fromJSON` or what?

roggervalf:
> > You can see how this is done in the job serialization code.
> 
> Yeah, but the `optsAsJSON` and `optsFromJSON` serialize `JobsOptions` as is. Renaming the field would also cause it to be renamed in the public interface. Should we rewrite these methods in the style of `fromJSON` or what?

I think @manast means to use this object https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L39-L43 for translating opts from public version, to the one that is saved in redis

mariuszste:
Ok, I have no idea how I didn't notice that :sweat_smile:

roggervalf (approved):
LGTM

PR #1641, opened 2023-01-23
fix(connection): apply console.warn in noeviction message

hmvs:
As stated in https://github.com/taskforcesh/bullmq/pull/1440 it should be a warning, not an error.


hmvs:
The reason for this change is that the console.error is monitored on production, and it spams our monitoring. I understand that a proper fix would be to have some refactoring and pass the `error` and `warn` functions.  So this is more like a quickfix.

manast (approved):

PR #1651, opened 2023-02-02
queueName is undefined.

twobitunicorn:


docs/gitbook/README (1).md
```diff
@@ -46,7 +46,7 @@ Jobs are added to the queue and can be processed at any time, with at least one
 ```typescript
 
-const worker = new Worker(queueName, async job => {
+const worker = new Worker('foo', async job => {
```
roggervalf, line 49: what about 'queueName'

twobitunicorn: It isn't defined in the working example.  If you check out the quick start page  you have the first two examples (producer, consumer).  https://docs.bullmq.io/readme-1.  The producer uses the queue name `'foo'`.  The consumer (this example) uses the variable `queueName` which is not defined.  I am guessing that since these are both working examples that work together, the queue name it should be using is `'foo'` but that is currently not the case.  It is undefined and the examples don't work without setting it.

roggervalf: oh I see, it is also defined before in the same section. In that case it looks good to me

roggervalf (approved):

PR #1653, opened 2023-02-03
feat(worker): add discardTtl option

roggervalf:


src/classes/worker.ts
```diff
@@ -139,6 +139,8 @@ export interface WorkerListener<
 
 const RATE_LIMIT_ERROR = 'bullmq:rateLimitExceeded';
 
+const DISCARD_TTL_ERROR = 'bullmq:discardTtlExceeded';
```
manast, line 142: I guess we will need to replace this constants by an enum soon.

src/classes/worker.ts
```diff
@@ -669,6 +670,12 @@ export class Worker<
     lockExtender();
```
manast, line 670: ok, seems a bit counterintuitive that we need to extend the lock if we are going to fail the job, but I guess this is necessary to avoid the risk of not having a lock when handling the failed job, which requires a valid lock.

manast (approved):

src/classes/worker.ts
```diff
@@ -669,6 +670,12 @@ export class Worker<
...
 
     try {
+      if (this.opts.discardTtl) {
+        if (this.opts.discardTtl < Date.now() - job.timestamp) {
+          return handleFailed(new Error(DISCARD_TTL_ERROR));
```
manast, line 675: Something hit me about this. Are we really sure we want to fail the job here? I am thinking that discarded jobs would probably just pollute the failed set where there could be legitimate jobs that failed for some reason that needs to be debugged. Maybe it is better to just remove the job while sending an event for it.

manast (changes requested):

valeeum:
Hi team!  Are there any plans to merge this functionality?

manast (approved):
LGTM

manast (commented):
Unfortunatelly I think this functionality cannot be implemented after the job has been fetched from the queue, as it will affect rate limiting. So it should instead be moved to moveToActive, where it can be discarded before it affects the rate limiter. Also it will allow for discarding several jobs in one go, cannot be too many though or it will keep Redis busy for too long.

PR #1675, opened 2023-02-10
avoid using completedIdx

wenq1:
This is an experiment to solve a long-term mystery of: why use completedIdx at all since we are already using Maps?

Also if this kinda works, why not swap the key/value of this.processing?

Also as a suger I added type to const processing so vscode seems more happy.


manast:
Thanks for the PR. The code in the PR does not seem simpler in my opinion than the current one 🤔

Issue #1676, opened 2023-02-10
Jobs get lost/stuck if `job.saveStacktrace()` happens just after a `job.remove()`


hyperair:
`job.saveStacktrace()` does a `hmset` on the job key without validating the lock or ensuring that the job key exists, meaning that it can result in a specific jobId getting stuck forever if it happens just after a `job.remove()`:
 - will not be picked up by workers
 - doesn't exist in any state (`active`, `wait`, etc.)
 - queue.add with the same jobId does nothing because the job key already exists
 - job gets stuck forever, and depending on usecase, can be either just a memory leak, or it may prevent the job from being scheduled ever again, even after obliterating the queue.

One sequence of events that can cause this to happen is:
 1. Job with a specific `jobId: "foo"` gets scheduled
    - `foo` gets added to the `wait` list.
 2. Worker 1 picks up `foo` and starts working on it:
    - `foo` moves into the `active` list
    - `foo:lock` is set to the worker's token
 3. Worker 1 hangs for a while on a synchronous loop and the lock expires
 4. Worker 2 detects the stall, and eventually picks it up
    - `foo` moves to the `active` list again
    - `foo` completes and is deleted (e.g. due to `removeOnComplete: true`)
 5. At this point, `foo` no longer exists anywhere in the queue -- its job key doesn't exist, and no references to it exist in the queue data structures.
 6. Worker 1 finishes its synchronous loop and continues to completion, whereupon in `worker.handleCompleted`, it throws an error due to either a missing lock, lock mismatch, or the job not being in active state
 7. This error is caught [here](https://github.com/taskforcesh/bullmq/blob/v3.6.4/src/classes/worker.ts#L680), and then it goes to `handleFailed()` where it tries to call `job.moveToFailed()`, which eventually leads to `job.saveStacktrace()` [here](https://github.com/taskforcesh/bullmq/blob/v3.6.4/src/classes/job.ts#L569)
 8. `job.saveStacktrace()` reintroduces the job key by blindly calling `hmset` [here](https://github.com/taskforcesh/bullmq/blob/v3.6.4/src/classes/job.ts#L1088)
 9. The `hmset` is pipelined with an evalsha of a lua script, but all of them fail and noop because the job is not in the correct state, or the lock is missing.

Test case to reproduce it is at https://github.com/hyperair/bullmq-savestacktrace-race. You can run it after installing dependencies with `npm start`:

```js
% npm start

> bullmq-savestacktrace-race@1.0.0 start
[...]

PR #1681, opened 2023-02-11
fix(job): check jobKey when saving stacktrace

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1676

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -566,7 +566,18 @@ export class Job<
 
     let command: string;
     const multi = client.multi();
-    this.saveStacktrace(multi, err);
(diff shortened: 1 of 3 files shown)


src/classes/job.ts
```diff
@@ -1070,7 +1081,10 @@ export class Job<
     );
   }
 
-  protected saveStacktrace(multi: ChainableCommander, err: Error): void {
+  protected saveStacktrace(
+    multi: ChainableCommander,
```
manast, line 1085: I think it is more elegant to call the saveStacktrace lua script inside this method.

PR #1682, opened 2023-02-13
avoid using completedIdx

wenq1:
I changed the PR a bit. Just so that

`const token = processing.get(completed);` and `const completed = promises[completedIdx];` can be avoided (since they are readily available).

Feel free to review or close. Just a minor change only, which I don't think matter that much.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -363,7 +363,7 @@ export class Worker<
 
           this.runStalledJobsCheck();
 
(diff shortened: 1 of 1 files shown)


manast:
For the record, I think what we will end doing is simplifying the run loop so that it is more similar to the python version.
https://github.com/taskforcesh/bullmq/blob/master/python/bullmq/worker.py#L47

PR #1688, opened 2023-02-16
feat(python): add retryJobs method

roggervalf:


python/bullmq/queue.py
```diff
@@ -57,6 +57,19 @@ async def obliterate(self, force: bool = False):
             if cursor == 0 or cursor == None or cursor == "0":
                 break
 
+    """
+        Retry all the failed jobs.
+    """
+    async def retryJobs(self, opts = {}):
```
manast, line 63: I think we need to start providing typings for options and also improve the doc string so that we can soon generate the API documentation automatically.

manast (approved):

Issue #1690, opened 2023-02-16
[Bug]: incorrect typing for worker's "failed" event?


okovpashko:
### Version

v3.4.2

### Platform

NodeJS

### What happened?

The [type](https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L95) for the "failed" event handler states that the `job` argument can be `undefined` but in the [docs](https://docs.bullmq.io/guide/events) (see screenshot) and in the source code of the worker [here](https://github.com/taskforcesh/bullmq/blob/42be85bf339615c9cd6ec6a92e3d1cc11f030228/src/classes/worker.ts#L863) (notice types of the filter's callback) and [here](https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L664) (notice the job's method call in the previous line), the job seems can not be `undefined`.

<img width="886" alt="image" src="https://user-images.githubusercontent.com/6490265/219371082-1a667254-bbd7-470e-8180-645f47a8f379.png">

The possible `undefined` in types causes developers to use type guards in the event's handler with the useless alternative branch that I suppose would never be executed.

Could you please consider fixing the types in case I am right or highlighting the cases when the `job` can be undefined in the docs?

### How to reproduce.

N/A

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

okovpashko:
UPD: the API reference also [tells us](https://api.docs.bullmq.io/interfaces/WorkerListener.html#failed) that the job is not optional.
<img width="1069" alt="image" src="https://user-images.githubusercontent.com/6490265/220148878-7f1b06f5-dd11-4204-8863-463f2092d984.png">

manast:
@roggervalf do you remember why the job can be undefined? I have a vague recollection there was a case where it could be indeed undefined.

manast:
But actually, even if that could be the case, it is a bit pointless since you cannot do anything with the event at all.

roggervalf:
hey guys, it could be undefined when the job is stalled and it reaches the limit of stalled jobs, if this job has an option to be removed, the worker does not get the instance of the job on the failed event https://github.com/taskforcesh/bullmq/compare/v1.81.1...v1.81.2

PR #1702, opened 2023-02-21
fix(worker): failed event receives an optional job parameter

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1690

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -33,13 +33,6 @@ const maximumBlockTimeout = 10;
 // note: sandboxed processors would also like to define concurrency per process
 // for better resource utilization.
 
-export type OptionalJob<
-  DataType = any,
-  ResultType = any,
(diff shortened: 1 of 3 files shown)


src/classes/worker.ts
```diff
@@ -92,7 +92,10 @@ export interface WorkerListener<
    * This event is triggered when a job has thrown an exception.
    */
   failed: (
-    job: Job<DataType, ResultType, NameType> | undefined,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
manast, line 95: what are these eslints rules for?

feihan:
Is it possible that this PR courses the following error:

error TS1016: A required parameter cannot follow an optional parameter.
51     failed: (job?: Job<DataType, ResultType, NameType>, error: Error, prev: string) => void;

Thanks

PR #1703, opened 2023-02-22
feat(worker): add remove on complete and fail options

roggervalf:


src/interfaces/worker-options.ts
```diff
@@ -48,12 +49,26 @@ export interface WorkerOptions extends QueueBaseOptions {
    */
   stalledInterval?: number;
 
+  /**
+   * You can provide an object specifying max
+   * age and/or count to keep.
+   * Default behavior is to keep the job in the completed set.
+   */
+  removeOnComplete?: KeepJobs;
```
manast, line 57: I think it should also clarify that the removeOnComplete/Fail on the job overrides whatever setting is used in the worker.

manast (approved):

PR #1729, opened 2023-03-07
Add Changelog link to the original.

arsonik:
Allow us (or maybe just me) to lazy find the actual changelog :)


arsonik:
oh cause its a symlink ... ! so its more like a GitHub issue then ... they should just follow the symlink instead of showing the location :\

Issue #1752, opened 2023-03-24
[Bug]: Parent job not executed if child is removed using `removeOnFail`


ppati000:
### Version

BullMQ Pro 5.2.0

### Platform

NodeJS

### What happened?

Setting `removeOnFail` on a child job does not seem to notify the parent job that it was deleted. The parent job does not execute.

**Expected behavior:** When a failed child job is deleted using `removeOnFail`, the parent job should execute once all other child jobs complete.

**Actual behavior:** The parent job does not execute. It only executes if `removeOnFail` is disabled and the child job is deleted manually after it has failed. See the example code below.

### How to reproduce.

Run the following code to reproduce the bug (= parent job is not executed). Then set `REMOVE_ON_FAIL = false` and run it again. This time the parent job will be executed.

```ts
import { WorkerPro, FlowProducerPro, QueuePro } from '@taskforcesh/bullmq-pro'

const REMOVE_ON_FAIL = true // Set to false => Parent job is executed correctly.

const flow = new FlowProducerPro()
const queue = new QueuePro('failParentTest')

const w = new WorkerPro('failParentTest', async (job) => {
  console.log('processing', job.name)
  if (job.name === 'child0') {
    console.log('failing', job.name)
    setTimeout(() => {
      console.log(`attempting to remove child0 manually (${REMOVE_ON_FAIL ? 'should already be deleted by removeOnFail' : 'should still exist because removeOnFail is disabled'})`)
      queue.remove(job.id)
    }, 1000)
    throw new Error('fail')
  }
}, {});

(async () => {
  flow.add({
    name: 'parent',
    data: {},
    queueName: 'failParentTest',
    children: [{
      queueName: 'failParentTest',
      name: 'child0',
      data: {},
      opts: {
        removeOnFail: REMOVE_ON_FAIL
      }
    }, {
      queueName: 'failParentTest',
      name: 'child1',
      data: {}
    }]
  })
})()
```

### Relevant log output

```shell
Output with REMOVE_ON_FAIL=true (bug)

processing child0
failing child0
processing child1
attempting to remove child0 manually (should already be deleted by removeOnFail)

Output with REMOVE_ON_FAIL=false (works as expected)

processing child0
failing child0
processing child1
attempting to remove child0 manually (should still exist because removeOnFail is disabled)
processing parent
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ppati000:
[...]

PR #1753, opened 2023-03-25
fix(flow): consider removing dependency on removeOnFail true

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1752

### DIFF
--- a/tests/test_flow.ts
+++ b/tests/test_flow.ts
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { Resolver } from 'dns';
 import { default as IORedis } from 'ioredis';
 import { beforeEach, describe, it } from 'mocha';
 import { v4 } from 'uuid';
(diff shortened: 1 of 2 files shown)


tests/test_flow.ts
```diff
@@ -28,6 +28,61 @@ describe('flows', () => {
     await removeAllQueueData(new IORedis(), queueName);
   });
 
+  describe('when removeOnFail is true in last pending child', () => {
+    it('moves parent to wait without getting stuck', async () => {
```
manast, line 32: do you mean move to "wait" or move to "completed" here?

manast: Also, I am wondering, if the child fails permanently and is removed, should the parent fail or complete? I always find these cases a bit confusing...

roggervalf: that may depends on fpof option, if it is true and the child is failed, parent will fail as well https://github.com/taskforcesh/bullmq/blob/00cebda77fdd160c7e41b0a6db372d34341c6061/src/commands/moveToFinished-12.lua#L139-L141

manast (approved):

PR #1756, opened 2023-03-27
feat(upstash): don't throw an error when detecting an upstash host

chronark:
Upstash is currently rolling out redis streams to all remaining regions and will be compatible with bullmq.
This draft is just a preparation for when the time comes.

Please let me know if you're happy with these changes.


manast:
LGTM. But I think we need to wait to merge until the support has reached all regions.

sayaad:
It seems like upstash fully supports streams now.
https://docs.upstash.com/redis/overall/rediscompatibility
I'm patiently waiting for this commit so I can switch redis hosts to upstash.

PR #1784, opened 2023-04-04
refactor(python): replace string concatenation with string formatting

nullndr:
## Changelog

This PR replaces the all string concatenation in the python folder with string formatting with the `f""` literal.

They were introduced in [python 3.6](https://peps.python.org/pep-0498/) so there shouldn't be any compatibility issue with oldest interpreters.

Also, formatted string have been seen to perform better than string concatenation, and they are also more easily to read.


manast (approved):
LGTM

Issue #1778, opened 2023-04-02
[Bug]: Flow, job.getChildrenValues returns an empty object


felixmosh:
### Version

v3.10.3

### Platform

NodeJS

### What happened?

When you have a flow with 2 jobs, that depends on each other (parent<-child), if the child job returns a value, the parent call of `await job.getChildrenValues()` returns an empty object, in previous version (v3.10.2) it returns an object with child id that contains child return value.

### How to reproduce.

```js
import { FlowProducer } from 'bullmq';
const flowProducer = new FlowProducer();

const queueName = 'assembly-line';
const chain = await flowProducer.add(
  {
    name: 'car',
    data: { step: 'engine' },
    queueName,
    children: [
      {
        name: 'car',
        data: { step: 'wheels' },
        queueName,
      },
    ],
  },
  {
    queuesOptions: {
      [queueName]: {
        defaultJobOptions: {
          removeOnComplete: true,
        },
      },
    },
  },
);
```

```js
import { Worker, Job } from 'bullmq';

const worker = new Worker(queueName, async (job: Job) => {
  const step = job.data.step;
  if(step === 'wheels') {
     return {model: 'test'}
  } else if(step === 'engine) {
     const childValue = await job.getChildrenValues();
     // -------^ this will be an empty object
  }
  // Do something with job
  return 'some value';
});
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

roggervalf:
hi @felixmosh, thanks for reporting this, we are taking a look on this

PR #1788, opened 2023-04-05
fix(flow): do not remove completed children results

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1778


manast (approved):
LGTM. Is there a test that verifies that results are removed when the parent is removed?

roggervalf:
> LGTM. Is there a test that verifies that results are removed when the parent is removed?

yeah, there is one https://github.com/taskforcesh/bullmq/pull/1788/files#diff-96ab08f0fa33777b0584d91abbb2f51700a0270f46fb51d0123fd3bce80e0d74R1669-R1780

PR #1789, opened 2023-04-05
chore(python): add missing typings in python

nullndr:
## Changelog

This PR updates the type signatures for the python functions and methods where they were missing or incomplete


manast (approved):
Looks great!

manast:
Btw, since we need to have (python) on all the commits I will squash this PR (the last chore is lacking the python context)

PR #1795, opened 2023-04-07
chore(python): update worker class

nullndr:
## Changelog

This PR adds the `WorkerOptions` typed dict class in the initializer for the python `Worker` class instead of a plain `dict`.


manast (commented):
Very nice 👏

PR #1796, opened 2023-04-07
chore(python job): add JobOptions

nullndr:
## Changelog

This PR follows #1795, adding the typed dict `JobOptions` class for the python class `Job`.

In the previous, I forgot to add `total=False` to the `WorkerOptions` class, so I added it here. Sorry!

## Notes

Please tell me if it is better to move these typed dict classes in a subfolder like `types` or something, since these types will grown once the python library will add more things to match the node version

### DIFF
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -13,7 +13,7 @@ optsDecodeMap = {
 optsEncodeMap = {v: k for k, v in optsDecodeMap.items()}
 
 
-class KeepJosb(TypedDict, total=False):
+class KeepJobs(TypedDict, total=False):
     """
     Specify which jobs to keep after finishing. If both age and count are
     specified, then the jobs kept will be the ones that satisfies both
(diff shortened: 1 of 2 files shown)


manast:
> Please tell me if it is better to move these typed dict classes in a subfolder like `types` or something since these types will grow once the Python library will add more things to match the node version

In the TS version of the library, we have a directory "interfaces", I am not sure what is the standard naming for Python maybe it is "typing" if so then sure, it is a good way to organize the files as they will grow in number.

manast (approved):

python/bullmq/job.py
```diff
@@ -13,6 +13,77 @@
 optsEncodeMap = {v: k for k, v in optsDecodeMap.items()}
 
 
+class KeepJosb(TypedDict, total=False):
```
ccollie, line 16: Should be KeepJobs

nullndr: Good catch

nullndr:
@manast can we merge this? I'm going to open another PR to add the `QueueOptions` as well and I need this to update the `add` method signature

PR #1798, opened 2023-04-09
chore(python queue): add JobOptions to add method

nullndr:
## Changelog

This PR follows #1797 and #1796, adding the typed dict class `JobOptions` to the `add` method of the `Queue` class


manast:
Thanks!. By the way, currently commits for python need to have the "python" context (so no "python context"), this is so that it does not trigger an automatic release of the NodeJS version every time a fix of a new feat is added to the python version.

PR #1801, opened 2023-04-09
Removing unnecessary semicolons

lehmann:
Removing unnecessary semicolons


manast:
Thanks. Some semicolons sneaked in when porting the code from typescript 😅

PR #1809, opened 2023-04-11
Now, a `Worker` knows how to deal with concurrency. The `getFirstComp…

lehmann:
Now, a `Worker` knows how to deal with concurrency. The `getFirstCompleted` function was replaced with `getCompleted`. In some extreme conditions of concurrence (e.g. a single Worker dealing with hundreds of events), the `asyncio.wait` function was returning multiple Jobs as 'Completed'. The original function was picking the very first one of the jobs, ignoring all others, and returning to a tuple `(very_first_completed, all_still_pending)`.
The proposed solution is just about to iterate over ALL completed jobs, fetching the available results, and processing all jobs together, linearly.
I've tested locally, and using the proposed change, the BullMQ Python client was able to deal with a load like 100,000 events in Redis with an average time to process around 7 ms (`concurrency=sqrt(100_000)`). Running the same benchmark code using `concurrency=1`, each event has been fully processed (time between push msg in `Queue` and the `processor` to be called) was around (avg) 50 **seconds**.

### DIFF
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -228,7 +228,7 @@ class Worker(EventEmitter):
 
 async def getCompleted(task_set: set) -> [List[Job], List]:
     job_set, pending = await asyncio.wait(task_set, return_when=asyncio.FIRST_COMPLETED)
-    jobs = [extract_result(job_task) for job_task in jobSet]
+    jobs = [extract_result(job_task) for job_task in job_set]
     # we filter `None` out to remove:
     # a) an empty 'completed jobs' list; and
     # b) a failed extract_result
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -1,4 +1,4 @@
-from typing import Callable, TypedDict, Any
+from typing import Callable, TypedDict, Any, List
 from uuid import uuid4
 from bullmq.scripts import Scripts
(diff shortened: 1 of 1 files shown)


python/bullmq/worker.py
```diff
@@ -224,13 +226,8 @@ async def close(self, force: bool = False):
...
-        try:
-            job = jobTask.result()
-            return (job, pending)
-        except Exception as e:
-            print("ERROR:", e)
-            traceback.print_exc()
-            return pending
+async def getCompleted(task_set: set) -> [List[Job], List]:
+    job_set, pending = await asyncio.wait(task_set, return_when=asyncio.FIRST_COMPLETED)
+    jobs = [jobTask.result() for jobTask in job_set]
```
manast, line 231: What happens if ```jobTaks.result()``` throws an exception? In the previous version we were doing a try/catch so that if the call to ```result``` throws an exception the worker will continue working after that event. 🤔

manast (commented):
Looks great, very nice optimization. However I wrote a small comment.

python/bullmq/worker.py
lehmann: Good catch!
I've removed the error handling for ... reasons, and forgot to put it back.
Added it back and pushed few code comments about the way it works.

manast:
Looks good now!

PR #1810, opened 2023-04-11
fix(worker-close): prevent setting timers after close() has been called

ppati000:
This pull request should fix the following race condition: If `close()` is called very shortly after a Worker is created, its timers are not initialized yet, so `close()` thinks it does not need to close any timer. However, the timers will then be initialized **after** `close()`, leaking the timers and preventing a clean process exit. This issue is particularly relevant in test code that might spawn and close workers very quickly.

---

Regarding `startStalledCheckTimer`: Here we even do an async call to `runStalledJobsCheck` before setting the new timer, and in the meantime `close()` may be called. This could lead to the same problem even in a fully initialized worker (you'd have to be unlucky enough to call `close()` while the stalled jobs check is running)

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -178,7 +178,7 @@ export class Worker<
 
   private blockingConnection: RedisConnection;
 
-  private stalledCheckTimer: NodeJS.Timeout;
+  private stalledCheckTimer: NodeJS.Timeout | null = null;
 
   private asyncFifoQueue: AsyncFifoQueue<void | Job<
     DataType,
@@ -786,6 +786,11 @@ export class Worker<
       clearTimeout(this.stalledCheckTimer);
 
       await this.runStalledJobsCheck();
+
+      if (this.closing) {
+        return;
+      }
+
       this.stalledCheckTimer = setTimeout(async () => {
         this.startStalledCheckTimer();
       }, this.opts.stalledInterval);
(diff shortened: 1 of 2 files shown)


src/classes/worker.ts
```diff
@@ -178,7 +178,7 @@ export class Worker<
 
   private blockingConnection: RedisConnection;
 
-  private stalledCheckTimer: NodeJS.Timeout;
+  private stalledCheckTimer: NodeJS.Timeout | null = null;
```
ppati000, line 181: (just for consistency with `extendLocksTimer`)

manast:
I have checked this issue and I think the only change necessary is to swap the order in which ```startStalledCheckTimer``` and ```startLockExtenderTimer```. The reason is that ```startLockExtenderTimer``` despite being defined as async is not really an async method, it just sets the timer, the ```startStalledCheckTimer``` on the other hand is a real async method, so we need to execute it afterwards, if we do it this way, then the close method will clean the timers correctly and no extra checks needs to be performed inside said methods.

ppati000:
@manast You're right regarding `startLockExtenderTimer`, swapping would fix this function and in that case, the closing check would not be necessary in `startLockExtenderTimer`.

Switching the order would not fix `startStalledCheckTimer` though, because it always runs an async call before setting the next timer. During the async call it is generally possible that `close()` is called and that it attempts to clear the timer before the new one is set.

So all in all... I'd say checking `closing` before setting a timer is the most robust approach and guaranteed to eliminate the issue.

ppati000:
@manast is your solution already applied / do we still need this one?

roggervalf:
hi @ppati000, you can try this fix since version https://github.com/taskforcesh/bullmq/releases/tag/v3.12.1

PR #1818, opened 2023-04-13
refactor the code by removing white space, semicolon etc..

poting-lin:
there are several improvements as below:

- boolean identity to True and False
- type hinting
- trailing whitespace
- there should be two lines before class
- remove the semicolon in the test file
- remove unused imports


manast:
Thanks for these nice additions.

PR #1819, opened 2023-04-13
chore(python): replace is operator

nullndr:
## Changelog

Please, in Python you should not compare `True` and `False` with the `is` operator. 

You can find recomendations for this in the [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations).

> Don’t compare boolean values to True or False using ==:
> 
> Correct:
> ```py
> if greeting:
> ```
> 
> Wrong:
> ```py
> if greeting == True:
> ```
> Worse:
> ```py
> if greeting is True:
> ```

### DIFF
--- a/python/bullmq/scripts.py
+++ b/python/bullmq/scripts.py
@@ -149,16 +149,16 @@ class Scripts:
         keys.append(metricsKey)
 
         def getKeepJobs(shouldRemove: bool | dict | int | None):
-            if shouldRemove is True:
-                return {"count": 0}
-
             if type(shouldRemove) == int:
                 return {"count": shouldRemove}
 
             if type(shouldRemove) == dict:
                 return shouldRemove
 
-            if shouldRemove is False or shouldRemove is None:
+            if shouldRemove:
+                return {"count": 0}
+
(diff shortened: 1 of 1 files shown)


poting-lin:
@nullndr  good catch!
please allow me to explain my understanding here and correct me if I'm wrong:

regarding `== `and `is`

what the document explains regarding the boolean: 
`if greeting:`
means "greeting has a True value?", such as True

`if greeting == True:`
means "greeting has a same value as True", such as True, 1

btw, I think we may need to refactor a little more to the function: `shouldRemove`, because the argument `shouldRemove` is not only a boolean.

nullndr:
There is also another reason that comes from the different implementations of `True` and `False`.

In CPython, they are just singleton, so `True == True` and `True is True` will result in the same exactly value, but in other implementations they could not be singleton, so comparing `True is True` will eventually results in `False`.

Remember that `is` compares the identity of two objects, not their value.

This is not true with the `==` operator, that actually compares the value of two objects.

Also, with `is` you can't compare values that could be cast to boolean (`0` and `1`):

```py
a = 1 is True # false
b = 1 == True # true
```

PR #1820, opened 2023-04-13
refactor(python): update getCompleted return type

nullndr:
## Changelog

This PR updates the return type of `getCompleted` since it was incorrectly typed as `[List[Job], List]`


manast (approved):
LGTM

Issue #1815, opened 2023-04-13
[Bug]: Scheduled delayed job documentation typo


coopbri:
### Version

v3.10.4

### Platform

NodeJS

### What happened?

The docs for [delayed jobs](https://docs.bullmq.io/guide/jobs/delayed) indicate that job execution at a specific point in time can be scheduled with the following code (taken straight from docs):

```ts
const targetTime = new Date("03-07-2035 10:30");
const delay = (targetTime - new Date()).getTime();

await myQueue.add('house', { color: 'white' }, { delay });
```

However, this code is incorrect, as `getTime` does not exist on a number.

### How to reproduce.

Copy code from documentation (above), execute said code.

### Relevant log output

```shell
(targetTime - (intermediate value)).getTime is not a function
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
Good catch, the call to ```getTime``` is not needed at all. Fixed now.

coopbri:
Thanks for updating that! I think the code is still erroneous, however. Dates cannot be subtracted like that in JS -- the code results in `NaN`.

![Screenshot_9](https://user-images.githubusercontent.com/20056195/231690586-9269e6e7-3375-436d-bd38-e86afe2b5f90.png)

manast:
Are you sure? it works in node:
<img width="279" alt="image" src="https://user-images.githubusercontent.com/95200/231691700-909686da-74c6-4a76-ac5b-0553182059fe.png">

coopbri:
Good catch, looks to be TypeScript catching the issue, I mistakingly thought it was a JS issue. Casting the date to a number fixes the issue.

![Screenshot_10](https://user-images.githubusercontent.com/20056195/231692307-a56509a7-e40c-4b1b-b942-b624051e0bec.png)

PR #1823, opened 2023-04-13
docs: fix type error in scheduled job example code

coopbri:
Related: #1815

Fixed type issue in scheduled job example code (works in JS due to coercion, but not TS with sufficiently strict config).


manast:
Thanks!

Issue #1822, opened 2023-04-13
Add an optional repeatable key to the repeat options


manast:
It would be pretty useful to be able to assign a custom repeatable key (i.e. the key where the repeatable metadata is stored), instead of relying on the automatic generation of the key based on the repeat options.

**Describe the solution you'd like**
Add an optional "key" field to the RepeatOptions interface. If this option is provided, instead of generating the key based on the repeat options it should use the key as the base for the actual key stored in Redis.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

tybro0103:
It seems like this would go a long way in making it easier to update the schedule of an existing repeatable without unintentionally creating a new repeatable... or am I missing something? I landed here by trying to figure out this particular problem.

tybro0103:
So for my use I want to be able to define a list of things in a single place that need to run on a regular schedule. Something like:
```js
const jobScheduleMap = {
  'refresh-the-cache': '0 3 * * *',
  'check-for-expired-things': '0 4 * * *',
  'process-subscriptions': '0 5 * * *',
};
```
And then I need to be able to ensure the repeatable jobs in Bull are in sync with this. So for now I'm resorting to this:
```js
// removing an item from this map will NOT cause it to be removed from the queue
// set value to 'DELETE' to cause it to be removed from the queue
const jobScheduleMap = {
  'refresh-the-cache': '0 3 * * *',
  'foo': 'DELETE',
  'check-for-expired-things': '0 4 * * *',
  'process-subscriptions': '0 5 * * *',
};

// reconciles the actual repeatables with the desired repeatables according to jobScheduleMap
const syncRepeatables = async () => {
  // STEP 1: loop through all the existing repeatable jobs
  //   remove those that are set for deletion or where the schedule isn't correct
  //   ignore those that aren't in the jobScheduleMap
  //   compile a list of those that are already correct; used in the next step
  const repeatables = await queue.getRepeatableJobs();
  const alreadyCreatedList = [];
  await Promise.all(repeatables.map(async (repeatable) => {
    const expectedPattern = jobScheduleMap[repeatable.name];
    if ((expectedPattern === 'DELETE') || (expectedPattern && (expectedPattern !== repeatable.pattern))) {
      await queue.removeRepeatableByKey(repeatable.key);
    } else if (expectedPattern === repeatable.pattern) {
      alreadyCreatedList.push(repeatable.name);
    }
  }))
[...]

PR #1824, opened 2023-04-14
feat(repeatable): allow saving custom key

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1822


src/classes/repeat.ts
```diff
@@ -201,13 +203,20 @@ export class Repeat extends QueueBase {
...
+  jobId,
+  key,
+}: {
+  name?: string;
+  nextMillis: number | string;
+  namespace?: string;
+  jobId?: string;
+  key?: string;
+}) {
+  const checksum = key ? key : md5(`${name}${jobId || ''}${md5(namespace)}`);
```
JefferyHus, line 219: Same here 

```typescript
const checksum = key ?? md5(`${name}${jobId ?? ''}${md5(namespace)}`);
```

EinfachHans:
This feature is really needed for me and this pr is open for quite long. Is it still planned to support this feature?

manast (approved):
LGTM

PR #1825, opened 2023-04-14
chore(python): use isinstance for object type comparison

nullndr:
## Changelog

As [PEP8](https://peps.python.org/pep-0008/#programming-recommendations), you should use `isinstance` instead of `type(a) is B`.


manast:
It seems like there is an issue with pytest as the github action seems to have been running for ever 🤔

nullndr:
@manast is it possible to stop the test and get the logs?

nullndr:
Sorry @manast, but the action seems to be stuck again

manast:
Seems like after the third test it hangs for ever: https://github.com/taskforcesh/bullmq/actions/runs/4761883795/jobs/8463539900?pr=1825

PR #1828, opened 2023-04-14
chore(python): move types into a subfolder

nullndr:
## Changelog

This PR moves the first types we introduced into a subfolder called `types`.

## Notes

I used the name `types` and not `typing` since python already has a builtin library called `typing`.

Sure, we could just do `bullmq.typing`, I don't have any strong opinion about this, feel free to tell me if you think another name will be better.


nullndr:
@manast sorry, I can't understand the error output of the tests for node, is the error related from the PR?

manast:
It is just that the coverage check things that the amount of test coverage has decreased with this PR, which is not really true.

nullndr:
@manast oh thanks

nullndr:
Hey @manast, is there something else that blocks this pr to be merged?

manast:
Sorry for the delay in merging this nice PR.

PR #1830, opened 2023-04-15
feat(python): add updateProgress method

roggervalf:


python/tests/job_tests.py
```diff
@@ -0,0 +1,42 @@
+"""
+Tests for the queue class.
```
manast, line 2: wrong comment?

manast (approved):
LGTM

python/tests/job_tests.py
roggervalf: good catch, fixed now

PR #1837, opened 2023-04-18
fix(python): stop processes when force stop

roggervalf:


python/bullmq/worker.py
```diff
@@ -220,12 +220,20 @@ async def close(self, force: bool = False):
...
             self.forceClosing = True
+            self.cancelProcessing()
 
         self.closing = True
 
         await self.blockingRedisConnection.close()
         await self.redisConnection.close()
 
+    def cancelProcessing(self):
+        print('hola')
```
manast, line 231: Seems to be some debug code left here.

python/bullmq/worker.py
```diff
@@ -220,12 +220,20 @@ async def close(self, force: bool = False):
...
+        for job in self.processing:
+            print('1u')
+            hola = job.cancel()
```
manast, line 234: What is the effect of job.cancel()?

roggervalf: it will stop the execution of that job as in https://www.pythontutorial.net/python-concurrency/python-cancel-tasks/

manast: Interesting.

manast (approved):

PR #1841, opened 2023-04-19
feat(python): accept redis options as string

H4ad:
Like `BullMQ` for typescript, we can pass the connection options of the Redis as string.

- [Reference for redis.from_url](https://redis-py.readthedocs.io/en/stable/connections.html#redis.Redis.from_url)

### DIFF
--- a/python/bullmq/redis_connection.py
+++ b/python/bullmq/redis_connection.py
@@ -6,14 +6,17 @@ class RedisConnection:
     RedisConnection class
     """
 
-    def __init__(self, redisOpts: dict = {}):
-        host = redisOpts.get("host") or "localhost"
-        port = redisOpts.get("port") or 6379
-        db = redisOpts.get("db") or 0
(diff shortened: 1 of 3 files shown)


python/bullmq/redis_connection.py
```diff
@@ -6,14 +6,17 @@ class RedisConnection:
     RedisConnection class
     """
 
-    def __init__(self, redisOpts: dict = {}):
-        host = redisOpts.get("host") or "localhost"
-        port = redisOpts.get("port") or 6379
-        db = redisOpts.get("db") or 0
-        password = redisOpts.get("password") or None
+    def __init__(self, redisOpts: dict | str = {}):
+        if not isinstance(redisOpts, str):
```
nullndr, line 10: This could be rewritten as 

```py
if isinstance(redisOpts, dict):
```

In order to delete the `not` operator

Issue #632, opened 2021-07-06
Stop/Abort a running job


Wolvan:
Queue processes we are running have a very long time to complete. Sometimes it turns out we made a mistake and need to abort the running task. Is there a way to signal the worker to stop its work/clean up and then discard the job?
If not, how would one best go about implementing one? Does the worker still read queue data or something that we can use as some kind of abort list?

manast:
This is a highly requested feature actually and BullMQ does not yet provide a way to do this natively, I would be happy to work on a feature like this if we can come up with a design that works well in javascript, currently the only "cancelable" asynchronous object that I know of are observables (since promise cancellation is not part of the official specification).

Wolvan:
We can design promises in a way that allows cancelling, just like bluebird does. In case of cancellation, we reject the job promise. I can probably try and provide a code snippet of a promise like that later.
Important is also that we extend the job object to include an onCancel callback of some kind which allows teardown of resources that the worker is using at the moment. Maybe have job extend EventListener and have a "cancel" event, or have a `if (typeof job.onCancel === "function") await job.onCancel()` somewhere in the code that handles cancelling in the queue.
I am not sure how to best notify the worker that the job should be aborted, but I can help looking into having the worker actually do the cancellation.

manast:
yeah, I know bluebird's cancellation mechanism and also I am familiar with the early cancellation specs for promises. The good thing was that you could cancel the promise easily and it would propagate all the way up and down in the promise chain. However AFAIK observables are currently the only "established" cancellation mechanism for async calls...

Wolvan:
Now that I think about it, I don't think you even need to bother with cancelling the promise. The core issue here is that once the job has been started, there is no way to signal the worker anymore. Aborting/Cancelling should be handled by the processor function, it can throw/resolve as needed depending on how the queue user wants to handle it.
We just need a way to actually notify the worker of a requested change.

Could we turn job into an eventemitter? That way we could potentially also send custom messages and the like to the workers.

manast:
@Wolvan an eventemitter would probably be a bit overkill, but a "cancel" method could do it, however we need a new "state" for cancelled jobs then, since currently they can only be completed or failed, of course they could be placed as "failed" with reason "cancelled", but that would be equivalent to just throwing an exception with that reason as message.
In any case the hard part to solve is how to cancel the job from the outside, and I think the currently more javascript idiomatic way to achieve it would be to make the job processor return an observable instead of a promise.

[...]

PR #1844, opened 2023-04-20
feat(job): add timeout option

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/632


rosslavery:
This is exciting!

Not sure from the diff but will this support sandboxed jobs?

manast:
The problem with this is that if a job times out you will end up having a zombie job process that will continue working regardless. That's why in the Pro version we had to use Observables so that the process can indeed be cancelled in the event of a timeout.

roggervalf:
@manast is right, the execution of the main function is not stopped so we will end with zombie jobs, so we should use the Observable solution from pro version

PR #1849, opened 2023-04-22
feat(python): add moveToDelayed script

roggervalf:


manast (commented):
Looks good so far, but I see some out-commented code.

roggervalf:
Yeah, I still need to add some tests too

DanielNetzer:
@manast @roggervalf is this something usable?
We have a python service that can utilize a worker implementation for Python if so.

manast:
@DanielNetzer the python lib uses the same lua scripts as the nodejs version, so apart from lacking some features, it should be quite safe to use in prod environments.

DanielNetzer:
As always, thanks for the super fast response.  @manast 
Is it documented somewhere, the Python usage?

manast:
https://docs.bullmq.io/python/introduction

manast:
It is not a lot but hopefully enough. And the warning is 2 months ago, but still, it can be rough on the edges.

DanielNetzer:
@manast this should be enough, again appreciate all the hard work invested in this library which is amazing.
We will start ASAP utilizing the python implementation and get back with feedback 👍🏽

PR #1857, opened 2023-04-27
refactor(finished-on): save finishedOn on job methods

roggervalf:


src/classes/scripts.ts
```diff
@@ -303,39 +303,17 @@ export class Scripts {
...
-      shouldRemove,
-      target,
-      token,
-      timestamp,
-      fetchNext,
-    );
-
     const result = await (<any>client).moveToFinished(args);
     if (result < 0) {
-      throw this.finishedErrors(result, job.id, 'finished', 'active');
```
roggervalf, line 335: here we are using the name of the command, that in this case is moveToFinished, no finished, so this is trying to be consistent with the other command error messages

manast (approved):
LGTM 👍

Issue #1312, opened 2022-07-08
Worker keeps handles open after calling close()


fredgreer:
I am seeing open handle warnings from Jest when creating (and closing) BullMQ workers in an integration test:

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

A minimal reproduction of the issue is:

```
import { Worker } from "bullmq";

describe("BullMQ worker", () => {
  it("leaves handles open after awaiting close()", async () => {
    const worker = new Worker("queue-name", async (job) => {}, {
      connection: {
        maxRetriesPerRequest: null,
      },
    });
    await new Promise((res) => setTimeout(res, 100));

    await worker.close();
  });
});
```

Interestingly the issue does not occur with the `setTimeout()` call removed.

- BullMQ version:  1.86.4
- jest version: 28.1.2
- ts-jest version: 28.0.5

I've uploaded the full reproduction here: https://github.com/PropellerAero/bullmq-open-handle-bug-repro

MeoMix:
Heyo. Just chiming in that I also experience this issue. I spent a good amount of time trying to track down what was happening in the BullMQ source before finding this open ticket.

I have one QueueScheduler, one Worker, and no jobs. I have no issues with tearing down QueueScheduler, but I do have issues with tearing down my Worker.

If I add a generous wait after teardown, all is good:

```
  async function stop() {
    if (worker) {
      await worker.close();
      await workerRunPromise;
      console.debug(`STOP: Worker for ${queue} isRunning: ${worker.isRunning()}`);
    }

    await queueScheduler.close(true);
    await queueSchedulerRunPromise;
    console.debug(`STOP: QueueScheduler for ${queue} isRunning: ${queueScheduler.isRunning()}`);

    await new Promise((resolve) => setTimeout(resolve, 1000));
  }
  ```
  
  but if I remove the timeout - Jest complains.
  
Digging through the source a bit, I was able to track it down to the processing loop:
  
  ```
                      const processing = (this.processing = new Map());
                    while (!this.closing) {
                        if (processing.size < this.opts.concurrency) {
                            const token = (0, uuid_1.v4)();
                            processing.set(this.retryIfFailed(() => this.getNextJob(token), this.opts.runRetryDelay), token);
                        }
                        /*
                         * Get the first promise that completes
                         */
                        const promises = [...processing.keys()];
[...]

PR #1861, opened 2023-04-28
fix(worker): open handle

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/1312

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -883,6 +883,7 @@ export class Scripts {
       Date.now(),
       opts.stalledInterval,
     ];
+    console.log('ble');
     return (<any>client).moveStalledJobsToWait(keys.concat(args));
   }
 
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -883,7 +883,6 @@ export class Scripts {
       Date.now(),
       opts.stalledInterval,
     ];
-    console.log('ble');
     return (<any>client).moveStalledJobsToWait(keys.concat(args));
   }
 
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -784,12 +784,10 @@ export class Worker<
     if (!this.opts.skipStalledCheck) {
       clearTimeout(this.stalledCheckTimer);
 
+      this.stalledCheckTimer = setTimeout(async () => {
+        await this.startStalledCheckTimer();
+      }, this.opts.stalledInterval);
       await this.runStalledJobsCheck();
-      if (!this.closing) {
-        this.stalledCheckTimer = setTimeout(async () => {
-          await this.startStalledCheckTimer();
-        }, this.opts.stalledInterval);
-      }
     }
   }
 
@@ -913,25 +911,23 @@ export class Worker<
 
     stalled.forEach((jobId: string) => this.emit('stalled', jobId, 'active'));
 
-    let jobPromises: Promise<Job<DataType, ResultType, NameType>>[] = [];
-    for (let i = 1; i <= failed.length; i++) {
+    const jobPromises: Promise<Job<DataType, ResultType, NameType>>[] = [];
+    for (let i = 0; i < failed.length; i++) {
       jobPromises.push(
         Job.fromId<DataType, ResultType, NameType>(
           this as MinimalQueue,
-          failed[i - 1],
+          failed[i],
         ),
       );
 
-      if (i % chunkSize === 0) {
-        const jobs = await Promise.all(jobPromises);
-        this.notifyFailedJobs(jobs);
-        jobPromises = [];
+      if ((i + 1) % chunkSize === 0) {
+        this.notifyFailedJobs(await Promise.all(jobPromises));
+        jobPromises.length = 0;
       }
     }
 
     if (jobPromises.length > 0) {
-      const jobs = await Promise.all(jobPromises);
-      this.notifyFailedJobs(jobs);
+      this.notifyFailedJobs(await Promise.all(jobPromises));
     }
   }
 
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -922,9 +922,7 @@ export class Worker<
       }
     }
 
-    if (jobPromises.length > 0) {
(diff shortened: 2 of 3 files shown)


src/classes/scripts.ts
```diff
@@ -883,6 +883,7 @@ export class Scripts {
       Date.now(),
       opts.stalledInterval,
     ];
+    console.log('ble');
```
manast, line 886: comment

manast (commented):
What is the purpose of this PR?

src/classes/worker.ts
```diff
@@ -910,21 +913,26 @@ export class Worker<
...
           this as MinimalQueue,
-          failed[i],
+          failed[i - 1],
         ),
       );
 
       if (i % chunkSize === 0) {
-        this.notifyFailedJobs(await Promise.all(jobPromises));
-        jobPromises.length = 0;
+        const jobs = await Promise.all(jobPromises);
```
manast, line 926: why is this better?

src/classes/worker.ts
```diff
@@ -910,21 +913,26 @@ export class Worker<
...
+        this.notifyFailedJobs(jobs);
+        jobPromises = [];
       }
     }
-    this.notifyFailedJobs(await Promise.all(jobPromises));
+
+    if (jobPromises.length > 0) {
```
manast, line 932: Should be needed, if jobPromises is an empty array, Promise.all will just return an empty array.

src/classes/worker.ts
manast: I spent some time trying to reduce allocations as much as possible, so thats why I prefer a length = 0 instead of allocating a new array.

src/classes/worker.ts
```diff
@@ -910,21 +913,26 @@ export class Worker<
...
 
-    const jobPromises: Promise<Job<DataType, ResultType, NameType>>[] = [];
-    for (let i = 0; i < failed.length; i++) {
+    let jobPromises: Promise<Job<DataType, ResultType, NameType>>[] = [];
+    for (let i = 1; i <= failed.length; i++) {
       jobPromises.push(
         Job.fromId<DataType, ResultType, NameType>(
           this as MinimalQueue,
-          failed[i],
+          failed[i - 1],
```
manast, line 921: why - 1?

src/classes/worker.ts
```diff
@@ -786,9 +785,11 @@ export class Worker<
       clearTimeout(this.stalledCheckTimer);
 
       await this.runStalledJobsCheck();
-      this.stalledCheckTimer = setTimeout(async () => {
-        this.startStalledCheckTimer();
-      }, this.opts.stalledInterval);
+      if (!this.closing) {
```
manast, line 788: I think that this code can be rearranged and then no need for checking this.closing here like this:
```ts
this.stalledCheckTimer = setTimeout(async () => {
  await this.startStalledCheckTimer();
}, this.opts.stalledInterval);
await this.runStalledJobsCheck();
```

The reason is that before the first time ```startStalledCheckTimer``` is called, we check if the worker is closing, so we can assume that. When close is called the stalledCheckTimer will be cleared, so it will not run this timer after that either.

src/classes/worker.ts
roggervalf: it was part of the first approach of trying to fix this issue, now I restore the length = 0 logic

src/classes/worker.ts
roggervalf: yeah, when it is empty we don't need to notify failed jobs

src/classes/worker.ts
roggervalf: this change was restored at the first version

src/classes/worker.ts
roggervalf: changed to the suggested form

src/classes/worker.ts
```diff
@@ -785,47 +784,49 @@ export class Worker<
...
       this.stalledCheckTimer = setTimeout(async () => {
-        this.startStalledCheckTimer();
+        await this.startStalledCheckTimer();
       }, this.opts.stalledInterval);
+      await this.runStalledJobsCheck();
     }
   }
 
-  private async startLockExtenderTimer(
+  private startLockExtenderTimer(
```
roggervalf, line 794: async is not needed as we are not awaiting any async function, clearTimeout and setTimeout are synchronous

src/classes/worker.ts
```diff
@@ -919,12 +920,15 @@ export class Worker<
         ),
       );
 
-      if (i % chunkSize === 0) {
+      if ((i + 1) % chunkSize === 0) {
```
roggervalf, line 923: this change it is because for example, at first execution (i = 0) at that point jobPromises is having 1 promise, the intention of this should be to execute notifyFailedJobs when jobPromises is having chunkSize equal to the length of this array, no when having just 1 value

src/classes/worker.ts
```diff
@@ -785,47 +784,49 @@ export class Worker<
...
-          const { job, ts } = item;
-          if (!ts) {
-            item.ts = now;
-            continue;
-          }
-
-          if (ts + this.opts.lockRenewTime / 2 < now) {
-            item.ts = now;
-            jobsToExtend.push(job);
+      if (!this.closing) {
```
roggervalf, line 800: at the end what is the real fix is this line to prevent open handles on executing startLockExtenderTimer

src/classes/worker.ts
roggervalf: so this test https://github.com/taskforcesh/bullmq/blob/master/tests/test_repeat.ts#L913 doesn't like this change, so I'm restoring to:

```ts
      await this.runStalledJobsCheck();
      this.stalledCheckTimer = setTimeout(async () => {
        await this.startStalledCheckTimer();
      }, this.opts.stalledInterval);

```

src/classes/worker.ts
manast: Ok

manast (commented):
Just a small comment regarding defensive programming.

src/classes/worker.ts
roggervalf: done

manast (approved):

PR #1867, opened 2023-05-02
Update repeatable.md

vibhanshu-pmi:
Fixed Invalid Cron expression

The job was running every second at 3:15 not just once


roggervalf (approved):
Lgtm

PR #1879, opened 2023-05-05
feat(worker): add worker threads support

manast:
This PR adds "experimental" support for worker threads. The new support runs the same tests as the previous "child fork" based sandboxed processors. It is now possible to choose worker threads by specifying the following option when instantiating a Worker:

```ts
      const worker = new Worker(queueName, processFile, {
        connection,
        useWorkerThreads,
      });
```

Note that this option is only relevant if you specify a "processFile". 

I am not expecting a huge improvement by this option, as worker threads do not seem to be much more lightweight than a standard process, at least in terms of memory consumption, although up start time may be a bit faster, I have not made any benchmarks yet. If the Node team makes improvements in the way worker threads work, things may get better for BullMQ as well.


roggervalf (approved):
Lgtm

PR #1887, opened 2023-05-09
fix(worker): use named import

Sladuca:
As of the [3.13.2](https://github.com/taskforcesh/bullmq/releases/tag/v3.13.2), this library no longer builds when using `@types/node` version >= `13.6` due to removed support for importing `EventEmitter` via default export.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41353#issuecomment-579381008 for more details

This PR fixes this by importing via named export instead of default export.


roggervalf (approved):
Lgtm

PR #1889, opened 2023-05-10
perf(rate-limit): call pttl in script moveJobFromActiveToWait

roggervalf:


src/commands/takeLock-1.lua
```diff
@@ -1,17 +0,0 @@
---[[
```
roggervalf, line 1: not used at all

Issue #1899, opened 2023-05-15
Flow: bigger priority or fifo for jobs after going back from waiting-children to waiting when using the WaitingChildrenError to generate children


sven-codeculture:
**Is your feature request related to a problem? Please describe.**
We have currently refactored our workflow to initialize a customer to use the flow as described here:
https://docs.bullmq.io/patterns/process-step-jobs#waiting-children
We first run a job, which adds the required children and then throws a WaitingChildrenException. This jobs all have children too and use the same mechanic too. (There are usally 3 steps, create customer, fetch all data for a specific data type, do a job for each fetched data and add it to the database)
This all works like a charm when only initializing a small bunch of customers at the same time, however when doing something like a re-initialize of all customers (which we sometimes require to do) this comes up to about 80 million jobs ran in the end.
![Flow](https://github.com/taskforcesh/bullmq/assets/3930643/fe27ca56-57c1-4816-b3b4-4db7ea8a3b67)

Due to the fact, that the parent gets added at the end of the queue with the same priority as the parents who still have to create its children, this causes to swell the required redis memory over time and at some point redis (we use AWS elasticache) runs out of memory and everything starts failing.

**Describe the solution you'd like**
The best way would be a possibility for flow parents to optionally change their priority when they move from waiting-children back to waiting. Also putting them as fifo when they get into waiting queue a second time. 

**Describe alternatives you've considered**
As a current workaround we had to limit the amount of customers initializing at the same time. This works as a workaround, however we always lose some time at the end of the number x customers initializing due to the fact that our cron checking if update/initializing is required only runs once every 5 minutes due to API limits.

**Additional context**
Using fifo directly is not an option. We run jobs with different priorities to make sure all child-data are processed before the next parent is touched, to prevent a memory overflow in redis.

PS: if required i can try to find some time implementing this and creating a PR, however want the OK that this logic would be ok for the bullmq project owners

roggervalf:
hi @sven-codeculture, reading your case I'm thinking on a new method, something like **changePriority** that could be call in your case before moving the parent to waiting-children for example

roggervalf:
[changePriority](https://api.docs.bullmq.io/classes/Job.html#changePriority) method is available since version 3.15.0

PR #1901, opened 2023-05-16
feat(job): add changePriority method

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1899


manast (commented):
Looks good. I guess we need some piece of documentation on how to use this method and also explaning that it can only be used when the job is waiting.

docs/gitbook/guide/jobs/delayed.md
```diff
@@ -20,9 +20,26 @@ await myQueue.add('house', { color: 'white' }, { delay: 5000 });
...
+If you want to change the delay after inserting a delayed job, just use **changeDelay** method. For example, let's say you want to change the delay from 2000 to 4000 milliseconds:
+
+```typescript
+const job = await Job.create(queue, 'test', { foo: 'bar' }, { delay: 2000 });
+
+await job.changeDelay(4000);
+```
+
+{% hint style="warning" %}
+Take in count that your job must be in delayed state when you change the delay.
```
manast, line 40: in -> into

docs/gitbook/guide/jobs/prioritized.md
```diff
@@ -24,3 +24,25 @@ await myQueue.add('wall', { color: 'blue' }, { priority: 7 });
 ```
 
 If several jobs are added with the same priority value, then the jobs within that priority will be processed in FIFO (First in first out) fashion.
+
+## Change priority
+
+If you want to change the priority after inserting a job, just use **changePriority** method. For example, let's say you want to change the priority from 16 to 1:
```
manast, line 30: -> just use *the* **changePriority** method.

-> let's say *that* you want to change...

manast (commented):
Just a few grammar issues

PR #1931, opened 2023-06-01
fix(rate-limit): consider paused queue

roggervalf:


src/commands/includes/getRateLimitTTL.lua
```diff
@@ -1,10 +1,14 @@
...
+local function getRateLimitTTL(maxJobs, rateLimiterKey)
   if maxJobs then
-    local jobCounter = tonumber(rcall("GET", limiterKey))
-    if jobCounter ~= nil and jobCounter >= maxJobs then
-      local pttl = rcall("PTTL", limiterKey)
-      if pttl > 0 then 
+    local pttl = rcall("PTTL", rateLimiterKey)
+
+    if pttl <= 0 then
+      rcall("DEL", rateLimiterKey)
```
roggervalf, line 6: we can delete the key here is it is 0

src/commands/includes/moveJobFromWaitToActive.lua
```diff
@@ -39,13 +31,13 @@ local function moveJobFromWaitToActive(keys, keyPrefix, jobId, processedOn, opts
     end
 
     -- check if we passed rate limit, we need to remove the job and return expireTime
-    if jobCounter > maxJobs then      
+    if expireTime > 0 then
       -- remove from active queue and add back to the wait list
       rcall("LREM", keys[2], 1, jobId)
-      rcall("RPUSH", keys[1], jobId)
+      rcall("RPUSH", targetKey, jobId)
```
roggervalf, line 37: consider paused queue, not only wait list

src/commands/includes/moveJobFromWaitToActive.lua
```diff
@@ -39,13 +31,13 @@ local function moveJobFromWaitToActive(keys, keyPrefix, jobId, processedOn, opts
...
 
       -- Return when we can process more jobs
-      return {0, 0, expireTime}
+      return {0, 0, expireTime, 0}
```
roggervalf, line 40: return our 4 values to prevent python errors when try to destructure

src/commands/moveToActive-9.lua
```diff
@@ -67,6 +67,11 @@ end
 if jobId then
   if string.sub(jobId, 1, 2) == "0:" then
     rcall("LREM", KEYS[2], 1, jobId)
+
+    if expireTime > 0 then
+      return { 0, 0, expireTime, 0 }
```
roggervalf, line 72: as we delete the previous marker, if expireTime is greater than 0 (in rate-limit), we can return expireTime

src/commands/moveToFinished-12.lua
```diff
@@ -192,20 +192,25 @@ if rcall("EXISTS", jobIdKey) == 1 then -- // Make sure job exists
...
         -- If jobId is special ID 0:delay, then there is no job to process
         if jobId then
             if string.sub(jobId, 1, 2) == "0:" then
                 rcall("LREM", KEYS[2], 1, jobId)
+
+                if expireTime > 0 then
+                    return { 0, 0, expireTime, 0 }
+                end              
             else
-                opts = opts or cmsgpack.unpack(ARGV[4])
```
roggervalf, line 206: here I'm fixing an issue, as ARGV[4] is the result of failedReason, not opts. opts is declare at the top of this script, we don't need to reassign a new value

tests/test_rate_limiter.ts
```diff
@@ -107,6 +107,60 @@ describe('Rate Limiter', function () {
...
+        queueEvents.once('completed', async () => {
+          resolve();
+        });
+
+        queueEvents.on('failed', async err => {
+          reject(err);
+        });
+      });
+
+      await delay(500);
```
roggervalf, line 138: in order to get drained attribute as true for all workers

PR #1940, opened 2023-06-02
fix: add a 0.1 sleep to resolve 100% cpu when redis disconnects

gonzafirewall:
Added a 0.1 sleep because this issue makes servers that I use like workers and scale using CPU measure scale to max when redis has some issue.

### DIFF
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -78,6 +78,7 @@ class Worker(EventEmitter):
                 # This should never happen or we will have an endless loop
                 print("ERROR:", e)
                 traceback.print_exc()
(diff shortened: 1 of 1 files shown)


python/bullmq/worker.py
```diff
@@ -199,3 +200,4 @@ def extract_result(job_task):
             # print error message and ignore the job
             print("ERROR:", e)
             traceback.print_exc()    
+        time.sleep(0.1)
```
roggervalf, line 203: there is a retry strategy in node version https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L862-L876 you can use it as reference

Issue #1941, opened 2023-06-02
[Bug]: not exists reprocessJob-6.lua 


gonzafirewall:
### Version

v0.4.2

### Platform

Python

### What happened?

A bug happened!

FileNotFoundError: [Errno 2] No such file or directory: '<PATH>/.venv/lib/python3.10/site-packages/bullmq/commands/reprocessJob-6.lua'

I investigate a little and efectivelly in the code was trying to get the 6 but in the commands folder only exists the 4.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

roggervalf:
same as https://github.com/taskforcesh/bullmq/issues/1952

PR #1942, opened 2023-06-02
feat: resolve commands dinamically based in a list of availables

gonzafirewall:
maybe this can help to have less errors because of typos or forgets

related to issue #1941


roggervalf:
Thank you @gonzafirewall, our problem was related with our manual way of copying lua scripts, but with our new automated pipeline this should be solved

PR #1953, opened 2023-06-07
feat(job): add removeDependencyOnFailure option

roggervalf:
https://github.com/taskforcesh/bullmq/pull/1100 ref #800


docs/gitbook/guide/flows/remove-dependency.md
```diff
@@ -0,0 +1,48 @@
+# Remove Dependency
+
+In some situations, you need to move a job to wait when one of its children fail.
+
+The pattern to solve this requirement consists on using **removeDependencyOnFailure** option.
```
manast, line 5: on using *the* **removeDependencyOnFailure** option.

docs/gitbook/guide/flows/remove-dependency.md
```diff
@@ -0,0 +1,48 @@
+# Remove Dependency
+
+In some situations, you need to move a job to wait when one of its children fail.
```
manast, line 3: Is this sentence accurate? I had the impression that the parent job will only be moved to wait when either *all* of its dependencies complete or fail, not just when one of its children fails.

src/classes/scripts.ts
```diff
@@ -89,8 +89,17 @@ export class Scripts {
       queueKeys.events,
     ];
 
-    const fpof = opts.fpof ? { fpof: true } : {};
-    const parent = job.parent ? { ...job.parent, ...fpof } : null;
+    const parent: Record<string, any> = job.parent ? { ...job.parent } : null;
+
+    if (parent) {
```
manast, line 94: why are the options applied to the parent?

src/classes/job.ts
```diff
@@ -39,6 +39,7 @@ const logger = debuglog('bull');
 const optsDecodeMap = {
   fpof: 'failParentOnFailure',
   kl: 'keepLogs',
+  rdof: 'removeDependencyOnFailure',
```
manast, line 42: since the options are mutually exclusive, we should probably need to throw an exception if the user sets both to true...

src/classes/scripts.ts
roggervalf: instead of using spread operator, we can just assign those attributes into parent object that will be faster

docs/gitbook/guide/flows/remove-dependency.md
roggervalf: I tried to rephrase it

docs/gitbook/guide/flows/remove-dependency.md
```diff
@@ -0,0 +1,48 @@
+# Remove Dependency
+
+In some situations, you may have a parent job and need to ignore when one of its children fail.
+
+The pattern to solve this requirement consists on using the **removeDependencyOnFailure** option.
+
```
manast, line 6: add: "This option will make sure that when a job fails, the dependency is removed from the parent, so the parent will complete without waiting for the failed children."

docs/gitbook/guide/flows/remove-dependency.md
```diff
@@ -0,0 +1,48 @@
...
+      name,
+      data: { idx: 3, foo: 'foo' },
+      queueName: 'childrenQueueName',
+    },
+  ],
+});
+```
+
+{% hint style="info" %}
+As soon as a _child_ with this option fails, the parent job will be moved to waiting state only if there are not more pending children.
```
manast, line 43: Correct grammar: "As soon as a child with this option fails, the parent job will be moved to a waiting state only if there are no more pending children."

src/classes/job.ts
```diff
@@ -1077,6 +1078,12 @@ export class Job<
       throw new Error(`Delay and repeat options could not be used together`);
     }
 
+    if (this.opts.removeDependencyOnFailure && this.opts.failParentOnFailure) {
+      throw new Error(
+        `RemoveDependencyOnFailure and failParentOnFailure options could not be used together`,
```
manast, line 1083: could -> can

src/classes/scripts.ts
manast: as long as the parent object is not passed as an argument it is ok, otherwise we should never mutate input arguments.

manast (approved):

PR #1957, opened 2023-06-07
Python release

roggervalf:


python/requirements.txt
```diff
@@ -5,8 +5,10 @@ filelock==3.12.0
 msgpack==1.0.5
 pipenv==2023.5.19
 platformdirs==3.5.1
+pre-commit==2.20.0
+python-semantic-release==7.34.3
 redis==4.5.5
-semver==3.0.0
+semver==2.13.0
```
manast, line 11: Why an older version?

manast (commented):
Looks good, but don't we also need a build step for the lua scripts?

python/requirements.txt
roggervalf: there is a conflict with python semantic release that is using version 2, as we are using the same method for both version is working for us as well

roggervalf:
> Looks good, but don't we also need a build step for the lua scripts?

in line 36 we are copying lua scripts

manast (approved):

PR #1958, opened 2023-06-07
fix: type of queueEvents for waitUntilFinished

MichalSajdik:
We were debugging this for around 3 days, for us to find out that `Queue`, which extends `MinimalQueue`, does not work and only `QueueEvents` does. 

You can also refer to your documentation: https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/api/bullmq.job.waituntilfinished.md

### DIFF
--- a/.madgerc
+++ b/.madgerc
@@ -1,0 +1,7 @@
+{
+  "detectiveOptions": {
+    "ts": {
+      "skipTypeImports": true
+    }
+  }
+}
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -33,6 +33,7 @@ import {
 import { Backoffs } from './backoffs';
 import { Scripts } from './scripts';
 import { UnrecoverableError } from './unrecoverable-error';
+import { QueueEvents } from "@src/classes/queue-events";
 
 const logger = debuglog('bull');
 
@@ -917,7 +918,7 @@ export class Job<
    * @param ttl - Time in milliseconds to wait for job to finish before timing out.
    */
   async waitUntilFinished(
-    queueEvents: MinimalQueue,
+    queueEvents: QueueEvents,
     ttl?: number,
   ): Promise<ReturnType> {
     await this.queue.waitUntilReady();
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -33,7 +33,7 @@ import {
 import { Backoffs } from './backoffs';
 import { Scripts } from './scripts';
 import { UnrecoverableError } from './unrecoverable-error';
-import { QueueEvents } from "@src/classes/queue-events";
+import type { QueueEvents } from '@src/classes/queue-events';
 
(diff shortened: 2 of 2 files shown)


manast:
hmm, but how would this type make any difference in working or not working? Are you getting compiler errors or what is the issue?

manast:
@roggervalf this type was changed in this commit: https://github.com/taskforcesh/bullmq/commit/f607ec7530fb4430e8cab7ed325583bd9d171ccf
It looks like a typo or what is the reason behind it?

roggervalf:
> @roggervalf this type was changed in this commit: [f607ec7](https://github.com/taskforcesh/bullmq/commit/f607ec7530fb4430e8cab7ed325583bd9d171ccf) It looks like a typo or what is the reason behind it?

yes I remember this change, it was made in order to prevent a circular reference between job and queueEvents, if you execute yarn circular:references you will get an error

manast:
@roggervalf there must be another way of breaking this dependency, otherwise, the code as it is is incorrect, it really only works if passing a QueueEvents instance.

roggervalf:
also this is our new api documentation https://api.docs.bullmq.io/classes/Job.html#waitUntilFinished, the one in the description is deprecated and we don't have complains in our tests https://github.com/taskforcesh/bullmq/blob/master/tests/test_obliterate.ts#L61 by using this method and passing queueEvent instances

roggervalf:
what I'm thinking is creating another type similar to MinimalQueue but for QueueEvents https://github.com/taskforcesh/bullmq/blob/master/src/types/minimal-queue.ts and use it here in order to keep it without circular references

capponMatteo:
Couldn't using `import type { QueueEvents } from './queue-events';` be a solution? That should just be used for type checking and the import should be removed at compile time.

capponMatteo:
Just tried adding this in a new `.madgerc` file and using `import type { QueueEvents } from './queue-events';` and `yarn circular:references` is now happy 😁
Not sure this is the correct way of handling it though.
```
{
  "detectiveOptions": {
    "ts": {
      "skipTypeImports": true
    }
  }
}
```

roggervalf:
I think we can try it, taking as reference https://github.com/pahen/madge/issues/232

roggervalf (approved):
Thank you @capponMatteo

jRexhmati:
I think this pull request is causing errors in code building

node_modules/bullmq/dist/esm/classes/job.d.ts:5:34 - error TS2307: Cannot find module '@src/classes/queue-events' or its corresponding type declarations.

jRexhmati:
When reverting back it works again

roggervalf:
hi @jRexhmati, please try version [3.15.4](https://github.com/taskforcesh/bullmq/releases/tag/v3.15.4) it should be fixed

jRexhmati:
Thanks for the quick fix

PR #1964, opened 2023-06-08
Fix queue-events import in Job

quentin-sommer:
Following recently merged https://github.com/taskforcesh/bullmq/pull/1958
There is a path alias config in the `tsconfig.json` that make the alias work for the typescript compiler. However the output will still contain the alias and it breaks the import.

see this [SO](https://stackoverflow.com/questions/43324209/typescript-do-not-replace-non-relative-paths-defined-in-tsconfig) for another example


roggervalf (approved):
Lgtm, thank you @quentin-sommer

PR #1969, opened 2023-06-09
fix(retry-job): consider priority when moving job to wait (python)

roggervalf:


.github/workflows/release.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           flake8 ./python --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
       - name: Release Python
-        if: ${{ contains(env.commitmsg, '(python):') }}
+        if: ${{ contains(env.commitmsg, '(python)') }}
```
roggervalf, line 55: we can just add (python)  at the end of our commits when changes affects both packages (node and python), so the release pipeline will be trigger for both stacks

PR #1975, opened 2023-06-13
fix(python): add retry strategy in connection

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/1940 and https://redis.readthedocs.io/en/latest/retry.html


python/bullmq/redis_connection.py
```diff
@@ -11,16 +18,20 @@ class RedisConnection:
...
         if isinstance(redisOpts, dict):
             host = redisOpts.get("host") or "localhost"
             port = redisOpts.get("port") or 6379
             db = redisOpts.get("db") or 0
             password = redisOpts.get("password") or None
 
             self.conn = redis.Redis(
-                host=host, port=port, db=db, password=password, decode_responses=True)
+                host=host, port=port, db=db, password=password, decode_responses=True,
+                retry=retry, retry_on_error=[BusyLoadingError, ConnectionError, TimeoutError])
```
manast, line 31: Since the array with the connection errors is used two times I think it could be refactored out.

manast (approved):

jwilm:
This change breaks graceful shutdown. The Worker run loop gets stuck for an extended period in redis.bzpopmin and doesn't get an opportunity to handle the canceling flag raised by Worker.close.

roggervalf:
hi @jwilm could you pls create a different issue to report it. Also would be great to have a test case to reproduce it

PR #1978, opened 2023-06-13
fix(python): fix 'ResponseError: timeout is not an integer or out of range'

wannabewayno:
python: v3.10.12
bullmq(python): v0.5.2
redis: v5.0.4

Encountered when connecting to redis with no existing jobs in the queue, a number of type float (5.0) is passed to the brpoplpush command which causes an error from the redis server that expects an Integer (5).

TL;DR
I'm new to Python (first time coding in python actually)
`yarn test` and `yarn cm` from the contributing guide didn't run for me so I have not performed these checks.
I read the contributing guide and the took inspiration from other commits to comply with the commit linter.

Let me know if you need anything else, I use bullmq at work for Node.js and glad I could give something back.

### DIFF
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -104,7 +104,7 @@ class Worker(EventEmitter):
         if job is None:
             timeout = min(delay_until - int(time.time() * 1000)
                           if delay_until else 5000, 5000) / 1000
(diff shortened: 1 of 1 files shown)


wannabewayno:
> hi @wannabewayno please try version 0.5.3, that should have this fix

Yep! It sure does!
Thanks!

PR #1981, opened 2023-06-13
fix(worker): use timeout as integer for redis lower than v6.0.0 (python)

roggervalf:


manast:
need to use (python) as commit context to avoid making a new release for node's bullmq.

roggervalf:
In this case is ok because in node there is also a change, as we were not using the util function

PR #1984, opened 2023-06-14
perf(priority): change priority as a new state

roggervalf:
decouple priority to keep one zset

BREAKING CHANGE: priority is separeted in its own zset, no duplication needed


manast (commented):
Looks pretty good. Some small comments only.

src/commands/includes/pushBackJobWithPriority.lua
```diff
@@ -1,17 +1,16 @@
...
+
+  if #firstPrioritizedPattern > 0 then
+    local firstTimestamp = string.sub(firstPrioritizedPattern[1], 1, 13)
+    local pprefix = (tonumber(firstTimestamp)-1)
+    rcall("ZADD", priorityKey, priority,  pprefix .. ":" .. jobId)
+    rcall("HSET", jobKey, "pprefix", pprefix)
   else
-    rcall("RPUSH", targetKey, jobId)
+    rcall("ZADD", priorityKey, priority, timestamp .. ":" .. jobId)
+    rcall("HSET", jobKey, "pprefix", timestamp)
```
roggervalf, line 14: what about **priority-prefix** as the name of this new attribute, suggestions?

src/classes/queue-getters.ts
```diff
@@ -287,7 +288,12 @@ export class QueueGetters<
       if (asc && multiCommands[index] === 'lrange') {
         results = results.concat(result.reverse());
       } else {
-        results = results.concat(result);
+        if(types[index] === "priority"){
+          const jobIds = result.map(pattern => pattern.substring(14));
```
manast, line 292: 14 here will not always be correct. It would be better to use a regexp that catches the jobId:
```js
const regex = /^[\d]+:(.*)$/;
```

src/commands/includes/pushBackJobWithPriority.lua
manast: a shorter name, pp if we want to call it priority-prefix, maybe there is a better name 🤔

docs/gitbook/guide/jobs/job-data.md
```diff
@@ -34,15 +34,15 @@ job.data # { color: 'red' }
...
+If you want to change the data after inserting a job, just use the **updateData** method. For example:
 
 {% tabs %}
 {% tab title="TypeScript" %}
 
 ```typescript
 const job = await Job.create(queue, 'wall', { color: 'red' });
 
-await job.update({
+await job.updateData({
```
roggervalf, line 45: more descriptive name

src/commands/moveToActive-10.lua
```diff
@@ -85,7 +93,17 @@ if jobId then
 
   if jobId then
     -- this script is not really moving, it is preparing the job for processing
-    return moveJobFromWaitToActive(KEYS, ARGV[1], target, jobId, ARGV[2], maxJobs, expireTime, opts)
+    return moveJobFromWaitToActive(KEYS, ARGV[1], target, jobId, ARGV[2], maxJobs, expireTime, paused, opts)
```
manast, line 96: This function name "moveJobFromWaitToActive" is starting to become annoying, specially now when we see it mixed with "moveJobFromPriorityToActive" which really is moving the job to active. I think we could rename it to a more correct name like "prepareJobForProcessing" or something like that. What do you think?

manast (commented):
Starts to look ready, just a few comments.

src/commands/moveToActive-10.lua
roggervalf: sounds good to me

src/classes/queue.ts
```diff
@@ -454,4 +455,12 @@ export class Queue<
     const client = await this.client;
     return client.xtrim(this.keys.events, 'MAXLEN', '~', maxLength);
   }
+
+  /**
+   * Delete old priority helper key.
+   */
+  async removePriorityKey(): Promise<number> {
```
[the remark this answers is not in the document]
roggervalf: this comment was addressed few commits ago

manast (approved):

PR #1999, opened 2023-06-19
docs(update): add job data section

roggervalf:


docs/gitbook/guide/jobs/job-data.md
```diff
@@ -0,0 +1,70 @@
+All jobs save data as an attribute.
```
manast, line 1: Every page needs to start with a title, in this case:
```
# Job Data
```

manast, line 1: We can make the text a bit more verbose:
Every job can have its own custom data. The data is stored in the ```data``` attribute of the job:

manast (commented):
Just a small comment.

Issue #1977, opened 2023-06-13
Get repeatable job by job key


ElFisherMan:
We are trying to be as much resilient as possible and to allow re-registering jobs (from mongoDB backup we keep since our redis can fail and we need to make sure we can recover from this).

Our solution is to have a re-registering code when redis is UP again. So we want to go over each job we kept in mongo collection backup and ask if it already exists in BullMQ.

But we can only get all the repeatable jobs and can't fetch them by job key.

We do see that there is a removeRepeatableByKey function so we want to be able to fetch a repeatable job by key.
Can you please add this ability?

geeksilva97:
I can work on a pr if that makes sense.

roggervalf:
base on last refactorization of repeatable jobs, it should be possible

Issue #1735, opened 2023-03-13
How to change repeatable job payload (data)?


ElFisherMan:
I want to use BullMQ as a scheduling solution for all our services. Meaning - each service can register a repeateable job with the following fields: ID, pattern (cron), data.
I would like to allow changing the data, while keeping the current scheduling.
AFAIK this is not possible in BullMQ (right?).
So the second best option - tell them to remove and re-add the reapeatable job with different data.
Having said that - I would like to have a check that checks if the data is different from the one configured for the repeatable job (with the same id and cron) and throw an error in such case.
Is this possible?

(When I call getRepeatableJobs, I only get the following fields (data is missing):
key, name, id, endDate, tz, pattern, next)

manast:
Removing and adding a new updated repeatable job is currently the only viable option. We could add an atomic API that makes this for you though, I will mark it as an enhancement.

manast:
Basically a "updateRepeatable" that atomically removes and creates a new repeatable with the new options.

ElFisherMan:
Sounds great - I think it will be great if you could also add a check that verifies if no relevant field was changed.
In such case - it can ignore the update request.
It will help us a lot (when scaling out takes place and a new service will try upsert the same scheduled task, it won't have to worry for the existance of the same task [id, pattern, payload])

abarbosa-godaddy:
@manast are you working on that already? This feature would be great in a project I am working on. I could open a pr for if there's no objection.

roggervalf:
Hey @abarbosa-godaddy, there is no work in progress about this new feature, still working on changes on the pro version, but you can open a pr for it if you have time and we can take a look.

abarbosa-godaddy:
Will do @roggervalf . Thank you,

geeksilva97:
Btw, I'm `abarbosa` as well (I just didn't want to change account three days ago :see_no_evil: )

manast:
With the new API "upsertJobScheduler" you can update the data payload of your repeatable jobs: https://docs.bullmq.io/guide/job-schedulers

PR #2022, opened 2023-06-27
feat: add getRepeatableJobByKey method to the queue - issue 1997

geeksilva97:
This PR adds the method `getRepeatableJobByKey` method to the Queue class to make it possible to retrieve repeatable job by key.

It addresses the issue #1977

### DIFF
--- a/src/classes/repeat.ts
+++ b/src/classes/repeat.ts
@@ -165,6 +165,15 @@ export class Repeat extends QueueBase {
     return this.scripts.removeRepeatable(repeatJobId, repeatJobKey);
   }
 
+  async getJobIdFromRepeatKey(repeatJobKey: string): Promise<string> {
+    const data = this.keyToData(repeatJobKey);
+    const nextMillis = await (
+      await this.client
+    ).zscore(this.keys.repeat, repeatJobKey);
+
+    return getRepeatJobId(data.name, nextMillis, md5(repeatJobKey), data.id);
+  }
+
   private keyToData(key: string, next?: number) {
     const data = key.split(':');
     const pattern = data.slice(4).join(':') || null;
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -307,6 +307,21 @@ export class Queue<
   }
 
   /**
+   * Gets a repeatable job from its key
+   *
+   * @param key - Job repeatable key
+   */
+  async getRepeatableJobByKey(
+    key: string,
+  ): Promise<Job<DataType, ResultType, NameType> | undefined> {
+    const repeatJobId = await (await this.repeat).getJobIdFromRepeatKey(key);
+
+    if (repeatJobId) {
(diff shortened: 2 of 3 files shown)


geeksilva97:
This change can be useful for solving this issue too https://github.com/taskforcesh/bullmq/issues/1735

src/classes/repeat.ts
```diff
@@ -165,6 +165,15 @@ export class Repeat extends QueueBase {
     return this.scripts.removeRepeatable(repeatJobId, repeatJobKey);
   }
 
+  async getJobIdFromRepeatKey(repeatJobKey: string): Promise<string> {
+    const data = this.keyToData(repeatJobKey);
+    const nextMillis = await (
+      await this.client
+    ).zscore(this.keys.repeat, repeatJobKey);
+
+    return getRepeatJobId(data.name, nextMillis, md5(repeatJobKey), data.id);
```
ElFisherMan, line 174: @geeksilva97 I'm now changing the md5 algorithm to a more genaral and customizable function.
See https://github.com/taskforcesh/bullmq/pull/2023
Anyway - thanks for this change (we really need it!), let's just sync and see which one of them is merged first.

geeksilva97: All right. It sounds good.

manast: Thanks for the PR. Ok so what ```getRepeatableJobByKey``` does is to find the latest job that has been created by the meta repeatable job so to speak. As it is documented somewhere, in the repeatable jobs context there is a "meta" key, that stores the information regarding the repetition configuration and then there is the jobs themselves that are created by the repeatable job mechanism. So in this case, the method is returning the last iteration of the repeatable job, the last job that has not yet been processed, right? So there are 2 things that I think need to be addressed, the first one is the name of the method, which right now can be confusing, I do not have a good suggestion yet... The other thing that is important is that the method needs to do what it does atomically, not in two steps like now, because there is a risk that by the time the job is fetched from the given job id, that job does not exist anymore (or it is not the last job anymore), since the job could complete between the call to ```getRepeatJobId``` and ```getJob```.

manast: Here is the better explanation we have regarding repeatable jobs: https://docs.bullmq.io/guide/jobs/repeatable
For atomicity you will need a LUA script that does both operations at the same time, that's the only way I can think of right now... so not completely trivial to do.

geeksilva97: So that's it. It's a meta job. I can give it a try - cause I'd like to know this concurrency capability of Lua-  but I don't know whether it is worth doing. At least not in this way> Maybe we could have a `RepeatableJob` class that could handle repeatable related things.

It was great to work on that I could walk through the current implementation, and I'm now ready to work on the issue of updating repeatable job data - that seems more feasible and will be useful in a project I'm working on using Bull MQ.

Thank you @manast .

Issue #1932, opened 2023-06-01
Change redis key and job id generation format


ElFisherMan:
As part of the effort of being fips(The Federal Information Processing Standards) complaint, we can't afford to use the md5 algorithem. This algorithem is used to save the job info under redis (part of the key is a md5 string).

Is there a way to change this? (supply an alternative or pass a function that will generate the key)

manast:
The MD5 algorithm is used to create a unique hash as a key for repeatable jobs. The alternative would be to now hash at all, but how would that be better or safer? I am not familiar with FIPS, but I guess what they mean by not using MD5 is for hashing passwords or other type of sensible information, the keys do not have any sensible information so I would be very surprised if this would be necessary.

manast:
@ElFisherMan I guess nobody tried to use BullMQ with FIPS before. As I mentioned, MD5 as it is used by BullMQ does not cause any security risk, so it is unfortunate that they are enforcing the rule for all use cases. We cannot just replace MD5 as that would be a breaking change for the rest of the users. I wonder if copy-pasting an MD5 implementation instead of using the built-in one would work? The reason we use MD5 is because it is small and fast for our needs.

ElFisherMan:
@manast I think copying the implementation will work.
It would help us a lot

manast:
What about changing this (https://github.com/taskforcesh/bullmq/blob/2e7c90d2a5d533d23ec6a9f525259211492097f1/src/classes/repeat.ts#L258):
```ts
function md5(str: string) {
  return createHash('md5').update(str).digest('hex');
}
```

to this:
```ts
const hashMethod = ['m','d','5'].join("")
function hashString(str: string) {
  return createHash(hashMethod).update(str).digest('hex');
}
```

Would it work?

ElFisherMan:
@manast unfortunately no - this algorithm is missing in the fips (infra level - AFAIK OpenSSL FIPS provider).
I think a better approach that does not cause breaking changes is a string value configuration that will be used in these lines.

 return createHash(crypto_hash_algorithm).update(str).digest('hex'); 

for instance - crypto_hash_algorithm='sha1'

I can create a pull request for that in this repo
WDYT?

manast:
@ElFisherMan ok, that sounds good to me.

ElFisherMan:
@manast Great!
I'll start working on it.
Do you have some guidelines for a pull request? (in terms of tests, coding standards, timeline for release)

PR #2023, opened 2023-06-27
feat(common): add option to change repeatable jobs redis key hash algorithm

ElFisherMan:
This PR enables to change the hash algorithm from the default (md5 - non fips complaint) to any string that represent a crypto supported hash algorithm. 

It addresses the issue https://github.com/taskforcesh/bullmq/issues/1932

### DIFF
--- a/src/classes/repeat.ts
+++ b/src/classes/repeat.ts
@@ -8,6 +8,7 @@ import { RedisConnection } from './redis-connection';
 
 export class Repeat extends QueueBase {
   private repeatStrategy: RepeatStrategy;
+  private repeatKeyHashAlgorithm: String;
 
   constructor(
     name: string,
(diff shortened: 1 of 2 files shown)


ElFisherMan:
@manast Can you please review?

manast (commented):
Looks good. Just a couple of tests missing that tests the new functionality and we are good to go.

manast (approved):
LGTM

ElFisherMan:
@manast the test that failed here in the pipeline does not seem to fail locally.
Is it possible that this test is flaky?

ElFisherMan:
@manast Since we want to use this ability - when can we expect a release with these changes?

ElFisherMan:
@manast can you pleae review again so we can merge?

PR #2029, opened 2023-06-29
fix(python): add pyproject.toml configuration

roggervalf:


python/pyproject.toml
```diff
@@ -0,0 +1,61 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "bullmq"
+version = "1.2.0"
```
manast, line 7: will the version be updated automatically by semantic release?

manast (commented):
LGTM but once again I am not sure how this is supposed to work 😅

manast (approved):

python/pyproject.toml
roggervalf: yeah, I'm taking as an example this project https://github.com/JesseMaitland/superglue/commit/99d5e92712eee23ae832785251104b156d9bd730, and semantic-release is updating this line

Issue #1990, opened 2023-06-16
[Bug]: getJobs with pagination returns incorrect number of jobs


roemhildtg:
### Version

v3.12.0

### Platform

NodeJS

### What happened?

I'm querying the queue using the [getJobs](https://api.docs.bullmq.io/classes/Queue.html#getJobs) method and am trying to fetch a limited subset of jobs (10 jobs with start/end values).

```ts
const status = ['completed', 'failed', 'waiting', 'active', 'delayed'];
const limit = 10
const start =  0;
const end = start + limit - 1;
const total = await this.queue.getJobCountByTypes(...status); // returns 28 total jobs
const jobs = await this.queue.getJobs(status, start, end, true); // returns 14 jobs
```

as you can see in the debugger, the number of jobs returned is 14:

<img width="427" alt="image" src="https://github.com/taskforcesh/bullmq/assets/5471079/c3c78187-df87-4cd3-abc0-5705999a1ec6">

### How to reproduce.

```ts


const {Queue} = require('bullmq');

const q = new Queue('jobs');

async function main() {
    for(let i = 0; i < 28; i ++){
        await q.add('http', {});
    }

    const status = ['completed', 'failed', 'waiting', 'active', 'delayed'];
    const total = await q.getJobCountByTypes(...status); // returns 28 total jobs
    const limit = 10;
    let start =  0;
    let end;
    console.log('total', total);
    while(start < total){
        end = start + limit - 1;
        console.log(`query: ${start} - ${end}`);
        const jobs = await q.getJobs(status, start, end, true); // returns 14 jobs
        console.log('jobs returned: ', jobs.length);
        start += limit;
    }
}

main();

```

I think the issue occurs when you have mixed status categories for the jobs. For example, if you have some completed, and some delayed, etc. the counts will be off. 

### Relevant log output

```shell
node queue-test.js
BullMQ: DEPRECATION WARNING! Optional instantiation of Queue, Worker and QueueEvents without providing explicitly a connection or connection options is deprecated. This behaviour will be removed in the next major release
total 224
query: 0 - 9
jobs returned:  24
query: 10 - 19
jobs returned:  16
query: 20 - 29
jobs returned:  10
query: 30 - 39
jobs returned:  10
query: 40 - 49
jobs returned:  10
query: 50 - 59
jobs returned:  10
query: 60 - 69
jobs returned:  10
query: 70 - 79
jobs returned:  10
query: 80 - 89
jobs returned:  10
[...]

PR #2030, opened 2023-06-29
Feature/get jobs async ignore status - #1990

fbalhawan:
This PR addresses issue #1990. I have created `getJobsAsync` method to get a list of jobs in a certain range, despite the job types.

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -345,6 +345,37 @@ export class QueueGetters<
   }
 
   /**
+   * Returns jobs regardless of statuses (note that JobType is synonym for job status)
+   * @param start - zero based index from where to start returning jobs.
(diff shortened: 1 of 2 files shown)


manast:
Thanks for the PR. Unfortunately this approach will not work in general. You are assuming that the job ids are all increasing integers, that may or may not be the case as it is possible to use custom job ids. Furthermore, as the queues grow in age, the number of jobs could be really big, while old jobs are removed as they are completed, or failed... Since there is no possible fix to correct this PR I am afraid I have to close it. In any case thanks for trying.

fbalhawan:
Thanks for the review @manast, I see your concerns. So the function should handle:
1- Custom job ids
2- Big number of jobs with consecutive sequences of ids
I would appreciate it if you can direct me to any other points or concerns to resolve the issue.

Issue #2019, opened 2023-06-26
Version the Docs and API Reference


tybro0103:
**Is your feature request related to a problem? Please describe.**
Just had a hell of a time because I'm seeing `job.updateData()` in the docs and the API reference, but it's nowhere to be found on my job object. Turns out that's because I installed the package last week, but v4 has since came out and `updateData` is a brand new fn name that didn't exist in v3.

**Describe the solution you'd like**
A dropdown in both the docs and API ref to select the version would be great.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

manast:
There seems to be some option in TypeDoc to display the name of the library and the version but I am not able to find how to  do it (https://github.com/TypeStrong/TypeDoc)
<img width="666" alt="image" src="https://github.com/taskforcesh/bullmq/assets/95200/d1218750-b3ae-4885-af6a-24644b1f1e15">

roggervalf:
https://github.com/citkane/typedoc-plugin-versions something to evaluate

manast:
@roggervalf does not look good... maybe better to wait for official support from typedoc.

roggervalf:
they answered me in https://github.com/TypeStrong/typedoc/issues/2325

roggervalf:
with our new api docs https://api.docs.bullmq.io/ we can close this issue

PR #2041, opened 2023-06-30
docs(typedoc): generate api versioning

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2019


manast (approved):
LGTM, I do not fully understand how the api versioning works so it looks ok I guess.

PR #2076, opened 2023-07-19
feat(queue): add promoteJobs to promote all delayed jobs

manast:


src/commands/retryJobs-6.lua
```diff
@@ -1,10 +1,13 @@
 --[[
-  Attempts to retry all failed jobs
+  Attempts to retry all jobs
+
+  Note: as this script now can be used also for completed and delayed jobs, the name "retry"
```
roggervalf, line 4: What about moveJobsToWait?

src/commands/retryJobs-6.lua
```diff
@@ -26,28 +29,44 @@ local rcall = redis.call;
...
-  end
+
+    if KEYS[3]:match("failed$") then
+        for i, key in ipairs(jobs) do
+            local jobKey = KEYS[1] .. key
+            rcall("HDEL", jobKey, "finishedOn", "processedOn", "failedReason")
+        end
+    end
+
+    if KEYS[3]:match("completed$") then
```
roggervalf, line 45: we can use elseif expression here

src/classes/scripts.ts
```diff
@@ -781,6 +781,14 @@ export class Scripts {
     return (<any>client).retryJobs(args);
   }
 
+  async promoteJobs(count = 1000): Promise<number> {
```
roggervalf, line 784: we can pass timestamp as a optional parameter, imagine that people may want to limit only the ones that may be promoted before that time and put as the default value Number.MAX_VALUE.

src/commands/retryJobs-6.lua
manast: hmm, maybe. Btw, I am now thinking this will not work so well with jobs with priorities 🤔

manast: But that problem was there before this PR anyway.

src/classes/scripts.ts
manast: The problem is that in the delayed case the timestamp is always in the future, so I am not sure it makes so much sense that's why I did not include it.

src/commands/retryJobs-6.lua
manast: The problem with priorities is that it would be quite slow too.

manast: I think it would be ok. The main use of this script is to quickly move all the jobs from one state to wait so that they get processed as quickly as possible. It should be an acceptable limitation.

roggervalf: Probably we need a warning in our documentation

src/classes/scripts.ts
roggervalf: yeah I was considering that this parameter will be passed in the future, imagine that you only want to retry the ones that are supposed to be retry the same day, instead of retrying all, you can pass the end of date

src/commands/retryJobs-6.lua
roggervalf: btw we dont' have any piece of documentation about retryJobs method, we have https://docs.bullmq.io/guide/retrying-failing-jobs, but we also can include it in a new section as part of https://docs.bullmq.io/guide/queues section

manast: I can add it directly in gitbook

Issue #1571, opened 2022-12-06
Failed jobs are attempting more than "attempts" value


wootwoot1234:
We are running BullMQ 3.2.4 and are seeing some failed jobs continue to retry more than the attempts value.

<img width="1598" alt="Screen Shot 2022-12-06 at 12 57 36 PM" src="https://user-images.githubusercontent.com/419609/205998662-89998e68-65aa-44fb-9eb7-8610aeb35bd4.png">

tmhao2005:
In our code, we set a dynamic limiter as having exception:

```ts
worker.rateLimit(5_000);
throw Worker.RateLimitError();
```

then the job was moved to `waiting` list and the `attemptsMade` was increased greater than its max attempts value. 

Basically we can handle this sort of logic from our end but we think the expected result is the job getting moved to failed list instead.

manast:
Not sure if that is the best thing to do in all cases. Since this is a special error that for example does not trigger backoffs, etc. The job has not really failed, it has just been moved back to wait as it is now considered rate-limited.

manast:
I will think more about it...

benmkramer:
I am having some similar issues. I see the job run both more than the defined attemps (seems to continuously retry no matter what... maybe that is expected since the rate limit is kind of a "try again later" more than a ... something went wrong and just want to try again)

Here is some code that can reproduce for me:
```javascript
const { Worker, Queue } = require("bullmq");
const Redis = require("ioredis");

const redis = new Redis("redis://localhost:6379", {
  maxRetriesPerRequest: null,
  enableReadyCheck: false,
});

const worker = new Worker("myQueue", async () => {
  console.log("running job...");
  const [isRateLimited, duration] = doSomething(); // always true, 1_000
  if (isRateLimited) {
    await worker.rateLimit(duration);
    throw Worker.RateLimitError();
  }
  console.log("done!"); // never reaches this point
}, {
  connection: redis,
  // WARNING: turning this on seems to run the job MANY MANY times
  // concurrency: 2, 
});

function doSomething() {
  return [true, 1_000];
}

const queue = new Queue("myQueue", { connection: redis });
queue.add("someJob", { foo: "bar" }, {
  attempts: 3,
});
```

Corresponding package.json dependencies:
```json
  {
    "bullmq": "^3.12.0",
    "ioredis": "^5.3.2"
  }
```

[...]

PR #2080, opened 2023-07-19
docs(retry): add stop retrying jobs pattern

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1571

### DIFF
--- a/docs/gitbook/patterns/stop-retrying-jobs.md
+++ b/docs/gitbook/patterns/stop-retrying-jobs.md
@@ -1,0 +1,54 @@
+# Stop retrying jobs
+
+When a processor throws an exception that is considered unrecoverable, you should use the `UnrecoverableError` class. In this case, BullMQ will just move the job to the failed set without performing any retries overriding any attempts settings used when adding the job to the queue.
+
+```typescript
+import { Worker, UnrecoverableError } from 'bullmq';
+
+const worker = new Worker('foo', async job => {doSomeProcessing();
+throw new UnrecoverableError('Unrecoverable');
+}, {
+  connection
+  },
+});
+
+await queue.add(
+  'test-retry',
+  { foo: 'bar' },
+  {
+    attempts: 3,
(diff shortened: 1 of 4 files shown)


docs/gitbook/patterns/stop-retrying-jobs.md
```diff
@@ -0,0 +1,54 @@
...
+```typescript
+import { Worker, UnrecoverableError } from 'bullmq';
+
+const worker = new Worker(
+  'myQueue',
+  async job => {
+    const [isRateLimited, duration] = await doExternalCall();
+    if (isRateLimited) {
+      await worker.rateLimit(duration);
+      if (job.attemptsMade >= job.opts.attempts) {
```
manast, line 38: Can't we do this check *before* ```await worker.rateLimit(duration)``` ?

tests/test_rate_limiter.ts
```diff
@@ -377,6 +383,56 @@ describe('Rate Limiter', function () {
...
+    describe('when reaching max attempts and we want to move the job to failed', () => {
+      it('should throw Unrecoverable error', async function () {
+        const dynamicLimit = 550;
+        const duration = 100;
+
+        const worker = new Worker(
+          queueName,
+          async job => {
+            await worker.rateLimit(dynamicLimit);
+            if (job.attemptsMade >= job.opts.attempts!) {
```
manast, line 395: Same comment as above.

docs/gitbook/patterns/stop-retrying-jobs.md
roggervalf: We can, but it's better to set the queue in rate limit before so other jobs after this one Will not be picked

tests/test_rate_limiter.ts
roggervalf: Same as above

manast (approved):

PR #2096, opened 2023-07-25
docs(bullmq-pro): add prioritized section

roggervalf:


docs/gitbook/bullmq-pro/groups/prioritized.md
```diff
@@ -0,0 +1,16 @@
+# Prioritized intra-groups
+
+BullMQ Pro supports priorities per group. A job is prioritized in a group when group and priority options are provided together.
+
+```typescript
+await myQueue.add('paint', { foo: 'bar' }, {
+    group: {
+        id: 'groupId',
+    },
+    priority: 10
```
manast, line 10: Shouldn't the priority setting be inside the group?

manast: Semantically makes more sense to me. If the user uses both priorities (group and standard) we could throw an exception.

manast (approved):

PR #2100, opened 2023-07-27
docs(readme): fix novu logo in used by section

michaldziuba03:
### Changes:

- fix broken logo
- correct "alt" value

## Before

before

## After

after


manast:
Thanks for the fix!

PR #2108, opened 2023-08-02
feat: add username option to redisOpts

f11y11:
This change adds username option to redisOpts. currently, you're unable to set a custom username for the Redis connection.


roggervalf (approved):
Lgtm

PR #2109, opened 2023-08-03
refactor(queue-keys): reuse getQueueQualifiedName method

roggervalf:


src/classes/queue-keys.ts
```diff
@@ -9,9 +9,8 @@ export class QueueKeys {
       '',
       'active',
       'wait',
-      'waiting',
```
roggervalf, line 12: some names were removed as we don't use or save those key names

manast (approved):
LGTM

PR #2116, opened 2023-08-05
docs(rate-limit): add getRateLimitTtl usage

roggervalf:


docs/gitbook/guide/rate-limiting.md
```diff
@@ -74,7 +74,8 @@ const worker = new Worker(
     if (isRateLimited) {
       await worker.rateLimit(duration);
       // Do not forget to throw this special exception,
-      // since the job is no longer active after being rate limited.
```
roggervalf, line 77: worker.rateLimit method only sets ttl, it does move the job to from active to wait, so this text should be changed. The movement of the job happens after throwing the error

Issue #2118, opened 2023-08-05
job.moveToDelayed is not a function


ashishnetworks:
### Version

3.15.1 and 4.7.0

### Platform

NodeJS

### What happened?

I'm not sure if it's a bug or what but I'm trying to use the `moveToDelayed` method but it's giving me the error that `job.moveToDelayed is not a function`  I tried checking if the method is written in the code and found that its written but somehow during the runtime, the method is not available. Is there any option that needs to be activated before using the method?

Here is the sample code

> const delayMs = 1000 * 10;
            const delayedTimestamp = Date.now() + delayMs;
            try {
                const delayResult = await job.moveToDelayed(delayedTimestamp, token);
                console.log({ delayresult: delayResult })
            } catch (err) {
                console.log('There was an error in calling delay method', err)
            }
            throw new DelayedError()

And it gives me following ourput

> There was an error in calling delay method TypeError: job.moveToDelayed is not a function

I also tried printing the entire job object and here is the output of the entire job object

> {
  job: {
    id: 'xBqj7gxBrK/874K63AGA6',
    name: '874K63AGA6',
    data: { userContext: [Object], request: [Object], headers: [Object] },
    opts: {
      attempts: 0,
      delay: 0,
      priority: 2,
      jobId: 'xBqj7gxBrK/874K63AGA6',
      timestamp: 1691249682619
    },
    progress: 0,
    attemptsMade: 1,
    processedOn: 1691249712474,
    timestamp: 1691249682619,
    stacktrace: '[]',
    returnvalue: 'null',
    queueName: 'reportQueue',
    prefix: 'bull',
    returnValue: null,
    updateProgress: [AsyncFunction: updateProgress],
    log: [AsyncFunction: log],
    updateData: [AsyncFunction: updateData]
  }
}

It can be clearly seen the method is not available in the object. But the question is why?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

roggervalf:
hi @ashishnetworks, we have a test case where moveToDelayed is part of job instance https://github.com/taskforcesh/bullmq/blob/master/tests/test_worker.ts#L2372, if you can provide a minimal reproducible test case that shows your issue we can take a look.

ashishnetworks:
- I could not build the project locally so was not able to run test cases but yeah I tried running a similar code locally with installed bullMq version (just like your UTs) and I see that moveToDelayes method is working prefectly in normal function workers.
[...]

PR #2122, opened 2023-08-08
feat(sandbox): emulate moveToDelayed method

roggervalf:
https://github.com/taskforcesh/bullmq/issues/2118


manast (approved):
LGTM

aliscott:
@roggervalf @manst I've just been testing this a bit more and one thing I noticed if the processor function I defined takes both the `job` and `token` args then it never completes.

e.g: this works:
```js
module.exports = function (job) {
   console.log('inside job')
}
```

but this doesn't:
```js
module.exports = function (job, token) {
   console.log('inside job')
}
```

From what I can tell, this is because the `processor.length` is now 2, so it's hitting this condition which is promisifying the function : https://github.com/taskforcesh/bullmq/blob/master/src/classes/child-processor.ts#L47

hugorut:
@roggervalf do we have an update on when this will go in? This PR is blocking an important upgrade to our production systems and that we'd like to get in ASAP. Thanks

roggervalf:
> @roggervalf do we have an update on when this will go in? This PR is blocking an important upgrade to our production systems and that we'd like to get in ASAP. Thanks

hi @hugorut, this pr was not merged as @aliscott has a good point, so I opened another pr to address this issue first https://github.com/taskforcesh/bullmq/pull/2142

src/classes/worker.ts
```diff
@@ -651,7 +651,7 @@ export class Worker<
 
           if (
             err instanceof DelayedError ||
-            err.name == 'DelayedError' ||
+            err.message == 'DelayedError' ||
```
andrhamm, line 654: Might this change have been made in error? After switching some `Job`s to be `SandboxedJob`s I'm seeing `DelayedError` exceptions logged. The log object shows `"name": "DelayedError", "message": ""` making me think it is emitting the error later because this condition is not met.

Issue #2124, opened 2023-08-08
[Bug]: Queue.clean on `delayed` job is not working


felixmosh:
### Version

v4.x.0

### Platform

NodeJS

### What happened?

Cleaning all delayed jobs is not working, it returns an empty array (which should contain removed job ids).

Related to https://github.com/felixmosh/bull-board/issues/615

### How to reproduce.

1. Create a job with a delay of `1 min`
2. Call ` queue.clean(500, 1000, 'delayed')` on the queue after few secs.
3. Nothing happens, the waiting job is still in the queue, an empty array returned.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
I think this works as designed actually, has to do with the grace period: https://github.com/taskforcesh/bullmq/blob/master/tests/test_clean.ts#L202

manast:
Can you write a test case that does not pass? this one does:
https://github.com/taskforcesh/bullmq/blob/master/tests/test_clean.ts#L192C2-L200

felixmosh:
https://github.com/taskforcesh/bullmq/pull/2130

PR #2130, opened 2023-08-10
fix: add a failing test cleaning a delayed queue with limit

felixmosh:
Added a failing test for #2124


roggervalf:
hey @felixmosh, it should be fixed in https://github.com/taskforcesh/bullmq/releases/tag/v4.11.1

PR #2133, opened 2023-08-12
fix(queue): different score purpose per state

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/2130
fixes https://github.com/taskforcesh/bullmq/issues/2124


src/commands/includes/cleanSet.lua
```diff
@@ -21,8 +21,9 @@ local function cleanSet(setKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp, l
 
     local jobKey = jobKeyPrefix .. job
     -- * finishedOn says when the job was completed, but it isn't set unless the job has actually completed
+    -- TODO: check if in completed/failed this is needed, as ZRANGEBYSCORE could bring all the expected jobs
     jobTS = getTimestamp(jobKey, attributes)
-    if (not jobTS or jobTS < timestamp) then
+    if (not jobTS or jobTS <= timestamp) then
```
roggervalf, line 26: I think better to consider <= as for completed and failed we can bring all the expected jobs in ZRANGEBYSCORE

src/commands/includes/getJobsInZset.lua
```diff
@@ -2,11 +2,17 @@
 -- of items in a sorted set only run a single iteration. If we simply used
 -- ZRANGE, we may take a long time traversing through jobs that are within the
 -- grace period.
-local function getJobsInZset(zsetKey, rangeStart, rangeEnd, maxTimestamp, limit)
+local function getJobsInZset(zsetKey, rangeStart, rangeEnd, maxTimestamp, limit, useTimestampAsScore)
```
manast, line 5: I am confused, why aren't we using rangeEnd anymore?
Btw, instead of sending useTimeAsScore, couldn't we just set the proper rangeEnd argument before calling this function?

manast: rangeStart is not used anymore either 🤔

src/commands/includes/cleanSet.lua
manast: I need more context to understand this 😅

roggervalf:
https://github.com/taskforcesh/bullmq/pull/2133#discussion_r1292151758 just to be consistent with a perf change in last commit, when getting finished states, we are getting all the jobs in the period of time to be removed also including the time passed parameter

src/commands/includes/cleanSet.lua
```diff
@@ -20,12 +20,18 @@ local function cleanSet(setKey, jobKeyPrefix, rangeStart, rangeEnd, timestamp, l
     end
 
     local jobKey = jobKeyPrefix .. job
-    -- * finishedOn says when the job was completed, but it isn't set unless the job has actually completed
-    jobTS = getTimestamp(jobKey, attributes)
-    if (not jobTS or jobTS < timestamp) then
+    if isFinished then
```
roggervalf, line 23: when clean is made for completed or failed states, we should be ok deleting job keys without timestamp check, as when we get these records we uses timestamp as one of the limits, this is why I'm using <= to be consistent on the other states

manast (approved):
LGTM

PR #2141, opened 2023-08-16
fix(worker): abort rate-limit delay when closing worker

manast:
fixes https://github.com/taskforcesh/bullmq-pro-support/issues/56

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -2,6 +2,10 @@ import * as fs from 'fs';
 import { Redis } from 'ioredis';
 import * as path from 'path';
 import { v4 } from 'uuid';
+
+// Note: this Polyfill is only needed for Node versions < 15.4.0
+import { AbortController } from 'node-abort-controller';
+
 import {
   GetNextJobOptions,
   IoredisListener,
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -166,6 +166,7 @@ export class Worker<
   private running = false;
   private blockUntil = 0;
   private limitUntil = 0;
+  private abortDelayController: AbortController | null = null;
 
   protected processFn: Processor<DataType, ResultType, NameType>;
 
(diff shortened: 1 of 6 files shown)


src/classes/worker.ts
```diff
@@ -474,8 +479,9 @@ export class Worker<
       }
     } else {
       if (this.limitUntil) {
-        // TODO: We need to be able to break this delay when we are closing the worker.
-        await this.delay(this.limitUntil);
+        this.abortDelayController?.abort();
```
roggervalf, line 482: what about deleting this line and the next one, and add the initialization of abortDelayController in the worker constructor, looks like the test pass

src/classes/worker.ts
```diff
@@ -567,8 +573,11 @@ export class Worker<
    *
    * This function is exposed only for testing purposes.
    */
-  async delay(milliseconds?: number): Promise<void> {
-    await delay(milliseconds || DELAY_TIME_1);
+  async delay(
+    milliseconds?: number,
+    abortController?: AbortController,
+  ): Promise<void> {
+    await delay(milliseconds || DELAY_TIME_1, abortController);
```
roggervalf, line 580: could we instead of passing aboutController as a param, just use this.abortDelayController

src/classes/worker.ts
manast: We could but if we did that, it assumes that you can only abort once, as an abort controller can only be used once... currently, we only can abort when closing, but if we wanted to abort the delay in a different scenario we would need to go back to this.

src/classes/worker.ts
manast: That would imply that every time we call delay it adds a new listener to the abortController, in other words, it would keep increasing memory until the worker is closed.

src/classes/worker.ts
manast: Also I added a comment below regarding memory consumption...

PR #2142, opened 2023-08-18
fix(sandbox): ignore extra params on processor

roggervalf:


src/classes/child-processor.ts
```diff
@@ -44,8 +43,15 @@ export class ChildProcessor {
       });
     }
 
+    const origProcessor = processor;
     if (processor.length > 1) {
```
roggervalf, line 47: this path was not tested, when there are more than 1 parameter, we lose completed event

src/classes/child-processor.ts
```diff
@@ -44,8 +43,15 @@ export class ChildProcessor {
...
-      processor = promisify(processor);
+      processor = function (arg: any) {
```
manast, line 48: I am confused about this, if there are more than 1 argument, why do we only define 1 argument (arg)

manast: In fact I do not understand why we need to separate between 1 and more than 1 🤔

manast (approved):
LGTM

PR #2143, opened 2023-08-18
fix(python): job getReturnValue not returning returnvalue

saulane:
function getReturnValue() for json loading the string returnvalue from a completed Job is not returning the value


manast:
oh, that was a miss from our part 😂

roggervalf (approved):
Lgtm

PR #2158, opened 2023-08-29
Improve documentation grammar, format, etc.

sambostock:
This makes minor changes to the documentation, mostly consisting of small grammar and spelling fixes, as well as formatting, and some rewording.

---

I was reading through the documentation and noticed a few things, so I figured I'd fix them. I kept thinking I was close to the end of the documentation, so I might as well keep going... There was more than I thought, but I did skim through all of it.

Note that I am a new user of BullMQ, so I can not be sure if the content is actually correct (except in cases where it was obvious). This just aims to improve the language used.


docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -78,13 +78,13 @@ worker = Worker("queueName", process, {"connection": connection})
...
+There are situations when it is useful to delay a job when it is being processed.
 
-This can be handled using the `moveToDelayed` method. However, it is important to note that when a job is being processed by a worker, the worker keeps a lock on this job with a certain token value. For the `moveToDelayed` method to work, we need to pass said token so that it can unlock without error. Finally, we need to exit from the processor by throwing a special error `DelayedError` that will signal the worker that the job has been delayed so that it does not try to complete (or fail the job) instead.
+This can be handled using the `moveToDelayed` method. However, it is important to note that when a job is being processed by a worker, the worker keeps a lock on this job with a certain token value. For the `moveToDelayed` method to work, we need to pass said token so that it can unlock without error. Finally, we need to exit from the processor by throwing a special error (`DelayedError`) that will signal to the worker that the job has been delayed so that it does not try to complete (or fail the job) instead.
 
 ```typescript
 import { DelayedError, Worker } from 'bullmq';
```
sambostock, line 90: I think this section needs to be audited by someone with more familiarity with `moveToDelayed` and `DelayedError`.

The code example uses `moveToDelayed` in one branch and throws `DelayedError` in the other branch, but this is different from the following example of `moveToWaitingChildren` and `WaitingChildrenError`, which are used in the same branch. Moreover, the description does not make it clear what the intended usage is.

docs/gitbook/bullmq-pro/groups/rate-limiting.md
```diff
@@ -28,9 +28,9 @@ const worker = new WorkerPro('myQueue', processFn, {
 
 ### Manual rate-limit
 
-Sometimes is useful to rate-limit a group manually instead of based on some static options. For example, if you have an API that returns 429 (Too many requests), and you want to rate-limit the group based on that response.
+Sometimes is useful to rate-limit a group manually instead of based on some static options. For example, if you have an API that returns `429 Too Many Requests`, and you want to rate-limit the group based on that response.
```
rdmchr, line 31:
```suggestion
Sometimes it's useful to rate-limit a group manually instead of based on some static options. For example, if you have an API that returns `429 Too Many Requests`, and you want to rate-limit the group based on that response.
```

Not sure, but I feel like this isn't quite right.

sambostock: Good catch!

manast (approved):

manast:
Thank you for this gigantic effort, very much appreciated!

PR #2159, opened 2023-08-29
feat(worker.ts): adds new sequential job fetch mode

### LINKED ISSUE #2157: [Bug]: Excessive Redis polling when worker has high concurrency
### Version

4.8.0

### Platform

NodeJS

### What happened?

Workers with high `concurrency` can end up in state where they execute `getNextJob` X times for every new added job.
Where X is `concurrency` number.

This is causing excessive load on Redis. 

This can be remediated by lowering `concurrecy`, but there are use cases where we want to keep `concurrency` high to ensure few slow jobs don't block the queue.

### How to reproduce.

Have a worker with concurrency 100 and quick process function and push a job to worker's queue every 10ms.
BullMQ will end up executing >50k commands/seconds on Redis.

### Relevant log output

_No response_

### Code of Conduct

urossmolnik:
Workers with high concurrency can end up in state where they execute `getNextJob` X times for every new added job. Where X is concurrency number. This is causing excessive load on Redis.

This can be remediated by lowering concurrency, but there are use cases where we want to keep concurrency high to ensure few slow jobs don't block the queue.

This PR tries to solve this by adding extra setting `jobFetchMode`. If explicitly setting this setting to `sequential`, worker does not issue `concurrency` fetch jobs queries to Redis in parallel, and instead loads them sequentially. This does not mean that jobs are processed sequentially!

fix #2157

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -223,6 +223,7 @@ export class Worker<
       stalledInterval: 30000,
       autorun: true,
       runRetryDelay: 15000,
+      jobFetchMode: 'concurrent',
       ...this.opts,
     };
 
@@ -409,6 +410,9 @@ export class Worker<
               this.opts.runRetryDelay,
             ),
           );
+          if (this.opts.jobFetchMode === 'sequential') {
+            break;
+          }
         }
 
         const job = await asyncFifoQueue.fetch();
--- a/src/interfaces/worker-options.ts
+++ b/src/interfaces/worker-options.ts
@@ -137,6 +137,19 @@ export interface WorkerOptions extends QueueBaseOptions {
    * @default false
    */
   useWorkerThreads?: boolean;
+
+  /**
+   * Worker jobs fetch mode.
+   *
+   * In concurrent mode, jobs are fetched concurrently up to `concurrency` in
+   * parallel. This makes queue processing faster, but also puts more pressure
+   * on the redis server.
+   * In sequential mode, jobs are fetched sequentially. This does not mean that
+   * jobs are processed sequentially!
+   *
+   * @default "concurrent"
+   */
+  jobFetchMode?: 'concurrent' | 'sequential';
 }
 
 export interface GetNextJobOptions {
--- a/tests/test_worker.ts
+++ b/tests/test_worker.ts
@@ -1865,6 +1865,59 @@ describe('workers', function () {
     });
   });
 
+  describe('Job fetch mode', () => {
+    it('should default to concurrent', async () => {
+      const worker = new Worker(
+        queueName,
+        async () => {
+          // No-op
+        },
+        {
+          connection,
+        },
+      );
+      expect(worker.opts.jobFetchMode).to.equal('concurrent');
+      await worker.close();
+    });
+
+    it('should process job in parallel in sequential fetch mode', async () => {
+      const concurrency = 4;
+      let processing = 0;
+      let processingMax = 0;
+
+      const worker = new Worker(
+        queueName,
+        async () => {
+          processing++;
+          processingMax = Math.max(processing, processingMax);
+          await delay(Math.round(Math.random() * 10));
+          processingMax = Math.max(processing, processingMax);
+          processing--;
+        },
+        {
+          connection,
+          concurrency,
+          jobFetchMode: 'sequential',
+        },
+      );
+      expect(worker.opts.jobFetchMode).to.equal('sequential');
+      await worker.waitUntilReady();
+
+      const waiting = new Promise((resolve, reject) => {
+        worker.on('completed', after(8, resolve));
+        worker.on('failed', reject);
+      });
+
+      await Promise.all(times(8, () => queue.add('test', {})));
+
+      await waiting;
+      expect(processingMax).to.be.lessThanOrEqual(concurrency);
+      expect(processingMax).to.be.greaterThan(1);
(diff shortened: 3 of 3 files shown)


src/classes/resolvable-promise.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Resolvable promise is a Promise with exposed resolve and reject methods to
+ * explicitly resolve or reject the promise.
+ */
+export type ResolvablePromise<T = void> = Promise<T> & {
```
manast, line 5: This has traditionally been called "Deferred" in the NodeJS/Javascript ecosystem.

src/classes/async-fifo-queue.ts
```diff
@@ -1,3 +1,5 @@
+import { ResolvablePromise } from './resolvable-promise';
+
```
manast, line 2: I could not determine if the changes in this file change the functionality or if it just refactors the implicit "deferred" promise in the current implementation into a separate module 🤔 The new code is in any case longer so most likely we want to keep the shorter version of it.

manast:
I am a bit confused about this PR and the reported issue. The way it is supposed to currently work is that getNextJob is called X times per concurrency factor **as long as there are jobs waiting in the queue**, but not X times per job waiting, it either fetches one job or issues a BRPOPLPUSH command (one), and sets the worker status to waiting so no more calls are performed until there are more jobs in the queue. In fact, the whole idea of having the AsyncQueue module is to avoid calling BRPOPLPUSH more than once, as that would introduce other issues.

manast:
@urossmolnik I understand the problem. Maybe we could add a new setting where the concurrency is applied differently, like instead of trying to fetch all the jobs it can according to the concurrency, it will try to fetch one, start processing, try to get the next one, and so on, until there are no more jobs and then issue a BRPOPLPUSH. This will be slower in busy queues but faster in your case.

urossmolnik:
> Maybe we could add a new setting where the concurrency is applied differently, like instead of trying to fetch all the jobs it can according to the concurrency, it will try to fetch one, start processing, try to get the next one, and so on, until there are no more jobs and then issue a BRPOPLPUSH. This will be slower in busy queues but faster in your case.

@manast thanks for suggestion that would be a good option. Curious why not just doing `BRPOPLPUSH` on every job get in this mode?

Also (a little of topic), but should outer loop also wait for `waiting` if `drained` so it does not keep looping? 🤔

manast:
> @manast thanks for suggestion that would be a good option. Curious why not just doing `BRPOPLPUSH` on every job get in this mode?

Because since BRPOPLPUSH is a blocking command, subsequent calls to this command will just be queued in IORedis, so we loose control on when they will actually be executed, and some extra bookkeeping is necessary to clean the IORedis command queue when you want to close the queue, etc. Overall it is just simpler to issue one such command at a time.

manast:
> Also (a little of topic), but should outer loop also wait for `waiting` if `drained` so it does not keep looping?

As long as there is an issued BRPOPLPUSH it will not keep looping but waiting for a job to arrive, otherwise it will loop as there should be jobs in the queue.

urossmolnik:
> > Also (a little of topic), but should outer loop also wait for `waiting` if `drained` so it does not keep looping?
> 
> As long as there is an issued BRPOPLPUSH it will not keep looping but waiting for a job to arrive, otherwise it will loop as there should be jobs in the queue.

👍 It loops only through undefines pushed to queue.

urossmolnik:
> > @manast thanks for suggestion that would be a good option. Curious why not just doing `BRPOPLPUSH` on every job get in this mode?
> 
> Because since BRPOPLPUSH is a blocking command, subsequent calls to this command will just be queued in IORedis, so we loose control on when they will actually be executed, and some extra bookkeeping is necessary to clean the IORedis command queue when you want to close the queue, etc. Overall it is just simpler to issue one such command at a time.

Thanks for explanation. I overlooked that we only have 1 connection and not a separate one for fetching jobs.

urossmolnik:
@manast went with your suggested approach and added new setting. Changes are pretty minimal now.

urossmolnik:
@manast just checking in if this is something that can eventually land or should we be looking at other workarounds.

manast:
Sorry for the late response. We are reviewing this PR.

urossmolnik:
> Sorry for the late response. We are reviewing this PR.

Thanks. Thinking this again, should I change `jobFetchMode` to `fetchConcurrency` and make this a little more flexible? 🤔

manast:
The reason for the slow pace of handling this issue is that we are considering a different mechanism. Although your fix will solve this particular problem, we think the setting will be difficult to understand for most users. The best would be if BullMQ could handle concurrency in such a way that this setting is not needed and no unnecessary calls to Redis are performed.

urossmolnik:
> The reason for the slow pace of handling this issue is that we are considering a different mechanism. Although your fix will solve this particular problem, we think the setting will be difficult to understand for most users. The best would be if BullMQ could handle concurrency in such a way that this setting is not needed and no unnecessary calls to Redis are performed.

Hey @manast just checking in if you have any rough timeline on this? We can use our BullMQ fork temporary, but our end goal is to upgrade to BullMQ PRO and I assume we'll hit same problems there.

manast:
I am working on this issue... hopefully something ready very soon.

manast:
Superseded by this PR: https://github.com/taskforcesh/bullmq/pull/2242

PR #2174, opened 2023-09-07
feat(flow-producer): add addBulk method (python)

roggervalf:


manast (approved):
LGTM, and I assume that since "python" is in the message it will not trigger a NPM release.

roggervalf:
> LGTM, and I assume that since "python" is in the message it will not trigger a NPM release.

Yeah, it should ignore it

PR #2175, opened 2023-09-09
fix(remove): change error message when job is locked

roggervalf:


python/bullmq/job.py
```diff
@@ -83,7 +83,7 @@ async def remove(self, opts: dict = {}):
         removed = await self.scripts.remove(self.id, opts.get("removeChildren", True))
         
         if not removed:
-            raise Exception(f"Could not remove job {self.id}")
+            raise Exception(f"Job {self.id} is locked")
```
manast, line 86: Maybe a better message would be: "Job {self.id} could not be removed because it is locked by another worker"

manast (commented):
Just a small comment.

manast (approved):

manast (approved):

PR #2184, opened 2023-09-16
fix(worker): throw exception with NaN as concurrency

manast:
I got hinted that if you use NaN as a concurrency value the worker goes bananas. Here is a fix for it.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -568,6 +568,8 @@ export class Worker<
       if (!this.closing) {
         await this.delay();
       }
+    } finally {
+      this.waiting = null;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/classes/worker.ts
```diff
@@ -564,8 +568,6 @@ export class Worker<
       if (!this.closing) {
         await this.delay();
       }
-    } finally {
-      this.waiting = null;
```
roggervalf, line 568: What's behind this deleted line?

manast: that's a good question... must be the leftover of some testing I was doing.

roggervalf (approved):

PR #2187, opened 2023-09-20
Avoid importing from '..'

synchrone:
Import commands from their originating module, not from a re-export module.

This is important for compatibility with import-in-the-middle and other import-altering code, otherwise this results in a partially-loaded module being passed into instrumentation code.

### DIFF
--- a/src/classes/child.ts
+++ b/src/classes/child.ts
@@ -1,7 +1,7 @@
 import { ChildProcess, fork } from 'child_process';
 import { Worker } from 'worker_threads';
 import { AddressInfo, createServer } from 'net';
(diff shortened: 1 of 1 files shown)


src/classes/child.ts
```diff
@@ -1,7 +1,7 @@
 import { ChildProcess, fork } from 'child_process';
 import { Worker } from 'worker_threads';
 import { AddressInfo, createServer } from 'net';
-import { ChildCommand, ParentCommand } from '../';
+import { ChildCommand, ParentCommand } from '../commands';
```
roggervalf, line 4: In commands are our lua scripts

PR #2188, opened 2023-09-21
fix:Batched unpack now uses range

SemvandenBroek:
The `drain-4` command used in the `queue.drain()` function still used a non-batching variant of the `unpack` function in the `removeZSetJobs.lua` file. This resulted in the drain function failing when the unpack had too many values to unpack. 

This PR adds a range to the unpack function so that it has to unpack fewer values for each batched iteration.


manast:
Good catch! thanks for the fix.

PR #2149, opened 2023-08-22
feat(connection): provide skipVersionCheck option for shared connections

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2148

PR #2189, opened 2023-09-21
fix: forward skipVersionCheck to blockingConnection

Marsup:
Hi,

It seems that #2149 missed a spot when passing around this flag, as workers still emit this warning even with a true flag.


roggervalf (approved):
Lgtm, thank you!

PR #2190, opened 2023-09-22
Search

ccollie:
Adds job search functionality using `mongo` syntax.


src/classes/scripts.ts
```diff
@@ -405,6 +412,81 @@ export class Scripts {
...
+
+    const queue: MinimalQueue = this.queue;
+
+    function addJobIfNeeded() {
+      if (currentJob && !isEmpty(currentJob) && jobId) {
+        // TODO: verify this
+        const trace = currentJob['stacktrace'];
+        if (!Array.isArray(trace)) {
+          if (typeof trace === 'string') {
+            currentJob['stacktrace'] = JSON.parse(trace);
```
roggervalf, line 456: curious about this logic, https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L316 fromJSON method is handling stacktrace values

src/classes/scripts.ts
```diff
@@ -405,6 +412,81 @@ export class Scripts {
...
+          } else {
+            currentJob['stacktrace'] = [];
+          }
+        }
+        const raw = currentJob as JobJsonRaw;
+        const job = Job.fromJSON(queue, raw, jobId);
+        const ts = currentJob['timestamp'];
+        job.timestamp = ts ? parseInt(ts) : Date.now();
```
roggervalf, line 464: timestamp too https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L299

src/commands/getJobsByFilter-1.lua
```diff
@@ -0,0 +1,1729 @@
...
+    end
+  end
+  return false
+end
+
+local function isObject(val)
+  return type(val) == 'table'
+end
+
+local function isArray(t)
```
roggervalf, line 102: we can use includes to split this command as it's kind of big

roggervalf (commented):
we may need also a new section in our docs for this feature. I didn't review so deeper but it looks like a good beginning for this feature. I'll revisit it

src/commands/getJobsByFilter-1.lua
ccollie: Agreed. I also intend to split out state list iteration methods using `lua's` `pairs` prrotocol.

src/commands/getJobsByFilter-1.lua
```diff
@@ -0,0 +1,1729 @@
...
+      return val[1]
+    end)
+  end
+
+  local result = { newCursor }
+
+  for _, jobId in pairs(scannedJobIds) do
+
+    local jobIdKey = keyPrefix .. jobId
+    local jobHash = redis.pcall('HGETALL', jobIdKey)
```
manast, line 1699: Maybe we could just restrict the search to the "data" field.

manast: Even though we later need all the fields of the job, if we assume the search would just match a narrow set of jobs, it would be more efficient to read the remaining fields only for the jobs that actually match the search.

manast (commented):
Impressive job overall. I could not review in detail all the .lua code, I will try to spend more time analyzing it as the PR is more ready.
A couple of remarks:
- This script can potentially put a lot of pressure on Redis, as it will be CPU intensive. I think it should be wise to try to move as much logic as possible to the JS side. Maybe preprocess the query so that the job to do on lua side would be minimal. A crazy extreme of this would actually be to generate the lua code in JS that actually is used to check every job and then use something like this: https://www.lua.org/manual/5.1/manual.html#pdf-loadstring
- The method must allow to limit for amount of returned jobs per call, and also the amount of jobs to test for. As without any of these two constraints the call to this script could potentially hang the Redis server for too long. More than 100ms per call will probably be too slow in practice.

src/commands/getJobsByFilter-1.lua
ccollie: Good catch. The code actually collects the fields that are accessed in the query so that's possible with not too much problem.

OrionWambert:
Any update ?

ccollie:
@OrionWambert Ping me. I may make some time over the weekend to pick this back up.

bfbechlin:
Any news on this topic? @manast @roggervalf It would be extremely helpful to have a wildcard search on jobs based on id, data and logs or even only in the id to start with. Awesome project btw, congratz 👏

PR #2195, opened 2023-09-25
refactor: simplify code in moveToActive

manast:


src/commands/moveToActive-10.lua
```diff
@@ -43,73 +47,59 @@ local rcall = redis.call
...
+local jobId = nil
+if ARGV[3] ~= "" then
+    jobId = ARGV[3]
 end
 
--- If jobId is special ID 0:delay, then there is no job to process
 if jobId then
-  if string.sub(jobId, 1, 2) == "0:" then
-    rcall("LREM", KEYS[2], 1, jobId)
+    -- clean stalled key
```
roggervalf, line 65: this line and the next one could be moved after line 61 as we are making the same validation on that block:
```
local jobId = nil
if ARGV[3] ~= "" then
    jobId = ARGV[3]
    -- clean stalled key
    rcall("SREM", KEYS[5], jobId)
end
```

src/commands/moveToActive-10.lua
```diff
@@ -43,73 +47,59 @@ local rcall = redis.call
...
+        jobId = rcall("RPOPLPUSH", waitKey, activeKey)
     end
-  end
+end
 
-  if jobId then
-    return prepareJobForProcessing(KEYS, ARGV[1], target, jobId, ARGV[2], maxJobs, expireTime, opts)
-  else
-    jobId = moveJobFromPriorityToActive(KEYS[3], KEYS[2], KEYS[10])
+-- If jobId is special ID 0:delay, then there is no job to process
```
roggervalf, line 89: this comment could be moved before line 69:
`if not jobId or (jobId and string.sub(jobId, 1, 2) == "0:") then`

src/commands/moveToActive-10.lua
manast: True... fixing.

PR #2201, opened 2023-09-29
fix(python): introduction doc

iliasmentz:
`job_token` is a required parameter for the process function otherwise the function breaks. 

Another solution would be to use `*args` but I am not sure if it's worth being in the documentation


manast:
Thanks for the fix. I am mostly thinking as a JS developer so this kind of errors are a result of that.

roggervalf (approved):
LGTM

Issue #2152, opened 2023-08-24
Ability to reset attempts of a job


rileyai-dev:
**Is your feature request related to a problem? Please describe.**
We use a mechanism (moveToDelay) to restart a job with a delay right after it was successfully performed. We can't use cron jobs because the delay is picked randomly. We can't either create a new job because we want to be sure we keep the unicity of the jobID (make sure there are no duplicates). The problem we have is that each time we moveToDelay, we would like the attempts made and stalled counters to be reseted.

**Describe the solution you'd like**
Functions to manually reset the attempts made and the stalled counters of a job.

**Describe alternatives you've considered**
An alternative in our case would be to create two different queues. Once the job is complete (queue A), post to the other queue (queue B) with 2 seconds delay. Job in queue A would complete and job in queue B will be able to re-create a new job in queue A   making sure unicity of job is kept.
With this solution there might be other challenges to handle...

**Additional context**
None

roggervalf:
hey @rileyai-dev, I've been thinking about this for a while, let me see if I can work on one option for this

rileyai-dev:
Cool, that would be great! In the meantime, we have implemented two queues, with one posting to the other with a delay.

juhanakristian:
I would also be interested in this feature. Currently I'm pausing the queue when a job fails enough times and I would like to restart queue with all the attempt counts reset.

manast:
I hope v5 resolves this issue for you (https://bullmq.io/news/231221/bullmqv5-release/), let me know.

lovre-nc:
Hi @manast, adding this option would be very valuable to us.

You might remember from previous [conversations](https://github.com/taskforcesh/bullmq/issues/2942#issue-2699562493) that we have a queue where individual jobs exist and repeat forever (using moveToDelayed). 
We would like to be able to reset the attemptsMade count before moving the job to delayed. 

Is there any chance of getting this implemented?

_We also thought about migrating to the new job scheduler system, but we like the simplicity of normal jobs and the job scheduler doesn't seem to support groups._

Simon-Hayden-Dev:
Adding to this, when a job fails more than it's allowed `attempts`, we want to be able to retry those jobs manually. But because the `Job.retry()` method does not reset the `attemptsMade` counter, those jobs are retried only one time.

Luckily, we did not have to bulk-retry jobs, because some external API was down or similar. But eventually this case might come and we don't really have a nice recovery strategy for those cases then.

[...]

PR #2203, opened 2023-09-30
feat(job): provide decrementAttemptsMade option in moveToDelayed

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2152


roggervalf:
It's desired to keep the attemptMade up today, se may need to think more about it

src/classes/scripts.ts
```diff
@@ -683,6 +685,7 @@ export class Scripts {
       jobId,
       token,
       delay,
+      opts.decrementAttempt ? '1' : '0',
```
manast, line 688: I find this a bit awkward, in the sense that we must decrement to effectively achieve a non-increment of attemptsMade. I think it would be better to not increment instead of increment and then decrement.

manast: The problem is probably that we increment attemptsMade as soon as the job is moved to active, but maybe we should increment this property when the job is completed or failed.

manast: When a job is moved to active we have not "made an attempt" we have "started an attempt", but it is not until the job finishes we should consider the attempt to be made, what do you think?

roggervalf: we may not only consider when it completes or fails, also when it's moved to delayed, retried script and movedToWaitingChildren

manast: We should think that "attemptsMade" main use is for the error retries, and the backoff function. So actually increasing it when completing the job or failing would work well in this regard. On the other hand, it is also good to know how many times a job has been started, but may for some reason not finished, like it was considered stalled and moved back to wait, or it has been considered rate-limited, and so on.

src/classes/job.ts
```diff
@@ -1027,14 +1040,25 @@ export class Job<
...
-    return this.scripts.moveToDelayed(
+    const movedToDelayed = await this.scripts.moveToDelayed(
       this.id,
       timestamp,
       delay > 0 ? delay : 0,
       token,
+      opts,
     );
+
+    if (!opts.skipAttempt) {
```
manast, line 1057: What is the scenario where skipAttempt is necessary?

PR #2225, opened 2023-10-11
fix: max event limit when removing jobs

jasonmiraglia:
Impose the configured max length of events limit when adding events for job removal. This prevents unbounded event creation on this stream which can lead to unexpected runaway memory usage.

### DIFF
--- a/tests/test_events.ts
+++ b/tests/test_events.ts
@@ -657,7 +657,7 @@ describe('events', function () {
       });
     });
 
-    describe('when jobs are retried inmediately', function () {
+    describe('when jobs are retried immediately', function () {
       it('should trim events so its length is at least the threshold', async () => {
         const numJobs = 80;
         const trimmedQueue = new Queue(queueName, {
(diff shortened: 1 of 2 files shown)


manast (approved):
LGTM

tests/test_events.ts
```diff
@@ -712,6 +712,34 @@ describe('events', function () {
...
+          events: {
+            maxLen: 20,
+          },
+        },
+      });
+
+      const client = await trimmedQueue.client;
+
+      for (let i = 0; i < numRemovals; i++) {
+        await trimmedQueue.remove(i.toString());
```
roggervalf, line 730: hi @jasonmiraglia, I cannot see where jobs are added in this test case

manast: Good catch, it is removing un-existing jobs.

jasonmiraglia: In this particular issue's case, it doesn't matter if the job exists or not, an event will still be added to the event stream for every removal attempt.

manast: Ah, it seems like you found a bug then :)

roggervalf:
closing as this change was introduced in another pr

PR #2242, opened 2023-10-23
feat(worker): better handling of concurrency when fetching jobs

manast:
Handles this issue: #2157

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -1,3 +1,4 @@
+import { type } from 'os';
 import { ParentCommand } from '../enums';
 import { JobJson, SandboxedJob } from '../interfaces';
 import { errorToJSON } from '../utils';
@@ -71,10 +72,10 @@ export class ChildProcessor {
     this.currentJobPromise = (async () => {
       try {
         const job = this.wrapJob(jobJson, this.send);
-        const result = (await this.processor(job, token)) || {};
+        const result = await this.processor(job, token);
         await this.send({
           cmd: ParentCommand.Completed,
-          value: result,
+          value: typeof result === 'undefined' ? null : result,
         });
       } catch (err) {
         await this.send({
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -1,4 +1,3 @@
-import { type } from 'os';
 import { ParentCommand } from '../enums';
 import { JobJson, SandboxedJob } from '../interfaces';
 import { errorToJSON } from '../utils';
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -200,7 +200,6 @@ export class Worker<
       name,
       {
         ...opts,
-        sharedConnection: isRedisInstance(opts.connection),
         blockingConnection: true,
       },
       Connection,
@@ -824,6 +823,7 @@ export class Worker<
         .finally(() => client.disconnect())
         .finally(() => this.connection.close())
         .finally(() => this.emit('closed'));
+      this.closed = true;
     })();
     return this.closing;
   }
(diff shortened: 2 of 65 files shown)


src/classes/worker.ts
```diff
@@ -505,20 +549,29 @@ export class Worker<
...
     jobId?: string,
   ): Promise<Job<DataType, ResultType, NameType>> {
     // If we get the special delayed job ID, we pick the delay as the next
     // block timeout.
     if (jobId && jobId.startsWith('0:')) {
       this.blockUntil = parseInt(jobId.split(':')[1]) || 0;
+
+      // Remove marker from active list.
+      await client.lrem(this.keys.active, 1, jobId);
+      if (this.blockUntil > 0) {
```
roggervalf, line 563: but could it be the case that when doing this logic, one delayed job with a lower delay value was added, so if we break here, blockUntil will have a higher value than expected, on the other hand if we continue, we will call moveToActive and get the most updated value if this case happens

src/classes/worker.ts
```diff
@@ -529,30 +582,35 @@ export class Worker<
...
 
       if (!this.closing) {
-        const client = await this.blockingConnection.client;
-
         let blockTimeout = Math.max(
           this.blockUntil
             ? (this.blockUntil - Date.now()) / 1000
             : opts.drainDelay,
-          0.01,
+          0,
```
roggervalf, line 589: what's behind this change? :thinking:

src/commands/includes/addDelayMarkerIfNeeded.lua
```diff
@@ -6,10 +6,22 @@
...
 local function addDelayMarkerIfNeeded(targetKey, delayedKey)
-  if rcall("LLEN", targetKey) == 0 then
+  local waitLen = rcall("LLEN", targetKey)
+  if waitLen <= 1 then
     local nextTimestamp = getNextDelayedTimestamp(delayedKey)
     if nextTimestamp ~= nil then
-      rcall("LPUSH", targetKey, "0:" .. nextTimestamp)
+      -- Check if there is already a marker with older timestamp
+      -- if there is, we need to replace it.
+      if waitLen == 1 then
```
roggervalf, line 15: so we already remove markers when calling moveToActive, from my first comment, if we decide to call moveToActive after detecting a marker, this logic won't be necessary I think

src/classes/worker.ts
manast: Hmm, if a delayed job with lower delay was added, then when calling BRPOPLPUSH we will move the new marker right? and update the timeout accordingly. We must block sometimes, otherwise we will enter in an endless loop calling moveToActive.

src/classes/worker.ts
manast: Means not blocking at all. I do not remember but some edge case was not working with 0.01.

manast: But in any case, it is more correct, we want to block only if the value is > 0.

src/commands/includes/addDelayMarkerIfNeeded.lua
manast: My plan is to move the markers logic to a different key, so then I can revisit this case and simplify it.

manast: Using BZPOPMAX instead of BRPOPLPUSH

src/classes/child-processor.ts
```diff
@@ -1,3 +1,4 @@
+import { type } from 'os';
```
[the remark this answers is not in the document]
manast: seems to be an old code, not sure where it comes from but it is not there anymore.

PR #2259, opened 2023-11-02
fix(worker): keep extending locks while closing workers

### LINKED ISSUE #2258: [Bug]: Lock mismatch and Missing lock for job with graceful worker shutdown in Kubernetes environment
### Version

v4.12.4

### Platform

NodeJS

### What happened?

We are seeing some strange behavior with job locks and double job processing in our Kubernetes environments. We have some jobs, all of which are Sandboxed, that take upwards of 1 hour to complete. We have setup our workers to "gracefully shutdown" by calling `worker.close()`. We can observe that the worker continues to process the job after `worker.close()` has been called, but when it completes it will throw either a `Error: Missing lock for job <jobID>. failed` error or `Error: Lock mismatch for job <jobID>. Cmd failed from active` error.

The pod terminates and the new pod starts processing the job that was just processed.

My understanding is that since we are using Sandboxed jobs, the main event loop is not blocked, therefore updating job locks should not be an issue here.

### How to reproduce.

[Here](https://github.com/godinja/bullmq-graceful-shutdown) is a repository outlining how to reproduce.

### Relevant log output

_No response_

### Code of Conduct

manast:
Fixes #2258

### DIFF
--- a/src/classes/queue-base.ts
+++ b/src/classes/queue-base.ts
@@ -20,6 +20,7 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
   keys: KeysMap;
   closing: Promise<void> | undefined;
 
+  protected closed: boolean = false;
   protected scripts: Scripts;
   protected connection: RedisConnection;
   public readonly qualifiedName: string;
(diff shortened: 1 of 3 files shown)


roggervalf (approved):
LGTM

PR #2260, opened 2023-11-04
feat(queue): clean works iteratively

manast:


src/classes/queue.ts
```diff
@@ -423,14 +423,28 @@ export class Queue<
       | 'delayed'
       | 'failed' = 'completed',
   ): Promise<string[]> {
-    const jobs = await this.scripts.cleanJobsInSet(
-      type,
-      Date.now() - grace,
-      limit,
-    );
+    limit = limit || Infinity;
```
roggervalf, line 426: instead of mutating limit here, it could be assign as a default value in line 416 limit=Infinity

manast: The thing is that it is not really mutating it, as it becomes local inside the function, but I can make it more explicit.

PR #2261, opened 2023-11-04
fix: update delay job property when moving to delayed set

manast:


tests/test_worker.ts
```diff
@@ -1984,6 +1984,38 @@ describe('workers', function () {
...
+          } catch (err) {
+            reject(err);
+          }
+        });
+      });
+
+      await failed;
+
+      await worker.close();
+    });
```
roggervalf, line 2018: Just one styled comment, new line after this one

roggervalf (approved):

tests/test_worker.ts
manast: Fixed

PR #2265, opened 2023-11-07
fix(python backoff): fix bug when using builtin backoff type

HalfdanJ:
When a builtin backoff type is used, the lookup was expecting and object with a property type, but the passed value from the job is a dict. This fixes the lookup and adds a test for it.

Before the fix, an exception would be thrown: 

```
Error moving job to failed 'dict' object has no attribute 'type'
```

### DIFF
--- a/python/bullmq/backoffs.py
+++ b/python/bullmq/backoffs.py
@@ -31,7 +31,7 @@ class Backoffs:
 def lookup_strategy(backoff: BackoffOptions, custom_strategy):
     backoff_type = backoff.get("type")
     if backoff_type in Backoffs.builtin_strategies:
-        Backoffs.builtin_strategies[backoff_type](backoff.delay)
+        Backoffs.builtin_strategies[backoff_type](backoff.get("delay"))
     elif custom_strategy:
         return custom_strategy
     else:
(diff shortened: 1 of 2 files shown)


python/tests/worker_tests.py
```diff
@@ -209,6 +209,34 @@ async def process2(job: Job, token: str):
         await worker2.close(force=True)
         await queue.close()
 
+    async def test_retry_job_after_delay_with_fixed_backoff(self):
```
roggervalf, line 212: looks like this change is taking more time than expected in our tests

python/bullmq/backoffs.py
```diff
@@ -31,8 +31,8 @@ async def calculate(backoff: BackoffOptions, attempts_made: int, err, job, custo
 def lookup_strategy(backoff: BackoffOptions, custom_strategy):
     backoff_type = backoff.get("type")
     if backoff_type in Backoffs.builtin_strategies:
-        Backoffs.builtin_strategies[backoff.type](backoff.delay)
```
roggervalf, line 34: I think it should be changed to:
`return Backoffs.builtin_strategies[backoff_type](backoff.get("delay"))`

python/tests/worker_tests.py
HalfdanJ: I will be honest, i wrote the test in blind. Wasn't able to succesfully run the test suite for python in the repo

python/bullmq/backoffs.py
HalfdanJ: Ah yes

HalfdanJ:
I managed to get my local tests working, discovered another bug in the lookup code that i've fixed. The tests are now passing locally, so 🤞

roggervalf (approved):
Lgtm

PR #2275, opened 2023-11-11
docs: delete page for now removed backoffStrategies Queue settings

lbennett-stacki:
I realised these settings are no longer available after poking around for a while. I wasn't sure if something was wrong with the types, my code or the docs. A quick source search revealed this setting is no longer available.


manast:
Thanks for the PR, however here you are removing documentation that refers to Bull, not to BullMQ.

lbennett-stacki:
Sorry @manast 🤦

PR #2276, opened 2023-11-11
fix: narrow the BackoffOptions "type" union

lbennett-stacki:
Seems the ability to define custom-named backoff strategies has been dropped, so we can narrow the `type` type.

Related:
https://github.com/taskforcesh/bullmq/pull/2275

### DIFF
--- a/src/interfaces/backoff-options.ts
+++ b/src/interfaces/backoff-options.ts
@@ -7,7 +7,7 @@ export interface BackoffOptions {
   /**
    * Name of the backoff strategy.
    */
-  type: 'fixed' | 'exponential' | (string & {});
+  type: 'fixed' | 'exponential' | 'custom';
   /**
    * Delay in milliseconds.
    */


roggervalf:
hi @lbennett-stacki, we can define custom backoff strategies

roggervalf:
pls take a look https://docs.bullmq.io/guide/retrying-failing-jobs

lbennett-stacki:
@roggervalf Thanks, I was getting confused between bull and bullmq custom backoff strategy definitions. I think this is still a valid change that may direct others away from some similar confusion.

In bull you can define a custom backoff strategy in one place and reference it by its unique key in the backoff settings object. However in bullmq you cannot define named/keyed backoff strategies as you did in bull, the value of `type` in the backoff settings object is only defined by how it is neither `fixed` nor `exponential`, the string value is never used to lookup a strategy? An explicit `"custom"` value indicates this better than _any_ string IMO.

Let me know if I've misread something!

roggervalf:
hey @lbennett-stacki, but custom won't be always a valid value, you are restricting to use fixed, exponential and custom specific values, instead of any other names as custom and ts will detect it as an error if someone define a different custom name apart of 'custom'

roggervalf:
as my comment above, this change will cause an error in ts

PR #2281, opened 2023-11-15
docs(pattern): add timeout section

roggervalf:


manast (commented):
I don't see a lot of value in this pattern, I am worried it will lead users in the wrong direction, as you hint at the end, it is better to divide the job in smaller parts. Basically cancelling an async operation is going to result in a side effect that in the general case can result in any undesired effect.

src/utils.ts
```diff
@@ -58,12 +58,11 @@ export function delay(
   return new Promise(resolve => {
     let timeout: ReturnType<typeof setTimeout> | undefined;
     const callback = () => {
-      abortController?.signal.removeEventListener('abort', callback);
       clearTimeout(timeout);
       resolve();
     };
     timeout = setTimeout(callback, ms);
-    abortController?.signal.addEventListener('abort', callback);
+    abortController?.signal.addEventListener('abort', callback, { once: true });
```
roggervalf, line 65: https://github.com/southpolesteve/node-abort-controller/blob/master/index.d.ts#L17 autoremove abort listener

PR #2282, opened 2023-11-15
fix(python): uniyfing redis connection args for Queue and Worker

saulane:
Currently the NodeJS version has the same connection arguments for the worker and queue whereas python as two different format.

It's a breaking change but using the same format might prevent bugs in the future


manast (approved):
LGTM

roggervalf (approved):

PR #2293, opened 2023-11-23
refactor(addJob): divide addJob into specific scripts

manast:


python/bullmq/scripts.py
```diff
@@ -97,25 +98,82 @@ def addJobArgs(self, job: Job, waiting_children_key: str|None):
...
-
-        return (keys,args)
+        return [packedArgs, jsonData, packedOpts]
 
     def addJob(self, job: Job, pipe = None):
         """
         Add an item to the queue
         """
-        keys, args = self.addJobArgs(job, None)
+        # if (parentOpts.waitChildrenKey) {
```
roggervalf, line 107: Looks like this code is almost the same as below

python/bullmq/utils.py
```diff
@@ -2,7 +2,7 @@
 import traceback
 
 def isRedisVersionLowerThan(current_version, minimum_version):
-    return semver.compare(current_version, minimum_version) == -1
+    return semver.VersionInfo.parse(current_version).compare(minimum_version) == -1
```
roggervalf, line 5: nice, no more warnings

roggervalf (approved):
Lgtm

PR #2295, opened 2023-11-23
fix(worker): do not wait for slow jobs fixes #2290

### LINKED ISSUE #2290: [Bug]: Job fetching stopped
### Version

v4.14.0

### Platform

NodeJS

### What happened?

We often end up in scenario (mostly in development) when job fetching is stopped and new scheduled jobs are not processed.
This problem is really visible when you have a long running job(s). In our case, we have one job that is constantly running - for max 30 minutes and then re-schedules itself.

We can optimize our setup and not run a job for that long, but job fetching should still not be blocked by single longer running job.

### How to reproduce.

- Schedule a long running job
- `AsyncFifoQueue` ends up in state where `this.pending.size > 0` and `this.queue.length === 0`
- Worker's call to `job = await asyncFifoQueue.fetch();` ends up waiting forever (until only pending job completes)

### Relevant log output

_No response_

### Code of Conduct

manast:


urossmolnik:
@manast with this change, Redis usage by BullMQ is much higher again. Example of bump that happened to us after upgrading from 4.14.0 to 4.14.1
CleanShot 2023-11-24 at 07 02 19@2x

Any reason why now that jobs are fetched sequentially, blocking job fetch (`waitForJob`) can't be used more extensively?

manast:
This fix is to avoid this issue: https://github.com/taskforcesh/bullmq/issues/2290
It has proven extremely difficult to get the main loop in BullMQ to fulfil all requirements 😅
Is it rpoplpush the command that has resulted in a large spike? was what the scenario where this happens, a lot of works waiting to be processed or a sparsed queue with burst of jobs?

urossmolnik:
> Is it rpoplpush the command that has resulted in a large spike?

Yes `rpoplpush` has the biggest spike. 

> what was the scenario where this happens, a lot of works waiting to be processed or a sparsed queue with burst of jobs?

We have quite constant steady works to be processed. But we also have quite many instances up. Not exactly sure why this change causes the spike thou as queue should be mark as drained and blocking waiting operation should kick in 🤔

manast:
I am trying to figure out why this happens. Clearly it is too eager trying to fetch the next job when there are no jobs to fetch.

manast:
@urossmolnik I have created 2 test cases for the scenarios I think matches your setup. The first case shows that "moveToActive" (the method that produces the calls to rpoplpush), is called a maximum of 10 times, which is the concurrency factor. This is because once a job has been fetched, that same job when completed will fetch the next, so only 10 calls to moveToActive are actually needed. The second case shows that if the number of jobs is smaller than the concurrency factor, at most, the number of jobs (+ 1 to detect there are no more jobs) calls to moveToActive are performed.
So as far as I can see the behaviour is correct and no unnecessary calls to rpoplpush are performed. Feel free to experiment with this cases to see if you can show a case where this is not true so can we look deeper into it.

```ts
  it('do not call moveToActive more than concurrency factor', async () => {
    const numJobs = 50;
    const concurrency = 10;
    let completedJobs = 0;
    const worker = new Worker(
      queueName,
      async job => {
        expect(job.data.foo).to.be.equal('bar');
        await delay(250);
      },
      { connection, prefix, concurrency },
    );

    // Add spy to worker.moveToActive
    const spy = sinon.spy(worker, 'moveToActive');

    for (let i = 0; i < numJobs; i++) {
      const job = await queue.add('test', { foo: 'bar' });
      expect(job.id).to.be.ok;
    }

    await new Promise<void>((resolve, reject) => {
      worker.on('completed', (job: Job, result: any) => {
        completedJobs++;
        if (completedJobs == numJobs) {
          resolve();
        }
      });
    });

    // Check moveToActive was called only concurrency times
    expect(spy.callCount).to.be.equal(concurrency);

    await worker.close();
  });

  it('do not call moveToActive more than number of jobs + 1', async () => {
    const numJobs = 50;
    let completedJobs = 0;
    const worker = new Worker(
      queueName,
      async job => {
        expect(job.data.foo).to.be.equal('bar');
        await delay(250);
      },
      { connection, prefix, concurrency: 100 },
    );

    // Add spy to worker.moveToActive
    const spy = sinon.spy(worker, 'moveToActive');

    for (let i = 0; i < numJobs; i++) {
      const job = await queue.add('test', { foo: 'bar' });
      expect(job.id).to.be.ok;
    }

    await new Promise<void>((resolve, reject) => {
      worker.on('completed', (job: Job, result: any) => {
        completedJobs++;
        console.log(completedJobs)
        if (completedJobs == numJobs) {
          resolve();
        }
      });
    });

    // Check moveToActive was called numJobs + 1 times
    expect(spy.callCount).to.be.equal(numJobs + 1);

    await worker.close();
  })
```

manast:
Btw, it could be the case that in your charts above what you see is actually normal, and that the low usage before was actually not normal, indicating that you were not maximizing the number of jobs that could be processed by the workers. What we need to understand is if unnecessary calls to Redis are performed or not. So far it looks ok from BullMQ side.

urossmolnik:
@manast thanks for this. Let us dig further.

> Btw, it could be the case that in your charts above what you see is actually normal, and that the low usage before was actually not normal, indicating that you were not maximizing the number of jobs that could be processed by the workers. What we need to understand is if unnecessary calls to Redis are performed or not. So far it looks ok from BullMQ side.

This could be the case. Queues lengths were not going up with `4.14.0` though.

We are also comparing load on Redis of `bullmq` vs `bull`. We do have about 2x CPU usage on Redis with same setup once we migrated to `bullmq`.

manast:
CPU load in isolation is not a metric that tells us if there is some performance issue, it needs to be compared with the amount of jobs processed per second. There is no reason why bull would be less performant than BullMQ, unless there is a issue that is creating an unnecessary amount of calls to Redis.

Issue #2288, opened 2023-11-20
ESM build contains __dirname references


AndriyAntonenko:
I am trying to use bullmq worker in my application. My application have limitation, that build artifact should be only ESM module.  When I am running my application it failed cause Worker class using __dirname global variable. Is it possible to make fix and use `import.meta.url` or smth like that?

PR #2296, opened 2023-11-24
fix(worker): get dirname by using module.filename

roggervalf:
ref #2288


manast:
Which issue is this PR addressing?

manast (approved):

Issue #2286, opened 2023-11-20
[Bug]: `getJobs` return `undefined` jobs


mkykadir:
### Version

v4.7.0

### Platform

NodeJS

### What happened?

Getting jobs within BullMQ queue sometimes returns "undefined" jobs within list, job count is correct but job itself is "undefined". Querying same queues after 100ms, job can be identified and list is not returning any "undefined" jobs.

I am querying my jobs with this

```TypeScript
      const [timeIntervalJobs, pubsubAwaitingJobs] = await Promise.all([
        bullmq
          .getQueue(BullMQWorker.Queues.TimeInterval(chainId))
          .getJobs(["active", "delayed", "prioritized", "waiting"]),
        bullmq
          .getQueue(BullMQWorker.Queues.PubsubTimeout(chainId))
          .getJobs(["active", "delayed", "prioritized", "waiting"]),
      ]);
```

I am mapping "job.name" to a task identifier I am managing, thus I need all jobs to be defined to identify where are tasks at the moment. I should be either find task identifier related job within TimeInterval or PubsubTimeout queue but since `getJobs` return "undefined" jobs sometimes, cannot find related job. 

### How to reproduce.

There is no deterministic way to reproduce the issue

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
We will need some test code that reproduces the issue, even though the behaviour would not happen all the time, we need to be able to reproduce it somehow to have any chance of fixing the issue.

mkykadir:
@manast will try to provide a deterministic test code for this

mkykadir:
@manast created [this repo](https://github.com/mkykadir/bullmq-undefined-case) you can clone and run it, you will see 

```Task name undefined for``` 

logs over time. With this code I am expecting task to be defined in one of the queues, when task is added to TaskUpdates queue its still in TimeInterval queue so there should not be any time point that task is not inside either of the queues. Actually `getJobs` returns tasks as well but they're `undefined` unfortunately.

manast:
Thanks for the repo. Actually this behaviour could be resolved by this PR: https://github.com/taskforcesh/bullmq/pull/2309
After a positive review we will merge it.

mkykadir:
> Thanks for the repo. Actually this behaviour could be resolved by this PR: #2309 After a positive review we will merge it.

Tried with this PR as well but on my "test" repo it still occurs time to time

pozhega:
[...]

Issue #2268, opened 2023-11-08
[Bug]: The active job count is sometimes wrong


haqadn:
### Version

4.11.4

### Platform

NodeJS

### What happened?

The active job count is inflated.

### How to reproduce.

It's not reproducable 100% of the time.

However, it happens when there is 20+ concurrent workers consuming the queue and a prioritized job fails. The `bull:<queue-name>:active` list gets appended with a `0:0` value. The active count would include items with this value.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
Actually, this is by design. For a very small amount of time it could happen that a "marker" is moved to the active set, and then quickly removed. We need this marker in order to signal workers that there are  priority jobs available. It should not be off more than 1 AFAIK.

haqadn:
It does go beyond 1. This screenshot is from a live database.
<img width="381" alt="image" src="https://github.com/taskforcesh/bullmq/assets/3737780/3fe875e3-c393-4426-b099-1315364fed82">

It poses a problem in my use-case since we scale up and down the number of workers based on the `active+waiting+prioritized` job count.

manast:
Yeah, we cannot use sets unfortunatelly, since the only command that blocks and moves an item from a list to another is BRPOPLPUSH. But I think it is strange that you get so many markers, that does not look right. By any chance could you create a test where this happens?

haqadn:
Unfortunately, I don't know a sure way of reproducing it. I was trying to find a pattern of it happening but couldn't find any. But, how the queuing is handled in my case is the follows:

1. There is an HTTP server that receives requests and creates jobs.
2. The jobs are pushed to a redis cluster with 1 read/write and a few read-only replicas.
3. Some workers that can horizontally scale. We try to keep the number of workers same as `active+waiting/prioritized` so that the jobs get processed in minimum wait time.
4. If a job fails, it is retried a few times.

Just trying to give you an idea of the architecture in case you can get any clues.



Btw, where does it make sure only one marker gets inserted at a time?

roggervalf:
hi @haqadn, that logic is [here](https://github.com/taskforcesh/bullmq/blob/master/src/commands/includes/addPriorityMarkerIfNeeded.lua#L9-L11) from https://github.com/taskforcesh/bullmq/blob/master/src/commands/includes/addJobWithPriority.lua

manast:
[...]

PR #2309, opened 2023-12-02
Feat/better handling of queue markers

manast:
This PR introduces a new mechanism for handling queue markers which results in a simpler implementation and addresses issues like thes ones: https://github.com/taskforcesh/bullmq/issues/2268 #2286 
However as it introduces a new key it is also a breaking change and all workers should be upgraded to this new version at once to avoid any issues.

https://bullmq.io/news/231204/better-queue-markers/#migrating-from-v4-to-v5


roggervalf (approved):
Lgtm

PR #2313, opened 2023-12-05
chore: address todos

roggervalf:


src/commands/addDelayedJob-6.lua
```diff
@@ -112,8 +112,6 @@ if not isPaused then
 end
 
 -- Check if this job is a child of another job, if so add it to the parents dependencies
--- TODO: Should not be possible to add a child job to a parent that is not in the "waiting-children" status.
```
roggervalf, line 115: we have a pattern where parent can be in active state when creating children dynamically https://docs.bullmq.io/patterns/process-step-jobs#waiting-children so this TODO is not needed

src/commands/moveToFinished-14.lua
```diff
@@ -29,10 +29,9 @@
       ARGV[3]  msg property returnvalue / failedReason
       ARGV[4]  return value / failed reason
       ARGV[5]  target (completed/failed)
-      ARGV[6]  event data (? maybe just send jobid).
```
roggervalf, line 32: unused argv

manast (approved):
Great cleanups

PR #2324, opened 2023-12-12
fix(flows): update constructor and methods to match queue base

lohart13:
I noticed that the flow producer wasn't closing correctly which was leading to my application crashing when passing the flow Redis connection options.

After making the following changes, the flow can now close nicely.

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -95,7 +95,6 @@ export class FlowProducer extends EventEmitter {
   closing: Promise<void> | undefined;
   queueKeys: QueueKeys;
 
-  protected closed: boolean = false;
   protected connection: RedisConnection;
 
   constructor(
@@ -172,13 +171,6 @@ export class FlowProducer extends EventEmitter {
    */
   get client(): Promise<RedisClient> {
     return this.connection.client;
-  }
-
-  /**
-   * Returns the version of the Redis instance the client is connected to,
-   */
-  get redisVersion(): string {
-    return this.connection.redisVersion;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/classes/flow-producer.ts
```diff
@@ -453,13 +478,22 @@ export class FlowProducer extends EventEmitter {
...
+   *
+   * Closes the connection and returns a promise that resolves when the connection is closed.
+   */
+  async close(): Promise<void> {
     if (!this.closing) {
       this.closing = this.connection.close();
     }
-    return this.closing;
+    await this.closing;
+    this.closed = true;
```
roggervalf, line 490: closed value is only used in Worker class, so here shouldn't be needed :thinking:

src/classes/flow-producer.ts
```diff
@@ -156,6 +167,31 @@ export class FlowProducer extends EventEmitter {
...
+   * Returns a promise that resolves to a redis client. Normally used only by subclasses.
+   */
+  get client(): Promise<RedisClient> {
+    return this.connection.client;
+  }
+
+  /**
+   * Returns the version of the Redis instance the client is connected to,
+   */
+  get redisVersion(): string {
```
roggervalf, line 180: also this method is not used in flow-producer class

src/classes/flow-producer.ts
lohart13: I've removed this now.

src/classes/flow-producer.ts
lohart13: I've removed this now.

roggervalf (approved):
Lgtm

PR #2335, opened 2023-12-15
fix(connection): require connection to be passed

roggervalf:


manast (approved):

void-nick:
The connection requirement is not optimized for NestJS integration (should I need to copy the connection for every worker in my app if it's already set in BullModule?)

leoperria:
> The connection requirement is not optimized for NestJS integration (should I need to copy the connection for every worker in my app if it's already set in BullModule?)

Same here

vsamofal:
same for me :)

manast:
If you do not pass a connection it will use the local connection, so please can you explain how this requirement affects NestJS?

manast:
(the reason for requiring the connection is to avoid bugs where the connection was indadvertedly left undefined so it used the local connection instead in production where there weren't any)

vsamofal:
NestJS do use types from bullmq, I already made a fix in nestjs https://github.com/nestjs/bull/pull/1960, waiting for PR approval

manast:
> the problem with nestjs, is that connection type become required and not optional, even tho it works, it breaks code for people and there is no easy way in nestjs to pass a connection because in nestjs these types are used in decorators

Yes, that is the reason this was introduced in a major release after more than 6 months of warnings that the option was deprecated.

void-nick:
Let me explain the problem a little better: in NestJS, I set up the connection with the `forRoot/forRootAsync` methods of BullModule, usually in the app top level module. But if I want to use BullMQ version 5 or higher, whenever I need to register a queue, the `RegisterQueueOptions` interface extends `QueueOptions->QueueBaseOptions` from BullMQ which will require the connection property again. Since I have already setup in the forRoot method, it doesn't make sense to keep repeating it in every queue registration.
For now, I only install @nestjs/bullmq, which will require BullMQ v4.
It may be just lacking an optional interface for NestJS integration.

PR #2342, opened 2023-12-19
perf(worker): fetch next job on failure

roggervalf:


manast (commented):
This is a small but a breaking change 😅

src/classes/worker.ts
```diff
@@ -715,8 +715,14 @@ export class Worker<
             return;
           }
 
-          await job.moveToFailed(err, token);
+          const failed = await job.moveToFailed(err, token, true);
           this.emit('failed', job, err, 'active');
+
+          if (failed) {
+            const [jobData, jobId, limitUntil, delayUntil] = failed || [];
```
manast, line 722: || [] is redundant as it is already checking if failed is truthy

roggervalf: I refactored to be similar to completed handler

src/classes/worker.ts
```diff
@@ -766,8 +766,14 @@ will never work with more accuracy than 1ms. */
             return;
           }
 
-          await job.moveToFailed(err, token);
+          const failed = await job.moveToFailed(err, token, true);
```
manast, line 769: I think "failed" here is misleading, as this includes the next job to process. Should be called probably "result" or similar.

src/classes/worker.ts
```diff
@@ -786,7 +792,7 @@ will never work with more accuracy than 1ms. */
       const result = await this.callProcessJob(job, token);
       return await handleCompleted(result);
     } catch (err) {
-      return handleFailed(<Error>err);
+      return await handleFailed(<Error>err);
```
manast, line 795: await here will imply an unhandled exception if handleFailed throws an exception, is this the intent?

roggervalf: it was placed when testing, removing it now

src/classes/worker.ts
roggervalf: sounds good to me

manast (approved):
LGTM.  I see this PR more from a consistency side, as failed jobs are a deviation of the normal functioning of the queue and therefore it is not so important to optimize for this path, however having the same code for both completed and failed jobs is beneficial in my opinion.

PR #2343, opened 2023-12-20
refactor(retries): remove skipAttempt when step jobs pattern

roggervalf:


manast (approved):
I like this, simpler and more consistent.

PR #2350, opened 2023-12-22
fix(worker): fix so that worker can be closed if Redis is down

manast:
Fixes https://github.com/taskforcesh/bullmq/discussions/372

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -41,6 +41,8 @@ export class RedisConnection extends EventEmitter {
     canDoubleTimeout: false,
   };
 
+  status: 'initializing' | 'ready' | 'closing' | 'closed' = 'initializing';
+
   protected _client: RedisClient;
 
   private readonly opts: RedisOptions;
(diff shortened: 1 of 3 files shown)


roggervalf (approved):
Lgtm

SrdjanCosicPrica:
Hi @manast @roggervalf 

My project was on 5.1.0 before this, and after upgrading to 5.1.1, my jest tests started reporting
```
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Adding `--detectOpenHandles` does not show anything so I believe I'm closing everything as expected. I'm not sure where the issue is since it worked fine before this patch

manast:
@SrdjanCosicPrica it would be great if you could isolate a test case where this happens so that we can reproduce it and fix it.

SrdjanCosicPrica:
Ok I'll do my best to find some time to replicate it.
I solved my issue by adding a sleep of 10 ms `Promise(resolve => setTimeout(resolve, 10))`.

I'm using docker-compose during tests and my best guess is that the JS thread doesn't get enough time to establish a connection and mark it as `ready` before the test wants to complete. Adding the sleep gave enough time for the status to change.

SrdjanCosicPrica:
Hey @manast I managed to reproduce it with minimal code that is close to how we are using it in the project where I discovered the issue.
https://github.com/SrdjanCosicPrica/bullmq-5.1.1-open-handles

PR #2357, opened 2023-12-29
refactor(commnads): reuse getOrSetMaxEvents when missing

roggervalf:


tests/test_flow.ts
```diff
@@ -2469,16 +2469,10 @@ describe('flows', () => {
 
   it('should add meta key to both parents and children', async () => {
```
roggervalf, line 2470: this test was modified in order to prevent a failure in pro version as meta object also includes bullmq-pro value

PR #2365, opened 2024-01-06
fix(backoff): change optional BackoffStrategy params to use union

roggervalf:


src/classes/scripts.ts
```diff
@@ -73,12 +73,12 @@ export class Scripts {
     return Number.isInteger(result);
   }
 
-  private async addDelayedJob(
+  protected addDelayedJob(
```
roggervalf, line 76: protected for extension in pro version

manast (commented):
This seems to imply a breaking change right?

PR #2374, opened 2024-01-11
fix(retry-jobs): add marker when needed

roggervalf:


tests/test_events.ts
```diff
@@ -264,7 +264,7 @@ describe('events', function () {
       { name: 'test', data: { foo: 'baz' } },
     ]);
 
-    await delay(1000);
+    await delay(2000);
```
roggervalf, line 267: taking the opportunity to check if there are no more drained events in this test case

tests/test_queue.ts
```diff
@@ -468,7 +468,7 @@ describe('queues', function () {
     });
 
     describe('when completed state is provided', () => {
-      it('retries all completed jobs', async () => {
+      it('retries all completed jobs', async function () {
```
roggervalf, line 471: this is a flaky test, but with the marker addition, it should be fixed

manast (approved):
LGTM

PR #2384, opened 2024-01-17
fix(stalled): consider adding marker when moving job back to wait

roggervalf:


tests/test_stalled_jobs.ts
```diff
@@ -32,7 +32,7 @@ describe('stalled jobs', function () {
   });
 
   it('process stalled jobs when starting a queue', async function () {
-    this.timeout(10000);
+    this.timeout(5000);
```
roggervalf, line 35: these timeout changes are done because these tests will failed if markers are not added when stalled jobs are moved to wait

src/commands/moveToFinished-14.lua
```diff
@@ -57,6 +57,7 @@ local rcall = redis.call
 --- Includes
 --- @include "includes/collectMetrics"
 --- @include "includes/getNextDelayedTimestamp"
+--- @include "includes/getTargetQueueList"
```
roggervalf, line 60: this reference was missing

manast (approved):
LGTM

PR #2389, opened 2024-01-20
fix(marker): differentiate standard and delayed markers

roggervalf:
a marker from a delayed job could override an existing marker from a standard job as both have the same member value, that cause that a worker picks the marker thinking that all the remaining jobs are in delayed status, so it wait the delay even having standard jobs. In order to avoid that we can just use a different member value for the delayed jobs and bzpopmin will pick the lower value (standard marker "0" with score 0) later it could pick marker "1" with delayed score


manast:
I guess there is more code coming with changes on how to react differently depending on the marker type right?

roggervalf:
Yeah I think so, I added some todos to double check some commands as well

roggervalf:
But this pr is only related to this case, it should be ready

PR #2403, opened 2024-01-30
fix(rate-limit): move job to wait even if ttl is 0

roggervalf:
When returning a rate limit error we must move the job from active to wait, independently of the ttl value, in case we don't move , the job will get stalled or could fail if we have maxStalledCount as 0

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -222,9 +222,7 @@ export class Scripts {
     return <string>result;
   }
 
-  async pause(pause: boolean): Promise<void> {
-    const client = await this.queue.client;
-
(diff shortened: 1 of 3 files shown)


src/classes/scripts.ts
```diff
@@ -222,9 +222,7 @@ export class Scripts {
     return <string>result;
   }
 
-  async pause(pause: boolean): Promise<void> {
-    const client = await this.queue.client;
-
+  protected pauseArgs(pause: boolean): (string | number)[] {
```
roggervalf, line 225: this change is just to separate args from actual command call

PR #2417, opened 2024-02-09
fix(flow): parent job cannot be replaced

roggervalf:
while adding a duplicated job, we can replace the parent job if this job had one, while replacing a parent value, we will end with an orphan parent, we must throw an error in this case

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -446,7 +446,7 @@ export class Scripts {
         );
       case ErrorCode.ParentJobCannotBeReplaced:
         return new Error(
-          `Parent job from job ${jobId} cannot be replaced. ${command}`,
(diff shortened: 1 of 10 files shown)


src/classes/scripts.ts
```diff
@@ -444,6 +444,10 @@ export class Scripts {
         return new Error(
           `Lock mismatch for job ${jobId}. Cmd ${command} from ${state}`,
         );
+      case ErrorCode.ParentJobCannotBeReplaced:
+        return new Error(
+          `Parent job from job ${jobId} cannot be replaced. ${command}`,
```
manast, line 449: The grammar does not seem to be correct here, this should be better: "The parent job of job ${jobId} cannot be replaced."

manast (commented):
LGTM, just a small comment.

PR #2420, opened 2024-02-13
docs(guide): fix queueEvents.on failed description

kerolloz:


docs/gitbook/guide/workers/README.md
```diff
@@ -88,7 +88,7 @@ queueEvents.on('completed', ({ jobId: string, returnvalue: any }) => {
 });
 
 queueEvents.on('failed', ({ jobId: string, failedReason: string }) => {
-  // jobId received a progress event
+  // Called every time a job is failed in any worker.
```
roggervalf, line 91: what about:
`when a jobs is moved to failed`
as if there is a backoff strategy, `delayed` event will be emitted, only when all retries are consumed, a `failed` event is emitted

kerolloz: yeah sure

roggervalf (approved):
Lgtm

Issue #2385, opened 2024-01-17
[Bug]: Redis connection not correctly closed (regression)


edvinv:
### Version

5.1.2

### Platform

NodeJS

### What happened?

I have updated bullmq from 4.17.0 to 5.1.2 and noticed that connection was not closed correctly when running some tests. When I pin down the problem I found that version 5 is not closing connection if there is no delay between creating and closing.

### How to reproduce.

```typescript
  const queue = new Queue('CALLS_JOB_QUEUE_NAME', {
    connection: {
      host: 'localhost',
      port: 6379,
    },
  });

  await queue.close();
  await promiseDelay(1000);
  console.log("done");
```

If I run this sample with 5.1.2 I see at console.log("done"), that connection in redis is still opened. Running the same sample with 4.17.0, connection is closed as expected. If I add additional delay before close like this:

```typescript
  const queue = new Queue('CALLS_JOB_QUEUE_NAME', {
    connection: {
      host: 'localhost',
      port: 6379,
    },
  });

  await promiseDelay(1000); // ADDED!!!
  await queue.close();
  await promiseDelay(1000);
  console.log("done");
```
Than with both version connection is correctly closed.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

arunp0:
You are trying to close the connection even before the connection is established.
https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts#L292

When you add delay, it gives enough time for the Redis connection to establish.

Not a bug.

edvinv:
No, IMHO this is a bug. Connection can be correctly closed even in the case when it was not yet fully established. And it was working like this before [commit](https://github.com/taskforcesh/bullmq/commit/888dcc2dd40571e05fe1f4a5c81161ed062f4542).

So current code:
```typescript
        if (status === 'ready') {
          // Not sure if we need to wait for this
          await this.initializing;
          if (!this.shared) {
            await this._client.quit();
          }
        }
```

should be changed to sommething like:
```typescript
        if (status === 'initializing' || status === 'ready') {
          // Not sure if we need to wait for this
          await this.initializing;
          if (!this.shared) {
            await this._client.quit();
          }
        }
```

[...]

PR #2425, opened 2024-02-14
fix(redis-connection): close redis connection even when initializing

manast:
Fixes https://github.com/taskforcesh/bullmq/issues/2385


roggervalf (approved):
Lgtm

Issue #2092, opened 2023-07-24
Get failedChildren when using removeDependencyOnFailure 


M3talen:
**Is your feature request related to a problem? Please describe.**
After the v4.4.0 update including the removeDependencyOnFailure feature (#1953) we can start a flow and upon completion we can get the parent job from the flowProducer that consists of a job object and a array of children. When using the removeDependencyOnFailure when a job fails, that job is not anymore in the children array. 
![image](https://github.com/taskforcesh/bullmq/assets/1522781/764ea01e-a897-425e-95c3-4af9c4abf5e4)

**Describe the solution you'd like**
Would it be possible to include a new array of failed children ? Looking through the jobs i can see that the job is not removed from redis. 

**Describe alternatives you've considered**
Using the failed event i can get the top level parent of a job and update a failed array in its data field. This doesn't seem like a reliable way to handle this. 

**Additional context**
Any way to get the failed children from a flow would be great.

prolific:
I too would love to have this option. I am building a flow of jobs where each one of them runs a specific set of tests against our system on a daily basis. I am already using `removeDependencyOnFailure` which is really very helpful but I also want to know the list of all the failed children along with the error information stored inside them because I want to either store this data in database or trigger mails based on all this information.

Is there a way to at least get the original list of job ids (including those that failed)? Maybe we can simply loop through them and fetch the required information?

roggervalf:
hey folks, ignoreDependencyOnFailure feature is available since v5.2.0

roggervalf:
hi @lltr, this option ignore the failure of that specific child, it only affect the child and the direct parent no other components of the flow

roggervalf:
pls take a look https://docs.bullmq.io/guide/flows/ignore-dependency

PR #2426, opened 2024-02-15
feat(flow): add ignoreDependencyOnFailure option

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2092 https://github.com/taskforcesh/bullmq/issues/800

### DIFF
--- a/src/commands/removeJob-1.lua
+++ b/src/commands/removeJob-1.lua
@@ -19,6 +19,7 @@ local rcall = redis.call
 --- @include "includes/getOrSetMaxEvents"
 --- @include "includes/isLocked"
 --- @include "includes/removeJobFromAnyState"
+--- @include "includes/removeJobKeys"
 --- @include "includes/removeParentDependencyKey"
 
 local function removeJob( prefix, jobId, parentKey, removeChildren)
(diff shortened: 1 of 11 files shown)


src/commands/removeJob-1.lua
```diff
@@ -50,11 +51,21 @@ local function removeJob( prefix, jobId, parentKey, removeChildren)
                 removeJob( childJobPrefix, childJobId, jobKey, removeChildren )
             end
         end
+
+        local failed = rcall("HGETALL", jobKey .. ":failed")
```
manast, line 55: This call could potentially require a lot of memory and resources if the number of failed jobs is very large. A solution would be to use "https://redis.io/commands/hlen/" first, and if the len is larger than a specific size, maybe 100 or so, start a iteration process. Obviously this is going to be complicated in general as the iteration would need to be done by calling "removeJob" iteratively... Maybe it is not critical to get this done in this PR, but we should mark it in an issue as a known issue/future enhancement.

manast (approved):
Just a small comment.

src/commands/removeJob-1.lua
roggervalf: I referenced one issue to remember this improvement

PR #2434, opened 2024-02-19
feat(worker): add support for naming workers

manast:
This PR adds a new optional property to jobs, processedBy, as well as the possibility to pass a unique name to the workers. This allows to create a mapping between processed jobs and the workers that actually processed them. Furthermore, a new "rawname" property is added to the data structure returned by ```getWorkers``` that includes the worker name.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -21,7 +21,6 @@ import {
   DELAY_TIME_1,
   isNotConnectionError,
   isRedisInstance,
-  WORKER_SUFFIX,
 } from '../utils';
 import { QueueBase } from './queue-base';
 import { Repeat } from './repeat';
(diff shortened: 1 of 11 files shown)


src/classes/worker.ts
```diff
@@ -256,16 +255,21 @@ export class Worker<
...
 
-        const mainFile = this.opts.useWorkerThreads
-          ? 'main-worker.js'
-          : 'main.js';
-        let mainFilePath = path.join(
-          path.dirname(module.filename),
-          `${mainFile}`,
-        );
+        // Separate paths so that bundling tools can resolve dependencies easier
+        const dirname = path.dirname(module.filename || __filename);
```
roggervalf, line 259: I guess this is related to this issue :thinking: ?

roggervalf (approved):
Lgtm, just few styled comments

src/classes/worker.ts
manast: No its not :) its just a refactor though.

PR #2438, opened 2024-02-21
fix(deps): replaced glob by fast-glob due to security advisory

manast:
Since newest versions of "glob" are only compatible with "type": "module" which is basically imposible to enable in this repo, this PR replaces it by fast-glob.


roggervalf (approved):
Lgtm

PR #2449, opened 2024-02-24
feat(worker): add support for conditional connection closure

### LINKED ISSUE #2437: Provide an option to opt-out of automatic Redis connection closing on worker.close
## Is your feature request related to a problem? Please describe.

Currently it is easy to get `Error: Connection is closed` error when Redis instance is reused across multiple workers and is explicitly managed by the application. In case you close all the workers and try to close the Redis connection, you will get `Error: Connection is closed`, because BullMQ has already closed the connection by itself. It is also a potentially unwanted behaviour, because sometimes you may only stop one worker, but not destroy the connection for all other workers.

## Describe the solution you'd like

Extra flag on `worker.close()` method that would avoid terminating the Redis connection/closing client.

## Describe alternatives you've considered

New method  `worker.stop()` which does the same thing as `worker.close()` without terminating the Redis connection. If this is introduced, probably `worker.close()` should start invoking `worker.stop` as well to reduce the duplication.

If this is accepted, we can send a PR with the change.

This should also potentially address https://github.com/taskforcesh/bullmq/issues/2402

kibertoad:
fixes #2437

In case application manages lifecycle of Redis connection itself, this provides an option of opting out of BullMQ automatic connection close.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -807,7 +807,7 @@ export class Worker<
    *
    * @returns Promise that resolves when the worker has been closed.
    */
-  close(force = false): Promise<void> {
+  close(force = false, closeConnection = true): Promise<void> {
     if (this.closing) {
       return this.closing;
     }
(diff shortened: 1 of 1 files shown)


manast:
This is not needed. The underlying Redis connection is only closed if it is not "shared": https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts#L297

PR #2454, opened 2024-02-28
Replace lodash methods

roggervalf:


src/classes/main-base.ts
```diff
@@ -35,7 +34,7 @@ export default (
 
   process.on('uncaughtException', async (err: Error) => {
     if (!err.message) {
-      err = new Error(toString(err));
+      err = new Error(`${err}`);
```
roggervalf, line 37: I would expect that err gets into this [line](https://github.com/lodash/lodash/blob/main/src/toString.ts#L40) so converting it as a string should be sufficient

manast (commented):
Good initiative. I think we need to think more about toString, or worst case implement a similar function ourselves.

src/classes/main-base.ts
roggervalf: looks like this event is only receiving Error objects https://nodejs.org/api/process.html#event-uncaughtexception, do we have another example when this is not true? If not looks like `${err}` will be sufficient :thinking:

manast: Ok, but then why are we doing this:
```ts
if (!err.message) 
```
If err is guaranteed an Error, then it should not be needed to to that check I think.

roggervalf: yeah agree, let me remove that check

src/classes/main-base.ts
```diff
@@ -33,10 +32,7 @@ export default (
   process.on('SIGTERM', () => childProcessor.waitForCurrentJobAndExit());
   process.on('SIGINT', () => childProcessor.waitForCurrentJobAndExit());
 
-  process.on('uncaughtException', async (err: Error) => {
-    if (!err.message) {
-      err = new Error(toString(err));
```
manast, line 38: This makes me a bit nervous, not sure if this is the correct thing to do. So if the error does not include .message we assume it is not an error object, but more likely just a string, and thus we create a new error here, which will create also an error stack. If we do not do this, then we call errorToJSON with a string , will this yield this a correct "value"? This uncaughException is emitted in userland code, out of our control... can we track who did add this code to see if we can understand the original intend?

roggervalf: I can see that this change was added last year with worker threads support https://github.com/taskforcesh/bullmq/commit/0820985e073582fdf841affad38ecc7ab64691ec#diff-dba50fb04784a53f9826c6cb96929de2d71329d82f1d97418a13fc54a955f8d5R37-R39

manast: I think that in order not to create a regression (even though no test is testing this feature), we should implement our own toString then if we want to get rid of lodash.

manast: ah ok, you did already 👍

manast (approved):

PR #2455, opened 2024-02-28
fix(worker): consider block timeouts less than 50ms

roggervalf:


src/classes/worker.ts
```diff
@@ -603,29 +603,27 @@ export class Worker<
...
-          // function only.
-          const result = await bclient.bzpopmin(this.keys.marker, blockTimeout);
-
-          if (result) {
-            const [_key, member, score] = result;
-
-            if (member) {
-              return parseInt(score);
-            }
+        blockTimeout = this.blockingConnection.capabilities.canDoubleTimeout
```
manast, line 606: Actually we must have a blockTimeout that is a positive integer. The reason. being that a value of 0 represents infinity, and a negative number would yield an exception. So maybe 1ms would do it? i.e. replace 0.05 by 0.001

src/classes/worker.ts
```diff
@@ -641,6 +634,23 @@ export class Worker<
...
 
+  protected getBlockTimeout(blockUntil: number): number {
+    const opts: WorkerOptions = <WorkerOptions>this.opts;
+
+    // when there are delayed jobs
+    if (blockUntil) {
+      const blockDelay = blockUntil - Date.now();
+      // when we reach the time to get new jobs
+      if (blockDelay < 1) {
+        return 0.001;
```
roggervalf, line 645: looks like it will be only needed when blockDelay is less than 1 when getting a marker from a delayed job, wdyt @manast

manast (approved):
Ok, lets try this approach.

PR #2465, opened 2024-03-08
feat(job): add exclusive execution option

roggervalf:


manast:
Ok, interesting use of rate-limiter for achieving this effect. A couple of things:
- I am not sure the name of the option conveys the actual function, for me it is more like "sequential execution", although the queue already executes sequentially, except when we are doing retries. I think we need to think more about the name here.
- This option only makes sense with concurrency is 1, as otherwise we have no guarantee that the execution is "sequential", so maybe if concurrency > 1 and this option is chosen we should rise an error.

manast:
Another thing to think about, what happens if there are delayed jobs already in the queue, should we care about them?
Maybe not actually... as they are delayed it means they should be executed as soon as possible after they expire. Also if the user cares about order, then they probably will not have delayed jobs running side by side regular jobs...

src/classes/job.ts
```diff
@@ -38,6 +38,7 @@ import type { QueueEvents } from './queue-events';
 const logger = debuglog('bull');
 
 const optsDecodeMap = {
+  ee: 'exclusiveExecution',
```
roggervalf, line 41: preserveOrder must be used

docs/gitbook/guide/workers/preserve-order.md
```diff
@@ -0,0 +1,16 @@
+# Preserve Order
+
+BullMQ supports preserving execution order in jobs. When _preserveOrder_ option is provided as *true*, jobs will be processed in the same order, indendently of retry strategies. If the current job fails and has a retry strategy, queue will be in rate limit state until the delay is accomplish.
+
+```typescript
+const worker = new Worker('queueName', async (job: Job) => {
+    // do some work
+}, {
+    concurrency: 1
```
manast, line 9: As concurrency is by default 1, I assume it is not required to specify any when using "preserveOrder".

src/classes/worker.ts
```diff
@@ -222,13 +222,18 @@ export class Worker<
...
       runRetryDelay: 15000,
+      preserveOrder: false,
       ...this.opts,
     };
 
     if (this.opts.stalledInterval <= 0) {
       throw new Error('stalledInterval must be greater than 0');
     }
 
+    if (this.opts.preserveOrder && this.opts.concurrency > 1) {
```
manast, line 233: It would be nice to refactor this check as we are doing exactly the same on "set concurrency". In fact, couldn't we just do ```this.concurrency =  ... ``` here and let the checks from set concurrency take care of it?

manast: Another thing, shouldn't we have this check on the python version as well?

src/types/job-options.ts
```diff
@@ -21,6 +21,12 @@ export type JobsOptions = BaseJobOptions & {
  * These fields are the ones stored in Redis with smaller keys for compactness.
  */
 export type RedisJobOptions = BaseJobOptions & {
+  /**
+   * If true, it will rate limit the queue when moving this job into delayed.
+   * Will stop rate limiting the queue until this job is moved to completed or failed.
+   */
+  ee?: boolean;
```
manast, line 28: wouldn't this be better as "po" from Preserver Order?

docs/gitbook/guide/workers/preserve-order.md
roggervalf: yes, no necessary, I added it just to be clear, I can add a comment about the default value

manast:
I think we need more tests for cover the scenario when you have several workers running in parallel. For instance, let's say you have 10 workers (assume all have concurrency 1 AND preserver order flag to true, if any has another set of options then it will not work as expected, we should also write this in the documentation). Now, I assume that as soon one job is fetched, rate limit kicks in and the other workers will just be idling. We need tests to verify that this is the case, also, what happens if the job stalls, it may go back to wait but we need to make sure that no other job waiting is taking by any of the other workers that are idling when this happens, so order is preserved. We also need to document that if you add delayed jobs, then the order cannot be guaranteed either...

manast:
It is a lot, but if we want to guarantee order there are several edge cases to consider, probably others we have not thought about yet.

src/types/job-options.ts
roggervalf: that option comes from the old version of this pr, I'm going to delete it

roggervalf:
> I think we need more tests for cover the scenario when you have several workers running in parallel. For instance, let's say you have 10 workers (assume all have concurrency 1 AND preserver order flag to true, if any has another set of options then it will not work as expected, we should also write this in the documentation). Now, I assume that as soon one job is fetched, rate limit kicks in and the other workers will just be idling. We need tests to verify that this is the case, also, what happens if the job stalls, it may go back to wait but we need to make sure that no other job waiting is taken by any of the other workers that are idling when this happens, so order is preserved. We also need to document that if you add delayed jobs, then the order cannot be guaranteed either...

there is no global concurrency, so having more than 1 worker will break preserve order behavior, also need to document it.

src/classes/worker.ts
roggervalf: I was think on adding it later

docs/gitbook/guide/workers/preserve-order.md
```diff
@@ -0,0 +1,15 @@
...
+```typescript
+const worker = new Worker('queueName', async (job: Job) => {
+    // do some work
+}, {
+    preserveOrder: true
+    });
+```
+
+{% hint style="warning" %}
+This feature is only allowed when using concurrency 1, any greater value will throw an error:
```
manast, line 14: We should add another hint, when using retries and backoffs, for instance, a failed job will keep the queue idle during the time the job is being backed off until it is picked again.

docs/gitbook/guide/workers/preserve-order.md
```diff
@@ -0,0 +1,19 @@
+# Preserve Order
+
+BullMQ supports preserving execution order in jobs. When _preserveOrder_ option is provided as *true*, jobs will be processed in the same order, independently of retry strategies. If the current job fails and has a retry strategy, queue will be in rate limit state until the delay is accomplish.
```
manast, line 3: -> BullMQ supports preserving execution order in jobs. When the _preserveOrder_ option is provided as *true*, jobs will be processed in the same order, independently of retry strategies. If the current job fails and has a retry strategy, the queue will be in rate limit state until the retry delay has expired.

docs/gitbook/guide/workers/preserve-order.md
```diff
@@ -0,0 +1,19 @@
...
+    preserveOrder: true
+    });
+```
+
+{% hint style="info" %}
+when using retries and backoffs, for instance, a failed job will keep the queue idle during the time the job is being backed off until it is picked again.
+{% endhint %}
+
+{% hint style="warning" %}
+This feature is only allowed when using concurrency 1, any greater value will throw an error. Make sure to also set a [global concurrency](https://docs.bullmq.io/guide/queues/global-concurrency)
```
manast, line 18: -> This feature is only allowed when using concurrency 1, any greater value will throw an error. Make sure to also set a [global concurrency](https://docs.bullmq.io/guide/queues/global-concurrency) if you want to preserve order having multiple running workers.

manast (approved):
I just had a couple of comments on the documentation.

TomKaltz:
Any reason this can't be merged?

docs/gitbook/guide/workers/preserve-order.md
manast: "queue will be in..." -> "the queue will be in..."

docs/gitbook/guide/workers/preserve-order.md
manast: "until it is picked again." -> "until it is picked up again."

manast (commented):
I think we also need tests to cover the case where we have multiple workers running in parallel.

manast:
@TomKaltz mostly because it is a difficult feature to get right, so it gets often sidelined in favor of other new features and fixes.

shahshubh:
@manast any updates on this getting merged ?

roggervalf:
hi @shahshubh, this was an option as a first approach but currently I don't think it would be as general as I wanted to be. I have in mind a different approach but it would imply more changes. I don't have a timeline when I could get it done for now as we are focusing on scheduler functionality these days and then I'll be working on the next breaking change. Hopefully after that I would have time to retake this work

PR #2473, opened 2024-03-15
fix(connection): console connection closed error

roggervalf:


manast:
A bit confused about this change, why replacing the reject by a resolve?

manast:
I am thinking that if there is some specific place where we need to ignore this error it would be more elegant to ignore it just in that place instead of inside waitUntilReady.

roggervalf:
yes, I'm still testing how to get rid of it

manast (approved):

PR #2478, opened 2024-03-18
fix(worker): set blockTimeout as 0.001 when reach the time to get delayed jobs [python]

roggervalf:


manast:
I think we should document somewhere that 0.001 seconds which is 1 millisecond is chosen because the granularity of our timestamps are milliseconds. Obviously we can still process much faster than 1 job per millisecond but delays and rate limits will never work with more accuracy than 1ms... maybe I will write a blog post about it.

PR #2481, opened 2024-03-21
fix(queue): use QueueOptions type in opts attribute

roggervalf:


src/classes/queue.ts
```diff
@@ -116,11 +117,7 @@ export class Queue<
     this.waitUntilReady()
       .then(client => {
         if (!this.closing) {
-          client.hset(
-            this.keys.meta,
-            'opts.maxLenEvents',
-            opts?.streams?.events?.maxLen ?? 10000,
-          );
+          client.hmset(this.keys.meta, this.metaValues);
```
roggervalf, line 120: using hmset to be reused in pro version

PR #2483, opened 2024-03-22
fix(scripts): use command name in error message when moving to finished

roggervalf:


src/classes/job.ts
```diff
@@ -200,7 +200,7 @@ export class Job<
       : undefined;
 
     this.toKey = queue.toKey.bind(queue);
-    this.scripts = new Scripts(queue);
+    this.setScripts();
```
roggervalf, line 203: this method will be overwritten in pro version to set scripts once

PR #2486, opened 2024-03-23
fix(python): accept all parameters for redis connection

radicevicbranko:
fix the issue when initializing connection with "ssl": True


roggervalf (commented):
Lgtm. thank you

Issue #2487, opened 2024-03-23
[Bug]: Stacktrace with limit removes the last error stack.


nocodehummel:
### Version

5.4.5

### Platform

NodeJS

### What happened?

Our jobs have `stackTraceLimit` set, and we noticed that the error stack is not updated. 

### How to reproduce.

The problem is caused by an incorrect use of slice inside the Job class.

```
this.stacktrace.push(err.stack);
if (this.opts.stackTraceLimit) {
   this.stacktrace = this.stacktrace.slice(0, this.opts.stackTraceLimit);
}
```

Simple reproduction [example](https://codepen.io/nocodehummel/pen/KKYmoYw) to demonstrate the issue.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
yeah, I see, this actually a real bug as we are pushing the new stacks and then slicing them away if the length is larger stackTraceLimit.

roggervalf:
closing as there is a merged fix in v5.4.6

xNest:
hey, the solution does not allow stackTraceLimit to be set to 0, is this the desired behaviour?

nocodehummel:
@xNest how do you mean does not allow?

xNest:
If you pass 0 with the intention of not saving a stack trace, the limit will be ignored. 

`
if (err?.stack) {
      this.stacktrace.push(err.stack);
      if (this.opts.stackTraceLimit) {
        this.stacktrace = this.stacktrace.slice(-this.opts.stackTraceLimit);
      }
    }
`

PR #2488, opened 2024-03-23
fix(job): stack trace limit (#2487)

nocodehummel:
Changed the array slice for the stacktrace.

Resolution is verified with our BullMQ implementation (via npm link).

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1231,7 +1231,7 @@ export class Job<
     if (err?.stack) {
       this.stacktrace.push(err.stack);
       if (this.opts.stackTraceLimit) {
-        this.stacktrace = this.stacktrace.slice(0, this.opts.stackTraceLimit);
+        this.stacktrace = this.stacktrace.slice(-this.opts.stackTraceLimit);
       }
     }
 
--- a/tests/test_job.ts
+++ b/tests/test_job.ts
@@ -810,11 +810,18 @@ describe('Job', function () {
       const job = (await worker.getNextJob(token)) as Job;
       const isFailed = await job.isFailed();
       expect(isFailed).to.be.equal(false);
-      await job.moveToFailed(new Error('test error'), '0', true);
+      // first time failed.
+      await job.moveToFailed(new Error('failed once'), '0', true);
       const isFailed2 = await job.isFailed();
+      const stackTrace1 = job.stacktrace[0];
       expect(isFailed2).to.be.equal(true);
       expect(job.stacktrace).not.be.equal(null);
       expect(job.stacktrace.length).to.be.equal(stackTraceLimit);
+      // second time failed.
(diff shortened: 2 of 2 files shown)


manast:
What issue does this PR address?

manast:
ok, I found it: https://github.com/taskforcesh/bullmq/issues/2487
We are going to need a small test to avoid a regression for this one.

nocodehummel:
I adjusted the test scripts but do not have a local redis.

nocodehummel:
I see that my test case failed because cannot `moveToFailed` when job is not active.

nocodehummel:
> The change also affects a test case that is failing: https://github.com/taskforcesh/bullmq/actions/runs/8419326269/job/23056271205?pr=2488#step:7:950

That is the test case I extended with extra checks. I tried to use `moveToFailed` twice which causes the issue. Now I have retrieved the job again from the queue for the second `moveToFailed`. 

A bit frustrating to not have a local Redis. When this does not work. How can I run the test again a Cloud Redis?

manast:
> A bit frustrating to not have a local Redis. When this does not work. How can I run the test with a Cloud Redis?

What is preventing you to running a local Redis instance? supports all OSes and also works on Docker.

nocodehummel:
> > A bit frustrating to not have a local Redis. When this does not work. How can I run the test with a Cloud Redis?
> 
> What is preventing you from running a local Redis instance? supports all OSes and also works on Docker.

Test cases contain `const redisHost = process.env.REDIS_HOST || 'localhost';`, but I do not know how to set it.

nocodehummel:
I managed to run the tests using `.env`, but now Chilld pool tests are failing. 
Do the test run when Redis host is configured as environment variable?

````
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
````

manast:
> Test cases contain `const redisHost = process.env.REDIS_HOST || 'localhost';`, but I do not know how to set it.

If you start a redis server locally then localhost points to that server, not sure I understand the issue...

nocodehummel:
I managed to get the test case working and checking that the stacktrace actually changed.

````
> bullmq@5.4.5 test
> ts-mocha -p tsconfig-cjs.json --config ./.mocharc.js -r dotenv/config
  Job
    .moveToFailed
      ✔ applies stacktrace limit on failure (688ms)
  1 passing (975ms)
````

nocodehummel:
> > Test cases contain `const redisHost = process.env.REDIS_HOST || 'localhost';`, but I do not know how to set it.
> 
> If you start a redis server locally then localhost points to that server, not sure I understand the issue...

The point is do not have a local Redis. It looks to me there is an issue with the test cases supporting a configured Redis host. People cannot always install anything they want on their PC.

Issue #2243, opened 2023-10-24
[Bug]: Many connections to Redis


venticello:
### Version

1.15.2

### Platform

Python

### What happened?

I tried to get data from completed tasks from 2 queues, and in the end I got 34 additional connections to Redis.
Well, we also lack the ability to reuse an existing connection like in NodeJS.

### How to reproduce.

```python
import asyncio
from bullmq import Queue

elo_queue = Queue('elo', "redis://default:redispw@localhost:6379")
dna_queue = Queue('dna', "redis://default:redispw@localhost:6379")

async def get_data():
    jobs = await elo_queue.getCompleted(0, 20)
    jobs2 = await dna_queue.getCompleted(0, 20)
    print(jobs, jobs2)
    await asyncio.sleep(10)

if __name__ == "__main__":
    asyncio.run(get_data())
```

### Relevant log output

```shell
127.0.0.1:6379> CLIENT LIST
id=20886 addr=172.24.0.1:47940 laddr=172.24.0.2:6379 fd=56 name= age=4 idle=4 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1928 events=r cmd=hgetall user=default redir=-1 resp=2 lib-name= lib-ver=
id=20887 addr=172.24.0.1:47942 laddr=172.24.0.2:6379 fd=57 name= age=4 idle=4 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1928 events=r cmd=hgetall user=default redir=-1 resp=2 lib-name= lib-ver=
id=20850 addr=172.24.0.1:47868 laddr=172.24.0.2:6379 fd=20 name= age=4 idle=4 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1928 events=r cmd=hgetall user=default redir=-1 resp=2 lib-name= lib-ver=
[...]

PR #2504, opened 2024-04-02
feat(python): support reusable redis connections

mful:
Like `BullMQ` for typescript, support reusing the same redis connection. This is helpful in environments with caps on connection count.


mful:
Linking #2243 — this does not directly address the bug, but should solve the specific problem reported there.

manast (approved):
LGTM

mful:
Great, let me know if you need anything else to merge.

PR #2509, opened 2024-04-05
perf(worker): use abort controller with stalledCheckTimer

roggervalf:
Draft

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -887,6 +887,8 @@ export class Worker<
     if (!this.opts.skipStalledCheck) {
       if (!this.closing) {
         try {
+          await this.checkConnectionError(() => this.moveStalledJobsToWait());
+
           const callback = async () => {
             this.abortStalledController?.signal.removeEventListener(
               'abort',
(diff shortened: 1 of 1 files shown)


src/classes/worker.ts
```diff
@@ -881,14 +885,29 @@ export class Worker<
...
     if (!this.opts.skipStalledCheck) {
-      clearTimeout(this.stalledCheckTimer);
-
       if (!this.closing) {
         try {
-          await this.checkConnectionError(() => this.moveStalledJobsToWait());
-          this.stalledCheckTimer = setTimeout(async () => {
-            await this.startStalledCheckTimer();
-          }, this.opts.stalledInterval);
+          const callback = async () => {
```
manast, line 890: I am a bit confused about using the same callback for the timeout as well as for the abort signal. I think abort should only do abort and nothing else.

manast: yes but there is a slight difference, in the delay case the callback is actually a "cleanup" function, but here the callback is both cleanup and starting a new timer. So I think in this case it will be better to refactor to a "cleaup" function that can be used only for the abort signal, and that is also called inside the current callback.

roggervalf: I think I overlooked this, as I can see that we are cleaning the timer here https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L869

PR #2512, opened 2024-04-07
perf(stalled): remove jobId from stalled after removing lock when moved from active

roggervalf:
https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveStalledJobsToWait-9.lua#L77 we will save some extra calls


src/commands/includes/removeLockToken.lua
```diff
@@ -0,0 +1,19 @@
+local function removeLockToken(jobKey, stalledKey, token, jobId)
```
manast, line 1: I think the name should just be "removeLock"

manast (approved):
other than the small comment i LGTM

Issue #2472, opened 2024-03-12
[Python] Add a log method


Spoutnik97:
**Is your feature request related to a problem? Please describe.**
Currently, it is not possible to log a message with the Python API

**Describe the solution you'd like**
Same as the JS API

**Describe alternatives you've considered**
-

**Additional context**
 Useful to debug or track information like time to compute each task

roggervalf:
hi @Spoutnik97 logs methods should be available in version 2.7.0

Spoutnik97:
Great! Thanks!
I already tested it, and every things works fine

PR #2523, opened 2024-04-13
feat(queue): add getJobLogs method [python]

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2472

### DIFF
--- a/python/bullmq/queue.py
+++ b/python/bullmq/queue.py
@@ -116,6 +116,31 @@ class Queue:
         """
         return self.client.pttl(self.keys["limiter"])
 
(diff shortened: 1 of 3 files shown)


lhr0909:
@roggervalf Hi, I am wondering when these changes will be uploaded to PyPI, seems like there is only bullmq 2.3.3 as latest on it. Did we change the way it is downloaded?

manast:
Looking into it... I think it has to do with some new security improvements in pypi.

roggervalf:
pls @lhr0909 try v2.7.1

Issue #2466, opened 2024-03-08
[Bug]: Worker stopped processing jobs, and mostly delayed Jobs


wernermorgenstern:
### Version

v5.4.2

### Platform

NodeJS

### What happened?

We have a service, where a worker runs, and processes jobs.  After the processing is done, it will create another job, which is delayed (around 64 minutes).
Today, I noticed that the service and worker stopped processing jobs.  There were no error messages in the logs.  When I used BullBoard (I use it as a UI to see jobs), I saw the jobs were still in the delayed state, and like 24 hours overdue.

When I restarted the service, and the worker started, it immediately started processing those delayed jobs.
This is not the first it happened.  Today I though first checked the delayed jobs.

In today's incident, the service has been running for 4 days.

We run in EKS on AWS (NodeJS service, using Typescript).  I use BullMQ Pro. And we are using Groups and each Group has a concurrency set to 1.

### How to reproduce.

I don't have any test code for this

### Relevant log output

```shell
No Logs or error logs were produced
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
Do you know if the worker is still connected to Redis when this happens? for instance, does the workers appear list of workers that are online for the queue?

wernermorgenstern:
That I need to check.  That is a good idea.  Is there a method which shows that?  Or something I can look directly in Redis?

manast:
You can use https://api.docs.bullmq.io/classes/v5.Queue.html#getWorkers or Taskforce.sh, I don't know if BullBoard can also show this information.

manast:
Also important to read this if you haven't already: https://docs.bullmq.io/guide/going-to-production#maxretriesperrequest

wernermorgenstern:
> Also important to read this if you haven't already: https://docs.bullmq.io/guide/going-to-production#maxretriesperrequest

I have set that one as null

```
    const redis = new Redis(config.redisBullUrl as string, {
      keepAlive: 1,
      autoResubscribe: true,
      autoResendUnfulfilledCommands: true,
      connectionName: `${config.serviceName}-${config.agentId}-${redisConnection}-bullmq`,
      connectTimeout: helpers.period('10s'),
      enableOfflineQueue: true,
      enableReadyCheck: true,
      enableAutoPipelining: true,
      maxRetriesPerRequest: null,
      retryStrategy: function (times) {
        return Math.max(Math.min(Math.exp(times), 20000), 1000)
      },
[...]

PR #2529, opened 2024-04-19
perf(worker): reset delays after generating blockTimeout value

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2466

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -661,7 +661,7 @@ export class Worker<
       const blockDelay = blockUntil - Date.now();
       // when we reach the time to get new jobs
       if (blockDelay < this.minimumBlockTimeout * 1000) {
-        blockTimeout = this.minimumBlockTimeout;
+        return this.minimumBlockTimeout;
       } else {
         blockTimeout = blockDelay / 1000;
       }
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -661,7 +661,7 @@ export class Worker<
       const blockDelay = blockUntil - Date.now();
       // when we reach the time to get new jobs
       if (blockDelay < this.minimumBlockTimeout * 1000) {
-        blockTimeout = minimumBlockTimeout;
+        blockTimeout = this.minimumBlockTimeout;
       } else {
         blockTimeout = blockDelay / 1000;
       }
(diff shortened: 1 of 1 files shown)


src/classes/worker.ts
```diff
@@ -552,7 +552,7 @@ export class Worker<
       try {
         this.blockUntil = await this.waiting;
 
-        if (this.blockUntil <= 0 || this.blockUntil - Date.now() < 10) {
+        if (this.blockUntil <= 0 || this.blockUntil - Date.now() < 5) {
```
roggervalf, line 555: in case this operation is too fast, what about using 5 milliseconds

src/classes/worker.ts
```diff
@@ -626,6 +626,7 @@ export class Worker<
           ? blockTimeout
           : Math.ceil(blockTimeout);
 
+        this.updateDelays(); // reset delays to avoid reusing same values in next iteration
```
manast, line 629: I am concerned about this. Can you express is in a test case where this was a problem before?

src/classes/worker.ts
manast: Yeah. The issue here is that if the moveToActive is called too early it may not return any job. In theory, it should be <= 1 ms to be consistent with the BZPOPMIN minimum timeout value.

manast: Btw I did not remember we had this hardcoded value here...

src/classes/worker.ts
roggervalf: It's a perf change as we can see from these logs that https://github.com/taskforcesh/bullmq/issues/2466#issuecomment-2065125025 zpopmin is called 2 more times after consuming the last delayed marker, as the blockUntil is not reset

src/classes/worker.ts
roggervalf: it's coming from https://github.com/taskforcesh/bullmq/commit/b0a13e86285998ce0524eb443d732bce91e9b6c5

src/classes/worker.ts
manast: ok, so the next call of BZPOPMIN in that example would have been the default instead of 0.9 or something.

manast (approved):

src/classes/worker.ts
manast: ok, I see you updated already, sorry for the long comment 😅

PR #2533, opened 2024-04-20
feat(worker): add getNextRateLimitTTLPerName method

roggervalf:
sometimes there are more than one rate limit rule that jobs can reach, for that purpose we can move those jobs to delayed instead of rate limiting the entire queue. But how we can determine the most accurate ttl for those jobs, this is where getNextRateLimitTTLPerName can be used


roggervalf:
not ready yet, need more discussion

PR #2543, opened 2024-04-25
fix(worker): force timeout on bzpopmin command

manast:
Workaround for this IORedis bug: https://github.com/redis/ioredis/issues/1888

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -38,10 +38,6 @@ import {
 
 // 10 seconds is the maximum time a BRPOPLPUSH can block.
 const maximumBlockTimeout = 10;
-/* 1 millisecond is chosen because the granularity of our timestamps are milliseconds.
-Obviously we can still process much faster than 1 job per millisecond but delays and rate limits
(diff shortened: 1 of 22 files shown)


src/classes/worker.ts
```diff
@@ -625,10 +625,25 @@ will never work with more accuracy than 1ms. */
           ? blockTimeout
           : Math.ceil(blockTimeout);
 
+        // We restrict the maximum block timeout to 10 second to avoid
+        // blocking the connection for too long in the case of reconnections
+        // reference: https://github.com/taskforcesh/bullmq/issues/1658
+        blockTimeout = Math.min(blockTimeout, maximumBlockTimeout);
```
roggervalf, line 631: this line is the same as reverting this comment https://github.com/taskforcesh/bullmq/commit/96933217bf79658e5bb23fd7afe47e0b1150a40d, what about apply this logic into https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L670 that is the only case where blockTimeout could be greaten than maximumBlockTimeout

PR #2545, opened 2024-04-26
feat(python): wait for jobs to finalize on close

klaemo:
This makes the worker `close` method wait for jobs to be done processing in the python library. I think this brings it closer to the behavior of the JS library.

We have used this in production for a while and has worked out well.


roggervalf (approved):
Lgtm

Issue #2542, opened 2024-04-25
[Bug]: failed jobs due to random "Error UNKNOWN while writing to socket. Connection lost"


klaemo:
### Version

vpy2.7.2

### Platform

Python

### What happened?

We have a python worker that consumes jobs from a nodejs queue. It mostly works really well, but we do get random `Error UNKNOWN while writing to socket. Connection lost` errors on the python side. On the nodejs side we do not observer these connection losses.

Connection losses are fine and expected in a distributed system, but I'm surprised they bubble up to the job and make it fail 🤔 As far as I can tell there is some kind of retry logic in place also in the python client. But, does it work? Shouldn't the gracefully handle temporary connectivity issues?

- python `3.11.5`
- redis `7.0.15`, hosted on DigitalOcean
- the worker started with all the default settings, we only specify the redis connection details

### How to reproduce.

I am not sure. We're unable to reliably reproduce this.

### Relevant log output

```shell
ERROR: Error UNKNOWN while writing to socket. Connection lost.
Traceback (most recent call last):
  File "/usr/src/app/.venv/lib/python3.11/site-packages/redis/asyncio/connection.py", line 473, in send_packed_command
    await self._writer.drain()
  File "/usr/local/lib/python3.11/asyncio/streams.py", line 378, in drain
    await self._protocol._drain_helper()
  File "/usr/local/lib/python3.11/asyncio/streams.py", line 167, in _drain_helper
    raise ConnectionResetError('Connection lost')
ConnectionResetError: Connection lost

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/src/app/.venv/lib/python3.11/site-packages/bullmq/utils.py", line 9, in extract_result
    return job_task.result()
           ^^^^^^^^^^^^^^^^^
  File "/usr/src/app/.venv/lib/python3.11/site-packages/bullmq/worker.py", line 116, in getNextJob
    job_instance = await self.moveToActive(token)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/app/.venv/lib/python3.11/site-packages/bullmq/worker.py", line 125, in moveToActive
    result = await self.scripts.moveToActive(token, self.opts)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2546, opened 2024-04-26
fix(python): increase redis retry strategy backoff

klaemo:
As you can see in the [redis-py docs](https://redis.readthedocs.io/en/stable/backoff.html#redis.backoff.ExponentialBackoff) the `ExponentialBackoff` starts with 8**ms** and ramps up to only 512**ms**. Paired with only three retries this makes it fail very quickly.

I increased the values to start with 1**s** and ramp up to 20**s**. Also, increased the retries to 20. I think this brings it closer to the JavaScript library.

Related to my issue: #2542


roggervalf (approved):
Lgtm

PR #2551, opened 2024-05-01
docs(pro): add local group concurrency section

roggervalf:


docs/gitbook/bullmq-pro/groups/local-group-concurrency.md
```diff
@@ -0,0 +1,13 @@
+# Local group concurrency
+
+It is possible to set specific group concurrency. This can be useful if you want to differentiate your groups with different max councurrency rules.
```
manast, line 3: It is *also* possible to ...

differentiate -> run

I would rewrite as this:
It is also possible to set a specific concurrency value to a given group.   This is useful if you require that different groups should run with different concurrency factors.
Please keep in mind that when specifying a group's concurrency factor, you are storing this value in Redis, so it is your responsibility to remove it if you are not using it anymore.

manast (commented):
One thing, does obliterate remove the keys created by this functionality?

docs/gitbook/bullmq-pro/groups/local-group-concurrency.md
roggervalf: hey, obliterate method is not removing any groups now, but I'm adding this deletion when removing groups in https://github.com/taskforcesh/bullmq-pro/pull/226

manast (approved):

Issue #2552, opened 2024-05-01
[Bug]: Worker reconnection after explicit .close() call  due to Redis worker reconnection


ScreaminSauce:
### Version

v5.7.7

### Platform

NodeJS

### What happened?

Using the current release, it appears that Worker reconnection logic keeps the ioredis client open (and reconnects) after the .close() method has been called on both the Queue and the Worker

v5.7.6 does not hang after test completion, leading me to believe the cause may be the updated ioredis client including PR #2543 (https://github.com/taskforcesh/bullmq/pull/2543)

### How to reproduce.

```
import { Redis } from 'ioredis';
import { strict as assert } from "node:assert";
import { Queue, QueueEvents, Worker } from 'bullmq';
describe.only("BullMq Test", async function () {
    it.only("should fire a completed event with a job in the completed state!", function (done) {
        (async function () {
            const eventsInFiredOrder = [];
            const redisClient = new Redis({ disconnectTimeout: 0, maxRetriesPerRequest: null });
            const jobName = new String(Date.now());
            const myQueue = new Queue(jobName, { connection: redisClient });

            const worker = new Worker(jobName, async job => {
                return job.data;
            }, { connection: redisClient });

            async function cleanUpTest() {
                myQueue.removeAllListeners();
                await myQueue.obliterate({ force: true });
                await myQueue.close();
                worker.removeAllListeners();
                await worker.close();
                redisClient.disconnect();
            }

            try {
                worker.on('completed', async (job) => {
                    eventsInFiredOrder.push('completed');
                    try {
                        const isCompleted = await job.isCompleted();
                        assert.ok(isCompleted);
                    }
                    catch (e) {
                        await cleanUpTest();
                        done(e);
                        return;//no need to process any further if we encountered an issue here 
                    }
                    try {
                        await cleanUpTest()
                        assert.deepEqual(eventsInFiredOrder, [
                            'completed'
                        ]);
                        done();
                    }
                    catch (e) {
                        done(e);
                    }
                });

                //Start Queue/Worker
                await myQueue.waitUntilReady();
                await worker.waitUntilReady();

                //Add job
                const jobData = { foo: 'bar' };
                await myQueue.add('myJobName', jobData);
            }
            catch (e) {
                await cleanUpTest();
                done(e);
            }
        })();
    });
```

### Relevant log output

```shell
ioredis:redis status[localhost:6379]: wait -> connecting +0ms
[...]

PR #2553, opened 2024-05-01
fix(worker): make sure clearTimeout is always called after bzpopmin

manast:
Potentially fix for https://github.com/taskforcesh/bullmq/issues/2552


roggervalf (approved):
Lgtm

PR #2558, opened 2024-05-09
feat(python): clarify/simplify test setup

AIexanderDicke:
- Adds script to copy rawScripts to be used by the python package
- Simplify setting up the python tests
- `redis-cli` not needed in order to cleanup `redis` before starting tests

### DIFF
--- a/python/copy_scripts.sh
+++ b/python/copy_scripts.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+cp -r ../rawScripts/. ./bullmq/commands/ || echo "Raw scripts are not available. Create them by running ' yarn install'  and/or ' yarn generate:raw:scripts'"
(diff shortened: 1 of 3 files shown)


manast:
Thanks for the PR. I wonder, when is copy_scripts.py supposed to be called?

AIexanderDicke:
@manast This script should be called before running the tests. You could also add the command to `run_tests.sh`, but I didn't want to change the functionality of this script.

manast:
I think if generating the scripts is necessary to make sure the tests are running with the latest lua script, it should be called in run_tests.sh. What do you think @roggervalf ?

roggervalf:
this should be covered by #3594. Thank you @AIexanderDicke

AIexanderDicke:
@roggervalf While I'm glad to see the changes in, it's frustrating that this PR was closed and the majority of the  changes I proposed here were copied to #3594 rather than merged. I think this is not a good practice.

PR #2559, opened 2024-05-09
feat(python): makes `RedisConnection`'s close method async, use `aclose` when possible

AIexanderDicke:
`redis.asyncio.client.Redis.close` has been deprecated in Version `5.0.1` of `redis`. Therefore, `RedisConnection.close` now uses `redis.asyncio.client.Redis.aclose` whenever possible. To ensure backwards compatibility, `close` is still used if `aclose` is not available. 

More importantly, `self.conn.aclose()` will be awaited (resulting in `RedisConnection.close` needing to be async). In fact, this fixes `RuntimeError: Event loop stopped before Future completed.` errors I get in certain situations when trying to gracefully stop a worker by running `worker.close()`.

### DIFF
--- a/python/bullmq/flow_producer.py
+++ b/python/bullmq/flow_producer.py
@@ -134,8 +134,8 @@ class FlowProducer:
 
         return result
 
-    def close(self):
+    async def close(self):
         """
         Close the flow instance.
         """
-        return self.redisConnection.close()
+        return await self.redisConnection.close()
--- a/python/bullmq/queue.py
+++ b/python/bullmq/queue.py
@@ -308,11 +308,11 @@ class Queue:
             'waiting-children'
         ]
 
-    def close(self):
+    async def close(self):
         """
         Close the queue instance.
         """
-        return self.redisConnection.close()
+        return await self.redisConnection.close()
 
     def remove(self, job_id: str, opts: dict = {}):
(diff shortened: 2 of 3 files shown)


python/bullmq/flow_producer.py
```diff
@@ -134,8 +134,8 @@ async def addBulk(self, flows: list[dict]):
 
         return result
 
-    def close(self):
+    async def close(self):
         """
         Close the flow instance.
         """
-        return self.redisConnection.close()
+        return await self.redisConnection.close()
```
manast, line 141: Do you need to await if the result is a future if you are going to return it anyway? I wonder because in JS is not needed.

AIexanderDicke: Thanks for your comment! I thought about it a little bit and I think the best solution is to get rid of the return statement completely.

manast: Hmm, not sure why it is clearer, for instance, is it possible that the user calls close without awaiting it?

AIexanderDicke: When you see something like
```
async def foo():
    await bar()
```
it is immediately clear that you have to run `await foo()`. On the other hand, if you have a function 
```
def foo():
   return bar()
```
it's not so clear in my opinion.

klaemo:
Can this get merged so we get rid of the deprecation warning and proper async closing of workers?

PR #2574, opened 2024-05-21
fix(repeat): throw error when endDate is pointing to the past

roggervalf:


manast (approved):
LGTM. However a small risk somebody will think this change should be a breaking change. However, as it is today, it is just not working, just failing silently, so it should be OK.

PR #2578, opened 2024-05-23
feat: add Serialize type

nullndr:
## Changelog

Currently BullMQ is lying with the data type workers operate with.

Consider the following code:

```ts
type Payload = {
  date: Date;
};

new Worker<Payload>("queue", async ({ data: { date }}) => { 
  // Typescript is ok with this, but Javascript will happily throw a TypeError
  console.log(date.toISOString());
});
```

This is because the type we use does not correctly reppresent the `data` object we think we operate with. You can imagine `data` declared as `data = JSON.parse(JSON.stringify(...));`.

> This is the same issue the remix.run team had to deal with a while ago (Indeed the type I added is actually just a little refactor of it).

This PR adds the `Serialize` type used by the `Worker` class.

The `Serialize` type serializes the `DataType` we pass to the `Worker`. With this the previous example correctly raise a compilation error in ts.

## Notes

Please note that this is likely to be a breaking change, since some users may just pass around the `data` object without any checking.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -15,7 +15,7 @@ import {
   RedisClient,
   WorkerOptions,
 } from '../interfaces';
-import { MinimalQueue } from '../types';
+import { MinimalQueue, Serialize } from '../types';
 import {
   delay,
   DELAY_TIME_1,
@@ -52,7 +52,7 @@ export interface WorkerListener<
    *
    * This event is triggered when a job enters the 'active' state.
    */
-  active: (job: Job<DataType, ResultType, NameType>, prev: string) => void;
+  active: (job: Job<Serialize<DataType>, ResultType, NameType>, prev: string) => void;
 
   /**
    * Listen to 'closing' event.
@@ -74,7 +74,7 @@ export interface WorkerListener<
    * This event is triggered when a job has successfully completed.
    */
   completed: (
-    job: Job<DataType, ResultType, NameType>,
+    job: Job<Serialize<DataType>, ResultType, NameType>,
     result: ResultType,
     prev: string,
   ) => void;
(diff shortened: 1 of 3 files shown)


manast:
This is an interesting PR. I wonder if you have a link with more information regarding the type Serialize, as I guess this has been discussed in some other forums, it is not completely trivial to understand how it works. Also as you mentioned, will imply a breaking change so we need to plan this change carefully.

nullndr:
@manast the primary ref for this is the type `Jsonify` from [type-fest](https://github.com/sindresorhus/type-fest/blob/0f732371f607fe44e934d178eb97ad71eccda873/source/jsonify.d.ts#L93).

Other packages have also been proposed on [SO](https://stackoverflow.com/questions/16261119/typescript-objects-serialization), but these are primary focusing on serialization of the value, not the type.

nullndr:
The remix team uses a bunch of utilities that will raise a typescript error if something is wrong. I could use those as well.

manast:
Is the remix license incompatible with BullMQ? I wondering in case we could just use that code as is?

nullndr:
The licensing is not really my best field, but remix.run is under the [MIT](https://github.com/remix-run/remix?tab=MIT-1-ov-file), so it should be fully compatible with BullMQ.

> in case we could just use that code as is

I truly understand your worries but the version in this PR is a little better to the remix.run, primary on the `Prettify` type. (The remix.run one is declared erroneusly)

nullndr:
This latest commit fixes a previous typo for the serialization of `Promise`, sorry!

manast:
I think that before we can merge this we are going to need a pretty good test coverage to avoid breaking existing codebases if for instance typescript would complain on legitimate types and so on.

nullndr:
I completely agree, let me know how I can help

PR #2579, opened 2024-05-24
fix: remove print calls [python]

roggervalf:


python/bullmq/timer.py
```diff
@@ -17,8 +17,8 @@ async def _job(self):
             while self._ok:
                 await asyncio.sleep(self.interval)
                 await self.callback(*self.args, **self.kwargs)
-        except Exception as ex:
-            print(ex)
+        except Exception:
+            pass
```
manast, line 21: maybe we need to emit an error here?

python/bullmq/utils.py
```diff
@@ -10,8 +10,7 @@ def extract_result(job_task):
     except Exception as e:
         if not str(e).startswith('Connection closed by server'):
             # lets use a simple-but-effective error handling:
-            # print error message and ignore the job
-            print("ERROR:", e)
+            # ignore the job
```
manast, line 13: emit error?

manast (commented):
Looks good but I think errors should be emitted where before we had the prints

manast (approved):
LGTM

PR #2585, opened 2024-05-29
fix(worker): properly cancel blocking command during disconnections

manast:
The custom code for performing a disconnection and reconnection was proven to cause problems, whereas the IORedis builtin worked properly in all the scenarios that we know of.  This should fix issue #2466


roggervalf (approved):
Lgtm

roggervalf:
related to https://github.com/redis/ioredis/issues/1888

PR #2586, opened 2024-05-29
Fix Redis config typo in going-to-production.md

mbrookson:
Fix typo - change `maxmemory` to `maxmemory-policy` which is the correct configuration that should be referenced.

See docs:
> The exact behavior Redis follows when the maxmemory limit is reached is configured using the maxmemory-policy configuration directive.

https://redis.io/docs/latest/develop/reference/eviction/


manast:
Good catch, thanks!

PR #2587, opened 2024-05-30
fix(delayed): avoid using jobId in order to schedule delayed jobs

roggervalf:
ref https://github.com/OptimalBits/bull/discussions/2742


manast (commented):
I propose a different way to calculate the score for the delayed job in the ZSET by using ZCOUNT https://redis.io/docs/latest/commands/zcount/ using as range all the possible values that a delayed job could get within a timestamp, that is:
[timestamp * 4096, (timestamp + 1) * 4096]. The count + 1 would then be used as the jobCounter is used today.

manast (commented):
This looks good to me. Lets give it a bit more thought before merging just to make sure it really works well.

manast (approved):
I think this is a great PR, for me everything looks correct.

PR #2597, opened 2024-06-08
fix(priority): use module instead of bit.band to keep order

roggervalf:


manast (commented):
Interesting. Didn't bit.band work as expected?

roggervalf:
yeah, it don't. That test fail with previous changes

manast (approved):

PR #2600, opened 2024-06-09
feat(sandbox): add getChildrenValues method

### LINKED ISSUE #753: Flows: job.getChildrenValues is not a function for sandboxed workers
Version 1.46.6
The below line works as documented for normal workers, but throws the error `job.getChildrenValues is not a function` for sandboxed processors. 
``` typescript
const childrenValues = await job.getChildrenValues();
```

Below are code snippets used to test, sans imports for brevity.
A nested job added to a FlowProducer:
``` typescript
await flowProducer.add({
    name: "parent-job",
    queueName: "parent",
    children: [{
      name: "child",
      queueName: "child",
      data: { colour: "red" },
    }]
  });
```

Processed by sandboxed worker processes:
``` typescript
const parentWorker = new Worker(
  "parent", 
  path.join(__dirname, "../dist") + "/processor-parent.js",
);

const childWorker = new Worker(
  "child", 
  path.join(__dirname, "../dist") + "/processor-child.js",
);
```

Processor definitions:
``` typescript
import { processorParent } from "./processors";
export default processorParent;

import { processorChild } from "./processors";
export default processorChild;
```

processors.ts:
``` typescript
export const processorParent = async(job: Job) => {
  const childrenValues = await job.getChildrenValues();
  console.log(childrenValues);
}

export const processorChild = async (job: Job) => {
  return "Child Return Test Value"
}
```

### LINKED ISSUE #1306: How to get children results inside parent job using SandboxedProcessor and Flow?
Hello everyone! How can I get the result of all children jobs in parent job inside SandboxedProcessor? 
It seems `job.getChildrenValues()` is not available for SandboxedJob. Any suggestions?

anirudh1713:
right before starting a child, get and pass children values along side other job data. This should be okay as the parent job runs only after all children have finished. We then use this data to implement `getChildrenValues` function for sanboxed jobs.

"fix  #1306", "fix #753".

### DIFF
--- a/src/classes/sandbox.ts
+++ b/src/classes/sandbox.ts
@@ -12,9 +12,10 @@ const sandbox = <T, R, N extends string>(
     let msgHandler: any;
     let exitHandler: any;
 
+    const childrenValues = await job.getChildrenValues();
     await child.send({
       cmd: ChildCommand.Start,
-      job: job.asJSONSandbox(),
+      job: { ...job.asJSONSandbox(), childrenValues },
       token,
     });
 
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -1,6 +1,10 @@
 import { ParentCommand } from '../enums';
 import { SandboxedJob } from '../interfaces';
-import { JobJsonSandbox } from '../types';
+import {
+  ChildrenValues,
+  JobJsonSandbox,
+  JobJsonSandBoxWithChildrenValues,
+} from '../types';
 import { errorToJSON } from '../utils';
 
 enum ChildStatus {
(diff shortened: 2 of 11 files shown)


src/classes/sandbox.ts
```diff
@@ -12,9 +12,10 @@ const sandbox = <T, R, N extends string>(
     let msgHandler: any;
     let exitHandler: any;
 
+    const childrenValues = await job.getChildrenValues();
     await child.send({
       cmd: ChildCommand.Start,
-      job: job.asJSONSandbox(),
+      job: { ...job.asJSONSandbox(), childrenValues },
```
anirudh1713, line 18: This should be ok, considering parent only starts after all of its children are done. Also, ig its probably ok to assume that children values don't need to be serialized. :thinking: 

If bullmq ever introduces new functionalities around running parent job before child jobs finishes, this approach will probably produce unexpected behavior since it will only return this last known values (based on when it was started).

roggervalf:
hi @anirudh1713 sorry for the delay. Actually there is one use case where you can create children dynamically https://docs.bullmq.io/patterns/process-step-jobs#waiting-children. If waiting for children step returns false right away in the same first execution of the parent, children values won't be available

roggervalf:
available since https://github.com/taskforcesh/bullmq/pull/3023

Issue #2399, opened 2024-01-26
[Bug]: Duplicate Jobs Created Despite Same Custom Key with Different Repeat Intervals


erenkurnaz:
### Version

v5.1.5

### Platform

NodeJS

### What happened?

**Description:**
I've encountered an issue related to creating repeated jobs with a custom key using the BullMQ package. This functionality was introduced in [PR #1824](https://github.com/taskforcesh/bullmq/pull/1824).

**Issue:**
When I add a job with the same repeat interval and custom key, no duplicate jobs are created, which is the expected behavior:
```typescript
await queue.add(queueName, {}, {
  repeat: {
    every: 1 * 1000,
    key: 'custom-key'
  },
});

```

However, changing the repeat interval while keeping the same key results in the creation of duplicate jobs:
```typescript
await queue.add(queueName, {}, {
  repeat: {
    every: 2 * 1000,
    key: 'custom-key'
  },
});

```

**Expected Behavior:**
Even when the repeat interval is modified, as long as the custom key remains the same, it should not result in duplicate jobs. The uniqueness of a job should be maintained based on the custom key provided.

### How to reproduce.

**Steps to Reproduce:**

* Create a repeated job with a specific interval and a custom key.
* Create another job with a different interval but the same custom key.
* Observe that duplicate jobs are created.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #2612, opened 2024-06-18
Repeatable Jobs - Ability to Upsert/Remove Job via Job ID Alone 


valeeum:
**Is your feature request related to a problem? Please describe.**
I'm in the process of creating middleware for a microservices framework which allows services to run on a given schedule via BullMQ.  The repeat options are provided in the service metadata as JSON configuration and the repeatable job is created when the service broker is initialized via the middleware.  The issue I'm facing is that if the repeat job options are changed and service restarted, it would create two separate repeatable jobs.   

**Describe the solution you'd like**
1)  Create a removeRepeatableByJobID method which would remove all instances of repeatable jobs by jobID (which i understand is not unique for repeatable jobs)
2) Add optional upsert-like capability when adding repeatable jobs so that if an existing job is already defined (by jobId), it gets overwritten with the new repeat job options.  

**Describe alternatives you've considered**
I understand that I can create a database table and store the repeatJobKey provided after calling the queue add() method and then  use that value to call removeRepeatableByKey but it seems overkill for this purpose.

doaortu:
For anyone stumbled in this issue,
while waiting the #2617  to merge,
I have a simple hack using existing redis connection to store the `repeatJobKey` : 
```typescript
const job = queue.add('name', data, {
  repeat: {
    // repeat opts,
  }
})
const ok = await conn.set("repeatable-job-id", job.repeatJobKey)
```

and when you want to remove the job, query the key and `removeRepeatableByKey`
```typescript
const key = await conn.get("repeatable-job-id")
if(key) await queue.removeRepeatableByKey(key)
const ok - await conn.del("repeatable-job-id")
```

I don't know if this hack is truly safe, but I'm using it right now, and I haven't found a problem yet.

valeeum:
@doaortu 

This is not safe because based on repeat options, you can have multiple jobs for the same job id.

manast:
This has been now implemented using job schedulers: https://docs.bullmq.io/guide/job-schedulers

Issue #2596, opened 2024-06-07
[Bug]: opitmize compute repeatJobKey


Gumplefik:
### Version

v4.10.0

### Platform

NodeJS

### What happened?

getRepeatKey should support more options.Think about this example:
1. We have define a task to notify people in Monday and Tuesday and we have register it as a worker which named notify.
```typescript
function notify(receiver: string, message: string) {
 // ...
}
```
2. We add different job
```typescript
queue.add("notify", ["xxx", "hello"], {
  repeat: {
    pattern: "0 10 1,2 * *"
  }
})
queue.add("notify", ["zzz", "hello"], {
  repeat: {
    pattern: "0 10 1,2 * *"
  }
})
```
3. But actually we have one job because these jobs have same repeatKey. These jobs have different parameters, so maybe they should be treat as different job.

Maybe some considerations should have different style like bold, so we are easy to notice these things.
[repeatable](https://docs.bullmq.io/guide/jobs/repeatable)
Also we can support custom way to compute a repeatKey

### How to reproduce.

support custom getRepeatKey function
```javascript
function customGetRepeatKey(name, data, repeatOpts) {
return ''
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2617, opened 2024-06-22
feat(repeatable): new repeatables structure

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2612 fixes https://github.com/taskforcesh/bullmq/issues/2399 https://github.com/taskforcesh/bullmq/issues/2596

### DIFF
--- a/src/classes/repeat.ts
+++ b/src/classes/repeat.ts
@@ -75,10 +75,10 @@ export class Repeat extends QueueBase {
         repeatOpts.jobId = opts.jobId;
       }
 
-      const optionsConcat = getRepeatKey(name, repeatOpts);
+      const qualifiedName = getQualifiedName(name, repeatOpts);
 
       const repeatJobKey = await this.scripts.addRepeatableJob(
-        opts.repeat.key ?? this.hash(optionsConcat),
+        opts.repeat.key ?? this.hash(qualifiedName),
         nextMillis,
         {
           name,
@@ -89,7 +89,7 @@ export class Repeat extends QueueBase {
           pattern: repeatOpts.pattern,
           every: repeatOpts.every,
         },
-        optionsConcat,
+        qualifiedName,
         skipCheckExists,
       );
 
@@ -149,19 +149,19 @@ export class Repeat extends QueueBase {
     repeat: RepeatOptions,
     jobId?: string,
   ): Promise<number> {
-    const optionsConcat = getRepeatKey(name, { ...repeat, jobId });
+    const qualifiedName = getQualifiedName(name, { ...repeat, jobId });
     const repeatJobKey = repeat.key ?? this.hash(optionsConcat);
-    const oldRepeatJobId = this.getRepeatJobId({
+    const legacyRepeatJobId = this.getRepeatJobId({
       name,
       nextMillis: '',
-      namespace: this.hash(optionsConcat),
+      namespace: this.hash(qualifiedName),
       jobId: jobId ?? repeat.jobId,
       key: repeat.key,
     });
 
     return this.scripts.removeRepeatable(
-      oldRepeatJobId,
-      optionsConcat,
+      legacyRepeatJobId,
+      qualifiedName,
       repeatJobKey,
     );
   }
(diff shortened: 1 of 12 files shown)


src/classes/repeat.ts
```diff
@@ -31,7 +36,7 @@ export class Repeat extends QueueBase {
     skipCheckExists?: boolean,
```
manast, line 36: Do you remember the use case where this boolean was needed?

src/classes/repeat.ts
```diff
@@ -70,23 +75,28 @@ export class Repeat extends QueueBase {
         repeatOpts.jobId = opts.jobId;
       }
 
-      const repeatJobKey = getRepeatKey(name, repeatOpts);
+      const optionsConcat = getRepeatKey(name, repeatOpts);
```
manast, line 78: Maybe we should rename the function getRepeatKey if this is not going to be the key anymore.

src/classes/repeat.ts
```diff
@@ -109,17 +119,12 @@ export class Repeat extends QueueBase {
...
     hasImmediately: boolean,
   ) {
-    const client = await this.client;
-
     //
     // Generate unique job id for this iteration.
     //
-    const jobId = this.getRepeatJobId({
-      name,
+    const jobId = this.getRepeatDelayedJobId({
```
manast, line 125: It is important that this jobId is generated in the same way as in the previous version to avoid breaking changes.

src/classes/repeat.ts
```diff
@@ -146,29 +149,56 @@ export class Repeat extends QueueBase {
     repeat: RepeatOptions,
     jobId?: string,
   ): Promise<number> {
-    const repeatJobKey = getRepeatKey(name, { ...repeat, jobId });
-    const repeatJobId = this.getRepeatJobId({
+    const optionsConcat = getRepeatKey(name, { ...repeat, jobId });
+    const repeatJobKey = repeat.key ?? this.hash(optionsConcat);
+    const oldRepeatJobId = this.getRepeatJobId({
```
manast, line 154: Maybe the word "legacy" instead of "old" makes this more explicit.

src/classes/repeat.ts
manast: As it is returning a concatenation of options, maybe getRepeatCocatOptions ?

src/classes/repeat.ts
manast: In the edge case if it adds a delayed job with a different jobId but the same delay time, then you will end up having 2 reputable jobs at least in one iteration, but maybe this edge case cannot happen in practice.

src/classes/repeat.ts
roggervalf: sounds good

src/classes/repeat.ts
```diff
@@ -70,23 +75,28 @@ export class Repeat extends QueueBase {
...
-        repeatableExists = !!(await client.zscore(
-          this.keys.repeat,
-          repeatJobKey,
-        ));
-      }
       const { immediately, ...filteredRepeatOpts } = repeatOpts;
 
       // The job could have been deleted since this check
-      if (repeatableExists) {
+      if (repeatJobKey) {
```
manast, line 99: The comment ```The job could have been deleted since this check``` is a bit confusing now since the "repeatableExists" boolean is not used anymore.

manast (commented):
LGTM, however, it would be important to improve the documentation, specially to give answers to the issues https://github.com/taskforcesh/bullmq/issues/2399 https://github.com/taskforcesh/bullmq/issues/2596

EinfachHans:
@roggervalf i really need some assistance to this change, because for me i think it is a **breaking change**?

currently using `5.9.0` and dispatching events with a unique key and my pattern + timezone. I also set the `name` to the same as the key. `bullmq` then adds one entry into my redis, where the name is like `bull:{{queue}}:repeat:{{key}}:{{someHash}}`. It also sets a `rjk` with the key set, followed by pattern and timezone in your old format.

To be able to easily remove the job again i use the following logic:
- Get all job ids by calling redis client directly with the pattern: `bull:{{queue}}:repeat:{{key}}:*` and stripe the first part so that only the id's are left
- With that id i call `getJob` on the queue instance
- Then i call `removeRepeatableByKey` with the rjk from the previously loaded job

**Now i migrated to `5.10.0`**

What i experienced is:
1. When adding a job, the `rjk` now only is the key i put into, not added any pattern information. This is great, because it easily allows me to remove it by that key
2. `bullmq` seems to add *two* entries into the redis atm. One as before and one where the name has no hash added and with less information. Is this indeed? Because this doesn't match with my current delete implementation.

But the main reason why i think this is an breaking change is that after a update i will have two kind ob jobs: Previous already existing job's where the `rjk` has the old format and the new one's. The new one'c can be easily removed with the `removeRepeatableByKey`, the old one's not.

First i thought maybe an job will then have the new `rjk` format **after** it is processed once and the successor event is created but this seems not to be the case.

I hope i described everything clear, if not, please let me know. Can you please help me out with this and how to handle this change in the implementation?

Issue #2650, opened 2024-07-14
[Bug]: 


elyran:
### Version

v5.8.7

### Platform

NodeJS

### What happened?

`Worker` causes Jest to hang 30 seconds after the test ends.

### How to reproduce.

1. Spin up a Redis instance: `docker run --rm -p 30001:6379 redis:7.2`
2. Run the following Jest test with: `node --inspect ./node_modules/jest/bin/jest.js --runInBand src/queue/BullMQ.test.ts`

The content of `BullMQ.test.ts`:
```
import { Worker } from 'bullmq';
const QUEUE_NAME = 'queue1';

describe(__filename, () => {
  it(`performs operations on a queue`, async () => {
    const worker = new Worker(
      QUEUE_NAME,
      async (job) => {
        // TODO: Handle the job.
      },
      {
        connection: {
          host: 'localhost',
          port: 30001,
        },
      },
    );
    await worker.waitUntilReady();

    // TODO: Add a job to the queue.

    await worker.close();
  });
});
```

### Relevant log output

```shell
ken@ken-pc:~/tmp/proj1$ node --inspect ./node_modules/jest/bin/jest.js --runInBand src/queue/BullMQ.test.ts
Debugger listening on ws://127.0.0.1:9229/7d6260b6-bc10-4045-9f33-86ac62ebb6b2
For help, see: https://nodejs.org/en/docs/inspector
 PASS  src/queue/BullMQ.test.ts
  /home/ken/tmp/proj1/src/queue/BullMQ.test.ts
    ✓ performs operations on a queue (21 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        2.537 s, estimated 3 s
Ran all test suites matching /src\/queue\/BullMQ.test.ts/i.
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
Done in 33.17s.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

elyran:
Please notice the timing. It's 30 seconds. What could it be? Thank you

manast:
By default ```worker.close``` will wait for the current jobs to complete, maybe your jobs are not completing in the test, or taking too much time?

elyran:
The reproduction code I've added to the issue is the exact copy / paste
[...]

PR #2656, opened 2024-07-16
fix(worker): fix close sequence to reduce risk for open handlers

manast:
Fixes or mitigates https://github.com/taskforcesh/bullmq/issues/2650

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -298,7 +298,7 @@ export class RedisConnection extends EventEmitter {
     return client.connect();
   }
 
-  async close(): Promise<void> {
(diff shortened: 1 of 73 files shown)


roggervalf (approved):
Lgtm

src/classes/redis-connection.ts
```diff
@@ -310,7 +310,7 @@ export class RedisConnection extends EventEmitter {
           await this.initializing;
         }
         if (!this.shared) {
-          if (status == 'initializing') {
+          if (status == 'initializing' || force || process.env.CI) {
```
roggervalf, line 313: better to not add a process.env value here, in case users use the same name

PR #2658, opened 2024-07-17
fix(repeatable): consider removing legacy repeatable job

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/2661


manast (approved):
LGTM

PR #2666, opened 2024-07-20
feat(job): allow passing a debounce as option

roggervalf:


src/commands/addDelayedJob-6.lua
```diff
@@ -86,10 +88,20 @@ else
     end
 end
 
+local debounceId = opts['debo'] and opts['debo']['id']
```
manast, line 91: This code should be refactored as it is exactly the same on 4 different places.

src/commands/addDelayedJob-6.lua
```diff
@@ -86,10 +88,20 @@ else
...
+if debounceId then
+  local debouncedJobKey = debounceJob(args[1], debounceId, opts['debo']['ttl'],
+    jobId, eventsKey, maxEvents)
+  if debouncedJobKey then
+    return debouncedJobKey
+  end
+end
+
 -- Store the job.
 local delay, priority = storeJob(eventsKey, jobIdKey, jobId, args[3], ARGV[2],
```
manast, line 102: as we are storing the opts with the debounce options, why do we need to pass the debounceId again?

src/commands/includes/debounceJob.lua
```diff
@@ -0,0 +1,20 @@
+--[[
+  Function to debounce a job.
+]]
+
+local function debounceJob(prefixKey, debounceId, ttl, jobId, eventsKey, maxEvents)
```
manast, line 5: I think it is more Redis friendly to pass the debounceKey as a command key instead of building it inside lua, as all the components of the key are known before calling this script.

src/classes/job.ts
```diff
@@ -323,6 +330,10 @@ export class Job<
       job.repeatJobKey = json.rjk;
     }
 
+    if (json.deid) {
+      job.debounceId = json.deid;
```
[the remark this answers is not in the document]
roggervalf: I didn't save all the option because ttl is not reuse as debounceId, similar to repeatable jobs, we only save repeatJobKey instead of saving all repeat options in job instance.

src/commands/addDelayedJob-6.lua
roggervalf: yeah I can refactor it

src/commands/includes/removeJob.lua
```diff
@@ -3,11 +3,15 @@
 ]]
 
 -- Includes
+--- @include "removeDebounceKey"
 --- @include "removeJobKeys"
 --- @include "removeParentDependencyKey"
 
-local function removeJob(jobId, hard, baseKey)
+local function removeJob(jobId, hard, baseKey, shouldRemoveDebounceKey)
```
manast, line 10: In general, I think passing booleans as flags is worse than passing an object with optional flags in it, because the code is more legible, however, it would also be slower... so we can keep it like this I guess. But we should add a comment when used, for example:
```lua
    removeJob(key, hard, baseKey, true --[[remove debounce key]]--)
```

manast (commented):
Just a thought. What happens if you have a job with debounce id and some TTL, and the job fails but you have retries? should the TTL reset in this case?

roggervalf:
> Just a thought. What happens if you have a job with debounce id and some TTL, and the job fails but you have retries? should the TTL reset in this case?

No in case of fixed I think, as It's name implies to keep it for that periodo of time. Better to use extended mode if they want to keep it in retries

manast (approved):
LGTM

jamesholcomb:
👏 👏 👏 wow almost 6 years from the original https://github.com/OptimalBits/bull/issues/1034 but happy to finally see this!

PR #2682, opened 2024-08-01
fix(flow): fail parent on failure by default

roggervalf:
our current behavior keeps parent in waiting-children state when regular children fail, instead we should fail parent directly as regular children must be considered as hard dependencies. For soft dependencies we have other options like removeDependencyOnFailure or ignoreDependencyOnFailure.
fixes https://github.com/taskforcesh/bullmq/issues/800


manast (approved):
LGTM, but as we are changing a default behaviour this would indeed be a breaking change.

PR #2683, opened 2024-08-01
fix(python): make sure json.dumps return JSON compliant JSON (python)

saulane:
The Python version can return a non-JSON-compliant value to Redis, which breaks when trying to retrieve the value using the Node version of the library.

By adding "allow_nan=True", we raise an error before sending the value to Redis, which prevents having bad JSON in the job's return value.


roggervalf:
hi @saulane, thanks for this fix. Could you also add a test case where it would fail without this change?

saulane:
> hi @saulane, thanks for this fix. Could you also add a test case where it would fail without this change?

@roggervalf  done

roggervalf (approved):
Lgtm

PR #2710, opened 2024-08-15
fear: add onChildFailure option

roggervalf:


src/commands/moveToFinished-14.lua
```diff
@@ -126,12 +114,13 @@ if rcall("EXISTS", jobIdKey) == 1 then -- // Make sure job exists
     -- 2) move the job Id to parent "processed" set
     -- 3) push the results into parent "results" list
     -- 4) if parent's dependencies is empty, then move parent to "wait/paused". Note it may be a different queue!.
-    if parentId == "" and parentKey ~= "" then
-        parentId = getJobIdFromKey(parentKey)
-        parentQueueKey = getJobKeyPrefix(parentKey, ":" .. parentId)
```
roggervalf, line 131: this logic is not needed anymore as all parent data is saved in parent attribute

src/commands/includes/moveParentIfNeeded.lua
```diff
@@ -0,0 +1,62 @@
...
+]]
+
+-- Includes
+--- @include "moveParentFromWaitingChildrenToFailed"
+--- @include "moveParentToWaitIfNeeded"
+
+local function moveParentIfNeeded(parentData, parentKey, jobIdKey,
+    failedReason, timestamp)
+    if parentData['ocf'] then
+        if parentData['ocf'] == 'fail' then
```
roggervalf, line 12: we can possible have a mapping like:
```
fail: f
ignore: i
remove: r
wait: w
```
that way we can only compare 1 character

manast: sounds good to me

PR #2712, opened 2024-08-16
fix(flow): recursive ignoreDependencyOnFailure option

roggervalf:
We are missing to handle this option when a grandchild fails with failParentOnFailure, then if a parent has ignoreDependencyOnFailure we should consider this case.


manast (approved):
LGTM

Issue #2718, opened 2024-08-21
[Bug]: `FlowProducer#getFlow()` returns `undefined` children when colon characters are used in queue names


brncsk:
### Version

v5.12.9

### Platform

NodeJS

### What happened?

`FlowProducer#getFlow()` returns an array of `undefined` children when descendant jobs passed to `FlowProducer#add()` have `queueName`s containing colon (`:`) characters.

### How to reproduce.

```js
import { FlowProducer } from 'bullmq';

const flow = new FlowProducer();

const { job } = await flow.add({
  name: 'parent',
  queueName: 'parent',
  children: [{ name: 'child', queueName: 'parent:child' }]
});

const { children } = await flow.getFlow({
  id: job.id,
  queueName: 'parent'
});

console.log(children); // → [undefined]
```

It's probably this line:

https://github.com/taskforcesh/bullmq/blob/c60e6a3f45606c623ef6a2b16415d67eac9b0269/src/classes/flow-producer.ts#L419

...that's causing the problem.

Splitting/joining by `:` happens in a lot of places so it'd be nice to issue a warning if the user happens to choose a queue name or other identifier that contains a colon and might end up in those places to beware of the consequences.

Somewhat similar/related: https://github.com/taskforcesh/bullmq/issues/1334

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

brncsk:
Thanks for the rapid fix, folks!

PR #2719, opened 2024-08-22
fix(flows): throw error when queueName contains colon

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2718


manast (approved):
lgtm

PR #2721, opened 2024-08-22
feat(queue): observability for queue

fgozdz:
For the record, we are trying to adhere to the following guidelines: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#conventions


src/enums/opentelemetry-attributes.ts
```diff
@@ -0,0 +1,18 @@
+export enum OpenTelemetryAttributes {
```
manast, line 1: As this interface is generic, it should be called just TelemetryAttributes

src/interfaces/opentelemetry.ts
```diff
@@ -0,0 +1,25 @@
+export interface Telemetry {
```
manast, line 1: opentelemetry.ts -> telemetry.ts

src/interfaces/opentelemetry.ts
```diff
@@ -0,0 +1,25 @@
...
+  tracer: Tracer;
+}
+
+export interface Tracer {
+  startSpan(name: string): Span;
```
manast, line 6: I wonder if this method startSpan wouldn't need to be asynchronous?

manast: (same with end() method)

src/classes/queue.ts
```diff
@@ -220,18 +225,33 @@ export class Queue<
     data: DataType,
     opts?: JobsOptions,
   ): Promise<Job<DataType, ResultType, NameType>> {
+    let span;
+    if (this.tracer) {
+      const spanName = `${this.name}.${name} Queue.add`;
+      span = this.tracer.startSpan(spanName);
+      span.setAttributes({
+        [OpenTelemetryAttributes.QueueName]: this.name,
```
roggervalf, line 233: there are other attributes that can be added like:
```
span.setAttributes({
        [SemanticAttributes.MESSAGING_SYSTEM]: 'BullMQ', // it could be a constant
         [SemanticAttributes.MESSAGING_DESTINATION]: this.name,
         [OpenTelemetryAttributes.JOB_NAME]: name
});
```
You can use this package https://www.npmjs.com/package/@opentelemetry/semantic-conventions for semantic conventions

src/classes/queue.ts
```diff
@@ -250,6 +270,11 @@ export class Queue<
         },
       );
       this.emit('waiting', job);
+
+      if (this.tracer) {
```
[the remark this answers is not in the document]
roggervalf: id won't be available as traces are happening before addition

src/classes/queue.ts
```diff
@@ -264,7 +289,19 @@ export class Queue<
   addBulk(
     jobs: { name: NameType; data: DataType; opts?: BulkJobOptions }[],
   ): Promise<Job<DataType, ResultType, NameType>[]> {
-    return this.Job.createBulk<DataType, ResultType, NameType>(
+    let span;
+    if (this.tracer) {
+      const jobsInBulk = jobs.map(job => job.name);
```
roggervalf, line 294: it seems like there is a way to span different values when doing bulk operation https://github.com/appsignal/opentelemetry-instrumentation-bullmq/blob/main/src/instrumentation.ts#L285-L325

src/classes/queue.ts
```diff
@@ -299,12 +356,26 @@ export class Queue<
    *
    */
   async close(): Promise<void> {
+    let span;
+    if (this.tracer) {
+      const spanName = `${this.name} Queue.close`;
```
roggervalf, line 361: no sure if we need this span for now :thinking:. Maybe we can start with add operations and then expand with other values while receiving more requests.

src/interfaces/opentelemetry.ts
roggervalf: looks like it's synchronous so no need to use async

roggervalf (commented):
There are few other package/repos where you can base like:
https://github.com/appsignal/opentelemetry-instrumentation-bullmq
https://github.com/jenniferplusplus/opentelemetry-instrumentation-bullmq
Where authors were trying to follow openTel conventions

manast:
One thing that I forgot to mention, the interface should be properly documented using typedoc syntax, similarly as we document the rest of the API: https://typedoc.org/example/

src/classes/queue.ts
fgozdz: I will leave it for now i think and delete if it were to cause some problems

roggervalf: still I think to expose few metrics for now as needed, I don't see that the other packages are exposing so many. If we expose them now and then we want to remove, we will wait until a next breaking change. could you consider to spans same methods as in these packages https://github.com/taskforcesh/bullmq/pull/2721#pullrequestreview-2256324791 that looks like people are using now

manast: My two cents: maybe close in particular is not so useful, however I do not think we need to implement only the same methods as in the existing packages, we can make it more powerful right in the first version. In the future we may even be able to make it configurable, which methods and which not to add spans to.

src/classes/queue.ts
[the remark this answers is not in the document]
manast: @roggervalf yes, but it is an attribute that can be added to the span using setAttributes

src/classes/worker.ts
```diff
@@ -507,10 +537,14 @@ export class Worker<
       }
 
       this.running = false;
-      return asyncFifoQueue.waitAll();
+      return await asyncFifoQueue.waitAll();
     } catch (error) {
       this.running = false;
       throw error;
+    } finally {
```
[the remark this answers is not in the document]
manast: Also, in the case of error, shouldn't we store this information in the span? just wondering here... how does the current open telemetry for BullMQ package handle exceptions?

src/classes/worker.ts
```diff
@@ -1017,6 +1166,17 @@ will never work with more accuracy than 1ms. */
...
 
   protected async extendLocks(jobs: Job[]) {
+    let span;
+    if (this.tracer) {
+      const spanName = `${this.name} ${this.id} Worker.extendLocks`;
+      span = this.tracer.startSpan(spanName);
+      span.setAttributes({
+        [TelemetryAttributes.WorkerName]: this.name,
+        [TelemetryAttributes.WorkerId]: this.id,
+        [TelemetryAttributes.WorkerJobsInvolved]: JSON.stringify(jobs),
```
manast, line 1176: wouldn't we also want to store all the IDs of the jobs whose locks we are extending?

src/classes/worker.ts
```diff
@@ -520,12 +554,30 @@ export class Worker<
...
+    }
+
+    const nextJob = await this._getNextJob(
       await this.client,
       await this.blockingConnection.client,
       token,
       { block },
     );
+
+    if (this.tracer) {
```
manast, line 576: In order to match the job that was added to the queue we also need to specify the job.id here in the span attributes.

src/classes/worker.ts
```diff
@@ -1065,6 +1239,10 @@ will never work with more accuracy than 1ms. */
     }
 
     this.notifyFailedJobs(await Promise.all(jobPromises));
+
```
manast, line 1242: I think we need to store the ids of the failed/stalled jobs here because they need to be mapped to the jobs added to the queue so that we can trace all the way of a job added to the queue to the place where they complete/fail/stall

src/classes/queue-base.ts
```diff
@@ -175,4 +202,59 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+  /**
+   * Wraps the code with telemetry and provides span for configuration.
+   *
+   * @param spanType - type of the span: Producer, Consumer, Internal
+   * @param getSpanName - name of the span
+   * @param callback - code to wrap with telemetry
+   * @returns
+   */
+  protected trace<T>(
+    getSpanType: () => SpanKind,
```
manast, line 215: We do not need to have a function for getting the span type here, the reason for having it for the span name is because as the span name usually requires some computation (like concatenating strings, etc), by having it as a callback it does not perform those computations if telemetry is not enabled. For the spanType we are just passing a const, so this will not impact performance in any way.

src/classes/queue-base.ts
```diff
@@ -175,4 +202,59 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+    });
+
+    try {
+      if (activeTelemetryHeaders) {
+        const activeContext = this.propagation.extract(
+          this.contextManager.active(),
+          activeTelemetryHeaders,
+        );
+
+        return this.contextManager.with(activeContext, () => callback(span));
```
manast, line 242: I assume contextManager.with will (for the Otel case), create a parent-child relationship for the spans?

manast, line 242: Also, in this case we need to call it like awaiting like this:
```ts
        return await this.contextManager.with(activeContext, () => callback(span));
```
The reason being that we want to catch potential exceptions thrown by the callback, without this await the exception will just bubble up and we loose the ability to record the exception in the tracer.

src/classes/queue-base.ts
```diff
@@ -175,4 +202,59 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+      }
+
+      const telemetryHeaders: Record<string, string> = {};
+
+      this.propagation.inject(this.contextManager.active(), telemetryHeaders);
+
+      return this.contextManager.with(
```
manast, line 249: same issue here we need to await before returning.

src/classes/queue.ts
```diff
@@ -220,38 +221,55 @@ export class Queue<
     data: DataType,
     opts?: JobsOptions,
   ): Promise<Job<DataType, ResultType, NameType>> {
-    if (opts && opts.repeat) {
-      if (opts.repeat.endDate) {
-        if (+new Date(opts.repeat.endDate) < Date.now()) {
-          throw new Error('End date must be greater than current timestamp');
+    return this.trace<Job<DataType, ResultType, NameType>>(
+      () => 3,
```
manast, line 225: We should always use enums or constants instead of magic numbers.

src/classes/queue-base.ts
```diff
@@ -175,4 +202,59 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+          this.contextManager.active(),
+          activeTelemetryHeaders,
+        );
+
+        return this.contextManager.with(activeContext, () => callback(span));
+      }
+
+      const telemetryHeaders: Record<string, string> = {};
+
+      this.propagation.inject(this.contextManager.active(), telemetryHeaders);
```
manast, line 247: Ok, so we are going to always inject the headers, hmm, we may need to give it a bit more of thought, not sure this is 100% correct.

src/classes/queue.ts
```diff
@@ -220,38 +221,55 @@ export class Queue<
...
+      () => `${this.name}.${name} Queue.add`,
+      async (span, telemetryHeaders) => {
+        if (telemetryHeaders) {
+          data = { ...data, telemetryHeaders };
```
manast, line 229: The telemetry data should be stored in the options, as the data is user data which we should not tamper with.

src/classes/queue.ts
```diff
@@ -261,20 +279,34 @@ export class Queue<
...
+          [TelemetryAttributes.BulkCount]: jobs.length,
+        });
+
+        return await this.Job.createBulk<DataType, ResultType, NameType>(
+          this as MinimalQueue,
+          jobs.map(job => ({
+            name: job.name,
+            data: {
+              ...job.data,
+              ...(span && telemetryHeaders),
```
manast, line 300: Should go on the options object.

src/classes/queue.ts
manast: we probably need to define a "telemetryData" or similar optional option in JobOpts with type "any" so that we can store the telemetry data there.

src/interfaces/telemetry.ts
```diff
@@ -0,0 +1,108 @@
...
+export interface Propagation {
+  inject<T>(context: Context, carrier: T, setter?: TextMapSetter): void;
+  extract<T>(context: Context, carrier: T): Context;
+}
+
+interface TextMapSetter {
+  get<T>(carrier: T, key: string): undefined | string | string[];
+}
+
+export interface JobDataWithHeaders {
```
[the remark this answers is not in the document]
manast: or maybe a better name: ```telemetryPropagation``` not sure.

src/classes/queue.ts
```diff
@@ -290,21 +320,35 @@ export class Queue<
    * and in that case it will add it there instead of the wait list.
    */
   async pause(): Promise<void> {
-    await this.scripts.pause(true);
-    this.emit('paused');
+    await this.trace<void>(
+      SpanKind.PRODUCER,
```
manast, line 324: In this case I do not think the kind is PRODUCER, as we are just pausing the queue, not producing any jobs. Probably INTERNAL is the correct one, not sure.

src/classes/queue.ts
```diff
@@ -290,21 +320,35 @@ export class Queue<
...
    */
   async close(): Promise<void> {
-    if (!this.closing) {
-      if (this._repeat) {
-        await this._repeat.close();
-      }
-    }
-    return super.close();
+    await this.trace<void>(
+      SpanKind.PRODUCER,
```
manast, line 340: Same here, should not be PRODUCER

src/classes/queue.ts
```diff
@@ -313,8 +357,15 @@ export class Queue<
    * queue.
    */
   async resume(): Promise<void> {
-    await this.scripts.pause(false);
-    this.emit('resumed');
+    await this.trace<void>(
+      SpanKind.PRODUCER,
```
manast, line 361: Same, etc.

src/classes/queue.ts
```diff
@@ -569,8 +717,18 @@ export class Queue<
    * @param maxLength -
    */
   async trimEvents(maxLength: number): Promise<number> {
-    const client = await this.client;
-    return client.xtrim(this.keys.events, 'MAXLEN', '~', maxLength);
+    return this.trace<number>(
+      SpanKind.PRODUCER,
```
manast, line 721: I am doubting if it is correct or necessary to have trace on all these methods actually. Because the most important duty of the telemetry is to be able to trace a job from its creation to its completion. But sure there may be situations where a job gets removed manually and things like that, probably we want to trace them too. But "pausing" a queue, or "trimEvents", it does not look like this is something we want to have spans for.

src/classes/queue.ts
```diff
@@ -424,7 +510,18 @@ export class Queue<
     jobId: string,
     progress: number | object,
   ): Promise<void> {
-    return this.scripts.updateProgress(jobId, progress);
+    await this.trace<void>(
+      SpanKind.PRODUCER,
```
manast, line 514: Updating a job progress should somehow also map to the span that created the job. Also the SpanKind cannot be PRODUCER.

tests/test_telemetry_interface.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    spanContext(): SpanContext {
+      return { traceId: 'mock-trace-id', spanId: 'mock-span-id' };
+    }
+
+    end(): void {}
+  }
+
+  class MockPropagation implements Propagation {
```
manast, line 133: Should not be needed as Propagation is OTel implementation specific

tests/test_telemetry_interface.ts
```diff
@@ -0,0 +1,303 @@
...
+
+      const activeContext = telemetryClient.contextManager.active();
+      activeContext['getSpan'] = () => span;
+
+      try {
+        await queue.add('testJob', { foo: 'bar' });
+
+        expect.fail('Expected an error to be thrown');
+      } catch (error) {
+        span.recordException(error);
```
manast, line 206: We should not need to call recordException here as this is supposed to be called inside queue.add when an exception occurs.

manast: I see, the issue you have here is that you are throwing an exception as soon as queue.add is called, but by doing so we are not really testing the try/catch inside ```queue.add``` So you could try to throw an error when calling ```Job.create``` instead which is called inside ```queue.add```.

tests/test_telemetry_interface.ts
```diff
@@ -0,0 +1,303 @@
...
+      ];
+
+      const addBulkStub = sinon
+        .stub(queue.Job, 'createBulk')
+        .rejects(new Error('Simulated bulk error'));
+
+      const span = telemetryClient.trace
+        .getTracer('testtracer')
+        .startSpan('Queue.addBulk.error') as MockSpan;
+      const recordExceptionSpy = sinon.spy(span, 'recordException');
```
manast, line 253: Same comment as above, in order for this test to be meaningful it should throw an exception from inside addBulk.

manast (commented):
Looks good! But please address the new and previous comments.

src/classes/queue-base.ts
```diff
@@ -76,6 +96,13 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
     this.keys = queueKeys.getKeys(name);
     this.toKey = (type: string) => queueKeys.toKey(name, type);
     this.setScripts();
+
+    if (opts?.telemetry) {
+      this.tracer = opts.telemetry.trace.getTracer(opts.telemetry.tracerName);
```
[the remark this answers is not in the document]
manast: You could just keep a "telemetry" property instead:
```ts
this.telemetry = opts.telemetry
```

src/classes/queue.ts
```diff
@@ -557,10 +695,20 @@ export class Queue<
    * @returns
    */
   async promoteJobs(opts: { count?: number } = {}): Promise<void> {
-    let cursor = 0;
-    do {
-      cursor = await this.scripts.promoteJobs(opts.count);
-    } while (cursor);
+    await this.trace<void>(
+      3,
```
manast, line 699: Make sure to change the 3 to KindSpan enum.

src/classes/worker.ts
```diff
@@ -402,116 +403,127 @@ export class Worker<
...
-
-      const asyncFifoQueue = (this.asyncFifoQueue =
-        new AsyncFifoQueue<void | Job<DataType, ResultType, NameType>>());
-
-      let tokenPostfix = 0;
-
-      const client = await this.client;
-      const bclient = await this.blockingConnection.client;
+    await this.trace<void>(
+      SpanKind.CONSUMER,
```
manast, line 407: I do not think this is the correct type, as this span is not consuming but starting the main worker loop.

src/classes/worker.ts
```diff
@@ -520,11 +532,26 @@ export class Worker<
    * @returns a Job or undefined if no job was available in the queue.
    */
   async getNextJob(token: string, { block = true }: GetNextJobOptions = {}) {
-    return this._getNextJob(
-      await this.client,
-      await this.blockingConnection.client,
-      token,
-      { block },
+    return this.trace<Job<DataType, ResultType, NameType>>(
+      SpanKind.CONSUMER,
```
manast, line 536: I think this should still be kind INTERNAL, because I do not think we should have more than one kind CONSUMER per job, and in this case getNextJob would lead to processJob which also has kind CONSUMER

src/classes/worker.ts
```diff
@@ -745,101 +783,137 @@ will never work with more accuracy than 1ms. */
     fetchNextCallback = () => true,
     jobsInProgress: Set<{ job: Job; ts: number }>,
   ): Promise<void | Job<DataType, ResultType, NameType>> {
-    if (!job || this.closing || this.paused) {
-      return;
-    }
+    const { telemetryMetadata: dstPropagationMedatada } = job.opts;
```
manast, line 786: Shouldn't this actually be the "source propagation metadata" ?

src/classes/worker.ts
```diff
@@ -745,101 +783,137 @@ will never work with more accuracy than 1ms. */
...
+
+    return this.trace<void | Job<DataType, ResultType, NameType>>(
```
manast, line 788: Shouldn't we need to pass the source propagation metadata as last parameter here so that it maps the job added by the PRODUCER?

manast: As the job can be completed or failed in this case, shouldn't we mark the span with this information somehow? Could you please check the old Otel implementation to see how completed and failed jobs are handled?

src/classes/queue-base.ts
```diff
@@ -175,4 +193,66 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+        srcPropagationMetadata,
+      );
+    }
+
+    const span = this.tracer.startSpan(
+      getSpanName(),
+      {
+        kind: spanKind,
+      },
+      currentContext,
```
manast, line 229: If srcPropagationMetadata is not defined, nor will currentContext, is this correct?

src/classes/worker.ts
```diff
@@ -779,101 +817,137 @@ will never work with more accuracy than 1ms. */
...
-      if (!this.connection.closing) {
-        const completed = await job.moveToCompleted(
-          result,
-          token,
-          fetchNextCallback() && !(this.closing || this.paused),
-        );
-        this.emit('completed', job, result, 'active');
-        const [jobData, jobId, limitUntil, delayUntil] = completed || [];
-        this.updateDelays(limitUntil, delayUntil);
+        if (!job || this.closing || this.paused) {
```
manast, line 832: I think this is wrong, if there is no job, we should not create a span here. It is common that there is no job to process when this function is called, so this should be left as it was, i.e. returning before creating a new span.

src/classes/worker.ts
manast: Please check the above comment I think you missed it before.

src/classes/worker.ts
```diff
@@ -779,101 +817,137 @@ will never work with more accuracy than 1ms. */
...
-    }
+    this.trace<void>(
+      SpanKind.INTERNAL,
+      () => `${this.name} ${this.id} Worker.resume`,
+      span => {
+        span?.setAttributes({
+          [TelemetryAttributes.WorkerId]: this.id,
+        });
+
+        if (this.resumeWorker) {
```
manast, line 945: Same here, we should not create a span if we are not resuming (as this check if used to see if the worker is paused to begin with).

src/classes/worker.ts
```diff
@@ -907,42 +981,54 @@ will never work with more accuracy than 1ms. */
...
-      ];
-
-      // Run cleanup functions sequentially and make sure all are run despite any errors
-      for (const cleanup of asyncCleanups) {
-        try {
-          await cleanup();
-        } catch (err) {
-          this.emit('error', <Error>err);
+  async close(force = false): Promise<void> {
+    await this.trace<void>(
```
manast, line 985: We should not need to create a span if we are already closing (as somebody already called this method before and is wrongly calling it again).

src/classes/worker.ts
```diff
@@ -958,20 +1044,32 @@ will never work with more accuracy than 1ms. */
    * @see {@link https://docs.bullmq.io/patterns/manually-fetching-jobs}
    */
   async startStalledCheckTimer(): Promise<void> {
-    if (!this.opts.skipStalledCheck) {
```
manast, line 961: same here, do not create a span if we are skipping the stalled checker. The spans are not just for wrapping methods, but if there is actually something useful to do in them! :)

manast (commented):
I wrote some comments, please do not forget to address the older comments as well.

src/classes/queue-base.ts
fgozdz: yes, startSpan method will be called without existing context, later it is bond to a new active context with setSpan method

manast: But setSpan is only called if spanKind is of type PRODUCER.

fgozdz: I addressed it in newest commit

fgozdz (commented):
LGTM

PR #2730, opened 2024-08-24
perf: remove old marker logic

roggervalf:
No need to support old marker logic removal as markers are living in a different key now in recent versions
Also we should validate if a jobId contains : as this separator could break flow logic https://github.com/taskforcesh/bullmq/blob/c60e6a3f45606c623ef6a2b16415d67eac9b0269/src/classes/flow-producer.ts#L419

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1198,6 +1198,10 @@ export class Job<
       this.opts.sizeLimit &&
       lengthInUtf8Bytes(jobData.data) > this.opts.sizeLimit;
 
+    if (this.opts.jobId == '0' || this.opts.jobId?.includes(':')) {
+      throw new Error("JobId cannot be '0' or contain :");
+    }
+
     if (exceedLimit) {
       throw new Error(
         `The size of job ${this.name} exceeds the limit ${this.opts.sizeLimit} bytes`,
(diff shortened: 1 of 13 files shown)


manast:
What would be the proper way to migrate from a version older than v5 to v6? for instance, if we upgrade from v4, then to v5 and then to v6, there could still be markers left, and break the queue.

roggervalf:
> What would be the proper way to migrate from a version older than v5 to v6? for instance, if we upgrade from v4, then to v5 and then to v6, there could still be markers left, and break the queue.

I added a method to remove legacy markers in case someone come from an older version

manast:
> 
> I added a method to remove legacy markers in case someone come from an older version

I think we also need a code example on how to do the migration. For instance, when should this method be called, and also is there a risk that the queue gets stuck before calling it? maybe we need to do a pause/remove markers/resume cycle or something like this.

roggervalf:
> > I added a method to remove legacy markers in case someone come from an older version
> 
> I think we also need a code example on how to do the migration. For instance, when should this method be called, and also is there a risk that the queue gets stuck before calling it? maybe we need to do a pause/remove markers/resume cycle or something like this.

code example could be added in migration guide similar to what was done in previous breaking changes like https://bullmq.io/news/062123/faster-priority-jobs/ where removeDeprecatedPriorityKey was mentioned. This method is similar as it's also intended to be used one time. I also added a test case to demonstrate that queues won't get stuck and having queues in a paused state will be the best option to not process that legacy marker as a regular job.

manast (approved):
LGTM

PR #2739, opened 2024-08-27
Fix/properly update data

### LINKED ISSUE #2731: [Bug]: Sandbox `.updateData` unexpected behavior regarding job data
### Version

v5.12.6

### Platform

NodeJS

### What happened?

I cannot get `job.updateData` and `job.moveToDelayed` methods working without having them fail the first couple of times. Then when the jobs are retried (like 2 or 3 times) it magically succeeds.

I am using Sandboxed workers of course 😄.

I'm trying to utilize the Step Pattern in your docs. But the only difference is: each step "builds on top of the jobData" of previous steps. But I keep losing the fields I've added in the step before. Its like `job.updateData()` updates the data in a "delayed" fashion if it contains new keys.

### How to reproduce.

Here's a reproduction within `bullmq` tests in my forked repo: https://github.com/taskforcesh/bullmq/compare/master...tolgap:bullmq:chore/reproduce-issue-2731?expand=1

### Relevant log output

_No response_

### Code of Conduct

manast:
fixes #2731

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn npm-run-all pretty:quick lint:staged
+npx npm-run-all pretty:quick lint:staged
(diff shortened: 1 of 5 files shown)


roggervalf (approved):
Lgtm

PR #2746, opened 2024-09-01
fix(connection): allow passing connection string into IORedis

### LINKED ISSUE #1220: Feature Request: Allow to using redis connection URL to connect directly.
The `ioredis` can accept redis connection URL on the constructor parameter. But it seems the bullmq can only accept the redis options only.

I knew that I can create a redis instance before and pass it to the connection opt, but it is better to accept the redis connection URL directly.

lukashroch:
Allow option to use/pass connection URL into ioredis directly without having to create connection manually and consequently maintaining it, fixes #1220.

It does not intentionally parse the url (as per #1220 requirement) and delegates URL parsing to ioredis using appropriate constructor. Thus added tests only check established connections, leaving rest up to ioredis.

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -211,7 +211,8 @@ export class RedisConnection extends EventEmitter {
 
   private async init() {
     if (!this._client) {
-      this._client = new IORedis(this.opts);
+      const { url, ...rest } = this.opts;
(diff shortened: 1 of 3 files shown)


manast (approved):
LGTM, cannot be cleaner than this I think.

roggervalf (approved):
Lgtm

PR #2756, opened 2024-09-04
fix(pattern): do not save offset when immediately is provided

roggervalf:


src/classes/repeat.ts
```diff
@@ -340,7 +340,11 @@ export const getNextMillis = (
   });
 
   try {
-    return interval.next().getTime();
+    if(opts.immediately){
+      return interval.prev().getTime();
```
roggervalf, line 344: when immediately is provided, previous time should be provided as prevMillis

src/classes/repeat.ts
```diff
@@ -65,7 +65,7 @@ export class Repeat extends QueueBase {
     const hasImmediately = Boolean(
       (every || pattern) && repeatOpts.immediately,
     );
-    const offset = hasImmediately ? now - nextMillis : undefined;
+    const offset = (hasImmediately && every) ? now - nextMillis : undefined;
```
roggervalf, line 68: only needed when every is provided, with pattern we don't need to subtract this value as it should execute next delayed job in an specific time

src/classes/repeat.ts
manast: why is that?

roggervalf: when next getTime is returned, next delayed job won't be schedule in the first cron iteration but in the second one, so it add extra delay. In order to prevent that, this value should be less that the first millis value. Looks like returning the current time is sufficient and makes more sense as we are using immediately option, changed in last commit

manast (approved):
LGTM

PR #2762, opened 2024-09-07
docs(groups): add warning when local concurrency is set

roggervalf:


docs/gitbook/bullmq-pro/groups/local-group-concurrency.md
```diff
@@ -14,6 +14,10 @@ const groupId = 'my group';
 await queue.setGroupConcurrency(groupId, 4);
 ```
 
+{% hint style="warning" %}
+Make sure to set [Group Concurrency](concurrency.md) at worker instance level along side with local concurrency. It will work as a default concurrency for those groups that don't have a local concurrency. In this way workers will know about this feature.
```
manast, line 18: Rephrase like this:
Make sure to also set the [Group Concurrency](concurrency.md) at the worker instance level along side with the desired local concurrency, as it is required for this feature to function properly and also will act as a default concurrency value for the groups that have no local concurrency defined.

manast: I think we could also throw an exception if the local concurrency is set for a group but there is no global concurrency set.

roggervalf: yeah let me think about it, also that should be implemented in pro

PR #2766, opened 2024-09-10
fix(metrics): differentiate points in different minutes to be more accurate

roggervalf:


manast (approved):
This looks correct, first computing the 1 minute bin and then subtracting them.

PR #2769, opened 2024-09-12
perf: keep jobs in waiting list when queue is paused

roggervalf:
This PR removes the legacy code regarding the old paused queue functionality, which was based on renaming the wait list key to paused. As paused keys are now handled as a queue state, we do not need the old mechanism anymore and so the code can be removed and simplified. We need some migration steps for old queues that may be in paused status at the time of the upgrade.


docs/gitbook/guide/flows/fail-parent.md
```diff
@@ -16,13 +16,13 @@ const originalTree = await flow.add({
       name,
       data: { idx: 0, foo: 'bar' },
       queueName: 'childrenQueueName',
-      opts: { failParentOnFailure: true },
+      opts: { onChildFailure: 'fail' },
```
manast, line 19: This is a nice change.

python/bullmq/scripts.py
```diff
@@ -451,6 +462,15 @@ async def retryJobs(self, state: str, count: int, timestamp: int):
         result = await self.commands["moveJobsToWait"](keys=keys, args=args)
         return result
 
+    async def repairDeprecatedPausedKey(self, maxCount: int):
```
manast, line 465: We need proper documentation on how and when a call to this method is needed.

src/classes/job.ts
```diff
@@ -39,10 +39,8 @@ const logger = debuglog('bull');
 
 const optsDecodeMap = {
   de: 'debounce',
-  fpof: 'failParentOnFailure',
```
manast, line 42: Won't we need to keep this old mappings to not cause a data breaking change?

manast: Maybe we should introduce a "migration" mechanism here. Something like migration steps for going between versions that are run if required in Lua scripts for atomicity. But we would need to keep a version number in the meta key, adding complexity.

src/classes/job.ts
```diff
@@ -1210,27 +1208,6 @@ export class Job<
       throw new Error(`Delay and repeat options could not be used together`);
     }
 
-    if (this.opts.removeDependencyOnFailure && this.opts.failParentOnFailure) {
```
manast, line 1213: nice to be able to remove this

src/classes/queue.ts
```diff
@@ -455,6 +454,13 @@ export class Queue<
     return this.scripts.drain(delayed);
   }
 
+  /**
+   * Remove legacy markers before v5
+   */
+  removeLegacyMarkers(): Promise<void> {
```
manast, line 460: Documentation about this is also needed. Most importantly practical information in how to apply something like this in the context of a production deployment.

src/classes/queue.ts
```diff
@@ -580,4 +586,18 @@ export class Queue<
     const client = await this.client;
     return client.del(this.toKey('priority'));
   }
+
+  /**
+   * Repair deprecated paused key
+   *
+   * @param maxCount - Max quantity of jobs to be moved to wait per iteration.
+   */
+  async repairDeprecatedPausedKey(maxCount: number = 1000): Promise<void> {
```
manast, line 595: Should "repair" be the proper name or actually something like "migrate" would be more accurate?

manast (commented):
Some initial comments, more will come.

src/classes/job.ts
roggervalf: we can keep this mappings that will be only used to format options of jobs with these old values, but no migration is needed as we are evaluation old and new option in lua scripts, new jobs will use new option

src/classes/queue.ts
roggervalf: I can change it, let me do it

src/classes/queue.ts
roggervalf: I added a me migrations section for it

docs/gitbook/guide/migrations/v6.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Migration to v6
+
+Make sure to pass **skipMigrationsExecution** option in any of our instances as false in order to execute all necessary changes when coming from an older version
+
+## Migration of deprecated paused key
+
+If you have paused queues after upgrading to this version. These jobs will be moved to wait state when initializing any of our instances (Worker, Queue, QueueEvents or FlowProducer).
+
+Paused key is not longer needed as this state is already represented by queue meta key. It also improve the process of pausing or resuming a queue as we don't need to rename any key.
```
manast, line 13: by meta key -> inside the meta key

manast: "It also improves..."

docs/gitbook/guide/migrations/v6.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Remove legacy markers
```
manast, line 15: I think we are going to need more detailed information on how to safely perform the migration.

src/classes/queue-base.ts
```diff
@@ -80,9 +84,40 @@ export class QueueBase extends EventEmitter implements MinimalQueue {
...
+          if (hasPendingMigrations) {
+            if (
+              this.opts.skipMigrationsExecution ||
+              typeof this.opts.skipMigrationsExecution === 'undefined'
+            ) {
+              console.error(
+                'Queue has pending migrations. See https://docs.bullmq.io/guide/migrations',
+              );
+            } else {
+              return runMigrations(client, {
```
manast, line 107: I do not think we should silently run migrations here, this just opens a window to shoot ourselves in the foot.

src/classes/queue.ts
```diff
@@ -709,4 +709,16 @@ export class Queue<
     const client = await this.client;
     return client.del(this.toKey('priority'));
   }
+
+  /**
+   * Migrate deprecated paused key
+   *
+   * @param maxCount - Max quantity of jobs to be moved to wait per iteration.
+   */
+  async migrateDeprecatedPausedKey(maxCount = 1000): Promise<void> {
```
manast, line 718: should be moved to a migration step.

src/commands/migrateDeprecatedPausedKey-2.lua
```diff
@@ -0,0 +1,31 @@
...
+  Move paused job ids to wait state to repair these states
+
+  Input:
+    KEYS[1] paused key
+    KEYS[2] wait key
+
+    ARGV[1] count
+]]
+
+local maxCount = tonumber(ARGV[1])
```
manast, line 11: I do not understand this script. It should not be possible to have both wait and paused keys at the same time, either it is wait or paused but not both, so basically just renaming the key from paused to wait should be enough as a migration step + setting the queue status to paused.

manast (approved):
Just a small comment in the python version.

python/bullmq/scripts.py
```diff
@@ -465,6 +463,15 @@ async def retryJobs(self, state: str, count: int, timestamp: int):
         result = await self.commands["moveJobsToWait"](keys=keys, args=args)
         return result
 
+    async def migrateDeprecatedPausedKey(self, maxCount: int):
```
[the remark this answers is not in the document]
roggervalf: I'll add this logic in a following pr

Issue #2744, opened 2024-08-29
Omission of Delayed creation for RepeatJob


Feelw00:
Hi,

After prioritizing the execution of a RepeatJob through Promote, the creation of Delayed is omitted when the server restarts.

There is a RepeatJob scheduled to run every day at 09:00. However, if I promote this job to prioritize its execution through the Bull Dashboard and then restart the server, the job is not created in Delayed, resulting in an issue.

This issue arises because, due to the nature of RepeatJob, a Completed Job with the same ID already exists.

When the `getNextMillis` function in the `src/classes/repeat.ts` file calculates the next execution time, it generates an ID based on the current time without checking for existing Completed Jobs, leading to this issue

To resolve this issue, I modified the process to check for Completed Jobs using the `getCompleted` method of the `QueueGetters` class before executing the `add` method of the `Queue` class.

This allows us to retrieve the `prevMillis` value of the most recent job and include it in the `opts` object when creating a new job with the `add` method

When using RepeatJob, this issue may cause regularly scheduled jobs to be missed. I would appreciate it if you could consider adding a fix for this to the core

Here is an example code to resolve the issue.

```typescript
  // src/classes/queue.ts
  async add(
    name: NameType,
    data: DataType,
    opts?: JobsOptions,
  ): Promise<Job<DataType, ResultType, NameType>> {
    if (opts && opts.repeat) {
      if (opts.repeat.endDate) {
        if (+new Date(opts.repeat.endDate) < Date.now()) {
          throw new Error('End date must be greater than current timestamp');
        }
      }
      
      // Retrieve the prevMillis value of the most recent Completed Job using the getCompleted method.
      opts.prevMillis = opts.prevMillis || (await this.getCompleted())[0]?.opts.prevMillis || 0;

      return (await this.repeat).addNextRepeatableJob<
        DataType,
        ResultType,
        NameType
      >(name, data, { ...this.jobsOpts, ...opts }, true);
    } else {
      const jobId = opts?.jobId;

      if (jobId == '0' || jobId?.startsWith('0:')) {
        throw new Error("JobId cannot be '0' or start with 0:");
      }

      const job = await this.Job.create<DataType, ResultType, NameType>(
        this as MinimalQueue,
        name,
        data,
        {
          ...this.jobsOpts,
          ...opts,
          jobId,
        },
      );
      this.emit('waiting', job);
      return job;
    }
  }
```

Feelw00:
[...]

PR #2778, opened 2024-09-20
fix(repeatable): avoid delayed job deletion if next job already existed

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/2744 as if a repeatable job is promoted and it's not removed after completion, we can readd same repeatable. This operation removes current delayed job, but after that we are adding a delayed job that have the same jobId, as a duplicate we just ignore to add a new delayed instance. We should validate that this job exists before removing it. ref https://github.com/taskforcesh/bullmq/pull/2776


manast (approved):
LGTM

Issue #1555, opened 2022-11-29
How to set a memory limit for a sandboxed processor?


ppedziwiatr:

manast:
Currently they are implemented as spawn processes, so there is no such thing as memory configuration AFAIK.

ppedziwiatr:
Thanks for the answer. What is the advantage of using a spawned process (in comparison to worker threads)?

ppedziwiatr:
Ok, if I understand properly, they are implemented as forks, (not as spawned processes) -https://github.com/taskforcesh/bullmq/blob/master/src/classes/child-pool.ts#L114

Which is event better, because I believe that each fork has a new V8 instance created = better isolation from the "host" envrionment.

And if that's really the case - it should be possible to set the memory limit per forked process - with sth like:
```
require('child_process').fork("newProcess.js", { 
        execArgv: ['--max-old-space-size=128']
    });)
```    
https://stackoverflow.com/a/37596117

Would you consider adding the max available memory setting in the Worker configuration options?

manast:
Yeah well, a fork is also a spawn, the only difference is that it creates a new V8 for you, but it is still a new process so there is no practical difference (with spawn you will need to call the node cmd line tool instead).
But sure, we can add support for setting custom flags to child processes, the tricky part is how would such an API look like so that it is consistent with current design.

ppedziwiatr:
For me personally - the simplest solution (assuming, that this is a more of "advanced" feature) sth like this:
```js
new Worker(updateQueueName, updateProcessor, {
    concurrency: workersConfig.update,
    metrics: {
      maxDataPoints: MetricsTime.ONE_WEEK,
    },
    sandboxExecArgv: ['--max-old-space-size=128', '--foo=bar']
  });
```

would be awesome!

ppedziwiatr:
Hey, any updates? :-)

ppedziwiatr:
Hey - just asking while reviewing our internal issues ;-) - is there any timeline for implementing this feature?

ppedziwiatr:
Hey, do you plan to implement this feature?

manast:
It does not seem to be a lot of interest in this feature. However if someone creates a PR with a clean and simple implementation I do not have anything against merging it.

opexu:
i need this feature too

PR #2791, opened 2024-09-28
feat(worker-thread): allow passing Worker Options

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/1555

### DIFF
--- a/src/classes/child.ts
+++ b/src/classes/child.ts
@@ -86,7 +86,7 @@ export class Child extends EventEmitter {
         stdin: true,
         stdout: true,
         stderr: true,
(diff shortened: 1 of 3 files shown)


src/classes/child.ts
```diff
@@ -83,6 +86,7 @@ export class Child extends EventEmitter {
         stdin: true,
         stdout: true,
         stderr: true,
+        ...(this.opts.workerThreadsOptions?this.opts.workerThreadsOptions:{})
```
manast, line 89: This line is not properly formatted, missing spaces.

manast (approved):
LGTM. Just a small formatting issue.

PR #2796, opened 2024-10-03
feat: add deduplication logic

roggervalf:


docs/gitbook/guide/jobs/debouncing.md
```diff
@@ -1,66 +1,66 @@
...
 
-Note that you must provide a debounce id that should represent your job. You can hash your entire job data or a subset of attributes for creating this identifier.
+Note that you must provide a deduplication id that should represent your job. You can hash your entire job data or a subset of attributes for creating this identifier.
 
 ## Extended Mode
 
-The Extended Mode takes a different approach by extending the debouncing duration until the job's completion or failure. This means as long as the job remains in an incomplete state (neither succeeded nor failed), any subsequent job with the same debouncer ID will be ignored.
+The Extended Mode takes a different approach by extending the deduplication duration until the job's completion or failure. This means as long as the job remains in an incomplete state (neither succeeded nor failed), any subsequent job with the same deduplication ID will be ignored.
```
manast, line 28: Instead of "Extended" which does not have any semantic meaning in the actual use of this feature but more about how it is implemented, I would rename it to just: "Simple Deduplication". And move it at the top, as this mode is simpler than the throttle which requires an extra argument "ttl".

roggervalf: sounds good

manast (approved):
LGTM

PR #2800, opened 2024-10-05
fix(sandbox): catch exit errors

roggervalf:
When exit errors happened too soon, promises were never resolved. This PR prevents that behavior.

### DIFF
--- a/src/classes/child-pool.ts
+++ b/src/classes/child-pool.ts
@@ -27,7 +27,7 @@ export class ChildPool {
     };
   }
 
-  async retain(processFile: string): Promise<Child> {
+  async retain(processFile: string, rejectCb: Function): Promise<Child> {
     let child = this.getFree(processFile).pop();
 
     if (child) {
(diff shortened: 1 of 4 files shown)


src/classes/child-pool.ts
```diff
@@ -40,7 +40,8 @@ export class ChildPool {
       workerForkOptions: this.opts.workerForkOptions,
       workerThreadsOptions: this.opts.workerThreadsOptions,
     });
-    child.on('exit', this.remove.bind(this, child));
```
roggervalf, line 43: in the past, we are removing this child in this handler but in sandbox file, depending on exit error we are removing or releasing. Removing this logic and handling it in sandbox

tests/test_sandboxed_process.ts
```diff
@@ -1050,7 +1092,7 @@ function sandboxProcessTests(
       });
     });
 
-    it('should remove exited process', async () => {
+    it('should release exited process', async () => {
```
roggervalf, line 1095: as my comment above, this test case changes as it's not removing but releasing

manast (approved):
LGTM.

Issue #1606, opened 2022-12-27
repeat with immediately:true at second time not work


suhaotian:
Code with below, first time will run immediately. but restart the server, add the same job to queue, no more effect with second time run:

```ts
const jobs = await queue.getRepeatableJobs();
// remove repeat jobs
await Promise.all(
  jobs.map(i => {
    return queue.removeRepeatableByKey(i.key);
  }),
);
await queue.add('task-name', {}, {
 repeat: {
      every: 1000 * 60 * 60 * 1.6,
      immediately: true,
    },
});
```

manast:
I think this works as designed. Since you are adding the same repeatable job, only the first time it will be immediate, then it will just repeat according to the "every" option.

suhaotian:
But in my mind, it's already removed by `removeRepeatableByKey` right?

tjhiggins:
@manast I agree with @suhaotian.

1. add repeatable job with `immediately: true` -> runs without delay
2. remove repeatable via removeRepeatableByKey 
3. re-add repeatable job with `immediately: true` -> runs with delay

It's not intuitive or expected imo

manast:
I created a test to specifically cover this case. It works perfectly, the problem is that you are probably getting already the second iteration of the job before you remove the job scheduler. Probably by having a larger every setting you will not be able to reproduce it anymore: https://github.com/taskforcesh/bullmq/pull/2805

PR #2805, opened 2024-10-07
test: starts immediately after remove job scheduler

manast:
Proves that there is no issue here: https://github.com/taskforcesh/bullmq/issues/1606


roggervalf (approved):
Lgtm

PR #2813, opened 2024-10-09
fix(sandbox): serialize sandboxed job error

LEI:
This change preserves the original error when throwing from a sandboxed job with a circular reference or I think a function in the object.

The TypeError is reproducible with worker threads off and a [simple circular reference](https://github.com/LEI/bullmq-sandbox):
```
Worker failed TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property 'ref' closes the circle
    at stringify (<anonymous>)
    at writeChannelMessage (node:internal/child_process/serialization:159:20)
    at target._send (node:internal/child_process:851:17)
    at target.send (node:internal/child_process:751:19)
    at ./node_modules/bullmq/dist/cjs/utils.js:134:18
    at new Promise (<anonymous>)
    at asyncSend (./node_modules/bullmq/dist/cjs/utils.js:132:12)
    at childSend (./node_modules/bullmq/dist/cjs/utils.js:152:56)
    at ChildProcessor.send (./node_modules/bullmq/dist/cjs/classes/main.js:9:57)
    at ./node_modules/bullmq/dist/cjs/classes/child-processor.js:77:28
```

The DataCloneError occurs with axios errors and `useWorkerThreads: true`, see [this branch](https://github.com/LEI/bullmq-sandbox/tree/axios):
```
Worker failed DataCloneError: function transformRequest(data, headers) {
    const contentType = headers.getContentType() || '';
    const ha...<omitted>... } could not be cloned.
    at new DOMException (node:internal/per_context/domexception:53:5)
    at ChildProcessor.send (./node_modules/bullmq/dist/cjs/classes/main-worker.js:9:69)
    at ./node_modules/bullmq/dist/cjs/classes/child-processor.js:77:28
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Initially attempted to check the value type ignoring "function" and serializing only "object", but using the stringify/parse method on the whole object seems more consistent and fixed all cases.

PR #2815, opened 2024-10-10
fix(sandbox): fix serialization of error with circular references are present

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/2813

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -212,6 +212,20 @@ export const parseObjectValues = (obj: {
   return accumulator;
 };
 
+const getCircularReplacer = (rootReference: any) => {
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -212,14 +212,28 @@ export const parseObjectValues = (obj: {
   return accumulator;
 };
 
+const getCircularReplacer = (rootReference: any) => {
+  const references = new WeakSet();
+  references.add(rootReference);
```
roggervalf, line 217: need to pass the root reference as error object is not the same as value

manast (approved):
LGTM

PR #2816, opened 2024-10-10
feat(repeat): deprecate immediately on job scheduler

manast:
This PR set the immediately option in the job scheduler as deprecated. Immediately has been a hack in its introduction, to mitigate the way "every" works, which in some cases can result in an undesired behaviour if the "every" setting is a large amount of time, lets say one month, the jobs could not be produced until many days or in worst case scenario 29 days had passed. With this PR however, immediately becomes the standard. This is a small breaking change that we hope will not affect any users, specially as this only affects the recently added "upsertJobScheduler". I think it is a good trade-off instead of having to wait for a major version.

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -71,7 +71,7 @@ export class JobScheduler extends QueueBase {
     const prevMillis = opts.prevMillis || 0;
 
     // Check if we have a start date for the repeatable job
-    const { startDate } = repeatOpts;
+    const { startDate, ...filteredRepeatOpts } = repeatOpts;
     if (startDate) {
       const startMillis = new Date(startDate).getTime();
       now = startMillis > now ? startMillis : now;
@@ -104,8 +104,6 @@ export class JobScheduler extends QueueBase {
           nextMillis,
         );
       }
-
-      const { startDate, ...filteredRepeatOpts } = repeatOpts;
 
       return this.createNextJob<T, R, N>(
         jobName,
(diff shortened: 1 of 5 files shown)


tests/test_job_scheduler.ts
```diff
@@ -744,11 +747,11 @@ describe('Job Scheduler', function () {
       { data: { foo: 'bar' } },
```
roggervalf, line 747: in previous expression, immediately option can be removed as it's not used anymore

tests/test_job_scheduler.ts
```diff
@@ -744,11 +747,11 @@ describe('Job Scheduler', function () {
...
     );
 
-    this.clock.tick(100);
+    this.clock.tick(nextTick);
```
roggervalf, line 750: we can keep this 100 milliseconds as nextTick is 2 seconds. To test that first execution is processed immediately is better to have a first delay lower than 2 seconds

src/classes/job-scheduler.ts
```diff
@@ -69,7 +69,6 @@ export class JobScheduler extends QueueBase {
     }
 
     const prevMillis = opts.prevMillis || 0;
-    now = prevMillis < now ? now : prevMillis;
 
     // Check if we have a start date for the repeatable job
     const { startDate } = repeatOpts;
```
roggervalf, line 74: what about doing:
`const { startDate, ...filteredRepeatOpts } = repeatOpts;`
in this line instead of in line 108

fgozdz (commented):
with the changes suggested by Rogger it looks good

tests/test_job_scheduler.ts
manast: In fact it can be removed completely.

PR #2822, opened 2024-10-12
feat(queue): add queue version support

manast:


src/classes/scripts.ts
```diff
@@ -220,7 +225,7 @@ export class Scripts {
     }
 
     if (<number>result < 0) {
-      throw this.finishedErrors({
+      throw finishedErrors({
```
roggervalf, line 228: finishedErrors metros is part of scripts class to have a way to override it in pro version without having to to rewrite places where is calles. If it's outside of script, new version of finishedErrors need to be passed in all places where new pro errors need to be handled

src/utils.ts
```diff
@@ -10,6 +10,10 @@ import { ChildMessage, RedisClient } from './interfaces';
 import { EventEmitter } from 'events';
 import * as semver from 'semver';
 
+import { ErrorCode } from './enums';
```
roggervalf, line 13: Looks like this reference is not used anymore

roggervalf (approved):
Lgtm, justo a minor comment

PR #2825, opened 2024-10-13
feat(queue): add option to skip metas update

manast:


src/classes/queue.ts
```diff
@@ -120,8 +120,8 @@ export class Queue<
 
     this.waitUntilReady()
       .then(client => {
-        if (!this.closing) {
-          client.hmset(this.keys.meta, this.metaValues);
+        if (!this.closing && !opts?.skipMetasUpdate) {
```
roggervalf, line 123: maybe we may need some docs in our guides about when to use this option as true or false

roggervalf (approved):
Lgtm

src/classes/queue.ts
manast: I wrote about it in the API documentation.

PR #2826, opened 2024-10-13
feat(queue): add version support

manast:
Backport of getVersion to the v4.x branch

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -13,6 +13,7 @@ import { Job } from './job';
 import { QueueGetters } from './queue-getters';
 import { Repeat } from './repeat';
 import { RedisConnection } from './redis-connection';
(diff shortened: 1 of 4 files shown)


src/classes/queue.ts
```diff
@@ -116,12 +118,8 @@ export class Queue<
...
     this.waitUntilReady()
       .then(client => {
-        if (!this.closing) {
-          client.hset(
-            this.keys.meta,
-            'opts.maxLenEvents',
-            get(opts, 'streams.events.maxLen', 10000),
-          );
+        if (!this.closing && !opts?.skipMetasUpdate) {
+          return client.hmset(this.keys.meta, this.metaValues);
```
fgozdz, line 122: HMset is marked as deprecated: https://redis.io/docs/latest/commands/hmset/

PR #2830, opened 2024-10-14
fix: min of attempts made or truncate in custom backoff

OdedShimoni:


manast:
Good catch, thanks!

OdedShimoni:
You're welcome Manuel!

PR #2834, opened 2024-10-17
fix(redis): use version for naming loaded lua scripts

manast:
If you share a Redis connection between different queues and you happen to use different versions of BullMQ, then you may be using incompatible Lua scripts, as the last loaded script wins, replacing the previously loaded one. Note that this is only a problem if 1) you are reusing a ioredis connection, and 2) you are mixing different versions of the BullMQ library, in other words, not very common in most scenarios, but this PR fixes the issue by appending a version number to every command name.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -710,11 +710,13 @@ export class Job<
           token,
           delay,
         );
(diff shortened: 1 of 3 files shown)


roggervalf (approved):
Lgtm

PR #2841, opened 2024-10-19
perf(queue): allow to add more than one base marker

roggervalf:
when using bulk operations, only one marker is added, in this case only one worker is awake. Other worker instances will be waiting for 5 seconds to pick jobs from waiting

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -4,6 +4,7 @@ import {
   BackoffOptions,
   BulkJobOptions,
   DependenciesOpts,
+  JobBulkOptions,
   JobJson,
   JobJsonRaw,
   MinimalJob,
@@ -278,6 +279,7 @@ export class Job<
       data: T;
       opts?: BulkJobOptions;
     }[],
+    opts: JobBulkOptions,
   ): Promise<Job<T, R, N>[]> {
     const client = await queue.client;
 
@@ -297,8 +299,18 @@ export class Job<
       });
     }
 
+    if (opts.markerCount > 1) {
+      const markers: (number | string)[] = [];
+      Array.from(Array(opts.markerCount - 1).keys()).forEach(index => {
+        markers.push(0, index + 1);
+      });
+      pipeline.zadd(queue.toKey('marker'), ...markers);
+    }
+
     const results = (await pipeline.exec()) as [null | Error, string][];
-    for (let index = 0; index < results.length; ++index) {
+    const jobResultLength =
+      opts.markerCount > 1 ? results.length - 1 : results.length;
+    for (let index = 0; index < jobResultLength; ++index) {
       const [err, id] = results[index];
       if (err) {
         throw err;
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -299,11 +299,18 @@ export class Job<
       });
     }
 
-    if (opts.markerCount > 1) {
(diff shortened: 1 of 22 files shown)


manast:
I see the problem, but I feel the solution is quite complex, maybe there is another way to do it that is simpler, we could for example always add a marker after every work in a bulk even if it is not really needed, that will be a bit less efficient but it is much simpler to implement as you just do it in the multi/exec pipeline... maybe there are even better solutions.

src/classes/job.ts
```diff
@@ -297,8 +299,18 @@ export class Job<
       });
     }
 
+    if (opts.markerCount > 1) {
```
manast, line 302: why not using the number of jobs instead of an option?
Also, If the jobs are standard jobs, no markers are needed at all right? this should only be needed for prioritised or if they are delayed less than the blockTime unless I am missing something.

src/classes/job.ts
```diff
@@ -297,8 +299,18 @@ export class Job<
...
+      const markers: (number | string)[] = [];
+      Array.from(Array(opts.markerCount - 1).keys()).forEach(index => {
```
manast, line 304: You can use "fill" instead: const array = Array(length).fill(0);

src/classes/job.ts
roggervalf: I added as an option to not add more matkers than necessary, imagine that you add 1000 jobs and you only have 2 workers. If add 1000 matkers workers Will consume all of them unnecessarily. Only need 2 markes one pr each worker to wake up them. Also is a good point about limiting by the type of job, this should only consider standard and prioritized jobs. Delayed matkers will be added by addDelayedJob script so I can ignore this type in that case. Standard jobs are also using matkers, this is needed for this case as well

src/classes/job.ts
roggervalf: I'll try it

src/classes/job.ts
manast: Yes, but this option will never be used correctly anyway, as it is difficult to keep track how many actual workers are there. It would be possible to do a client list and find the workers also, but this is slower and it does not work in GCP as they do not support the CLIENT LIST command (but we could ignore this as it is GCPs fault so to speak)

roggervalf: markers are also needed for standard jobs as we are not using brpoplpush anymore

roggervalf: I think I can support both options as using the number of jobs by default and markerCount as an optional parameter

manast: How fast would a call to "getWorkers" be?

manast: what about this: a worker keeps a "getWorkersCount" cached for lets say 5 minutes, or some configurable value, and based on this count it will apply the number of markers to add?

manast: the number of workers should not change particularly often anyway

manast: Something like, in "add bulk", if we hace a fresh workers count we use that one, if it is not fresh we get a new count and use that count.

roggervalf: I don't think that it's necessary to call getWorkers every time as you said, this number should not change particularly often. Maximum marker count can be handled by users in my opinion or they can avoid of passing markerCount in bulk operations.

roggervalf:
in favor of https://github.com/taskforcesh/bullmq/pull/2904

PR #2850, opened 2024-10-21
fix: proper way to get version

manast:
In NodeJS it is not possible to read reliably the version from the library package.json, so we need to rely on this other method where in release time we update the version number and hard code it in a file that we can import.


manast (commented):
LGTM, however a proper documentation chapter will be required for this, maybe as a subchapter of global events.

roggervalf (approved):
Lgtm

PR #2851, opened 2024-10-22
fix(connection): do not allow to set blockingConnection option

roggervalf:
blockingConnection is an internal property that must not be overwritten by users


manast (approved):
LGTM

PR #2853, opened 2024-10-22
feat(python): add Job.getChildrenValues

Arvinje:
Adds getChildrenValues method to the Python package's Job class, used to query returning values of child jobs

### DIFF
--- a/python/tests/job_tests.py
+++ b/python/tests/job_tests.py
@@ -6,7 +6,8 @@ https://bbc.github.io/cloudfit-public-docs/asyncio/testing.html
 
 import unittest
 
-from bullmq import Queue, Job
+from asyncio import Future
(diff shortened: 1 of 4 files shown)


python/tests/job_tests.py
```diff
@@ -103,5 +104,75 @@ async def test_promote_delayed_job(self):
 
         await queue.close()
 
+    async def test_get_children_values(self):
```
roggervalf, line 107: better to move these tests into https://github.com/taskforcesh/bullmq/blob/master/python/tests/flow_tests.py

Arvinje: good idea. done

roggervalf (approved):
Lgtm. Thank you

PR #2854, opened 2024-10-22
feat: Fanout support with redis streams

orrgal1:
Leveraging Redis Streams to add pubsub/fanout support. This is a much needed feature that begins to actually close the gap between bull and other systems which do support pubsub or fanout like rabbitmq, kafka, google pubsub, sns+sqs, etc. And the beauty of it is that you can finally get that functionality without the over complexity of most other systems.

How it works:
- Use a `StreamProducer` to produce messages to a stream. These are just barebones data, no metadata or options.
- Use a `FanoutWorker` on the other end to consume from the stream and transfer the messages to queues. At this point you can also provide a function to apply `JobsOptions` to jobs as they are added to the queues.
- From this point continue processing via queues as usual.
- The fanout worker also internally handles trimming the stream to defined time based retention for memory management.
- Also available is the `QueueToStreamWorker` to support easily fanning out from existing queues.

Love to hear what the community thinks about it becoming a feature and specifically about the approach taken here.

Thanks!


manast:
Thanks for this effort. I have been reviewing this PR and also discussed with the team. Although an interesting use-case, I think that this does not currently fit with the short and medium term goals we have for BullMQ. However, as far as I can see, this code can be packaged in its own npm module, either as free from BullMQ dependencies, or if desired using BullMQ as a dependency and subclassing and/or extending as necessary. If we would get some indications that a feature like this integrated in BullMQ is of great value for the community we could evaluate again. Please do not feel discouraged by this, just because it does not currently fit with our goals it does not mean it could be a useful functionality.

orrgal1:
No problem totally understood. As we have a very strong case for this functionality at EverC I will start by packaging it separately or simply using the fork. Thanks for the review and feedback!

orrgal1:
Hi @manast so I went ahead and packaged it into https://www.npmjs.com/package/bullmq-extra. Hope you like :)

PR #2861, opened 2024-10-23
fix: proper way to get version

manast:


roggervalf (approved):
Lgtm

manast:
Merging even though node@current (NodeJS v23) is not passing the tests due to bugs in Node itself. All LTS versions are green and those are the ones we should care about.

PR #2864, opened 2024-10-24
docs(job): clarify per-queue scoping of Job Ids

clemens:
This adds a clarification to the docs that Job Id scoping is per queue, not globally.

We came across this question in a recent PR in one of our apps and research in the docs yielded no clear explanation, which required us to dig into the code and actually experiment with it, so I thought it was worth a contribution.


roggervalf (approved):
Lgtm, thank you

PR #2878, opened 2024-10-31
fix(queue-event.ts): modify waitUntilReady to work properly in queueEvent

### LINKED ISSUE #2874: [Bug]: Event missing during QueueEvent initializing
### Version

v5.21.2

### Platform

NodeJS

### What happened?

For testing, I created queue, worker, and queueEvent objects, and used queueEvent.on to listen for events. However, I observed that about 1-2 initial events are lost in 3-4 out of 5 attempts. To address this issue, I tried using the queueEvents.waitUntilReady() function to delay until initialization, but it didn’t work as expected. Only by using delay functions like setTimeout could I get it to function properly.

After inspecting the code, I found that waitUntilReady() returns the RedisClient object before client.xread is executed in queueEvent. (The issue seems to arise because the ID is initially set to $ during initialization, but before this, a job is added, causing the latest event to update and trigger the problem.)

In the attached log below, you can see that the waiting event for Job1 is missing. Debugging also revealed that the added event was omitted.

To address this issue, I propose modifying the waitUntilReady() method in queueEvent. I believe this can be resolved by having waitUntilReady() fulfill the Promise only when xread is executed within queueEvent.

If possible, I would like to take on this issue and work on resolving it.

### How to reproduce.

Run this script

```typescript
import { Queue, QueueEvents, Worker } from "bullmq";


async function main() {

    const queue = new Queue('footprint', {
        connection: {
            host: 'localhost',
            port: 10001
        }
    });

    const queueEvents = new QueueEvents('footprint', {
        connection: {
            host: 'localhost',
            port: 10001
        }
    });


    const worker = new Worker('footprint', async (job) => {
        await job.updateProgress(42);

        console.log(`Processing job with id ${job.id}`);
        return { result: 'Success' };
    }, {
        connection: {
            host: 'localhost',
            port: 10001
        }
    });

    await Promise.all([
        queue.waitUntilReady(),
        queueEvents.waitUntilReady(),
        worker.waitUntilReady()
    ]);

    queueEvents.on("waiting", (jobId) => {
        console.log('QueueEvent: Job waiting', jobId);
    });

    queueEvents.on('completed', ({ jobId }) => {
        console.log('QueueEvent: Job completed', jobId);
    });

    queueEvents.on('drained', () => {
        console.log('QueueEvent: Queue drained');
    });

    queueEvents.on('progress', ({ jobId, data }) => {
        console.log('QueueEvent: Job progress', jobId, data);
    });

    for (let i = 0; i < 2; i++) {
        await queue.add('test', { foo: 'bar' });
    }

    await new Promise(() => {});
}

main().catch((error) => console.error(error));

```

### Relevant log output

```shell
QueueEvent: Job waiting {
  jobId: "2",
}
Processing job with id 1
QueueEvent: Job progress 1 42
QueueEvent: Job completed 1
Processing job with id 2
QueueEvent: Job progress 2 42
QueueEvent: Job completed 2
QueueEvent: Queue drained
```

### Code of Conduct

chungjung-d:
Previously, waitUntilReady functioned as soon as the Redis client was ready, but in the modified version, it has been adjusted to work when queueEvent starts reading the stream.

fix #2874


chungjung-d:
This code sets up a clientReadyPromise that just hangs out until the run method kicks in and starts reading events.

Basically, clientReadyPromise starts off in a waiting state. Then, when run actually gets going and makes its first xread call to Redis, the Promise finally resolves. This way, any code that calls waitUntilReady is paused until the queue is fully up and running. Once run is actively reading events, everything waiting on waitUntilReady gets the "all set" signal and moves on.

In other words, it’s just a way to keep things on hold until the first event comes in, making sure everything’s ready to roll before moving forward 😄

manast:
yes but it should not be needed to do this, besides you are not really waiting as the call to XREAD is blocking and the code is not awaiting it.

chungjung-d:
So, as it stands, whether XREAD reads the Redis stream or not, waitUntilReady will still act as if it's true, which I thought didn't align with the intended function and could be confusing.

Do you think waitUntilReady should remain false until XREAD is truly listening, or do you have any better suggestions on how to handle this?

chungjung-d:
@manast 
As you suggested, I've modified it to check that XREAD operates correctly before transitioning to the Ready state. If you feel this step is unnecessary, I understand, but I would appreciate it if you could consider it for the sake of making the purpose of the waitUntilReady function clearer. Thank you very much for your consideration.

roggervalf:
closing it as part of last comment in https://github.com/taskforcesh/bullmq/issues/2874. Anyway thanks for your pr @chungjung-d

PR #2879, opened 2024-11-04
feat(flows): add telemetry support

manast:


fgozdz (commented):
LGTM but maybe we should also observe close method like in the queue and worker classes

Issue #2880, opened 2024-11-04
[Bug]: Setting a concurrency doesn't change the concurrency field on the worker


nlochschmidt:
### Version

v5.22.0

### Platform

NodeJS

### What happened?

When we set a concurrency on a worker in the constructor and then set a new concurrency using the setter `worker.concurrency = 10`, then `worker.concurrency` ~is still the concurrency of the worker set in the constructor~ is `undefined`.

We can use `worker.opts.concurrency` to get the current `concurrency` but this seems like a bug.

Relevant code: 

https://github.com/taskforcesh/bullmq/blob/092019f862194acf29933faf7d1f712f1e9852b9/src/classes/worker.ts#L241

and

https://github.com/taskforcesh/bullmq/blob/092019f862194acf29933faf7d1f712f1e9852b9/src/classes/worker.ts#L387-L396

### How to reproduce.

```
const worker = new Worker('test', async job => {}, { concurrency: 100 });
worker.concurrency = 10
assert(worker.concurrency == 10)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

nlochschmidt:
Oh, actually, it seems like the value of `worker.concurrency` is also not the one from the constructor but actually `undefined`, but TypeScript doesn't complain about this:

```
const maxTasks: number = worker.concurrency
```

nlochschmidt:
Related https://github.com/microsoft/TypeScript/issues/11596#issuecomment-253603478

PR #2883, opened 2024-11-05
fix(worker): allow retrieving concurrency value

fgozdz:


manast (approved):
LGTM

src/classes/worker.ts
```diff
@@ -395,6 +395,10 @@ export class Worker<
     this.opts.concurrency = concurrency;
   }
 
+  get concurrency() {
+    return this.opts.concurrency;
```
roggervalf, line 399: I think this is a bug where this.opts.concurrency should not be set, instead it should be this.concurrency = concurrency at line 395. opts should be immutable as we threat other options like priority or delay. If you change that line, this getter may not be needed

roggervalf (changes requested):

src/classes/worker.ts
fgozdz: This is resulting in unending recursion: "RangeError: Maximum call stack size exceeded"

fgozdz:
This issue fixes #2880

src/classes/worker.ts
manast: @fgozdz what @roggervalf means is that we should not replace this.opts.concurrency as this is mutating the opts object which was passed in the constructor. Instead you can define a private property like _concurrency, and use this instead of this.opts.concurrency. In the constructor you set _concurrency to a default value if this.opts.concurrency is not defined.

manast: Honestly speaking It would be far mote elegant to have a explicit setter like "setConcurrency" instead of an implicit setter, but this would imply a breaking change, so we cannot do that now.

roggervalf: @fgozdz you can try doing:
```js
set concurrency(concurrency: number) {
    if (
      typeof concurrency !== 'number' ||
      concurrency < 1 ||
      !isFinite(concurrency)
    ) {
      throw new Error('concurrency must be a finite number greater than 0');
    }
    this._concurrency = concurrency;
  }

  get concurrency(): number {
    return this._concurrency;
  }

```
on worker constructor you may need to replace https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L241 as:
`this._concurrency = this.opts.concurrency;
`
and add a private attribute type after this one https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L182 like:
`private _concurrency:number;
`
also in order to use this new attribute you should replace this.opts.concurrency to use this._concurrency at worker level

roggervalf (approved):
Lgtm, thank you

Issue #2860, opened 2024-10-23
[Bug]: `immediately` on `upsertJobScheduler` kicks in multiple times with `pattern`.


vorillaz:
### Version

5.21.2

### Platform

NodeJS

### What happened?

According to the docs the immediately option on `upsertJobScheduler` kicks in the worker instantly once registered, but once using a cron pattern along with immediately the worker starts processing multiple times in a row, from 20 to 30 times during my tests. As an additional information, according to my tests `limit` option actually the kick in.

### How to reproduce.

Take the [docs sample](https://docs.bullmq.io/guide/job-schedulers#using-job-templates) along with a test worker:
```ts
const worker = async () => {
  console.log('workkkkkkks again');
};
```

```ts
await myQueue.upsertJobScheduler(
  "end-soon-job",
  {
    every: 60000, // every minute
    immediately: true,
  },
  {
    name: "timed-end-job",
    opts: {
      attempts: 0,
      removeOnComplete: 5,
      removeOnFail: 20,
    },
  }
);
```

Once changing the `every` property to cron pattern the worker kicks in multiple times:
```ts
await myQueue.upsertJobScheduler(
  "end-soon-job",
  {
    pattern: '* * * * *', // every minute in cron
    immediately: true,
  },
  {
    name: "timed-end-job",
    opts: {
      attempts: 0,
      removeOnComplete: 5,
      removeOnFail: 20,
    },
  }
);
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
immediately should be ignored for cron expressions, as it really does not make a lot of sense in that case. Does it work properly without it?

vorillaz:
Hey @manast 
It makes sense to use immediately for recurring jobs like data rollups that they are not so time sensitive.

`immediately` works as expected along with `every`, triggered once before the scheduled call. Upon using a corn pattern though the worker kicks in multiple times in a row. 

And that’s a regression bug in my honest opinion cause it indicates something is not properly calculated, time wise.

manast:
In fact we are planing to deprecate "immediately" in the next major version of BullMQ. Every will work as it had immediately to true all the time, as that is what users expects of every. However as cron specifies specific points in time, having an option as immediately does not make a lot of sense. If you really need it, it would be possible to implement it as a custom strategy, but it is also as simple as adding an initial job when you upsert your cron expression for the first time.

vincentrolfs:
[...]

PR #2891, opened 2024-11-09
fix(repeatable): only apply immediately in the first iteration

manast:
Fixes https://github.com/taskforcesh/bullmq/issues/2860


roggervalf (approved):
Lgtm

Issue #2876, opened 2024-10-30
[Bug]: Duplicate job scheduler tasks


noe-charmet:
### Version

5.21.2

### Platform

NodeJS

### What happened?

We use BullMQ Job Scheduler to trigger frequent tasks (multiple per second). We leverage the `schedulerId` to ensure one task is only defined once. We have however observed that on some occasions tasks are not properly deduplicated and we end up with multiple instances of the repeated task. In our specific case we trigger `upsertJobScheduler` at startup and have seen this ocure when having crash loops causing many restarts in a short period of time.

### How to reproduce.

I have written this JS script which pushes BullMQ to its limits. It goes beyond what we expect from it, but provides a reliable way to reproduce the bug that we are reporting here. Run the script and wait a few dozen seconds, you should see the counter for `waiting` tasks  go up (restarting the script seems to help too).

I've tried deep diving into the BullMQ code and debug logging at various places, but I'm yet unsuccessful in identifying the root cause of this duplication.

When there is no `Worker` involved, the duplication does not happen. I'm therefore suspecting that it could be triggered when the task is being processed by the worker when the upsert happens.

```js
const {Worker, Queue} = require('bullmq');
const {setTimeout, setInterval} = require('timers/promises');

const connection = {
  host: '127.0.0.1',
  password: 'password',
};

async function run() {
  const queue = new Queue('test', {
    connection,
  });

  await queue.waitUntilReady();

  const worker = new Worker(
    'test',
    async () => {
      await setTimeout(250);
    },
    {connection},
  );

  await worker.waitUntilReady();

  for await (const _ of setInterval(50)) {
    console.log('###');
    await queue.upsertJobScheduler('test', {
      every: 100,
    });
    console.log('active', await queue.getActiveCount());
    console.log('waiting', await queue.getWaitingCount());
    console.log('delayed', await queue.getDelayedCount());
  }
}

run().catch((error) => {
  console.error(error);
  process.exit(1);
});

```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

noe-charmet:
Just tested with the deprecated Repeatable job interface using a `key` and the issue does not happen under these conditions.

```js
const {Worker, Queue} = require('bullmq');
const {setTimeout, setInterval} = require('timers/promises');

const connection = {
  host: '127.0.0.1',
  password: 'password',
};

async function run() {
  const queue = new Queue('test', {
    connection,
  });

  await queue.waitUntilReady();

[...]

PR #2892, opened 2024-11-10
Fix/avoid hazards when upserting job schedulers

manast:
Fixes https://github.com/taskforcesh/bullmq/issues/2876


src/classes/job-scheduler.ts
```diff
@@ -268,3 +295,13 @@ export const defaultRepeatStrategy = (
     // Ignore error
   }
 };
+
+function removeUndefinedFields(obj: Record<string, any>) {
```
roggervalf, line 299: Looks like this method is not being used

roggervalf, line 299: This is a duplicated method. It's added in utils but it was not removed from here

src/classes/job.ts
```diff
@@ -464,11 +465,11 @@ export class Job<
    * @returns
    */
   asJSON(): JobJson {
-    return {
+    return removeUndefinedFields<JobJson>({
```
roggervalf, line 468: ok I get the intention, what if instead of iteration yo all these option at the end, do something like
```js
const jobJson = {
id: this id,
// all the other attributes that don't be unfefined
};

if(this parent) {
jobJson.parent = this.parent;
}
// same idea with the other atrributes
```

roggervalf (approved):
Lgtm, added some non-blocker comments

src/classes/job.ts
manast: Because it is easier to forget and it will be more code to maintain. I discovered that we have a lot of empty fields which just consume space for no reason before I added this feature.

PR #2898, opened 2024-11-12
docs: telemetry note on close force in worker

fgozdz:


src/classes/worker.ts
```diff
@@ -995,7 +995,8 @@ will never work with more accuracy than 1ms. */
    * This method waits for current jobs to finalize before returning.
    *
    * @param force - Use force boolean parameter if you do not want to wait for
-   * current jobs to be processed.
+   * current jobs to be processed. When using telemetry, be mindful that it can 
+   * interfere with the proper closure of spans, potentially leading to their disappearance.
```
manast, line 999: maybe the correct phrasing here is not "disappearance" but that the span will not be exported as it has not been closed properly.

manast (approved):

PR #2908, opened 2024-11-15
fix: removeOn booleanish

semoal:


roggervalf:
hi @semoal. Workers intentionally only support KeepJobs options, for having the effect of boolean you can pass count as 0

PR #2919, opened 2024-11-21
feat(queue): refactor a protected addJob method allowing telemetry extensions

manast:
This PR is required by packages that extend the Queue class such as BullMQ Pro so that they can customise the telemetry data when adding jobs to the queue.


roggervalf (approved):
Lgtm

Issue #2875, opened 2024-10-30
Enhance management of Job Schedulers


lpoulin:
**Is your feature request related to a problem? Please describe.**
It is a little frustrating that you cannot get all of the data back on a job scheduler that it was created with. I would like to display job schedulers in our application so they can be modified, but there doesn't appear to be a way to view the job template, so I can't populate the form with the existing values before upserting. 

**Describe the solution you'd like**
It would be nice if JobScheduler could be its own type instead of trying to squeeze it into RepeatableJob, since the class lacks all of the properties that a job scheduler has, like the job template. Perhaps the API could also support getting jobs that were produced by a job scheduler, or at least have some indication on the Job itself that it was produced by a job scheduler.

Additionally, I would like to be able query for a single job scheduler, instead of fetching all of them from a queue and then filtering on the ID. 

**Additional context**
```
{"key":"ExampleJobScheduler","name":"ExampleJobScheduler","endDate":null,"tz":null,"pattern":"*/30 * * * * *","every":null,"next":1730314440000}
```
Above is all of the information you get back when retrieving a job scheduler. How am I supposed to know what the job template looks like so I can determine if I need to make changes to it?

manast:
@lpoulin sorry for not answering sooner, we have discussed a bit about the issues you bring up, which are very valid ones. The job scheduler concept is a new concept to improve the quite bad APIs we had for repeatable jobs until now, but we had to make some compromises not to require breaking changes. We will try to address your requests in later releases.

lpoulin:
thanks for the update @manast! and thank you for considering my feedback.

roggervalf:
hi @lpoulin this feature should be available since v5.32.0

PR #2929, opened 2024-11-23
feat(queue): enhance getJobScheduler method to include template information

roggervalf:
ref https://github.com/taskforcesh/bullmq/issues/2875


src/classes/job-scheduler.ts
```diff
@@ -241,18 +248,28 @@ export class JobScheduler extends QueueBase {
...
-        tz: jobData.tz || null,
-        pattern: jobData.pattern || null,
-        every: jobData.every || null,
+        name: schedulerAttributes.name,
+        endDate: parseInt(schedulerAttributes.endDate) || null,
+        tz: schedulerAttributes.tz || null,
+        pattern: schedulerAttributes.pattern || null,
+        every: schedulerAttributes.every || null,
+        ...(schedulerAttributes.data
+          ? {
```
manast, line 266: If you are checking if data is defined here, why having a default in the next line?:
```data: JSON.parse(schedulerAttributes.data || '{}')```
Also, could it potentially be the case that you have an empty data but options?

File: tests/test_job_scheduler.ts [hunk 2]
```diff
@@ -360,6 +360,12 @@ describe('Job Scheduler', function () {
       tz: null,
       pattern: '*/2 * * * * *',
       every: null,
+      template: {
```
manast, line 363: Maybe add cases where data / opts are undefined?

src/classes/job-scheduler.ts
roggervalf: yeah I refactored a little bit to not save this data if it's empty, same for template opts

tests/test_job_scheduler.ts
roggervalf: addressed as well

manast (approved):
LGTM

Issue #247, opened 2020-07-30
Promoting a recurring job does not schedule the next job


jasonkim:
I'm not sure if this is the expected behavior.
If I were to promote a recurring job, it should run it and then ideally queue the next job.

manast:
It should work, since a recurring job is waiting as a delayed job, can you provide a test where this does not hold?

jasonkim:
Finally got around to create a simple case.

```
import ioredis from 'ioredis'
import bullmq from 'bullmq'
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
async function test() {
  const { Queue, QueueScheduler, Worker, Job } = bullmq
  const connection = new ioredis()

  // bull-board
  const { setQueues } = bullboard

  const testq = new Queue('test', {connection})
  testq.add('recurring', {}, { repeat: { cron: '* * * * *' } })
  setQueues([testq])
  new QueueScheduler('test')

  new Worker('test', async job => {
    console.log(Date())
    console.log(job.id)
  })

  await sleep(90000)
  console.log(await testq.getDelayedCount())
  const jobs = await testq.getDelayed()
  const job = jobs[0]
  // console.log(job)
  job.promote()
  console.log("Promoted")
  await sleep(1000)
  console.log(await testq.getDelayedCount())

}
test()
```

I'm using a bull-board, but you can comment that out as that is not very important.
It should print the datetime and jobid every time the worker runs, but after promotion, it stops.
I've put in sleep function, but that's not very important and just shows that the recurring job works.

Khauri:
Also noticed this issue. Promoting the job would run the process function, but the job would not be rescheduled afterwards. 
I had to work around this by essentially duplicating the data of the recurring job into a new job and adding that to the queue.

manast:
I have reproduced this issue and know exactly why it happens, but I need some time to fix it.

Khauri:
Thanks! In the meantime I've just done this and it works for my purposes:

```js
if(/^repeat:/.test(job.id)) {
  await queue.add(job.name, job.data);
} else {
  await job.promote();
}
```

But I do have a small question. When promoting a repeating job, say one that runs hourly with the next execution time being 10am, is the intended behavior to keep the scheduled time for the job as its next execution time (still 10am), or would it be to schedule it for the next hour afterwards (11am?).

manast:
@Khauri it will be scheduled to the next run.

janfietz:
[...]

Issue #2405, opened 2024-01-30
[Bug]: Repeatable job is not fired off immediately after it's removed and added back


raymondfeng:
### Version

v4.14.0

### Platform

NodeJS

### What happened?

Hi, I noticed a strange behavior. We have a repeatable job. If we remove it or drain the queue for delayed jobs and add it back with immediately=true , the workers are not fired off.

### How to reproduce.

1. Create a queue and add a repeatable job with `immediately=true` option
2. Create worker for the queue to handle the repeatable job
3. Remove the job or drain the queue
4. Add the repeatable job back to the queue with `immediately=true` option

The worker is not fired off immediately.

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
Could you please produce the actual source code that reproduces the issue?

fabianboerner:
This issue is the case when the original repeated job is still in completed and not completely removed from redis.

manast:
This is probably fixed with the new ```upsertJobScheduler``` method. @fgozdz could you verify it indeed is resolved?

PR #2944, opened 2024-11-28
test(job-scheduler): add tests with drain and promote for job scheduler

fgozdz:
ref #247 #2405


tests/test_job_scheduler.ts
```diff
@@ -1896,4 +1896,42 @@ describe('Job Scheduler', function () {
...
+    await worker.waitUntilReady();
+
+    await queue.drain(true);
+
+    await queue.upsertJobScheduler('scheduler-test', {
+      every: 50000,
+      immediately: true,
+    });
+
+    expect(worker.isRunning()).to.be.true;
```
roggervalf, line 1931: instead of this check, you are looking for checking that worker is able to process jobs from scheduler after drain operation, you can use a queueEvent instance to listen for completed state or from worker instance like in https://github.com/taskforcesh/bullmq/blob/master/tests/test_worker.ts#L1835-L1851

tests/test_job_scheduler.ts
```diff
@@ -1896,4 +1896,49 @@ describe('Job Scheduler', function () {
...
     await worker.close();
   });
+
+  it('should schedule next repeatable job after promote', async function () {
+    await queue.upsertJobScheduler('scheduler-test', { every: 50000 });
+
+    await queue.promoteJobs();
+
+    const waitingCount = await queue.getWaitingCount();
+    expect(waitingCount).to.be.equal(1);
```
roggervalf, line 1906: to check that at least the next repeatable was scheduled you may need to check the delayed count as well, so after this line you can add:
```js
const delayedCount = await queue.getDeleayedCount();
expect(delayedCount).to.be.equal(1);
```

tests/test_job_scheduler.ts
```diff
@@ -1896,4 +1896,49 @@ describe('Job Scheduler', function () {
...
+          expect(job).to.be.ok;
+          expect(job.data.foo).to.be.eql('bar');
+        } catch (err) {
+          reject(err);
+        }
+        resolve();
+      });
+    });
+
+    const job = await queue.add('test', { foo: 'bar' });
```
roggervalf, line 1936: this addition looks like it's not needed but you may need to check the completion of the scheduled job from scheduler

manast (approved):

manast:
@fgozdz I think this PR could be easily fixed and merged.

tests/test_job_scheduler.ts
```diff
@@ -1896,4 +1896,42 @@ describe('Job Scheduler', function () {
...
+        (await queue.getActiveCount()) == 1,
+    ).to.be.true;
+    await queue.removeJobScheduler('scheduler-test');
+    await worker.close();
+  });
+
+  it('worker should start processing repeatable jobs after drain', async function () {
+    await queue.upsertJobScheduler('scheduler-test', {
+      every: 50000,
+      immediately: true,
```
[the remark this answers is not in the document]
manast: In this case we are using pattern.

manast:
Fixes #247

src/classes/job-scheduler.ts
```diff
@@ -86,7 +86,7 @@ export class JobScheduler extends QueueBase {
 
     // Check if we have a start date for the repeatable job
     const { startDate, immediately, ...filteredRepeatOpts } = repeatOpts;
-    let startMillis = now;
+    let startMillis = (now > prevMillis ? now : prevMillis) + (offset || 0);
```
roggervalf, line 89: with this change, when promoting a job, we won't generate same jobId, but we are skipping that iteration and adding more delay for next created job. I believe we should still try to schedule a job with same promotion delay. On the other hand, offset is only used in jobs using "every", not in jobs using patterns

roggervalf: I believe as was mentioned internally, that we can generate a jobId using startMillis or just current timestamp

manast: > with this change, when promoting a job, we won't generate same jobId, but we are skipping that iteration and adding more delay for next created job. I believe we should still try to schedule a job with same promotion delay. On the other hand, offset is only used in jobs using "every", not in jobs using patterns

When promoting a job we take current delayed job and move it to wait. When this job is moved to active we need to generate the next iteration, obviously since the promoted job has been started earlier (as it was promoted manually) the next iteration will be farther in time from the previous job than normal. This is expected, the alternative would be to reset the offset to this new point in time, but this is super messy and difficult to get right as the promoting operation is independent on the schedulers operation. Not sure what you refer with generating the same jobId, we cannot generate the same Id, thats actually the fix in this PR, generating the same ID would result in the job being ignored that's what happened before and did not work, now we generate the next iteration ID, as it has been for many years before this issue was introduced recently. I will add a new case for the pattern case though, to see if it also works as since you mentioned it does not have an offset value.

manast: > with this change, when promoting a job, we won't generate same jobId, but we are skipping that iteration and adding more delay for next created job. I believe we should still try to schedule a job with same promotion delay. On the other hand, offset is only used in jobs using "every", not in jobs using patterns

To clarify, it is not skipping any iteration, it is just running the promoted job earlier than what it would have done normally, the next iteration will be scheduled at its normal slot time independently on when the promoted job was actually processed.

roggervalf: I'm also comparing it with bull behavior and with this change is going to be the same. Next iterations are scheduled in they regular slots as previous jobs were executed at their regular time

roggervalf (approved):
Lgtm

PR #2945, opened 2024-11-28
feat(job-scheduler): add getJobSchedulersCount method

manast:
Added a missing method for the job schedulers api


roggervalf (approved):
Lgtm

Issue #2942, opened 2024-11-27
Telemetry: Add option for a separate trace for every job attempt/start


lovre-nc:
**Is your feature request related to a problem? Please describe.**
Hi, we've attempted to migrate from @appsignal/opentelemetry-instrumentation-bullmq to the bullmq-otel package. 

It seems that a single trace is re-used for the whole lifespan of a job. This is a problem for us because we have a queue where individual jobs exist and repeat forever. We have one job per user and it runs every 5 minutes by delaying it at the end of the worker instead of letting it complete. That means that these jobs have thousands or millions of starts.

When using the official bullmq telemetry this seems to result in one endlessly long trace with thousands or millions of spans inside it. 
This behavior may be useful for queues where the jobs are one-off jobs to see multiple job attempts together but it does not work at all for these kind of repeating jobs.

**Describe the solution you'd like**
Ideally this behavior would be configurable per queue. 
We'd like to have one trace per job attempt/start instead of one trace per job.

**Describe alternatives you've considered**
Staying with @appsignal/opentelemetry-instrumentation-bullmq, but we have some issues with it so that's not ideal.

**Additional context**
None

manast:
In your use case, do you mean that you have always one active job per user?

lovre-nc:
Basically a job in this queue exists as long as the user exists and it runs every 5 minutes and is never moved to completed. 

An illustrated example of our worker that should help explain:
```ts
const worker = new Worker(
  "pollMessagesForUser",
  async (job: Job, token?: string) => {
    const userId = job.id;

    await messagesService.fetchAndStoreUserMessages(userId);

    await job.moveToDelayed(Date.now() + 300_000, token);
    throw new DelayedError();
  },
  { connection }
);
```
---
In the meanwhile I was able to solve the issue myself for now by patching `bullmq` using `pnpm patch`:
```diff
diff --git a/dist/cjs/utils.js b/dist/cjs/utils.js
index 6caa7bae682a971ffac6f585aca1b54b9aaa3267..b40cd744c05ca81f8eecdca93b7086fed3c1568a 100644
--- a/dist/cjs/utils.js
+++ b/dist/cjs/utils.js
@@ -238,7 +238,7 @@ async function trace(telemetry, spanKind, queueName, operation, destination, cal
         const { tracer, contextManager } = telemetry;
         const currentContext = contextManager.active();
         let parentContext;
-        if (srcPropagationMetadata) {
+        if (false) {
[...]

PR #2946, opened 2024-11-29
feat(telemetry): add option to omit context propagation

fgozdz:
ref #2942


roggervalf:
Wondering if an extra option can be passed at job level as for the case that was reportes, only repeatables jobs in some cases need this logic but we can keep the base logic for all the other job types

fgozdz:
I have changed it to pass the option entirely in bullmq, inside the job. For now only in methods regarding enhancement issue.

fgozdz (commented):
LGTM

manast (approved):

manast:
I will squash this PR as there were a few commits on this one...

PR #2952, opened 2024-12-05
fix(worker): catch connection error when moveToActive

roggervalf:
when disconnection happens when moveToActive is called, connection errors are console logged because we are not executing this call in try block


tests/test_job_scheduler.ts
```diff
@@ -1211,7 +1211,7 @@ describe('Job Scheduler', function () {
   });
 
   it('should repeat 7:th day every month at 9:25', async function () {
-    this.timeout(8000);
+    this.timeout(12000);
```
roggervalf, line 1214: flaky test sometimes because of timing, when a disconnection happens when moving job to active

PR #2959, opened 2024-12-09
feat(telemetry): metrics

fgozdz:


.github/pull_request_template.md
```diff
@@ -1,30 +1,36 @@
 <!--
-  Thank you for submitting a PR! 
+  Thank you for submitting a PR!
```
manast, line 2: Please make sure to not reformat unrelated files.

docs/gitbook/SUMMARY.md
```diff
@@ -1,137 +1,137 @@
...
 
-* [What is BullMQ](README.md)
-* [Quick Start](<README (1).md>)
-* [API Reference](https://api.docs.bullmq.io)
-* [Changelogs](changelog.md)
-  * [v4](changelogs/changelog-v4.md)
-  * [v3](changelogs/changelog-v3.md)
-  * [v2](changelogs/changelog-v2.md)
-  * [v1](changelogs/changelog-v1.md)
+- [What is BullMQ](README.md)
```
manast, line 3: Do not reformat unrelated files.

src/interfaces/telemetry.ts
```diff
@@ -1,4 +1,19 @@
-import { SpanKind } from '../enums';
+import { SpanKind, ValueType } from '../enums';
+
+/**
+ * Configuration for bullmqOtel package.
```
manast, line 4: Even though this will be now used by the bullmq-otel package, it is should not be specifically for this package.

src/interfaces/telemetry.ts
```diff
@@ -1,4 +1,19 @@
...
+ * Used for tracing and metrics.
+ */
+export interface Configuration {
```
manast, line 7: Configuration is a way too generic name for an interface.

src/classes/flow-producer.ts
```diff
@@ -207,12 +203,7 @@ export class FlowProducer extends EventEmitter {
       ? `${parentKey}:dependencies`
       : undefined;
 
-    return trace<Promise<JobNode>>(
-      this.telemetry,
-      SpanKind.PRODUCER,
-      flow.queueName,
-      'addFlow',
-      flow.queueName,
+    return telemetry<Promise<JobNode>>(
```
manast, line 206: I am very confused about this name change, now we have both a telemetry object and a telemetry function. Besides, the best practice is to name functions as verbs, and in this case it is a substantive, so it is not clear what the function does by its name alone.

src/interfaces/telemetry.ts
manast: I could not find where Configuration is used in the code, so what is its purpose?

manast (commented):
I think this PR is not ready as the metrics are not being actually used in the code. I added some other comments.

src/interfaces/telemetry.ts
fgozdz: https://github.com/taskforcesh/bullmq-otel/pull/6#discussion_r1912559963 I will be removing configuration from here in response to that comment

roggervalf:
metrics is enabled as part of https://github.com/taskforcesh/bullmq/pull/3769

PR #2966, opened 2024-12-12
fix: guarantee every repeatable jobs are slotted

manast:
This PR adds a guarantee for repeatable jobs based on the "every" option so that only one delayed job per job scheduler can ever be produced by making sure the next iteration is "slotted". By "slotted" we mean that we only allow the repeatable jobs for a given every setting to be placed in evenly slots in the time space. So for example, if we have an every setting of 5000 (5 seconds), then the repetition will be slotted in time 0, 5, 10, 15, 20, 25, and so on. However we also compute an offset so that we can start the iterations at the same point in time the job scheduler is added to the queue. So for instance, if we add the job scheduler at second 10:23:03 (second 3), the iterations will be slotted at 3, 8, 13, 18, and so on.

Another improvement in this PR affects retried repeatable jobs. A repeatable job that has produced the next repeatable job iteration will be marked as such and will not create new iterations independently on how many times this job is retried.

### DIFF
--- a/tests/test_job_scheduler.ts
+++ b/tests/test_job_scheduler.ts
@@ -1636,20 +1636,24 @@ describe('Job Scheduler', function () {
 
       let isFirstRun = true;
 
-      const worker = new Worker(
-        queueName,
-        async () => {
-          this.clock.tick(177);
-          if (isFirstRun) {
-            isFirstRun = false;
-            throw new Error('failed');
-          }
-        },
-        {
-          connection,
-          prefix,
-        },
-      );
+      let worker;
+      const processingAfterFailing = new Promise<void>(resolve => {
+        worker = new Worker(
+          queueName,
+          async () => {
+            this.clock.tick(177);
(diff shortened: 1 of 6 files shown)


tests/test_job_scheduler.ts
```diff
@@ -773,54 +783,177 @@ describe('Job Scheduler', function () {
...
+
+      await queue.upsertJobScheduler(
+        'repeat',
+        {
+          every: 2000,
+        },
+        { data: { foo: 'bar' } },
+      );
+
+      //this.clock.tick(1000);
```
roggervalf, line 951: this line can be removed

roggervalf (approved):
Lgtm

michalvankodev:
Could this change cause the already-present repeated jobs to be duplicated (the jobId has been affected due to the offset parameter) after upgrade?

manast:
@michalvankodev we have thought about that case and it should not happen.

Issue #2962, opened 2024-12-10
[Bug]: DelayError throwing in Child Processor breaks without Message


jpulec:
### Version

v5.29.1

### Platform

NodeJS

### What happened?

# Description
Throwing `DelayedError` doesn't work properly with child processor sandboxing, if you don't pass a `message` to the thrown error. This appears to be because when sending a message to the parent that a child failed, it checks if there's an `err.message` property, and if not, wraps the error with `new Error(err)` which causes the `name` property to be overwritten as `Error` so that the checks for `err.name === 'DelayedError` in the `Worker` class are incorrect.

### How to reproduce.

```typescript
// worker.js
import { pathToFileURL } from 'url';

const processorUrl = pathToFileURL(__dirname + '/my_processor.js');

worker = new Worker(queueName, processorUrl);
```

```typescript
// my_processor.js
import { SandboxedJob } from 'bullmq';

module.exports = async (job: SandboxedJob) => {
   // Note the absense of a passed in message here
    throw new DelayedError();
};
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2967, opened 2024-12-13
fix(sandboxed): fix detecting special errors by sending default messages

roggervalf:
fixes https://github.com/taskforcesh/bullmq/issues/2962


tests/fixtures/fixture_processor_unrecoverable.js
```diff
@@ -13,7 +13,7 @@ module.exports = function (job) {
       throw new Error('Not yet!');
     }
     if (job.attemptsMade < 2) {
-      throw new UnrecoverableError('Unrecoverable');
+      throw new UnrecoverableError();
```
manast, line 16: Will this still work if the user passes a different message?

roggervalf: yes, it will

manast (approved):

PR #2970, opened 2024-12-14
fix(worker): mitigate job's locks extensions

manast:
This fix tries to resolve this issue: https://github.com/taskforcesh/bullmq/discussions/2965#discussioncomment-11549685
And this one:
https://github.com/taskforcesh/bullmq/issues/2969


src/classes/worker.ts
```diff
@@ -911,6 +911,8 @@ will never work with more accuracy than 1ms. */
           const failed = await handleFailed(<Error>err);
           return failed;
         } finally {
+          jobsInProgress.delete(inProgressItem);
```
roggervalf, line 914: shouldn't this line be sufficient and the other extra references can be deleted as this line will be executed after completed and failed handlers?

manast: No because there are other async calls in between, so the lock extender could run in between as well and the there will be no lock for this job.

roggervalf: before line 864 we also need to delete that inProgressitem as well

PR #2975, opened 2024-12-19
fix(dynamic-rate-limit): validate job lock cases

roggervalf:
Validate existente of job when dynamic rate límit is used. Upgrade cron-parser to fix a vulnerability as well


src/classes/scripts.ts
```diff
@@ -1387,13 +1387,11 @@ export class Scripts {
    */
   async moveJobFromActiveToWait(jobId: string, token: string) {
     const client = await this.queue.client;
-    const lockKey = `${this.queue.toKey(jobId)}:lock`;
```
manast, line 1390: why can't we generate the key here?

src/commands/moveJobFromActiveToWait-9.lua
```diff
@@ -24,37 +23,45 @@ local rcall = redis.call
...
 local jobId = ARGV[1]
 local token = ARGV[2]
-local lockKey = KEYS[4]
+local jobKey = ARGV[3]
 
-local lockToken = rcall("GET", lockKey)
-local pttl = rcall("PTTL", KEYS[7])
-if lockToken == token then
-  local metaKey = KEYS[6]
+if rcall("EXISTS", jobKey) == 1 then
```
manast, line 32: I think we should be able to skip checking for the key existence as we need to read the token anyway, and it will return an empty string or nil if the key does not exist.

roggervalf: as we don't expose this method to be used when using manual processing we should be good skipping it. For other methods like moveToComplete or moveToFailed is still needed as when passing token '0' we basically ignore that check

src/classes/scripts.ts
roggervalf: removeLock include concats lock suffix to jobKey that is passed as an argv, so no need to pass lockKey

manast (approved):
LGTM

PR #2978, opened 2024-12-20
feat(job): add moveToWait method for manual processing

roggervalf:


manast (commented):
Wouldn't we also need to update the documentation on how to use this method?

roggervalf:
> Wouldn't we also need to update the documentation on how to use this method?

Already added

manast (approved):
LGTM

PR #2979, opened 2024-12-20
fix(sandbox): fix issue where job could stay in active forever

manast:
This PR solves a critical issue where a sandbox processor could get stuck in active for ever if the child process is killed when the job is in the middle of the processing.


roggervalf (approved):
Lgtm

PR #2987, opened 2024-12-29
fix(events-producer): serialize and deserialize custom events

roggervalf:
### Why

### How

_Any extra info here._
Ignore events that do not need this deserialization as the ones that we handled internally in this package

### DIFF
--- a/src/classes/queue-events-producer.ts
+++ b/src/classes/queue-events-producer.ts
@@ -40,7 +40,7 @@ export class QueueEventsProducer extends QueueBase {
     const args: any[] = ['MAXLEN', '~', maxEvents, '*', 'event', eventName];
 
     for (const [key, value] of Object.entries(restArgs)) {
-      args.push(key, value);
+      args.push(key, JSON.stringify(value));
     }
 
     await client.xadd(key, ...args);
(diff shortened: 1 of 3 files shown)


src/classes/queue-events.ts
```diff
@@ -298,20 +299,40 @@ export class QueueEvents extends QueueBase {
...
+            case 'error':
+            case 'failed':
+            case 'paused':
+            case 'removed':
+            case 'resumed':
+            case 'retries-exhausted':
+            case 'stalled':
+            case 'waiting':
+            case 'waiting-children':
+              break;
```
roggervalf, line 324: avoid to do a deserialization for these events that do not need it

src/classes/queue-events-producer.ts
```diff
@@ -40,7 +40,7 @@ export class QueueEventsProducer extends QueueBase {
     const args: any[] = ['MAXLEN', '~', maxEvents, '*', 'event', eventName];
 
     for (const [key, value] of Object.entries(restArgs)) {
-      args.push(key, value);
+      args.push(key, JSON.stringify(value));
```
manast, line 43: I think we are digging ourselves more into this rabbit hole, as it would have been much simpler to just allow sending strings as payloads for the custom events and let the user serialize and deserialise as they see fit, this will never be satisfactory for all cases out there...

src/classes/queue-events.ts
manast: Fastest here would be to use a Set object that is global for this module and that provides fast check for membership.

Issue #2988, opened 2024-12-30
[Bug]: Job scheduler creates multiple delayed jobs when upserting in quick succession 


tjhiggins:
### Version

v5.34.5

### Platform

NodeJS

### What happened?

If you call upsertJobScheduler in quick succession it will create multiple delayed jobs for the same scheduler. 

### How to reproduce.

Took me awhile to reproduce. Needed to have a worker defined. I also struggled to get it to fail consistently with the fake clock...

Test: https://github.com/taskforcesh/bullmq/commit/d6117a0f34de9c15991b43784615040ae599e158#diff-a19149ca7b4f2a0cee7cff833e33245c005ef785fc0a7ac3616a11b79148a21fR213

### Relevant log output

Struggled to make it fail every time with the fake clock, but here is an example failure:

```shell
$ yarn test -g 'when job schedulers are upserted in quick succession'
yarn run v1.22.22
$ npm-run-all clean:scripts generate:raw:scripts transform:commands
$ rimraf rawScripts src/scripts
$ ts-node --project tsconfig-cjs.json scripts/generateRawScripts.ts
$ node ./scripts/commandTransform.js ./rawScripts ./src/scripts
$ ts-mocha -p tsconfig-cjs.json --config ./.mocharc.js -g 'when job schedulers are upserted in quick succession'


  Job Scheduler
    when job schedulers are upserted in quick succession
      1) should create only one job scheduler and one delayed job


  0 passing (58ms)
  1 failing

  1) Job Scheduler
       when job schedulers are upserted in quick succession
         should create only one job scheduler and one delayed job:

      AssertionError: expected [ Job{ …(22) }, Job{ …(22) } ] to have a length of 1 but got 2
      + expected - actual

      -2
      +1
      
      at Context.<anonymous> (tests/test_job_scheduler.ts:254:31)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)



error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tjhiggins:
If the test passes when you run try adding more schedulers

Ex.

```ts
await queue.upsertJobScheduler(jobSchedulerId, {
  every: ONE_MINUTE * 5,
});

await this.clock.tickAsync(1);

await queue.upsertJobScheduler(jobSchedulerId, {
  every: ONE_MINUTE * 5,
});

[...]

PR #2991, opened 2024-12-31
fix(job-scheduler): avoid duplicates when upserting in a quick sequence

manast:
### Why

This PR fixes this issue: https://github.com/taskforcesh/bullmq/issues/2988
### How

This PR makes sure that the "nextMillis" portion of every repeatable jobs is always calculated on well defined so called "time slots". The offset is only used to calculate the "delay" option of the repeatable job so that it repeats at the appropriate time, but is not used for the "nextMillis" calculation anymore.

### DIFF
--- a/tests/test_job_scheduler.ts
+++ b/tests/test_job_scheduler.ts
@@ -206,11 +206,10 @@ describe('Job Scheduler', function () {
   });
 
   describe('when job schedulers are upserted in quick succession', function () {
-    let worker: Worker;
-    beforeEach(async function () {
+    it('should create only one job scheduler and one delayed job', async function () {
       const date = new Date('2017-02-07 9:24:00');
       this.clock.setSystemTime(date);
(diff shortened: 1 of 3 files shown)


tests/test_job_scheduler.ts
```diff
@@ -205,6 +205,53 @@ describe('Job Scheduler', function () {
     });
   });
 
+  describe('when job schedulers are upserted in quick succession', function () {
+    let worker: Worker;
+    beforeEach(async function () {
+      const date = new Date('2017-02-07 9:24:00');
+      this.clock.setSystemTime(date);
+      worker = new Worker(
```
roggervalf, line 213: beforeEach and afterEach in this describe block is only affecting one test, they can be placed inside that test

roggervalf (approved):
Lgtm, only one styled comment

PR #2993, opened 2025-01-03
perf(job-scheduler): add delayed job and scheduler in same script

roggervalf:
### Why

Why is this change necessary? In order to handle addition of a job scheduler and delayed job in same script. Be able to execute in serverless redis as we don't need to use multi
### How

How did you implement this? Add a new include to handle delayed job additions

_Any extra info here._

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -144,7 +144,10 @@ export class JobScheduler extends QueueBase {
               jobSchedulerId,
               {
                 ...opts,
-                repeat: { ...filteredRepeatOpts, offset: newOffset },
+                repeat: {
+                  ...filteredRepeatOpts,
+                  offset: newOffset ? newOffset : undefined,
+                },
                 telemetry,
               },
               iterationCount,
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -116,18 +116,70 @@ export class JobScheduler extends QueueBase {
     const multi = (await this.client).multi();
     if (nextMillis) {
       if (override) {
-        this.scripts.addJobScheduler(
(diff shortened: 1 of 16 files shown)


src/classes/worker.ts
```diff
@@ -793,7 +793,7 @@ will never work with more accuracy than 1ms. */
         if (job.repeatJobKey) {
           const jobScheduler = await this.jobScheduler;
           await jobScheduler.upsertJobScheduler(
-            job.repeatJobKey,
+            job.jobSchedulerId,
```
roggervalf, line 796: wrong value was passed, it must be jobSchedulerId, not key that contains prefix + queueName

src/classes/job-scheduler.ts
```diff
@@ -118,18 +118,71 @@ export class JobScheduler extends QueueBase {
...
-          nextMillis,
-          JSON.stringify(typeof jobData === 'undefined' ? {} : jobData),
-          Job.optsAsJSON(opts),
-          {
-            name: jobName,
-            endDate: endDate ? new Date(endDate).getTime() : undefined,
-            tz: repeatOpts.tz,
-            pattern,
-            every,
+        return this.trace<Job<T, R, N>>(
```
roggervalf, line 121: copied for now, when update job scheduler job get refactored, we will reuse same wrapper

src/commands/includes/addDelayedJob.lua
```diff
@@ -0,0 +1,25 @@
+--[[
+  Add marker if needed when a job is available.
```
manast, line 2: The comment here seems to be wrong.

roggervalf: addressed in https://github.com/taskforcesh/bullmq/pull/2997

PR #3001, opened 2025-01-09
fix(job-scheduler): add next delayed job only when prevMillis matches with producerId

roggervalf:
### Why

1. Why is this change necessary? In order to only add delayed jobs while current producer matches with prevMillis

### How

1. How did you implement this? Do this validation when updating job scheduler, comparing prevMillis with producerId

_Any extra info here._

### DIFF
--- a/src/commands/updateJobScheduler-6.lua
+++ b/src/commands/updateJobScheduler-6.lua
@@ -1,23 +1,24 @@
 --[[
   Updates a job scheduler and adds next delayed job
 
-    Input:
-      KEYS[1] 'marker',
-      KEYS[2] 'meta'
-      KEYS[3] 'id'
-      KEYS[4] 'delayed'
-      KEYS[5] events stream key
-      KEYS[6] 'repeat' key
-      
-      ARGV[1] next milliseconds
-      ARGV[2] jobs scheduler id
-      ARGV[3] msgpacked delayed opts
-      ARGV[4] timestamp
-      ARGV[5] prefix key
-      ARGV[6] producer key
+  Input:
+    KEYS[1] 'marker',
+    KEYS[2] 'meta'
+    KEYS[3] 'id'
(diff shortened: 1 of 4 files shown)


src/commands/updateJobScheduler-6.lua
```diff
@@ -38,24 +39,31 @@ local nextDelayedJobKey =  schedulerKey .. ":" .. nextMillis
...
+
+  if producerId then
+    local currentDelayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. prevMillis
+
+    if producerId == currentDelayedJobId then
+      addDelayedJob(nextDelayedJobKey, nextDelayedJobId, delayedKey, eventsKey, schedulerAttributes[1],
+        schedulerAttributes[2] or "{}", delayedOpts, timestamp, jobSchedulerId, maxEvents, KEYS[1], nil, nil)
+  
+      if ARGV[6] ~= "" then
+        rcall("HSET", ARGV[6], "nrjid", nextDelayedJobId)
```
roggervalf, line 63: we can even potentially deprecated this attribute with this new validation

src/commands/updateJobScheduler-6.lua
```diff
@@ -1,23 +1,24 @@
...
+    KEYS[4] 'delayed'
+    KEYS[5] events stream key
+    KEYS[6] 'repeat' key
+
+    ARGV[1] next milliseconds
+    ARGV[2] jobs scheduler id
+    ARGV[3] msgpacked delayed opts
+    ARGV[4] timestamp
+    ARGV[5] prefix key
+    ARGV[6] producer key
```
manast, line 17: could we use a key here instead of an arg?

manast (commented):
Looks good to me, other than I think producerKey could be a key instead of an arg.

Issue #3000, opened 2025-01-08
[Question] DelayedError and attempts.


Stijn-Timmer:
Hey BullMQ team,

I have a question about the DelayedError error and its behavior regarding attempts.

Below, I have this example:

```typescript
if (await isLocked(domain)) {
  // Move job to the delayed queue.
  await job.moveToDelayed(randomDelay(0, 300), token);
  throw new DelayedError();
}

// Handle different fetch attempts with escalating strategies.
switch (job.attemptsStarted) {
  case 1: 
    // Plain o'l fetch
  case 2:
    // Puppeteer
  case 3: 
    // 3rd party
}
```
In this example, I'm locking a domain to not overload our target with many requests in the queue hitting the same domain.

But when I lock it and delay the job for a bit and handle some other jobs in the meantime, it appears that throwing the DelayedError increments the attempt count, always skipping a step in the case.

I am probably missing something or fundamentally doing something wrong with this locking mechanism, but I'm curious how I could simply delay a job without counting it as an 'attempt'.

roggervalf:
hi @Stijn-Timmer, attemptsStarted attribute is increased every time that a job is moved to active. What you may be looking for is attemptsMade attribute that is increased only when errors different than DelayedError, WaitingChildrenError or RateLimitError happen

PR #3003, opened 2025-01-10
docs(stop-retrying): clarify different between attemptMade and attemptsStarted

roggervalf:
### Why

  1. Why is this change necessary? We should clarify the difference between attemptMade and attemptStarted and when they can be used

### How

  1. How did you implement this? Fix example in our documentation and add a info section to clarify these concepts

ref https://github.com/taskforcesh/bullmq/issues/3000


manast (approved):
LGTM

PR #3015, opened 2025-01-16
fix(job-scheduler): validate if next delayed job exists

roggervalf:
### Why

2. What problem does it solve or improve? Jobs cannot be in more than 1 state, if we do not validate if a job already exists when a delayed job that is created by a job scheduler is promoted, we can have this job id in more than 1 state, i.e active and delayed or completed and delayed
### How

 1. How did you implement this? Validate job key existence before trying to add this record

_Any extra info here._

### DIFF
--- a/src/commands/addJobScheduler-6.lua
+++ b/src/commands/addJobScheduler-6.lua
@@ -84,17 +84,16 @@ local schedulerKey = repeatKey .. ":" .. jobSchedulerId
 local nextDelayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. nextMillis
 local nextDelayedJobKey =  schedulerKey .. ":" .. nextMillis
 
+local nextDelayedJobDoesNotExist = rcall("EXISTS", nextDelayedJobKey) ~= 1
 -- If we are overriding a repeatable job we must delete the delayed job for
 -- the next iteration.
 local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
 if prevMillis ~= false then
-  local delayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. prevMillis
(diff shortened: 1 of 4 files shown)


src/commands/addJobScheduler-6.lua
```diff
@@ -88,33 +50,34 @@ local nextDelayedJobKey =  schedulerKey .. ":" .. nextMillis
...
-  local delayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. prevMillis
+  local prevDelayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. prevMillis
 
-  if rcall("ZSCORE", delayedKey, delayedJobId) ~= false
-    and (rcall("EXISTS", nextDelayedJobKey) ~= 1 
-    or delayedJobId == nextDelayedJobId) then
-    removeJob(delayedJobId, true, prefixKey, true --[[remove debounce key]])
-    rcall("ZREM", delayedKey, delayedJobId)
+  if rcall("ZSCORE", delayedKey, prevDelayedJobId) ~= false then
+    removeJob(prevDelayedJobId, true, prefixKey, true --[[remove debounce key]])
```
manast, line 56: Instead of removing the job here and then adding a new one, couldn't we just set a flag here and if true not add the job again?

roggervalf: the problem is that if the delayed is not regenerated with new repeat opts, next delayed job will be scheduled with old repeat options. This is happening in that way because we use opts from current delayed job to schedule the next one.

manast: Ok, can you write a comment then to explain why it is needed so we not need to remember it next time? :)

roggervalf: sure, I'll add a description about it

PR #3018, opened 2025-01-17
feat(job-scheduler): save iteration count

roggervalf:
### Why

  1. Why is this change necessary? In order to show how many iterations are being made

### How

  1. How did you implement this? Add new ic attribute to save this value

_Any extra info here._


manast (approved):
LGTM

PR #3021, opened 2025-01-20
Add React .tsx and .jsx to supportedFileTypes in worker.ts

WilliamBlais:
### Why
This change is necessary to ensure support for React .tsx and .jsx file types in the worker logic, particularly in scenarios where files like EmailWorker.tsx render React components and manage tasks like sending emails.
### How
Updating the supportedFileTypes array in worker.ts to include .tsx and .jsx. This allows the worker to correctly use React files.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -259,7 +259,7 @@ export class Worker<
           }
           processor = processor.href;
         } else {
-          const supportedFileTypes = ['.js', '.ts', '.flow', '.cjs'];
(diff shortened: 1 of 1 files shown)


manast (commented):
Thanks for the PR. It would be great to have a minimal test case to prove that the change actually work.

WilliamBlais:
@manast Added a test case. 


By default, the code in worker.ts:265 appends .js to the processorFile path when the file extension is not found in supportedFileTypes.

Would it be better to throw an exception instead of appending .js when the file type is not supported? This would provide clearer feedback and prevent potential issues caused by processing unintended file.

What do you think?

PR #1417, opened 2022-09-16
feat(sandboxed-job): add support for getChildrenValues

rosslavery:
Hi @manast @roggervalf :D 

Looking to get this feature added for sandboxed workers. The only part I'm not certain on is how you'd like [this error](https://github.com/rosslavery/bullmq/blob/feat/sandbox-getchildrenvalues/src/classes/child-processor.ts#L229) to be handled, or if there is try/catch further up the chain so it's not needed here.

Fixes https://github.com/taskforcesh/bullmq-pro-support/issues/21
Fixes https://github.com/taskforcesh/bullmq/issues/1306
Fixes https://github.com/taskforcesh/bullmq/issues/753

PR #3023, opened 2025-01-21
feat(sandbox): add support for getChildrenValues

manast:
### Why

This PR adds support for using getChildrenValues from sandboxes, fixes https://github.com/taskforcesh/bullmq/issues/1306, https://github.com/taskforcesh/bullmq/issues/753, https://github.com/taskforcesh/bullmq-pro-support/issues/21 supersedes https://github.com/taskforcesh/bullmq/pull/1417

### How

Added a response mechanisms for operations that require returning values back to the child processor from the parent process.


fgozdz (approved):
LGTM

PR #3025, opened 2025-01-22
fix(worker): avoid possible hazard in closing worker

manast:
### Why

It seems like in some edge cases BZPOPMIN is issued after the closing procedure starts which in some backend like DragonFly can cause a longer time to actually close the worker which leads to some tests timing out.

### How

Make sure BZPOPMIN is only called if the closing process has not started yet.

We are not 100% sure this is the cause for the timing out of some tests but still this change is needed to rule out this possibility.


roggervalf (approved):
Lgtm

PR #3026, opened 2025-01-24
feat(queue-getters): add prometheus exporter

manast:
### Why

To provide built in support for Prometheus.

### How

Adding a new method ```exportPrometheusMetrics``` that returns metrics in prometheus open metrics format. 

An http server must be used to provide a HTTP endpoint to be consumed by prometheus. The current export only support gauges but we could add counters as well if metrics are enabled in the queue.

### DIFF
--- a/tests/test_getters.ts
+++ b/tests/test_getters.ts
@@ -4,6 +4,8 @@
 import { expect } from 'chai';
 import { after } from 'lodash';
 import { describe, beforeEach, it, before, after as afterAll } from 'mocha';
+import * as sinon from 'sinon';
+
 import { default as IORedis } from 'ioredis';
 import { v4 } from 'uuid';
 import { FlowProducer, Queue, QueueEvents, Worker } from '../src/classes';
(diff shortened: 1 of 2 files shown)


tests/test_getters.ts
```diff
@@ -1011,4 +1013,45 @@ describe('Jobs getters', function () {
...
   });
+
+  describe('#exportPrometheusMetrics', () => {
+    afterEach(() => {
+      sinon.restore();
+    });
+
+    it('should export all job states in Prometheus gauge format', async () => {
+      const counts = {
+        waiting: 5,
```
roggervalf, line 1024: prioritized and waiting-children states are missing, just for completeness

roggervalf (approved):
Lgtm

tests/test_getters.ts
manast: As the test is a mock up, I think it does not really matter what states we have test as in the future new can come and others can be deleted or removed.

atmosx:
Hello,

@manast thanks for adding prometheus metrics support. Why are these metrics of type `gauge`? Using gauges makes queries like "how many jobs failed the last three days?" hard to answer because the metric goes up and down. Looks like these metrics should be [counters](https://prometheus.io/docs/tutorials/understanding_metric_types/#counter) to be used properly in dashboards and alerts 🤔

PR #3030, opened 2025-01-26
feat: add a queue registry

manast:
### Why

To be able to efficiently get all the bullmq queues that are available in a Redis host. This feature makes
queue discovery a mostly O(1) operation instead of O(n).

### How

Added a new key with a fixed name: "bullmq:registry", that holds a ZSET with the fully queue qualified names. This queue names will be kept in the registry until the queue is completely obliterated.

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -21,6 +21,7 @@ import { RedisConnection } from './redis-connection';
 import { SpanKind, TelemetryAttributes } from '../enums';
 import { JobScheduler } from './job-scheduler';
 import { version } from '../version';
+import { BullMQRegistryKey } from '../consts/bullmq-registry-key';
 
 export interface ObliterateOpts {
   /**
@@ -179,13 +180,30 @@ export class Queue<
     this.waitUntilReady()
       .then(client => {
         if (!this.closing && !opts?.skipMetasUpdate) {
-          return client.hmset(this.keys.meta, this.metaValues);
+          const multi = client.multi();
+          multi.hmset(this.keys.meta, this.metaValues);
+          multi.zadd(BullMQRegistryKey, Date.now(), this.qualifiedName);
         }
       })
       .catch(err => {
         // We ignore this error to avoid warnings. The error can still
         // be received by listening to event 'error'
       });
+  }
+
+  /**
+   * Returns the queues that are available in the registry.
+   * @param start - zero based index from where to start returning jobs.
+   * @param end - zero based index where to stop returning jobs.
+   */
+  static getRegistry(
+    client: {
+      zrange: (key: string, start: number, end: number) => Promise<string[]>;
+    },
+    start = 0,
+    end = -1,
+  ): Promise<string[]> {
+    return client.zrange(BullMQRegistryKey, start, end);
   }
 
   emit<U extends keyof QueueListener<JobBase<DataType, ResultType, NameType>>>(
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -43,6 +43,7 @@ import {
 } from '../utils';
 import { ChainableCommander } from 'ioredis';
 import { version as packageVersion } from '../version';
+import { BullMQRegistryKey } from '../consts/bullmq-registry-key';
 export type JobData = [JobJsonRaw | number, string?];
 
 export class Scripts {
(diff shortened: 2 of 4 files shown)


src/consts/bullmq-registry-key.ts
```diff
@@ -0,0 +1 @@
+export const BullMQRegistryKey = 'bullmq:registry';
```
roggervalf, line 1: we use bull as prefix for all our keys, do we really need to change that pattern?

src/classes/scripts.ts
```diff
@@ -1464,6 +1465,7 @@ export class Scripts {
     const client = await this.queue.client;
 
     const keys: (string | number)[] = [
+      BullMQRegistryKey,
```
roggervalf, line 1468: we have this class for handling queue keys, https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-keys.ts we can add a new method to handle global keys

manast: hmm, this is not really Queue key, so I am not sure it fits so well on a class that generates queue keys :/

src/consts/bullmq-registry-key.ts
manast: yes, because this is a global key, if we used a prefix then we would not be able to auto-discover the queues defeating the whole purpose. The idea is that a UI can easily get all the queues available in a Redis instance.

roggervalf: no sure why we are not be able to auto-discover queues. I think we should reuse same prefix that users provides with prefix option. Imagine the scenario where 2 different prefixes are use in the same redis instance. With this key, auto-discovery will return a combination of all queues independent of these prefixes, I think discovery should depend on the prefix that is passed by users or if not passed, use bull as the default value that is used by our other keys

manast: If you want to segment the queues in prefixes you can still use this approach as the queue names are stored in the registry fully qualified. The problem we are having is that auto discovery is too slow as you must check every key in Redis to match a certain pattern, currently this one ```*:*:id ``` so some users complain that it takes too much time it can even time out.

roggervalf: > If you want to segment the queues in prefixes you can still use this approach as the queue names are stored in the registry fully qualified

That means that users need to get all queue qualified names from discovery, then filter by prefix and then they will get all the queues related to that prefix. 

> The problem we are having is that auto discovery is too slow as you must check every key in Redis to match a certain pattern, currently this one *:*:id  so some users complain that it takes too much time it can even time out.

I'm in favor of this functionality, my only concern is that we should respect the use of the prefix that could be provided by the user. I also use prefix for separation of queues and seeing some other users using custom prefixes. Please take in count that it would be the same functionality, the only consideration is to use the prefix that users provide when creating queues, our default value is bull

tests/test_queue.ts
```diff
@@ -715,4 +716,142 @@ describe('queues', function () {
...
+      for (const name of queueNames) {
+        const queue = new Queue(name, { connection: client, prefix });
+        queues.push(queue);
+      }
+
+      await Promise.all(queues.map(q => q.waitUntilReady()));
+
+      await delay(100);
+
+      const results = await Queue.getRegistry(client, 0, -1);
```
roggervalf, line 808: need some documentation about this new method. Also a warning that if users do only use FlowProducers, discovery won't work

tests/test_queue.ts
```diff
@@ -715,4 +716,142 @@ describe('queues', function () {
...
+      const queue = new Queue('test-registry-not-paused', {
+        connection: client,
+        prefix,
+      });
+      await queue.waitUntilReady();
+
+      let errorCode: number | null = null;
+      try {
+        const result = await queue.obliterate();
+        console.log(result);
```
roggervalf, line 841: these console logs can be removed

src/classes/queue.ts
```diff
@@ -188,6 +192,21 @@ export class Queue<
       });
   }
 
+  /**
+   * Returns the queues that are available in the registry.
+   * @param start - zero based index from where to start returning jobs.
+   * @param end - zero based index where to stop returning jobs.
+   */
+  static getRegistry(
```
roggervalf, line 200: following my previous comment, what if we pass custom prefix here. So we can get the discovery related to that custom value

roggervalf (approved):

tests/test_queue.ts
manast: Ok, it should work for flow producers also, this was an oversight from my part.

PR #3034, opened 2025-01-29
fix(queue-getters): filter undefined jobs from getJobs

DanFaudemer:
Ensure getJobs does not return undefined if a job is deleted between ID retrieval and fetching.

# Why

* getJobs could return undefined if a job was deleted between ID retrieval with a return type `Promise<JobBase[]>`
* This caused unexpected behavior when processing job lists.
* Fixing this ensures getJobs only returns valid job objects.

# How

* Updated getJobs to filter out undefined results.
* Ensured that the function always returns a valid JobBase[] array.

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -391,9 +391,13 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs: (JobBase | undefined)[] = await Promise.all(
+      jobIds.map(
+        jobId => this.Job.fromId(this, jobId) as Promise<JobBase | undefined>,
+      ),
     );
+
+    return jobs.filter(job => Boolean(job));
   }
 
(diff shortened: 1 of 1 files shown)


manast:
Thanks for the PR. There is a subtle issue here and that is that the number of returned jobs will not match the requested amount based on the start and end arguments, which for an application that maybe is doing pagination it could lead to believe that there are no more pages left as the last call returned less jobs than expected. Maybe the best solution would be to perform the whole operation using MULTI in one transaction.

DanFaudemer:
You are right! Just for reference, the Python version is already filtering out None jobs: [Python BullMQ - Queue](https://github.com/taskforcesh/bullmq/blob/master/python/bullmq/queue.py#L293).

I don’t know the codebase in-depth, but I believe that as soon as a job is missing, pagination will break—even without explicitly filtering out missing jobs—because job IDs will shift.

I see three possible solutions:
* I am not a redis expert but as you mentioned  we may use MULTI with WATCH on getRanges and have a retry mechanism in case jobs have been updated
* Fetch additional pages until the result contains the expected number of jobs
* Modify the return type of getJobs to include undefined jobs

Which solution do you think is the best approach?

roggervalf:
getJobs was refactored to retrieve job records using lua, this change is not longer needed

PR #3036, opened 2025-01-30
fix(retry-job): consider updating failures in job

roggervalf:
### Why

  1. Why is this change necessary? Errors are not updated when retrying a job

### How

  1. How did you implement this? Pass right object into retryJob script and reuse updateJobFields include 

_Any extra info here._

### DIFF
--- a/tests/test_worker.ts
+++ b/tests/test_worker.ts
@@ -2698,14 +2698,14 @@ describe('workers', function () {
       });
     });
 
-    describe('when job has been marked as discarded', () => {
-      it('does not retry a job', async () => {
+    describe('when job has been failed and moved to wait', () => {
+      it('saves failedReason', async () => {
         const worker = new Worker(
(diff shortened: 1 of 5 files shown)


tests/test_worker.ts
```diff
@@ -2698,14 +2698,14 @@ describe('workers', function () {
...
 
-    describe('when job has been marked as discarded', () => {
-      it('does not retry a job', async () => {
+    describe('when job has been failed and moved to wait', () => {
+      it('saves failedReason', async () => {
         const worker = new Worker(
           queueName,
           async job => {
             expect(job.attemptsMade).to.equal(0);
-            job.discard();
```
roggervalf, line 2707: replacing this test as discard method is not currently working as expected, also this method will be removed in next breaking change

manast (approved):
LGTM

PR #3042, opened 2025-02-01
fix(worker): process job when closing and it was moved to active

roggervalf:
### Why

  1. Why is this change necessary? When a job is moved to active, even when closing. We need to try to process it. If not, we will create stalled jobs

### How

  1. How did you implement this? Remove closing check when processing a job

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -819,7 +819,7 @@ will never work with more accuracy than 1ms. */
     fetchNextCallback = () => true,
     jobsInProgress: Set<{ job: Job; ts: number }>,
   ): Promise<void | Job<DataType, ResultType, NameType>> {
-    if (!job || this.closing || this.paused) {
+    if (!job) {
       return;
     }
 
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -189,6 +189,7 @@ export class Worker<
 
   private stalledCheckStopper?: () => void;
   private waiting: Promise<number> | null = null;
+  private waitingRun: Promise<void> = Promise.resolve();
   private _repeat: Repeat; // To be deprecated in v6 in favor of Job Scheduler
 
   protected _jobScheduler: JobScheduler;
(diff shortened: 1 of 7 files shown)


src/classes/worker.ts
```diff
@@ -819,7 +826,7 @@ will never work with more accuracy than 1ms. */
     fetchNextCallback = () => true,
     jobsInProgress: Set<{ job: Job; ts: number }>,
   ): Promise<void | Job<DataType, ResultType, NameType>> {
-    if (!job || this.closing || this.paused) {
+    if (!job) {
```
manast, line 829: not needed, as the signature of this method already guarantees that the job cannot be undefined or null

src/commands/addDelayedJob-6.lua
```diff
@@ -88,7 +88,7 @@ else
     end
 end
 
-local deduplicationJobId = deduplicateJob(args[1], opts['de'],
+local deduplicationJobId = deduplicateJob(opts['de'],
```
manast, line 91: unrelated fix?

src/commands/includes/deduplicateJob.lua
```diff
@@ -1,24 +1,24 @@
 --[[
-  Function to debounce a job.
+  Function to deduplicate a job.
 ]]
 
-local function deduplicateJob(prefixKey, deduplicationOpts, jobId, deduplicationKey, eventsKey, maxEvents)
+local function deduplicateJob(deduplicationOpts, jobId, deduplicationKey, eventsKey, maxEvents)
```
manast, line 5: this seems to be an unrelated fix.

src/commands/addDelayedJob-6.lua
roggervalf: Unused variable, not a fix, just re formatting

roggervalf:
Closed in favor of https://github.com/taskforcesh/bullmq/pull/3059

PR #3044, opened 2025-02-02
feat(job-scheduler): revert add delayed job and update in the same script

manast:
### Why

The feature introduced in this PR https://github.com/taskforcesh/bullmq/pull/2997 was causing some weird errors and regressions so we must revert it until we can re-implement the same functionality in a better way.

### How

Just reverting the code that was causing the issues.


roggervalf (approved):
Lgtm

manast:
@jnormore there are new tests now that I do not expect to pass with the version you are using. In particular when calling upserJobScheduler you could end having several delayed jobs and thus getting more iterations than the ones you have specified in your scheduler's settings.

PR #3045, opened 2025-02-02
fix: do not skip job to process when closing worker

manast:
### Why

When a job has been fetched and moved to active we should start to process it even if the worker has started the closing process.

### How

Just letting the worker to process the job that it has already moved to active.

Before this fix, jobs would end in a stalled state even when closing the worker gracefully.


roggervalf:
closed in favor of https://github.com/taskforcesh/bullmq/pull/3059

PR #3046, opened 2025-02-02
refactor(pause): stop run execution

roggervalf:
### Why

  1. Why is this change necessary? As part of trying to fix the processing of jobs when closing a worker. When pausing a job we don't need to wait for a promise, instead just set it as paused and stop run execution

### How

  1. How did you implement this? Remove promise from pause method and refactor resume method to re-execute run method

Notice that we wait for pending jobs in https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L954 when pausing a worker by default, but we can possibly add paused promise into async queue https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L586, that means that we can get pause method hanging

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -472,12 +472,10 @@ export class Worker<
          */
         while (
           !this.waiting &&
+          !this.paused &&
           numTotal < this._concurrency &&
           (!this.limitUntil || numTotal == 0)
         ) {
-          if (this.paused) {
-            break;
-          }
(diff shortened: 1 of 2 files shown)


roggervalf:
> I am not sure this is correct. When the worker is paused it must be awaiting something otherwise it will loop at maximum speed doing nothing consuming all the CPU.

I modified the test case to verify that runExecution is resolved after pausing

manast (approved):

PR #3052, opened 2025-02-05
perf(add-job): add job into wait or prioritized state when delay is provided as 0

roggervalf:
### Why

1. Why is this change necessary? When a job is added without delay, default value is 0. That means that it will be added into wait or prioritized state, no need to add a job into delayed when it will be moved it directly into one of those states when delay is provided as 0

### How

 1. How did you implement this? Evaluate that delay is greater than 0

Issue introduces in https://github.com/taskforcesh/bullmq/pull/2816

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -158,7 +158,7 @@ export class Queue<
 
   protected libName = 'bullmq';
 
-  private _repeat?: Repeat; // To be deprecated in v6 in favor of JobScheduler
+  protected _repeat?: Repeat; // To be deprecated in v6 in favor of JobScheduler
   protected _jobScheduler?: JobScheduler;
 
   constructor(
--- a/src/commands/includes/deduplicateJob.lua
+++ b/src/commands/includes/deduplicateJob.lua
@@ -2,7 +2,7 @@
   Function to debounce a job.
 ]]
 
-local function deduplicateJob(prefixKey, deduplicationOpts, jobId, deduplicationKey, eventsKey, maxEvents)
+local function deduplicateJob(deduplicationOpts, jobId, deduplicationKey, eventsKey, maxEvents)
   local deduplicationId = deduplicationOpts and deduplicationOpts['id']
   if deduplicationId then
     local ttl = deduplicationOpts['ttl']
--- a/src/commands/includes/isJobSchedulerJob.lua
+++ b/src/commands/includes/isJobSchedulerJob.lua
@@ -1,0 +1,15 @@
+--[[
+  Function to check if the job belongs to a job scheduler and
+  current delayed job matches with jobId
+]]
+local function isJobSchedulerJob(jobId, jobKey, jobSchedulersKey)
+  local repeatJobKey = rcall("HGET", jobKey, "rjk")
+  if repeatJobKey  then
+    local prevMillis = rcall("ZSCORE", jobSchedulersKey, repeatJobKey)
+    if prevMillis ~= false then
+      local currentDelayedJobId = "repeat:" .. repeatJobKey .. ":" .. prevMillis
+      return jobId == currentDelayedJobId
+    end
+  end
+  return false
+end
--- a/src/commands/removeJob-3.lua
+++ b/src/commands/removeJob-3.lua
@@ -5,6 +5,7 @@
     Input:
       KEYS[1] queue prefix
       KEYS[2] meta key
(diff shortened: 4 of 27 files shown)


src/commands/includes/deduplicateJob.lua
```diff
@@ -2,7 +2,7 @@
   Function to debounce a job.
 ]]
 
-local function deduplicateJob(prefixKey, deduplicationOpts, jobId, deduplicationKey, eventsKey, maxEvents)
```
roggervalf, line 5: unused variable

src/classes/queue.ts
```diff
@@ -158,7 +158,7 @@ export class Queue<
 
   protected libName = 'bullmq';
 
-  private _repeat?: Repeat; // To be deprecated in v6 in favor of JobScheduler
+  protected _repeat?: Repeat; // To be deprecated in v6 in favor of JobScheduler
```
roggervalf, line 161: needed in pro package

src/commands/removeJob-3.lua
```diff
@@ -79,10 +80,18 @@ local prefix = KEYS[1]
 local jobId = ARGV[1]
 local shouldRemoveChildren = ARGV[2]
 local jobKey = prefix .. jobId
-
--- Check if the job belongs to a job scheduler and it is in delayed state.
-if rcall("ZSCORE", prefix .. "delayed", jobId) and rcall("HGET", jobKey, "rjk") then
-    return -8 -- Return error code as the job is part of a job scheduler and is in delayed state.
+local repeatKey = KEYS[3]
+
+local repeatJobKey = rcall("HGET", jobKey, "rjk")
```
manast, line 85: I do not think these tests are necessary. if the job is in delayed state and the job belongs to a job scheduler, then that should enough to determine that the job cannot be removed.

manast (commented):
The PR seems to include fixes also in job scheduler unrelated to this PR description

src/commands/removeJob-3.lua
roggervalf: no really, because if a job is already processed, it does not longer has a dependency with scheduler. Also apart of delayed state, this job can be promoted to waiting and prioritized state, where we must not remove it as it didn't generate the next delayed job yet. It will generate next delayed job as soon as it's moved to active.

src/commands/includes/isJobSchedulerJob.lua
```diff
@@ -0,0 +1,15 @@
+--[[
+  Function to check if the job belongs to a job scheduler and
+  current delayed job matches with jobId
+]]
+local function isJobSchedulerJob(jobId, jobKey, jobSchedulersKey)
```
roggervalf, line 5: this method considers all possible states where we can consider a job as the current generated job from a job scheduler. While this job matches with the id of the currentDelayedJobId means that this job didn't pass through active state yet

src/commands/removeJob-3.lua
manast: We do not support priorities for job-scheculers jobs. The only case that needs to be handled would be when promoting delayed jobs, the question is if promote should imply moving to wait or just updating the timestamp to now. So that a delayed job that is promoted do not go to wait. Since we do not use BLPOPRPUSH anymore, moving to wait for promoting a delayed job does not really improve performance, it will still need to push a marker to notify workers.

roggervalf: I can demonstrate that we support priorities as in this test case https://github.com/taskforcesh/bullmq/pull/3064 in general promotion of jobs, being manual or automatic by include script considers all the other states. That means if delayed job to be promoted is not having priority greater than 0, it's moved to waiting or to paused (queue is paused), if delayed job has priority greater than 0, it's promoted to prioritized state. Delay markers only notify workers that there are pending delayed jobs to be promoted at some point in the future.

manast: One assumption we have is that a job-scheduler always must have a delayed job while the job scheduler is active, so we either have to relax that assumption and make sure still covers all edge cases or we must ensure that the assumption is always true.

roggervalf: to ensure that the assumption is always true is not possible as it would imply to break priority option when it's being used as part of the template, another case is when pausing a queue, as when these jobs are promoted, they will be moved to that state

manast (approved):
LGTM

PR #3059, opened 2025-02-06
fix(worker): wait fetched jobs to be processed when closing

manast:
### Why

This change is necessary to avoid possible stalled jobs when closing a worker, since the close operation just stops processing the jobs that may be already in active status but have not yet started to be processed.

### How

Making sure that we wait for the jobs to complete before resolving the close promise.

_Any extra info here._

### DIFF
--- a/tests/test_events.ts
+++ b/tests/test_events.ts
@@ -29,6 +29,7 @@ describe('events', function () {
 
   beforeEach(async function () {
     queueName = `test-${v4()}`;
+    console.log('queueName', queueName);
     queue = new Queue(queueName, { connection, prefix });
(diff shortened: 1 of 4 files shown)


tests/test_events.ts
```diff
@@ -29,6 +29,7 @@ describe('events', function () {
 
   beforeEach(async function () {
     queueName = `test-${v4()}`;
+    console.log('queueName', queueName);
```
roggervalf, line 32: this line can be removed

roggervalf (approved):
Lgtm, just a minor styled comment

Issue #3062, opened 2025-02-08
[Bug]: Should Queue.getJobScheduler() return null/undefined when a job scheduler does not exist


mbakiev:
### Version

v5.40.2

### Platform

NodeJS

### What happened?

Should Queue.getJobScheduler() return null/undefined if the job scheduler ID does not exist? Seems misleading to return an object with the id/name.

How does one tell if a job scheduler exists or not based on the result?

### How to reproduce.

    const queue = new Queue('queue')
    const result = await queue.getJobScheduler("does-not-exist")
    console.log(result)

Produces:

    {
      key: 'does-not-exist',
      name: 'does-not-exist',
      id: null,
      endDate: null,
      tz: null,
      pattern: null,
      next: null
    }

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3065, opened 2025-02-08
fix(job-scheduler): return undefined in getJobScheduler when it does not exist

roggervalf:
### Why

  1. Why is this change necessary? If job scheduler does not exist, getJobScheduler method tries to split the id because of the old format of repeatable jobs. But as old format requires to have this id separated by :, we can also validate that the id does not contains :. Take in count that transformation of this data is being used for getJobSchedulers too.

### How

  1. How did you implement this? Change retuned type to undefined and when validating old format, make sure id/ key contains : before trying to transform it to old repeatable format

fixes https://github.com/taskforcesh/bullmq/issues/3062

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -283,7 +283,9 @@ export class JobScheduler extends QueueBase {
       };
     }
 
-    return this.keyToData(key, next);
+    if (key.includes(':')) {
+      return this.keyToData(key, next);
+    }
   }
 
   private keyToData(key: string, next?: number): JobSchedulerJson {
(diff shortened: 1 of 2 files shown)


src/classes/job-scheduler.ts
```diff
@@ -283,7 +283,9 @@ export class JobScheduler extends QueueBase {
       };
     }
 
-    return this.keyToData(key, next);
+    if (key.includes(':')) {
```
manast, line 286: As this can only happen with legacy repeatable jobs I think we should write a comment here to clarify that this keyToData thing is only here for legacy purposes and could be removed in a major release.

manast (approved):

src/classes/job-scheduler.ts
roggervalf: Sounds good

PR #2707, opened 2024-08-14
feat: version 6 (#2682)

roggervalf:
Branch to collect all breaking changes for v6

PR #3066, opened 2025-02-08
fix(job-scheduler): consider removing current job from wait, paused o…

roggervalf:
…r prioritized

### Why

### How

_Any extra info here._

### DIFF
--- a/src/commands/addJobScheduler-6.lua
+++ b/src/commands/addJobScheduler-6.lua
@@ -4,6 +4,10 @@
     Input:
       KEYS[1] 'repeat' key
       KEYS[2] 'delayed' key
+      KEYS[3] 'wait' key
+      KEYS[4] 'paused' key
+      KEYS[5] 'meta' key
+      KEYS[6] 'prioritized' key
       
       ARGV[1] next milliseconds
       ARGV[2] msgpacked options
@@ -22,6 +26,7 @@
 ]] local rcall = redis.call
 local repeatKey = KEYS[1]
 local delayedKey = KEYS[2]
+local prioritizedKey = KEYS[6]
 
 local nextMillis = ARGV[1]
 local jobSchedulerId = ARGV[3]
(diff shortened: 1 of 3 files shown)


src/commands/addJobScheduler-6.lua
```diff
@@ -4,6 +4,10 @@
     Input:
       KEYS[1] 'repeat' key
       KEYS[2] 'delayed' key
+      KEYS[3] 'wait' key
+      KEYS[4] 'paused' key
```
manast, line 8: We do not have pause key anymore, so why is this key needed?

src/commands/addJobScheduler-6.lua
```diff
@@ -37,14 +43,28 @@ local prefixKey = ARGV[6]
...
+    if rcall("EXISTS", nextDelayedJobKey) ~= 1 or currentJobId == nextDelayedJobId then
+        if rcall("ZSCORE", delayedKey, currentJobId) ~= false then
+            removeJob(currentJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", delayedKey, currentJobId)
+        elseif rcall("ZSCORE", prioritizedKey, currentJobId) ~= false then
+            removeJob(currentJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", prioritizedKey, currentJobId)
+        else
+            if isQueuePaused(KEYS[5]) then
+                if rcall("LREM", KEYS[4], 1, currentJobId) > 0 then
```
manast, line 59: This should not be needed as there is no pause key.

manast (commented):
Just a small comment regarding the paused key.

manast (approved):
I am seeing that we actually still have the pause key even if it is not useful anymore. I guess we need to set it for deprecation then.

roggervalf:
> I am seeing that we actually still have the pause key even if it is not useful anymore. I guess we need to set it for deprecation then.

It's being removed as part of next breaking change https://github.com/taskforcesh/bullmq/pull/2707

PR #3067, opened 2025-02-10
refactor: remove redundant AC polyfill and use built-in UUID generator

v1rtl:
### Why

1. AbortController has been available since Node 14.17, there is no longer a need to polyfill it. Node 12 has been EOL for quite some time and is considered dangerous to use. There is no longer a need for `uuid` as a package, since Node 16.7.0 the `crypto` module provides a safe generator for UUIDs.
2. Smaller supply chain, avoidance of downloading and loading unnecessary polyfills.

### How

Removed both of the packages and replaced them with native functionality.

Tests pass locally on Node 18 / 20 / 22. Can't test on Node 16 because corepack is not compatible with it.


manast:
Thanks. Always great to get rid of dependencies!

talentlessguy:
oops, forgot to replace uuid in tests, gimme a moment

manast:
it seems like there are still some references to the uuid module. Btw, we will need to decide if this can be merged in current major version or if we will need to make a breaking release as there are potential users out there using lower than 16 in production.

talentlessguy:
> it seems like there are still some references to the uuid module. Btw, we will need to decide if this can be merged in current major version or if we will need to make a breaking release as there are potential users out there using lower than 16 in production.

I'd say a breaking change would be better, as well as settings `engines.node` to `>=16.7`

manast:
I think the engines.node is the most established way of doing this.

talentlessguy:
@manast just added engines.node

talentlessguy:
Is there anything to be done on this PR besides conflict merging?

manast:
> Is there anything to be done on this PR besides conflict merging?

No, but it is a breaking change so we must wait until next major release.

v1rtl:
Is it worth fixing the merge conflicts?

roggervalf:
this was addressed as part of v5.76.2

v1rtl:
@roggervalf abort controller polyfill is still in deps: https://github.com/taskforcesh/bullmq/blob/a9e0a806d4fbacc526e276e34c6facf91f98b9c4/package.json#L68

roggervalf:
I forgot that this was taking care of node-abort-controller, still cannot removed It since we are supporting node 12 in our current versión and this would be a breaking change. Would you like to address the merge conflicts and then I can mark It for next breaking change or do you prefer to create another pr?

v1rtl:
@roggervalf sure, i'll make a new PR

v1rtl:
@roggervalf i'm not allowed to make a PR, says that owner of the repo limited to who can create PRs (only collaborators)

PR #3074, opened 2025-02-12
fix(drain): delay queuekey for redis cluster

harperKKK:
### Why

fix crossslot error when running queue.drain() in redis cluster

### How
replace '' key with queue.toKey('') to include hash tag

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -656,12 +656,12 @@ export class Scripts {
     const keys: (string | number)[] = [
       queueKeys.wait,
       queueKeys.paused,
-      delayed ? queueKeys.delayed : this.queue.toKey(''),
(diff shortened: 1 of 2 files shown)


manast (commented):
This code does not seem correct to me. "baseKey" is not the delayed key, so it does not make sense to use it for deleting delayed jobs.

manast:
Ok; I get the intent now. I think the the solution would be to pass a new ARGV with a boolean to express if the delayed jobs should also be deleted or not, and always pass the delayed key, that should work. It was an odd pattern to have a key that can sometimes be undefined anyway.

harperKKK:
i update the drain script using a new ARGV.

roggervalf (approved):
Lgtm

PR #3076, opened 2025-02-13
feat(queue): support removeGlobalConcurrency method

roggervalf:
### Why

1. Why is this change necessary? In order to revert a global concurrency setting

### How

  1. How did you implement this? Remove global concurrency option from meta

_Any extra info here._


manast (approved):
LGTM

Issue #3073, opened 2025-02-12
[Bug]: job.processedBy is undefined after second worker run


dimonnwc3:
### Version

v5.40.2

### Platform

NodeJS

### What happened?

When passing `name` to worker options, it's supposed to show this name on the `job.processedBy` property. It works for the first time but doesn't for all sequentially runs. It can be reproduced only when there is a timeout inside worker function.

### How to reproduce.

`queue.ts`
```
import { Queue } from "bullmq"

const q = new Queue("super-task", { connection: { url: "localhost" } })

while (true) {
  await q.add("super-task", {})
  await new Promise((res) => setTimeout(res, 5000))
}
```

`worker.ts`
```
import { Worker } from "bullmq"

new Worker(
  "super-task",
  async (job) => {
    console.log(job.processedBy) // <- job.processedBy is undefined after second run
    await new Promise((resolve) => setTimeout(resolve, 5000))
  },
  {
    connection: { url: "localhost" },
    name: "processedBy-custom-name",
  },
)

```

### Relevant log output

```shell
processedBy-custom-name
undefined
undefined
...
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3077, opened 2025-02-14
fix(job): set processedBy when moving job to active in moveToFinished

roggervalf:
### Why

  1. Why is this change necessary? processedBy attribute is not being set when moveToFinished script is called

### How

  1. How did you implement this? Pass worker name as part of options, this arg is already passed

fixes https://github.com/taskforcesh/bullmq/issues/3073


manast (approved):
LGTM

PR #3080, opened 2025-02-15
fix(worker): do not resume when closing

roggervalf:
### Why

  1. Why is this change necessary? When resuming, run method is called again. If there is no processor method it throws an error outside of the main promise. Also resuming is not needed as we are waiting for pending jobs after.

### How

  1. How did you implement this? Remove resume call when closing

_Any extra info here._


manast (approved):
LGTM

PR #3092, opened 2025-02-19
Cast delay_until to fix error when using retries

keeganmccallum:
### Why

When using retries, things function largely as expected, but throw a type error due to string delay_until passed to max()

### How

just cast to int, it works well and if statement checks for existence already, worst case will throw a better error, best case actually works

_Any extra info here._


roggervalf:
hey @keeganmccallum could you pls add a test case for this change, you can see a similar test case in https://github.com/taskforcesh/bullmq/blob/master/python/tests/worker_tests.py#L243

keeganmccallum:
done! is pretty well the same just the exponential case seesm to trigger the bug

python/bullmq/worker.py
```diff
@@ -144,7 +144,7 @@ def nextJobFromJobData(self, job_data = None, job_id: str = None, limit_until: i
                 self.blockUntil = 0
 
         if delay_until:
-            self.blockUntil = max(delay_until, 0) or 0
```
roggervalf, line 147: I tested your change but even with an int cast, I'm still seeing delay_until as an integer

keeganmccallum: Happened to me using dragonfly, may not be string in all cases but as safety should have the cast

roggervalf: hi @keeganmccallum, sorry for the delay. I created a pr where I was able to reproduce this issue. Soon I'll merge it

roggervalf:
pls try v2.12.1

PR #3108, opened 2025-02-25
fix(scheduler): remove multi when updating a job scheduler

roggervalf:
### Why

 1. Why is this change necessary? As part of giving support to serverless dbs

### How

  1. How did you implement this? Move addition of delayed job when updating job scheduler

_Any extra info here._

### DIFF
--- a/src/commands/includes/addJobFromScheduler.lua
+++ b/src/commands/includes/addJobFromScheduler.lua
@@ -1,0 +1,41 @@
+--[[
+  Add delay marker if needed.
+]]
+
+-- Includes
+--- @include "addDelayedJob"
+--- @include "addJobWithPriority"
+--- @include "isQueuePaused"
+--- @include "storeJob"
+
+local function addJobFromScheduler(jobKey, jobId, rawOpts, waitKey, pausedKey, metaKey, prioritizedKey,
+  priorityCounter, delayedKey, markerKey, eventsKey, name, maxEvents, timestamp, data, jobSchedulerId)
+  local opts = cmsgpack.unpack(rawOpts)
+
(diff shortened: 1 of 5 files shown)


src/commands/includes/addJobFromScheduler.lua
```diff
@@ -0,0 +1,41 @@
...
+  Add delay marker if needed.
+]]
+
+-- Includes
+--- @include "addDelayedJob"
+--- @include "addJobWithPriority"
+--- @include "isQueuePaused"
+--- @include "storeJob"
+
+local function addJobFromScheduler(jobKey, jobId, rawOpts, waitKey, pausedKey, metaKey, prioritizedKey,
```
manast, line 11: In time we will need to figure out some way to reduce the amount of arguments passed to this utility functions.

src/commands/updateJobScheduler-11.lua
```diff
@@ -43,19 +51,16 @@ local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
 if prevMillis ~= false then
   local currentDelayedJobId =  "repeat:" .. jobSchedulerId .. ":" .. prevMillis
 
-  if producerId == currentDelayedJobId then
+  if producerId == currentDelayedJobId and rcall("EXISTS", nextDelayedJobKey) ~= 1 then
```
manast, line 54: This seems to be new code "and rcall("EXISTS", nextDelayedJobKey) ~= 1" not sure which issue it fixes but It would be nice with a test case that proves it is needed.

manast (commented):
LGTM AFAICS, but I wrote a comment regarding a small change that seems to be a fix that would require a test case.

manast (approved):

PR #3111, opened 2025-02-26
fix(telemetry): fix span name for moveToFailed command

yesmeck:
### Why

The span name for failed jobs was `undefined`.

Sreenshot from our monitoring tool.
CleanShot 2025-02-26 at 14 05 51@2x

### How

Fixed the switch statement.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -824,7 +824,7 @@ export class Job<
         return 'delay';
       case 'retryJob':
         return 'retry';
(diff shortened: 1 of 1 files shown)


roggervalf:
Thank you @yesmeck, I'm taking your pr as reference. There are other changes to take in count so I'm creating a pr to fix

roggervalf:
Fixed in v5.41.6

PR #3112, opened 2025-02-26
fix(flow): consider delayed state when moving a parent to failed

roggervalf:
### Why

  1. Why is this change necessary? When dynamically add children jobs, a parent can be in delayed state when failParentOnFailure is activated

### How

  1. How did you implement this? Consider delayed state when moving a parent to failed

_Any extra info here._


manast (approved):
LGTM

PR #3113, opened 2025-02-26
fix(telemetry): fix span name for moveToFailed logic

roggervalf:
### Why

  1. Why is this change necessary? As reference https://github.com/taskforcesh/bullmq/pull/3111

### How

  1. How did you implement this? Send different span names depending on which operation happens when failing a job

_Any extra info here._


manast (approved):
LGTM

PR #3119, opened 2025-03-01
build and publish wheels

dimbleby:
direct invocation of setup.py is [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).  Build and publish wheels.


manast (approved):
LGTM. The PR needs to be squashed though as it does not follow conventional commits.

dimbleby:
I do not find it easy to guess what y'all would like the commit message to say - but I think maintainers have permission to edit such things!

manast:
@dimbleby don't worry, my message was intended to @roggervalf

PR #3127, opened 2025-03-09
perf(worker): optimize job retrieval for failed jobs in chunks

YotamZiv298:
### Why
The current implementation of `moveStalledJobsToWait` processes failed jobs by making individual `Job.fromId` calls within a loop. This can lead to performance bottlenecks when dealing with many stalled jobs, as each job requires a separate Redis operation.

This change is necessary to:
* Reduce the number of Redis operations
* Improve memory usage by preventing large arrays of pending promises
* Enhance performance when handling stalled jobs at scale

### How

* Implemented chunk processing for failed jobs using Promise.all
* Reduced number of Redis operations by batching job retrievals

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -1243,22 +1243,19 @@ will never work with more accuracy than 1ms. */
         });
 
         // Todo: check if there any listeners on failed event
-        const jobPromises: Promise<Job<DataType, ResultType, NameType>>[] = [];
-        for (let i = 0; i < failed.length; i++) {
-          jobPromises.push(
-            Job.fromId<DataType, ResultType, NameType>(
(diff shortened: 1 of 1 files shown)


manast (approved):
LGTM

YotamZiv298:
Weird, tests fail… I'll check it out later today.

manast:
@roggervalf it seems this PR introduces new flakiness in some unrelated tests, we need to decide if merging it as is or trying to find the root cause.

roggervalf (approved):
Lgtm

roggervalf:
> it seems this PR introduces new flakiness in some unrelated tests, we need to decide if merging it as is or trying to find the root cause.

it seems that they are unrelated to this change

Issue #3136, opened 2025-03-12
[Bug]: Job.getDependenciesCount() does not include failed children count


mbakiev:
### Version

v5.42.0

### Platform

NodeJS

### What happened?

When children fail with `ignoreDependencyOnFailure`, the failed children are not accounted for in the results of `getDependenciesCount()`.

The failed children count can be included in the `processed` count or a new counter `failed` should be added to the result of `getDependenciesCount()`

My current work around is to use `getFailedChildrenValues()` to find the failed children count.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3137, opened 2025-03-13
feat(job): support ignored and failed counts in getDependenciesCount

roggervalf:
### Why

1. Why is this change necessary? getDependenciesCount is not returning number for ignored or failed child states

### How

  1. How did you implement this? Create a new script to handle these numbers that are missing

ref https://github.com/taskforcesh/bullmq/issues/3136


manast (approved):
LGTM

PR #3138, opened 2025-03-13
fix: QueueEvent requires a queueName

emmanuelgeoffray:
### Why
There is an issue in the doc, the Quick Start example is not working for QueueEvents.
It is missing a queueName, a required parameter.

### How
Provide a queueName


manast:
Thanks for the PR. I changed it to use "my-queue-name" to make it more explicit.

PR #3145, opened 2025-03-14
fix(job-scheduler): add marker when upserting job scheduler if needed

manast:
### Why

When adding a new job scheduler with "every", the first job should be processed immediately, however this was not the case, as it was just moved to the wait list without adding a marker to wake up the workers.

### How

Used the standard methods we have to add jobs to the wait list.

_Any extra info here._


src/commands/includes/addJobFromScheduler.lua
```diff
@@ -22,15 +25,9 @@ local function addJobFromScheduler(jobKey, jobId, rawOpts, waitKey, pausedKey, m
...
-        -- LIFO or FIFO
-        local pushCmd = opts['lifo'] and 'RPUSH' or 'LPUSH'
-        rcall(pushCmd, pausedKey, jobId)
-      else
-        -- LIFO or FIFO
-        local pushCmd = opts['lifo'] and 'RPUSH' or 'LPUSH'
-        rcall(pushCmd, waitKey, jobId)
-      end
+      local pushCmd = opts['lifo'] and 'RPUSH' or 'LPUSH'
+      local target, isPausedOrMaxed = getTargetQueueList(metaKey, activeKey, waitKey, pausedKey)
```
roggervalf, line 29: it can be moved in line 24 to replace isQueuePaused call, also because getTargetQueueList verify activeKey

roggervalf (approved):
Lgtm

PR #3153, opened 2025-03-17
fix(scheduler): remove next delayed job when possible

roggervalf:
### Why

  1. Why is this change necessary? When the next delayed job id already exists, we cannot add a new record if this job is in any state different than delayed or any of the waiting states. In this case a duplicated event should be emitted and no update scheduler.

### How

  1. How did you implement this? Make sure that next delayed job id can be removed if it exists. If it cannot be removed, emit duplicated state.

_Any extra info here._


tests/test_job_scheduler.ts
```diff
@@ -293,11 +293,57 @@ describe('Job Scheduler', function () {
 
       await this.clock.tickAsync(ONE_MINUTE);
       const count = await queue.getJobCountByTypes('delayed', 'waiting');
-      expect(count).to.be.equal(2);
+      expect(count).to.be.equal(1);
 
       await worker.close();
     });
 
+    describe('when next delayed job already existed and it is not in waiting or delayed states', function () {
```
manast, line 301: existed -> exists

tests/test_job_scheduler.ts
```diff
@@ -293,11 +293,57 @@ describe('Job Scheduler', function () {
...
+        const token = 'my-token';
+
+        await worker.waitUntilReady();
+
+        const jobSchedulerId = 'test';
+        await queue.upsertJobScheduler(jobSchedulerId, {
+          every: ONE_MINUTE * 1,
+        });
+
+        const duplicating = new Promise<void>(resolve => {
```
manast, line 315: shouldn't the duplicated event also be emitted by the Queue class that did the actual upsert?

src/commands/addJobScheduler-11.lua
```diff
@@ -61,8 +62,43 @@ local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
...
-    if rcall("EXISTS", nextDelayedJobKey) ~= 1 or currentJobId == nextDelayedJobId then
+    if rcall("EXISTS", nextDelayedJobKey) == 1 then
+        if rcall("ZSCORE", delayedKey, nextDelayedJobId) ~= false then
+            removeJob(nextDelayedJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", delayedKey, nextDelayedJobId)
+        elseif rcall("ZSCORE", prioritizedKey, nextDelayedJobId) ~= false then
+            removeJob(nextDelayedJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", prioritizedKey, nextDelayedJobId)
+        else
+            if isQueuePaused(metaKey) then
```
manast, line 75: There is code duplication here, this could be improved by first determining the key from where to remove the job:
```lua
local pausedOrWaitKey = waitKey
if isQueuePaused(metaKey) then
  pasuedOrWaitKey = pausedKey
end
... etc
```

src/commands/addJobScheduler-11.lua
```diff
@@ -61,8 +62,43 @@ local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
 
 if prevMillis ~= false then
     local currentJobId = "repeat:" .. jobSchedulerId .. ":" .. prevMillis
+    local currentDelayedJobKey = schedulerKey .. ":" .. prevMillis
 
-    if rcall("EXISTS", nextDelayedJobKey) ~= 1 or currentJobId == nextDelayedJobId then
+    if rcall("EXISTS", nextDelayedJobKey) == 1 then
+        if rcall("ZSCORE", delayedKey, nextDelayedJobId) ~= false then
+            removeJob(nextDelayedJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", delayedKey, nextDelayedJobId)
```
manast, line 70: What if the job is in the delayed key due to a retry ?

src/commands/addJobScheduler-11.lua
```diff
@@ -61,8 +62,43 @@ local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
...
+        elseif rcall("ZSCORE", prioritizedKey, nextDelayedJobId) ~= false then
+            removeJob(nextDelayedJobId, true, prefixKey, true --[[remove debounce key]] )
+            rcall("ZREM", prioritizedKey, nextDelayedJobId)
+        else
+            if isQueuePaused(metaKey) then
+                if rcall("LREM", pausedKey, 1, nextDelayedJobId) > 0 then
```
manast, line 76: needing to remove here in the hope that the job will be added later hurts my eyes :sweat_smile:

tests/test_job_scheduler.ts
roggervalf: we don't do that even for regular jobs, this event is only handled by QueueEvents

src/commands/addJobScheduler-11.lua
roggervalf: it should not be possible as when a job is moved to active for the first time, we add the next delayed job. That means retried jobs won't be affected

src/commands/addJobScheduler-11.lua
roggervalf: I know but a job that it's being added later can have different options, data, etc, so it can be in a different state as well

manast (approved):
LGTM

PR #3157, opened 2025-03-18
feat(prometheus export): expose global variables

martin-cycle:
expose global variables in prometheus export

### Why
Right now, it is not possible to identify the environment the prometheus data is extracted from. In aggregated Grafana dashboards, it makes the data impractical.

### How
I added a `globalVariables` input in the `exportPrometheusMetrics` function


manast:
Thanks for the PR, looks good. I will prepare the documentation for this new feature.

PR #3174, opened 2025-03-25
chore: fix osv-scanner workflow

manast:
### Why

The osv-scanner workflow was failing due to an error in the yaml file.

### How

Fixed the error (hopefully)

_Any extra info here._


package.json
```diff
@@ -82,43 +82,48 @@
...
+    "semantic-release": "^24.2.3",
     "sinon": "^18.0.1",
     "test-console": "^2.0.0",
-    "ts-mocha": "^10.0.0",
+    "ts-mocha": "^11.1.0",
     "ts-node": "^10.7.0",
     "typedoc": "^0.25.0",
     "typescript": "^4.7.4"
   },
+  "resolutions": {
```
roggervalf, line 122: why is this attribute needed? I don't see it as part of package.json properties https://docs.npmjs.com/cli/v11/configuring-npm/package-json

manast: Because there are some packages that have vulnerabilities but their parent packages have not been upgraded to address these vulnerabilities, so we must override those packages to eliminate the vulnerabilities.

roggervalf: ok I found it, [resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-how-to-use-it) is the same as [overrides](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides). They are use for the same purpose, but the one used here is for yarn, the second one when using npm when installing dependencies

roggervalf (approved):

PR #3177, opened 2025-03-26
feat: add deduplicated job id to the deduplicated event

manast:
### Why

So that it is possible to get a notification of the job id that was actually discarded due to deduplication.

### How

passing the job id to the deduplication event.

_Any extra info here._

### DIFF
--- a/tests/test_events.ts
+++ b/tests/test_events.ts
@@ -610,6 +610,49 @@ describe('events', function () {
   });
 
   describe('when job is deduplicated when added again with same debounce id', function () {
+    it('emits deduplicated event', async function () {
+      const testName = 'test';
+      const worker = new Worker(
+        queueName,
+        async () => {
+          await delay(50);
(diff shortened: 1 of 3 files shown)


tests/test_events.ts
```diff
@@ -610,6 +610,49 @@ describe('events', function () {
   });
 
   describe('when job is deduplicated when added again with same debounce id', function () {
+    it('emits deduplicated event', async function () {
+      const testName = 'test';
+      const worker = new Worker(
+        queueName,
+        async () => {
+          await delay(50);
+          await queue.add(testName, { foo: 'bar' }, { jobId: 'a1' });
```
roggervalf, line 619: is this addition needed? It's not using deduplication opts so the event triggered by this line would be duplicated.

roggervalf (approved):
Lgtm, just a minor comment on the test case

tests/test_events.ts
manast: thats not needed, some hallucination from Grok. Removing...

PR #3181, opened 2025-03-29
feat(flows): add continueParentOnFailure option

manast:
### Why

It is useful in some occasions to let a parent job be processed as soon as one child fails. This could be to do clean up operations or something else. It is just a robust way to allow "post-processing" for children that fail.

### How

Introduced a new job option "continueParentOnFail"

_Any extra info here._


src/commands/includes/moveParentToWait.lua
```diff
@@ -0,0 +1,46 @@
+--[[
+  Validate and move parent to a wait status (wait, prioritized or delayed)
+]]
+--- @include "addDelayMarkerIfNeeded"
+--- @include "addJobInTargetList"
+--- @include "addJobWithPriority"
+--- @include "isQueuePausedOrMaxed"
+--- @include "getTargetQueueList"
+local function moveParentToWait(parentQueueKey, parentKey, parentId, timestamp)
+    rcall("ZREM", parentQueueKey .. ":waiting-children", parentId)
```
roggervalf, line 10: this line is not needed as in moveParentToFailedIfNeeded is done in line 30

src/commands/moveStalledJobsToWait-9.lua
```diff
@@ -87,39 +88,37 @@ if (#stalling > 0) then
...
-                                    jobKey,
-                                    timestamp
-                                )
+
+                                if opts['fpof'] then
+                                    moveParentToFailedIfNeeded(parentData['queueKey'],
+                                        parentData['queueKey'] .. ':' .. parentData['id'], parentData['id'], jobKey,
+                                        timestamp)
+                                elseif opts['cpof'] then
+                                    moveParentToWait(parentData['queueKey'], parentKey, parentData['id'], timestamp)
```
roggervalf, line 110: in order to not try to remove the parent twice from waiting-children, we can pass a flag to let this method know that removal was already done as in moveParentToFailedIfNeeded we are doing that removal too. Also as this calls is separated from moveParentToFailedIfNeeded. When removing from an state, you should make sure that removal is == 1, because parent cannot only be in waiting-children state

src/commands/moveToWaitingChildren-8.lua
```diff
@@ -75,18 +76,22 @@ if rcall("EXISTS", jobKey) == 1 then
...
+        if opts['fpof'] then
+          moveParentToFailedIfNeeded(
+              parentData['queueKey'],
+              parentData['queueKey'] .. ':' .. parentData['id'],
+              parentData['id'],
+              jobKey,
+              timestamp
+          )
+        elseif opts['cpof'] then
+          moveParentToWait(parentData['queueKey'], parentKey, parentData['id'], timestamp)
```
roggervalf, line 93: same as above

tests/test_flow.ts
```diff
@@ -3187,6 +3187,337 @@ describe('flows', () => {
...
+          connection,
+          prefix,
+        },
+      );
+
+      await processing;
+
+      await parentWorker.close();
+      await childrenWorker.close();
+      await flow.close();
```
roggervalf, line 3249: need to call after this line
`await removeAllQueueData(new IORedis(redisHost), parentQueueName);`
to clean extra keys for that new queue

tests/test_flow.ts
```diff
@@ -3187,6 +3187,337 @@ describe('flows', () => {
...
+            prefix,
+          },
+        );
+      });
+
+      await processing;
+      await parentWorker.close();
+      await waitingChildren;
+      await childrenWorker.close();
+      await flow.close();
```
roggervalf, line 3340: same as above

tests/test_flow.ts
```diff
@@ -3187,6 +3187,337 @@ describe('flows', () => {
...
+        });
+      });
+
+      await waitingFailedChildren;
+
+      await processing;
+
+      await parentWorker.close();
+      await childrenWorker.close();
+      await flow.close();
```
roggervalf, line 3428: same as above

tests/test_flow.ts
```diff
@@ -3187,6 +3187,337 @@ describe('flows', () => {
...
```
roggervalf, line 3517: same as above to clean extra new keys

tests/test_flow.ts
```diff
@@ -3187,6 +3187,337 @@ describe('flows', () => {
...
+    });
+
+    it('should move the parent to prioritized after a child fails', async () => {
```
roggervalf, line 3431: this test is flaky, it could be related to one of my previous comment where need to check that worker was moved from waiting-children in moveParentToWait state

roggervalf (commented):
Something else to test, is when parent is dynamically adding children and need to call moveToWaitingChildren

src/commands/includes/moveParentToWait.lua
manast: Are you checking the new code? there is no such line in moeParentToFailedIfNeeded

src/commands/moveToFinished-14.lua
```diff
@@ -96,7 +98,8 @@ if rcall("EXISTS", jobIdKey) == 1 then -- // Make sure job exists
     local maxAge = opts['keepJobs']['age']
 
     -- Make sure it does not have pending dependencies
-    if ARGV[5] == "completed" and rcall("SCARD", jobIdKey .. ":dependencies") ~= 0 then
+    if ARGV[5] == "completed" and not rcall("HGET", jobIdKey, "igdp") and rcall("SCARD", jobIdKey .. ":dependencies") ~=
```
roggervalf, line 101: something that we need to think of is that children under dependencies set are considered hard dependencies. So a parent should wait for these children before moving it to completed. How we are avoiding this check, it's when children are moving to other states like parent:failed or parent:unsuccessful. When a parent is moved to active, purpose would be to do some clean up operations for example as mentioned in description. But if users forget to do this operations, pending dependencies will be processed anyway and kept in queues. This error should serve to let users know that there are still unprocessed children. So there are 2 options:
1: remove this skip operation and let parent fail if post processing did not do anything with hard dependencies. One extra case would be to do some operation after a child failure and then move the parent to waiting-children again to wait for the remaining hard dependencies.
2. clarify that hard dependencies will be soft dependencies, users should validate take this case entirely as parent won't let you know that they forgot to do clean up operations. Need to validate what will happen if users move job back to waiting children after some after operation when one of the child fail with this option

manast: The thing is that if you are using the new option continueParentOnFailure, it needs to consider the dependencies soft, as it would not make sense otherwise.

roggervalf (approved):
Lgtm

PR #3182, opened 2025-03-29
fix(flow): validate pending dependencies before removing lock

roggervalf:
### Why

  1. Why is this change necessary? When a parent throws an error because it has pending dependencies, it's removing the lock first. That causes that moveToFailed throws a different error than pending dependencies, it throws error that lock is missing.

### How

  1. How did you implement this? Move pending dependencies validation before releasing lock, that way the error is properly propagated to the parent

_Any extra info here._


manast (approved):
lgtm

PR #3187, opened 2025-04-01
feat(updateProgress): allow more types to be used as progress

nullndr:
### Why

Currently `Job.updateProgress()` requires his argument to be a `number | object` union, this does not let us use other primitives types like `string` and `boolean`.

Since the progress is serialized using `JSON.stringify()` we can have a more permissive type for the progress.

### How

All references to the `updateProgress()` method and the `progress` property are now a `number | object | string | boolean` union.

The implementation of the serialization process has not changed, since `JSON.stringify()` already handles both `string` and `boolean`.


manast (approved):
LGTM

PR #3190, opened 2025-04-02
feat: add removeUnprocessedChildren

manast:
### Why

This method is useful so that combined with "continueParentOnFailure" the user can remove the children that has not been processed in order to avoid processing jobs from a parent that will not care anymore of the children results.

### How

Similar to our current remove job, but ignoring active, completed and failed children.

_Any extra info here._

### DIFF
--- a/src/commands/includes/isLocked.lua
+++ b/src/commands/includes/isLocked.lua
@@ -5,6 +5,7 @@
   returns:
     boolean
 ]]
+--- @include "destructureJobKey"
 
 local function isLocked( prefix, jobId, removeChildren)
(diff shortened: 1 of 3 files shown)


src/commands/includes/isLocked.lua
```diff
@@ -5,6 +5,7 @@
   returns:
     boolean
 ]]
+--- @include "destructureJobKey"
```
roggervalf, line 8: how we didn't notice that this include was missing. Maybe because it was included in other includes where this one was included

roggervalf (approved):
Lgtm

PR #3194, opened 2025-04-04
feat(queue): add getIgnoredChildrenFailures method

roggervalf:
### Why

  1. Why is this change necessary? In order to use a better naming for this method

### How

  1. How did you implement this? Add new method, add deprecation warning. Also this pr fixes the configuration of ESLint, as v9 has breaking changes

_Any extra info here._


src/classes/job-scheduler.ts
```diff
@@ -77,7 +77,10 @@ export class JobScheduler extends QueueBase {
     // Check if we reached the end date of the repeatable job
     let now = Date.now();
     const { endDate } = repeatOpts;
-    if (!(typeof endDate === undefined) && now > new Date(endDate!).getTime()) {
+    if (
+      !(typeof endDate === 'undefined') &&
```
roggervalf, line 81: this is a current issue when evaluating this value

src/classes/repeat.ts
```diff
@@ -52,7 +52,10 @@ export class Repeat extends QueueBase {
     // Check if we reached the end date of the repeatable job
     let now = Date.now();
     const { endDate } = repeatOpts;
-    if (!(typeof endDate === undefined) && now > new Date(endDate!).getTime()) {
+    if (
+      !(typeof endDate === 'undefined') &&
```
roggervalf, line 56: same as above

src/classes/job.ts
```diff
@@ -952,7 +952,18 @@ export class Job<
   }
 
   /**
-   * Get this jobs children failure values if any.
+   * Get jobs children failure values that were ignored if any.
+   *
+   * @returns Object mapping children job keys with their failure values.
+   */
+  async getIgnoredChildrenValues(): Promise<{ [jobKey: string]: string }> {
+    return this.getFailedChildrenValues();
```
manast, line 960: I think we should also rename the underlying method as it looks a but confusing otherwise.

roggervalf: underlying method is just a call directly to Redis, no other method coming from custom scripts

roggervalf: ok, I just copy the same logic as the another method, so reusing a method is not needed. There are only 2 lines, it should be fine

src/classes/job.ts
```diff
@@ -984,7 +984,20 @@ export class Job<
   }
 
   /**
-   * Get this jobs children failure values if any.
+   * Get job's children failure values that were ignored if any.
```
manast, line 987: In this description it would be important to explain what are "ignored" children in this context, like what are ignored children, how are they generated to begin with?

manast (approved):
can be merged after improving the documentation.

manast (approved):
LGTM

PR #3195, opened 2025-04-04
fix(job-scheduler): fix endDate presence validation

roggervalf:
### Why

1. Why is this change necessary? Because typeof returns a string, it won't return undefined. This pr fixes endDate presence validation

### How

  1. How did you implement this? Instead of checking that is not undefined, just check it's presence

_Any extra info here._

### DIFF
--- a/src/classes/async-fifo-queue.ts
+++ b/src/classes/async-fifo-queue.ts
@@ -8,7 +8,7 @@ class Node<T> {
 }
 
 class LinkedList<T> {
-  length: number = 0;
+  length = 0;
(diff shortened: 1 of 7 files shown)


src/classes/async-fifo-queue.ts
```diff
@@ -8,7 +8,7 @@ class Node<T> {
 }
 
 class LinkedList<T> {
-  length: number = 0;
+  length = 0;
```
roggervalf, line 11: other changes were applied by ESLint

manast (approved):
good catch

PR #3198, opened 2025-04-04
fix(FlowProducer): fix the queueName otel attribute when using addNode

runia1:
### Why

1. This fixes a bug where the job name was getting set as the queue name in span attributes.

### How

1. I implemented this by switching it to use the correct value.

flow-producer_ts__Working_Tree___flow-producer_ts__—_bullmq

Example in Grafana UI where the attribute is incorrectly set to the job name:
Screenshot 2025-04-04 at 12 37 19 PM


wtkersting (approved):
Checks out - queueName should be queueName

roggervalf (approved):
Lgtm

PR #3201, opened 2025-04-05
fix(flow): remove job from dependencies when fpof or cpof

roggervalf:
### Why

  1. Why is this change necessary? When a job fails with fpof or copf, we were not removing child from dependencies. Children should live only in one of these parent child states: dependencies, failed, unsuccessful or processed

### How

  1. How did you implement this? Make sure to remove child from dependencies when it's moved to another parent child key

_Any extra info here._


manast (approved):
LGTM

Issue #3204, opened 2025-04-08
[Bug]: split redis version error


XianZhengquan:
### Version

v5.47.3

### Platform

NodeJS

### What happened?

```
private async getRedisVersion() {
    const doc = await this._client.info();
    const redisPrefix = 'redis_version:';
    const maxMemoryPolicyPrefix = 'maxmemory_policy:';
    const lines = doc.split('\r\n');
    let redisVersion;

    for (let i = 0; i < lines.length; i++) {
      if (lines[i].indexOf(maxMemoryPolicyPrefix) === 0) {
        const maxMemoryPolicy = lines[i].substr(maxMemoryPolicyPrefix.length);
        if (maxMemoryPolicy !== 'noeviction') {
          console.warn(
            `IMPORTANT! Eviction policy is ${maxMemoryPolicy}. It should be "noeviction"`,
          );
        }
      }

      if (lines[i].indexOf(redisPrefix) === 0) {
        redisVersion = lines[i].substr(redisPrefix.length);
      }
    }

    return redisVersion;
  }
```
The obtained doc is not correctly segmented by \r\n, resulting in the inability to obtain the corresponding version information.

to fix
```
const lines = doc.split(/\r?\n/);
```

### Relevant log output

```shell
TypeError: Invalid version. Must be a string. Got type "object".
    at new SemVer (/Users/yasuo/Desktop/workspace/jojo/qa-workbench-service/node_modules/.pnpm/semver@7.7.1/node_modules/semver/classes/semver.js:20:13)
    at compare (/Users/yasuo/Desktop/workspace/jojo/qa-workbench-service/node_modules/.pnpm/semver@7.7.1/node_modules/semver/functions/compare.js:3:3)
    at Object.lt (/Users/yasuo/Desktop/workspace/jojo/qa-workbench-service/node_modules/.pnpm/semver@7.7.1/node_modules/semver/functions/lt.js:2:29)
    at isRedisVersionLowerThan (/Users/yasuo/Desktop/workspace/jojo/qa-workbench-service/node_modules/.pnpm/bullmq@5.46.1/node_modules/bullmq/src/utils.ts:228:17)
    at RedisConnection.init (/Users/yasuo/Desktop/workspace/jojo/qa-workbench-service/node_modules/.pnpm/bullmq@5.46.1/node_modules/bullmq/src/classes/redis-connection.ts:256:34)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

[...]

PR #3208, opened 2025-04-10
fix: fix version lines split

XianZhengquan:
normalize line splitting for cross-platform compatibility

#3204

### Why

### How

_Any extra info here._

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -548,7 +548,7 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   }
 
   private parseClientList(list: string, matcher: (name: string) => boolean) {
-    const lines = list.split('\n');
+    const lines = list.split(/\r?\n/);
     const clients: { [index: string]: string }[] = [];
 
     lines.forEach((line: string) => {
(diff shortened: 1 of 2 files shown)


src/classes/queue-getters.ts
```diff
@@ -548,7 +548,7 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   }
 
   private parseClientList(list: string, matcher: (name: string) => boolean) {
-    const lines = list.split('\n');
+    const lines = list.split(/\r?\n/);
```
roggervalf, line 551: there are some test cases where we use getWorkers methods. It can be validated that actual array still has valid strings

roggervalf: https://github.com/taskforcesh/bullmq/blob/165b6a1fb9873c86e9455922ecc507bd736408e8/tests/test_getters.ts#L73 for reference

XianZhengquan: The test case has been run.

```
/tests/test_connection.ts
```

XianZhengquan:
@joebowbeer 
can merge?

Issue #3211, opened 2025-04-10
Connection type should allow string


jscheel:
https://github.com/taskforcesh/bullmq/blob/75d88780aec86465efd4ed5ad54f4c6b5afe0b49/python/bullmq/types/queue_options.py#L12

Should be `dict[str, Any] | redis.Redis | str` because the code currently supports all three: https://github.com/taskforcesh/bullmq/blob/75d88780aec86465efd4ed5ad54f4c6b5afe0b49/python/bullmq/redis_connection.py#L25

PR #3212, opened 2025-04-11
fix(connection): add str type in connection option [python]

roggervalf:
### Why

  1. Why is this change necessary? Connection is supporting strings

### How

  1. How did you implement this? Add str as supported type

fixes https://github.com/taskforcesh/bullmq/issues/3211


manast (approved):
LGTM

manast:
Maybe we need a small test case where we pass a string.

manast (changes requested):
Maybe we need a small test case where we pass a string.

manast (approved):
LGTM

manast (approved):

PR #3220, opened 2025-04-16
refactor: expose opts decode and encode maps for extension

roggervalf:
### Why

  1. Why is this change necessary? In order to allow extension when new options are added in pro versions

### How

_Any extra info here._


manast (approved):
LGTM

PR #3225, opened 2025-04-17
test(clean): fix flaky test when checking keys quantity

roggervalf:
### Why

  1. Why is this change necessary? This is a flaky test, where keys could be greater than 7. Instead we can check those keys that we are interested - those related to the parent job that it's being removed

### How

  1. How did you implement this? Filter keys to only include the ones related to the parent job

_Any extra info here._

### DIFF
--- a/tests/test_clean.ts
+++ b/tests/test_clean.ts
@@ -486,6 +486,8 @@ describe('Cleaner', () => {
           await queue.clean(0, 0, 'failed');
 
           const client = await queue.client;
+          // only checks if there are keys under job key prefix
+          // this way we make sure that all of them were removed
           const keys = await client.keys(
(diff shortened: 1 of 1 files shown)


tests/test_clean.ts
```diff
@@ -486,10 +486,11 @@ describe('Cleaner', () => {
           await queue.clean(0, 0, 'failed');
 
           const client = await queue.client;
-          const keys = await client.keys(`${prefix}:${queue.name}:*`);
+          const keys = await client.keys(
```
manast, line 489: I think a comment here would be necessary to understand the intend of this code.

manast (changes requested):

manast (approved):
LGTM

Issue #3215, opened 2025-04-14
[Bug]: removeOnFail on parent job only works if set during job publish


krukid:
### Version

v5.45.2

### Platform

NodeJS

### What happened?

`removeOnFail` on parent job only works if set during job publish - the same option that is set on worker level has no effect.
`removeOnComplete` set on worker level is respected in the same scenario.

A relevant semi-recent fix that might be missing support for this: https://github.com/taskforcesh/bullmq/pull/2947

### How to reproduce.

```js
import { Redis } from 'ioredis';
import { FlowProducer, Worker } from 'bullmq';

const redis = new Redis({ host: 'redis-host', maxRetriesPerRequest: null });

new Worker('demo-flow-leaf', async () => {
    // see comments below about how `removeOnFail` affects parent job removal
    throw new Error('fail');
}, {
    connection: redis,
    removeOnComplete: { count: 3 }, // this will be respected
    removeOnFail: { count: 2 }, // this will be respected
});

new Worker('demo-flow-child', async () => {
}, {
    connection: redis,
    removeOnComplete: { count: 3 }, // this will be respected
    removeOnFail: { count: 2 }, // will NOT be respected
});

new Worker('demo-flow-root', async () => {
}, {
    connection: redis,
    removeOnComplete: { count: 3 }, // this will be respected
    removeOnFail: { count: 2 }, // will NOT be respected
});

new FlowProducer({ connection: redis }).add(
    {
        name: 'demo-flow-root',
        queueName: 'demo-flow-root',
        opts: {
            // removeOnFail: { count: 2 }, // this will be respected
        },
        children: [
            {
                name: 'demo-flow-child',
                queueName: 'demo-flow-child',
                opts: {
                    failParentOnFailure: true,
                    // removeOnFail: { count: 2 }, // this will be respected
                },
                children: [
                    {
                        name: 'demo-flow-leaf',
                        queueName: 'demo-flow-leaf',
                        opts: {
                            failParentOnFailure: true
                        }
                    }
                ]
            },
        ],
    }
)
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

roggervalf:
[...]

PR #3228, opened 2025-04-17
perf(flow): change parent failure in a lazy way

roggervalf:
### Why

  1. Why is this change necessary? Instead of trying to move parent jobs recursively to failed status, a lazy failed process will be triggered. Failed parent jobs should instead be moved to be processed by the parent workers. In this way workers can also handle failed events from these jobs. This is not only more elegant and is a better design, but it is also a more scalable approach than our current implementation.
We introduce a new job internal attribute, which is called deferredFailure used to identify that a job must be moved to failed when it's eventually picked by a worker.

### How

1. How did you implement this? 
Move parent to wait when a child with failParentOnFailure true is triggered and set deferredFailure attribute with a special error message. When the worker picks this job, it will call detect the new attribute and call moveToFailed script with this deferredFailure as an UnrecoverableError. Something to take into account is that if the job is already in wait, prioritized, active or failed state, the job will not  be affected by this new logic. The job is moved only if it's in waiting-children or a delayed state (setting delay to 0 in this case to process this job as soon as possible)

ref https://github.com/taskforcesh/bullmq/issues/3215

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -139,6 +139,11 @@ export class Job<
   failedReason: string;
 
   /**
+   * Reason for failing in a lazy way.
+   */
+  lazyFailedReason: string;
+
+  /**
    * Timestamp for when the job finished (completed or failed).
    */
   finishedOn?: number;
@@ -384,6 +389,10 @@ export class Job<
     job.attemptsMade = parseInt(json.attemptsMade || json.atm || '0');
 
     job.stalledCounter = parseInt(json.stc || '0');
+
+    if (json.lfr) {
+      job.lazyFailedReason = json.lfr;
+    }
 
     job.stacktrace = getTraces(json.stacktrace);
 
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -139,7 +139,8 @@ export class Job<
   failedReason: string;
 
   /**
-   * Deferred failure to move a job to failed.
+   * Deferred failure. Stores a failed message and marks this job to be failed directly
+   * as soon as the job is picked up by a worker, and using this string as the failed reason.
    */
   deferredFailure: string;
 
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -139,9 +139,9 @@ export class Job<
   failedReason: string;
 
   /**
-   * Reason for failing in a lazy way.
+   * Deferred failure to move a job to failed.
    */
-  lazyFailedReason: string;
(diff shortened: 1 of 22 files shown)


tests/test_connection.ts
```diff
@@ -117,36 +117,40 @@ describe('RedisConnection', () => {
   });
 
   describe('Worker', () => {
-    it('initializes blockingConnection with blocking: true', () => {
+    it('initializes blockingConnection with blocking: true', async () => {
       const worker = new Worker('test', async () => {}, { connection: {} });
       expect((<any>worker).blockingConnection.extraOptions.blocking).to.be.true;
+      await worker.close();
```
roggervalf, line 123: need to close pending connection instances in order to prevent a flaky tests where stderr output is validated

src/classes/job.ts
```diff
@@ -138,6 +138,11 @@ export class Job<
    */
   failedReason: string;
 
+  /**
+   * Deferred failure to move a job to failed.
```
manast, line 142: Improve this documentation with:
Deferred failure. Stores a failed message and marks this job to be failed directly as soon as the job is picked up by a worker, and using this string as the failed reason.

src/classes/worker.ts
```diff
@@ -906,6 +907,12 @@ will never work with more accuracy than 1ms. */
 
         try {
           jobsInProgress.add(inProgressItem);
+          if (job.deferredFailure) {
```
manast, line 910: Shouldn't we skip adding the job to jobsInProgress? we should have a test case that covers this need. As it looks now the job is removed from this set only in handleFailed and handleCompleted

roggervalf: processJob method is wrapped into retryIfFailed method. That means if a disconnection happens before failing this job and we do not add job into jobsInProgress, lock can be lost and we won't try to renew it. So It should be fine to add it in jobsInProgress

manast: ok, but when is it removed from jobsInProgress?

roggervalf: I could validate it now, it's being removed just in the beginning of calling handleFailed https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L857. So actually, in this case is not necessary to add it in jobsInProgress as it would be removed just after handleFailed. I'm moving that line only to be executed when processor callback is used

manast (approved):
LGTM

PR #3229, opened 2025-04-17
docs: sync docs-api branch with master

roggervalf:
### Why

  1. Why is this change necessary? Keep docs-api branch updated with master to prepare to be fixed in next pre

### How

_Any extra info here._


manast:
something does not seem right with this PR.

roggervalf:
> something does not seem right with this PR.

It's still a draft, for now it's being updated with master

PR #3231, opened 2025-04-20
feat(job): implement moveToWaitingChildren for sandboxed jobs

### LINKED ISSUE #3230: Implementing moveToWaitingChildren in SandboxedJobs
I discovered this page (https://docs.bullmq.io/patterns/process-step-jobs#waiting-children) while looking for a solution to a problem I'm working on and thought the `moveToWaitingChildren` would be ideal for what I'm doing. I kept getting an error though saying `job.moveToWaitingChildren is not a function` and upon some digging discovered that the `SandboxedJob` interface does not define it and my worker is running a sandboxed processor.

I'm not sure if there is a technical limitation for why sandboxed jobs can not implement the `moveToWaitingChildren` function but it would be nice if it was possible to use this function from within a sandboxed processor/job.

I noticed that the implementation of `moveToWaitingChildren` is pretty straightforward so I imported `Scripts` into my job and then tried
```
const scripts = new Scripts(queue);
scripts.moveToWaitingChildren(jobId, token)
```
But it just hangs for a minute and then I get a permanent failure on the job.

shichongrui:
### Why
`moveToWaitingChildren` is not implemented on sandboxed jobs. The docs don't explain this and from what I could see there was no reason for it not to be other than that it just wasn't implemented there.

### How
I followed the same approach that the `moveToDelayed` function uses in a sandboxed job to send commands back to the parent process to actually execute the functionality.

I did not see any tests for `moveToDelayed` that tested from within a sandboxed process so I wasn't sure if there was somewhere to add tests for this functionality.

I've done a `yarn pack` and then untarred it in a local bullmq project and confirmed that it was working though.

Fixes #3230

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -156,6 +156,15 @@ export class ChildProcessor {
         });
       },
       /*
+       * Proxy `moveToWaitingChildren` function.
+       */
+      moveToWaitingChildren: async (token?: string) => {
+        await send({
+          cmd: ParentCommand.MoveToWaitingChildren,
+          value: { token },
+        });
+      },
+      /*
        * Proxy `updateData` function.
        */
       updateData: async (data: any) => {
--- a/src/classes/sandbox.ts
+++ b/src/classes/sandbox.ts
@@ -52,6 +52,9 @@ const sandbox = <T, R, N extends string>(
                       msg.value?.token,
                     );
                     break;
+                  case ParentCommand.MoveToWaitingChildren:
+                    await job.moveToWaitingChildren(msg.value?.token);
+                    break;
                   case ParentCommand.Update:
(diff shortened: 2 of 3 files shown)


manast:
Not having a test for moveToDelayed was definitely an overlook from out part, we will fix it asap.

manast:
I actually checked in the codebase and there is indeed a test for moveToDelayed, you can use it for inspiration for your moveToWaitingChildren: https://github.com/taskforcesh/bullmq/blob/master/tests/test_sandboxed_process.ts#L892

pkudinov:
just faced the same issue today. Is there any plan to finish this PR @shichongrui ? I see @ocke had plans to work on it as well

UPD: I went ahead and implemented it here: #3389

Issue #3213, opened 2025-04-13
job.getDependencies doesn't return ignored child jobs


bubike:
Maybe I am missing something but there are two ways to react to dependecy failure, if I don't want fail the parent:
1. use removeDependencyOnFailure
2. use ignoreDependencyOnFailure

If I use the latter, I expect the failed children to be included in the dependencies (using **job.getDependencies**), but they aren't. This has an effect on **getFlow**, which doesn't return the full tree and ignores failed jobs (these are usually the ones you are most interested in).

Also, **job.getDependenciesCount** returns **all** processed, unprocessed, ignored and failed and filter can be specified. So why doesn't the same apply for getDependencies? 

I want to be able to see the whole tree including the failed jobs in order to keep track of what went ok and what failed, while not failing the parent, as it may be able to finish without some of the failed children job data.

Am I understanding the framework incorrectly or can this be fixed? Or is there another way to achieve what I want?

bubike:
Thanks for the clarification @roggervalf. I achieved the same effect with a little workaround but this enhancement would be very useful.

PR #3238, opened 2025-04-23
feat(flow): support ignored children in getFlow and getDependencies methods

roggervalf:
### Why

1. Why is this change necessary? getFlorw and getDependencies methods are missing to retrieve ignored children jobs
  
### How

  1. How did you implement this? Add extra logic on these methods to retrieve data from jobKey:failed

ref https://github.com/taskforcesh/bullmq/issues/3213

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1008,8 +1008,9 @@ export class Job<
    * on processed/unprocessed dependencies, since v7.2 you must consider that count
    * won't have any effect until processed/unprocessed dependencies have a length
    * greater than 127
-   * @see https://redis.io/docs/management/optimization/memory-optimization/#redis--72
-   * @returns dependencies separated by processed and unprocessed.
+   * @see {@link https://redis.io/docs/management/optimization/memory-optimization/#redis--72}
+   * @see {@link https://docs.bullmq.io/guide/flows#getters}
+   * @returns dependencies separated by processed, unprocessed and ignored.
    */
   async getDependencies(opts: DependenciesOpts = {}): Promise<{
     nextIgnoredCursor?: number;
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1012,6 +1012,8 @@ export class Job<
    * @returns dependencies separated by processed and unprocessed.
    */
   async getDependencies(opts: DependenciesOpts = {}): Promise<{
+    nextIgnoredCursor?: number;
+    ignored?: Record<string, any>;
     nextProcessedCursor?: number;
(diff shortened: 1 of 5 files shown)


src/interfaces/minimal-job.ts
```diff
@@ -21,6 +21,10 @@ export interface MoveToWaitingChildrenOpts {
 }
 
 export interface DependenciesOpts {
+  ignored?: {
```
manast, line 24: This interface should be documented.

src/classes/job.ts
```diff
@@ -1005,32 +1005,41 @@ export class Job<
    * @returns dependencies separated by processed and unprocessed.
    */
   async getDependencies(opts: DependenciesOpts = {}): Promise<{
+    nextIgnoredCursor?: number;
```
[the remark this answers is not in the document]
manast: Maybe include a small example as it is quite an advanced method that supports pagination, etc.

src/classes/job.ts
```diff
@@ -1005,32 +1005,41 @@ export class Job<
...
+    ignored?: Record<string, any>;
     nextProcessedCursor?: number;
     processed?: Record<string, any>;
     nextUnprocessedCursor?: number;
     unprocessed?: string[];
   }> {
     const client = await this.queue.client;
     const multi = client.multi();
-    if (!opts.processed && !opts.unprocessed) {
+    if (!opts.processed && !opts.unprocessed && !opts.ignored) {
```
manast, line 1017: I find this condition difficult to reason about. So for example, if opts.ignored is true, then it will skip this code and go through the else case? so basically this if is saying, all options must be false?

roggervalf: so if you don't pass any of these options means to retrieve all types of jobs. If at least one of these option is provided, means that you want to be specific (else case)

manast (approved):
Looks great

src/classes/job.ts
```diff
@@ -1053,37 +1064,80 @@ export class Job<
...
+          }
+          case 'ignored': {
+            ignoredCursor = results[index][1][0];
+
+            const rawIgnored = results[index][1][1];
+            const transformedIgnored: Record<string, any> = {};
+
+            for (let ind = 0; ind < rawIgnored.length; ++ind) {
+              if (ind % 2) {
+                transformedIgnored[rawIgnored[ind - 1]] = rawIgnored[ind];
```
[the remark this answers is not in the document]
roggervalf: Failures is not an stringified string

PR #3245, opened 2025-04-27
fix(remove): pass correct children meta references

roggervalf:
### Why

 1. Why is this change necessary? children meta keys could be different than root parent meta key

### How

_Any extra info here._


src/commands/includes/removeJobWithChildren.lua
```diff
@@ -32,7 +32,7 @@ removeJobChildren = function(prefix, meta, jobKey, options)
             for i = 1, #processed, 2 do
                 local childJobId = getJobIdFromKey(processed[i])
                 local childJobPrefix = getJobKeyPrefix(processed[i], childJobId)
-                removeJobWithChildren(childJobPrefix, meta, childJobId, jobKey, options)
+                removeJobWithChildren(childJobPrefix, childJobPrefix .. "meta", childJobId, jobKey, options)
```
manast, line 35: Interestingly, the argument "meta" is not being used in this function AFAICS

manast:
Instead of passing the argument "meta" everywhere, better to avoid passing it and just compute the metaKey here if needed:
```ts
if removeJobKeys(jobKey) > 0 then
            local metaKey = prefix .. "meta"
            local maxEvents = getOrSetMaxEvents(meta)
```

manast (approved):
LGTM

PR #3246, opened 2025-04-27
ci: use pytest

roggervalf:
### Why

  1. Why is this change necessary? Tests do not break ci step in case of python. No need to call tests from a bash script but instead use purest

### How

_Any extra info here._


manast (approved):
LGTM

Issue #3248, opened 2025-05-05
[Bug]: Passing `undefined` for maxStalledCount causes the stall checking process to fail


matthieu-beteille:
### Version

5.34.9

### Platform

NodeJS

### What happened?

When maxStalledCount is explicitly set to undefined in the worker options, jobs that are stalled (e.g., after worker crashes) never get moved back to failed or waiting.

### Issue
After a runtime crash, many jobs were stuck in active state. In redis I observed them moving back and forth between the active and stalled sets but never moved to failed or waiting, indicating they were detected as stalled but the checker couldn't complete the process.

### Root cause

In my code, I had logic where maxStalledCount was explicitly set to `undefined` for some jobs:

```
const worker = new Worker(queueName, { 
  ...otherOpts, 
  maxStalledCount: undefined (this was an overridable parameter in my code which could be a number or undefined)
})
```

The `moveStalledJobsToWait` script fails when ARGV[1] (`maxStalledCount`) is null, when comparing stall counts:
- https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveStalledJobsToWait-9.lua#L81

### Solution

After setting an explicit default in my code and redeploying, the stalled jobs were properly moved to waiting and reprocessed as expected on restart.

### How to reproduce.

const worker = new Worker(queueName, { 
  ...otherOpts, 
  maxStalledCount: undefined
})

### Relevant log output

```shell
ReplyError: ERR user_script:456: attempt to compare nil with number script: [REDACTED]
on @user_script:456.
t parseError (/[REDACTED]/node_modules/.pnpm/redis-parser@3.0.0/node_modules/redis-parser/lib/parser.js:179:12)\\n    
at parseType (/[REDACTED]/node_modules/.pnpm/redis-parser@3.0.0/node_modules/redis-parser/lib/parser.js:302:14)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3249, opened 2025-05-06
fix(worker): maxStalledCount no less  than 0

roggervalf:
### Why

1. Why is this change necessary? Set default worker option values even when passing null

### How

_Any extra info here._
ref https://github.com/taskforcesh/bullmq/issues/3248


manast (approved):
LGTM. However a stalledInterval of 1 is probably quite bad as well. We use small stalledIntervals in tests to simulate job stalls, but in practice, anything less than maybe 5 or 10 seconds is too small.

PR #3251, opened 2025-05-07
feat(job): add moveToCompleted method [python]

roggervalf:
### Why

  1. Why is this change necessary? In order to provide a way to do manual processing. moveToComplete method is not exposed in job instance

### How

  1. How did you implement this? Add moveToCompleted method in job instance and allow manual processing

_Any extra info here._


manast (approved):
LGTM

PR #3255, opened 2025-05-08
fix(deps): upgrade cron-parser to v5.6.1

### LINKED ISSUE #3252: [Bug]: Problem with cron intervals
### Version

v5.51.0

### Platform

NodeJS

### What happened?

`cron-parser` that is used for cron expressions parsing has a bug parsing some interval expressions (https://github.com/harrisiirak/cron-parser/issues/156), so the `bullmq` also creates incorrect repeatable jobs in those cases.

The mentioned bug is fixed in newer versions (major version upgrade) of `cron-parser`. So the dependency of `bullmq` needs to be updated either. There is a breaking change (renamed function) in newer version of `cron-parser` which is trivial to change in `bullmq` also. Other than that, as far as I see, the new version will have same behavior for the `bullmq` usage.

### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

roggervalf:
### Why

 1. Why is this change necessary? Keep it up to date and fix https://github.com/taskforcesh/bullmq/issues/3252

### How

fixes https://github.com/taskforcesh/bullmq/issues/3252


src/interfaces/repeat-options.ts
```diff
@@ -1,11 +1,20 @@
...
  * Settings for repeatable jobs
  *
  * @see {@link https://docs.bullmq.io/guide/jobs/repeatable}
  */
-export interface RepeatOptions extends Omit<ParserOptions, 'iterator'> {
+export interface RepeatOptions {
+  /**
+   * End date when job should stop
+   */
+  endDate?: Date | string | number;
```
roggervalf, line 12: I'm avoiding to reuse type from cron parser because it does not have descriptions and it also contains an extra custom type in their option attributes that is not compatible with Date class. Justo keeping the primitive types

tests/test_job_scheduler.ts
```diff
@@ -1645,7 +1645,7 @@ describe('Job Scheduler', function () {
       const job = await queue.upsertJobScheduler('repeat', {
         pattern: '0 1 * * *',
         endDate: new Date('2017-05-10 13:13:00'),
-        tz: 'Europe/Athens',
+        tz: 'Etc/UTC',
```
manast, line 1648: wouldn't this be a breaking change?

roggervalf: I thought about it, no sure if someone is using utc option as we don't put it in our documentation. But if you prefer, it can be saved for next breaking change

roggervalf:

roggervalf:
@copilot fix all merge conflicts

PR #3257, opened 2025-05-10
ci: apply timeout when using pytest

roggervalf:
### Why

  1. Why is this change necessary? Instead of hanging if a test case did not resolve well, we can apply a timeout using pytest. Also this pr adds verbose logs for testing progress

### How

  1. How did you implement this? Add pytest-timeout module

_Any extra info here._


manast (approved):
LGTM

Issue #3250, opened 2025-05-06
Actual Debounce (vs Deduplication)


drewB:
**Is your feature request related to a problem? Please describe.**
There are many occasions where you want a debounce rather than a throttle.  For example, if there is going to be a rapid series of events and you want to run the job only after there is a pause in that event stream.

**Describe the solution you'd like**
In the current deduplication there is a simple mode and a throttle mode.  It would be great if there was a debounce mode where you give an id and a delay.  When a new job is queued with the same id as an existing job it is replaced with the new one (or its duration is reset to the original value).

**Describe alternatives you've considered**
I can think of 2 alternatives:

1. Listen for deduplication events and change the delay on the source job.
2. Allow all the jobs to run and at the start of the job check a shared memory key to see if it should still run.

**Additional context**
I know originally there was a debounce option that was depreciated in favor of deduplication.  I didn't see any discussion of why, particularly since deduplication doesn't actually do debouncing.

drewB:
For anyone needing to do this, here is how I implemented it with the current limitations.  In my case, I wanted a job to run after 10 mins without another job creation.

Scheduling job

```
queue.add(debouncedJobName,
            { data },
            {
                delay: 1000 * 60 * 10,
                deduplication: { id: `${debouncedJobName}:${data.uniqueId}` }
            }
        );
```
Note: data.uniqueId can be any piece of data that is unique to that job (mongoDB record id, user id, etc).  This allows multiple job related to different records to work.

Automate Rescheduling
```
queueEvents.on('deduplicated', async ({ jobId, deduplicationId }, id) => {
      const jobName = deduplicationId.split(':')[0];  // get the job name from the deduplicateId

      if (jobName === debouncedJobName) {
        const job = await queue.getJob(jobId);
        const delay = job.opts.delay;
        await job.changeDelay(delay);  // reset delay
      }
    });
```

manast:
@drewB thanks for the workaround. We are working on a native deduplication that provides real debounce capabilities, it will also make use of the delay functionality.

roggervalf:
available since v5.56.0

Issue #3151, opened 2025-03-17
Allow deduplicated jobs to queue when job is already running


nwalters512:
**Is your feature request related to a problem? Please describe.**

I want to run a certain task in response to some user activity. However, it's an expensive task, and users could potentially do the triggering activity many times in rapid succession. I can achieve this pretty well with deduplication (https://docs.bullmq.io/guide/jobs/deduplication).

However, I want to be absolutely sure that I'll always run the task after the latest user action such that I never fail to process the latest state. Based on the docs, this doesn't seem to be the behavior:

> While this job is not moved to completed or failed state, next jobs added with same deduplication id will be ignored and a deduplicated event will be triggered by our QueueEvent class.

I interpret this to mean that while my job is running, if I try to enqueue another job with the same deduplication ID, it will _not_ be queued and will never run after the current one is done executing. This means that sometimes, I'll fail to process the latest data from the user.

**Describe the solution you'd like**

Back when debouncing (which ultimately became deduplication) was first proposed in https://github.com/OptimalBits/bull/issues/1034, what I'm describing was actually spelled out explicitly as a requirement for this new functionality: https://github.com/OptimalBits/bull/issues/1034#issuecomment-734181580. However, as far as I can tell, it was just never implemented (or if it was, it's not documented). For context, the functionality was implemented here: https://github.com/taskforcesh/bullmq/pull/2796

**Describe alternatives you've considered**

I'm not sure if there is one. Maybe it's already possible with the right set of flags and options? Or maybe things already work how they were originally specced to and the docs just don't reflect that? Any guidance would be appreciated.

mattmcmanus:
Amazing timing. We ran into the need for this today and were surprised there was no support.

billouboq:
We do too, that would be such a nice improvement

matthieu-beteille:
big +1

I would also like to suggest two related/follow-up features for deduplication that would be extremely helpful:
- The ability to postpone the initial job execution for X ms when we expect a burst of similar events (already filed here https://github.com/taskforcesh/bullmq/issues/2767)
- The ability to combine/merge deduplicated job data through a custom "reducing' function. In the case described above, you might only care about the latest job data, but in more advanced scenarios, you might need finer control over what gets executed based on all the deduplicated job data (for example, merging fields from multiple events or applying custom prioritization logic)

Happy to create separate issues for these requests if that's better

nwalters512:
[...]

Issue #2767, opened 2024-09-11
Debounce: sliding mode


freiondrej-lokalise:
First of all, thank you for implementing debouncing! Much appreciated. 🚀 

We have a usecase which is not supported out of the box if I'm not mistaken - we have a refresh job that we want to have debounced, but differently from the current Fixed mode, it would be pushed N ms into the future every time the job tries to be scheduled. Let me illustrate with an example:

In the current Fixed mode, let's say we schedule a job A with 5s TTL. Then, after 3 seconds, we try to schedule a new job B with the same debounceKey. It is ignored, so in 5s from the start, job A runs.

In the suggested Sliding mode, we schedule job A with 5s TTL. Then, after 3 seconds, we schedule a new job B with the same debounce key and the same 5s TTL. Job A is discarded before it ever runs, and job B runs 8s from the start (3s from the beginning + its 5s TTL).

(We could also think of this mode as "overriding" mode, as the last job is the one that ends up being executed.)

---

There's also another question, how to behave when the latest job is still running and a new one is scheduled with the same debounceKey.

Example to illustrate what I just mentioned (building up on the example from the previous section):

1) we schedule job A with 5s TTL. 3 seconds in, we schedule job B with the same debounceKey and the same 5s TTL. On the 8th
second of the timeline, job B starts and it takes 10 seconds to process (so it ends on the 18th second in total). On the 12th second in total, job C is scheduled with the same debounce key and the same 5s TTL. Job C starts running on the 17th second in total, while job B is still active.

2) we schedule job A with 5s TTL. 3 seconds in, we schedule job B with the same debounceKey and the same 5s TTL. On the 8th
second of the timeline, job B starts and it takes 10 seconds to process (so it ends on the 18th second in total). On the 12th second in total, job C is scheduled with the same debounce key and the same 5s TTL. We ignore job C, as job B is still running (pretty much like in Extended mode).

3) we schedule job A with 5s TTL. 3 seconds in, we schedule job B with the same debounceKey and the same 5s TTL. On the 8th
second of the timeline, job B starts and it takes 10 seconds to process (so it ends on the 18th second in total). On the 12th second in total, job C is scheduled with the same debounce key and the same 5s TTL. We wait until job B finishes, then start the countdown, so job C starts running on 23rd second in total.

4) we schedule job A with 5s TTL. 3 seconds in, we schedule job B with the same debounceKey and the same 5s TTL. On the 8th
[...]

PR #3260, opened 2025-05-13
feat(deduplication): add support for replace and extend options

roggervalf:
### Why

  1. Why is this change necessary? Deduplication option is only allowing to use 2 different modes, simple and throtle https://docs.bullmq.io/guide/jobs/deduplication. While these modes are useful for those specific cases, there is also the case of debouncing, where we can replace a job while applying a ttl. This pr is creating that extension and adding a new mode in our documentation

### How

  1. How did you implement this? Add 2 options into deduplication option:
  * extend: in case users want to extend an specific ttl
  * replace: job saved in a deduplicated key will be replaced (deleted) by a new job. Take in count that this functionality will only take effect if this previous job is in delayed state

ref https://github.com/taskforcesh/bullmq/issues/3250 https://github.com/taskforcesh/bullmq/issues/3151 https://github.com/taskforcesh/bullmq/issues/2767


billouboq:
Just to be sure, does it work for this use case :

Let's say you have a "Compute" task, I want that task to be done once every 5 seconds, BUT if that task is active and a new one with same id is created, it will wait the 5 seconds and do it again. (throttle with leading ?)

Right now I fear that if I have an active job, if a new one with same id is created it's skipped because of the active one which can lead to some window where you can loose some computation if you task take time.

Like if I want to compute based on document update, if my computation take 10 seconds and that for those 10 seconds the new job are being skipped then we won't have the latests computation based on the 10 seconds window

Hope it's clear

roggervalf:
> Let's say you have a "Compute" task, I want that task to be done once every 5 seconds, BUT if that task is active and a new one with same id is created, it will wait the 5 seconds and do it again. (throttle with leading ?)

hi @billouboq, that would be the same as:
```js
await myQueue.add(
   'house',
   { color: 'white' },
   { deduplication: { id: 'customValue', ttl: 5000}, delay: 5000 }
 );
// after 5 seconds til will be finished, so you can add a new job with same dedup id
 await myQueue.add(
   'house',
   { color: 'red' },
   { deduplication: { id: 'customValue', ttl: 5000 }, delay: 5000 },
 );
```

src/types/job-options.ts
```diff
@@ -9,12 +10,12 @@ export type CompressableJobOptions = {
    * Debounce options.
    * @deprecated use deduplication option
    */
-  debounce?: DebounceOptions;
+  debounce?: DeduplicationOptions;
```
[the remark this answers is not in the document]
roggervalf: will be removed in next breaking change, we cannot do it right now as some people might be using it

docs/gitbook/guide/jobs/deduplication.md
```diff
@@ -38,6 +38,38 @@ await myQueue.add(
 
 In this example, after adding the house painting job with the deduplicated parameters (id and ttl), any subsequent job with the same deduplication ID customValue added within 5 seconds will be ignored. This is useful for scenarios where rapid, repetitive requests are made, such as multiple users or processes attempting to trigger the same job.
 
+## Debounce Mode
+
+In the Debounce Mode, deduplication works by assigning a delay to a job upon its creation. If a similar job (identified by a unique deduplication ID) is added during this delay period, it'll replace the previous job while this one is still in delayed state and TTL value from deduplication key will be updated as well. This prevents the queue from being overwhelmed with multiple instances of the same task, keeping job information up to date.
```
manast: This text needs to be proof read, as the  part about TTL does not seem to be coherent with the rest.

src/commands/includes/deduplicateJob.lua
```diff
@@ -1,23 +1,68 @@
...
-      end
-  end
+]]
+-- Includes
+--- @include "removeJobKeys"
+local function deduplicateJob(deduplicationOpts, jobId, delayedKey, deduplicationKey, eventsKey, maxEvents,
+    prefix, delay)
+    local deduplicationId = deduplicationOpts and deduplicationOpts['id']
+    if deduplicationId then
+        if deduplicationOpts['replace'] and delay and delay > 0 then
```
manast: maybe it is enough with ```if deduplicationOpts['replace'] and delay > 0 then```

src/commands/includes/deduplicateJob.lua
```diff
@@ -1,23 +1,68 @@
...
+            if currentDebounceJobId then
+                if rcall("ZREM", delayedKey, currentDebounceJobId) > 0 then
+                    removeJobKeys(prefix .. currentDebounceJobId)
+                    rcall("XADD", eventsKey, "*", "event", "removed", "jobId", currentDebounceJobId,
+                        "prev", "delayed")
+
+                    if deduplicationOpts['extend'] then
+                        rcall('SET', deduplicationKey, jobId, 'PX', delay)
+                    else
+                        rcall('SET', deduplicationKey, jobId, 'KEEPTTL')
```
manast, line 22: wouldn't this code be the same as not doing anything at all?

roggervalf: not really, because job will be replaced, so jobId needs to be updated. Just extension is not done

src/commands/includes/deduplicateJob.lua
roggervalf: this is an old logic, delay is not passed anymore here

docs/gitbook/guide/jobs/deduplication.md
roggervalf: I rephrased this section

docs/gitbook/guide/jobs/deduplication.md
```diff
@@ -38,6 +38,48 @@ await myQueue.add(
 
 In this example, after adding the house painting job with the deduplicated parameters (id and ttl), any subsequent job with the same deduplication ID customValue added within 5 seconds will be ignored. This is useful for scenarios where rapid, repetitive requests are made, such as multiple users or processes attempting to trigger the same job.
 
+## Debounce Mode
+
+Debounce Mode works by delaying a job upon creation and providing a TTL. If another job with the same deduplication ID is added during this delay period, it replaces the previous one. The deduplication key’s TTL is refreshed, ensuring only the most recent job is kept. This mechanism avoids flooding the queue with duplicates while maintaining the latest job data.
```
manast: "Debounce Mode works by delaying a job upon creation and providing a TTL" -> "Debounce Mode can be achieved by delaying a job upon creation while providing a matching TTL as well as having extend and replace options set as true."

manast: "If another job with the same deduplication ID is added during this delay period, it replaces the previous one." -> "Debounce is achieved because if another job with the same deduplication ID is added during this delay (and TTL period) it will replace the previous job with the new one, as well as reseting the TTL, thus ensuring that only the most recent job is kept".

docs/gitbook/guide/jobs/deduplication.md
```diff
@@ -38,6 +38,48 @@ await myQueue.add(
...
+
+```typescript
+import { Queue } from 'bullmq';
+
+const myQueue = new Queue('Paint');
+
+// Add a job that will be deduplicated for 5 seconds.
```
manast: For the example I think it would be more illustrative to have a for-loop, so instead of just adding one job directly after the other, add like 10 jobs with a delay of 500ms per job for example, and then show that only the latest job is processed at the end while the 9 first were ignored.

manast (commented):
I just fixed a couple of docs, also I think we should have a test simular to the doc improvement I mentioned, like having a loop where we add a bunch of jobs and only the last one is kept. Finally, not sure why the tests are placed in the test_events file, as this deduplication functionality should have its own dedicated test file.

docs/gitbook/guide/jobs/deduplication.md
```diff
@@ -38,6 +38,43 @@ await myQueue.add(
 
 In this example, after adding the house painting job with the deduplicated parameters (id and ttl), any subsequent job with the same deduplication ID customValue added within 5 seconds will be ignored. This is useful for scenarios where rapid, repetitive requests are made, such as multiple users or processes attempting to trigger the same job.
 
+## Debounce Mode
+
+Debounce Mode can be achieved by delaying a job upon creation while providing a matching TTL as well as having extend and replace options set as true. Debounce is achieved because if another job with the same deduplication ID is added during this delay (and TTL period) it will replace the previous job with the new one, as well as reseting the TTL, thus ensuring that only the most recent job is kept. This mechanism avoids flooding the queue with duplicates while maintaining the latest job data.
+
```
manast, line 44: "the latest job data" -> "the latest job's data"

manast (approved):
LGTM, but there is a small grammar error

manast (approved):

PR #3266, opened 2025-05-15
perf(stalled): fail stalled jobs in a lazy way

roggervalf:
### Why

1. Why is this change necessary? Now that we have a deferredFailure attribute, we can fail our stalled jobs in a lazy way. So failed event in workers will contain a job instance and not be undefined

### How

_Any extra info here._


manast (approved):
LGTM

lesderid:
Does this mean the Worker API could now emit a `stalled` [event](https://github.com/taskforcesh/bullmq/blob/5270deddda6bcc57b333c81d84abea0ccdd66f6e/src/classes/worker.ts#L1398) that includes the `job`, like [other events](https://api.docs.bullmq.io/interfaces/v5.WorkerListener.html), instead of just the `jobId`?

PR #3268, opened 2025-05-17
fix(flow): add new error code when parent has failed children

roggervalf:
### Why

  1. Why is this change necessary? When a parent is moved to completed, failed children are ignored. That cause that a parent could be completed when a hard dependency failed. This is wrong as a parent should moved to failed in this situation

### How

  1. How did you implement this? Validate that there are no unsuccessful children before trying to move it to completed

_Any extra info here._

### DIFF
--- a/src/commands/moveToFinished-14.lua
+++ b/src/commands/moveToFinished-14.lua
@@ -50,8 +50,9 @@
       -1 Missing key.
       -2 Missing lock.
       -3 Job not in active set
-      -4 Job has pending dependencies
+      -4 Job has pending children
       -6 Lock is not owned by this client
+      -9 Job has failed children
 
     Events:
       'completed/failed'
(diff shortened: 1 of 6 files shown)


src/commands/moveToFinished-14.lua
```diff
@@ -81,10 +82,14 @@ local jobIdKey = KEYS[12]
 if rcall("EXISTS", jobIdKey) == 1 then -- Make sure job exists
     -- Make sure it does not have pending dependencies
     -- It must happen before removing lock
-    if ARGV[5] == "completed" and
-        not rcall("HGET", jobIdKey, "igdp") and -- check if we should ignore this check
-        rcall("SCARD", jobIdKey .. ":dependencies") ~= 0 then
```
roggervalf, line 86: no need for this check as that breaks the hard dependency relationship with parent and children failures

manast (approved):
LGTM

Issue #3262, opened 2025-05-13
[Bug]: Repeating job stops working when using upsertJobScheduler and every


sailflip:
### Version

v5.52.2

### Platform

NodeJS

### What happened?

I have a node.js worker instance that creates a repeating job using upsertJobScheduler() and every: 1_000.

When I have 2 instance of my worker, running the 2nd instance can cause the job to stop being scheduled.

If I switch to using pattern instead it works.

### How to reproduce.

Here's a simple node.js app that I run against my local redis instance.
```
import { Queue, Worker } from 'bullmq'

const worker = new Worker<unknown>(
  'my_queue',
  async (job) => {
    console.log(`Running job ${job.name}`)
  },
  {
    connection: { url: 'redis://localhost:6379' }
  }
)

const queue = new Queue('my_queue', {
  connection: { url: 'redis://localhost:6379' }
})


async function main() {
  console.log('Adding job')

  await queue.upsertJobScheduler(
    'my_job_scheduler',
    {
      // pattern: '*/1 * * * * *',
      every: 1_000,
    },
    {
      data: {},
      name: 'my_job',
      opts: {
        removeOnComplete: true
      }
    }
  )
}

async function gracefulShutdown(signal: string) {
  console.log(`Received ${signal}, shutting down gracefully...`)
  await worker.close()
  process.exit(0)
}

process.on('SIGINT', () => gracefulShutdown('SIGINT'))
process.on('SIGTERM', () => gracefulShutdown('SIGTERM'))

main()
```

To reproduce I do this:
```
npm run dev
```
watch the jobs being run, in a new terminal window run:
```
npm run dev
```

About 1 in 2 times the jobs stop being scheduled.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

roggervalf:
thanks @sailflip I could reproduce it with your example. I opened a pr for it

PR #3269, opened 2025-05-18
fix(scheduler): remove current job when it is in delayed state

roggervalf:
### Why

  1. Why is this change necessary? Job iteration is broken as current job is not gracefully removed

### How

fixes https://github.com/taskforcesh/bullmq/issues/3262

### DIFF
--- a/src/commands/addJobScheduler-11.lua
+++ b/src/commands/addJobScheduler-11.lua
@@ -64,7 +64,7 @@ local maxEvents = getOrSetMaxEvents(metaKey)
 local function removeJobFromScheduler(prefixKey, delayedKey, prioritizedKey, waitKey, pausedKey, jobId,
     metaKey, eventsKey)
     if rcall("ZSCORE", delayedKey, jobId) then
-        removeJob(nextDelayedJobId, true, prefixKey, true --[[remove debounce key]] )
+        removeJob(jobId, true, prefixKey, true --[[remove debounce key]] )
         rcall("ZREM", delayedKey, jobId)
         return true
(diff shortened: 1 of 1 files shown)


tests/test_job_scheduler_stress.ts
```diff
@@ -53,34 +54,47 @@ describe('Job Scheduler Stress', function () {
...
+        );
+        expect(count).to.be.gte(1);
+        // previous job can be active while a delayed or waiting job is added
+        expect(count).to.be.lte(2);
+      }
+      previousJob = await queue.upsertJobScheduler(
         jobSchedulerId,
         {
-          every: ONE_HOUR,
+          every: 200,
```
manast, line 85: magic number?

manast (commented):
Can you please provide more background to what chase this is fix resolving?

tests/test_job_scheduler_stress.ts
roggervalf: no magic, just using a lower value so I don't have to wait 1 hour for a completion in this test case

roggervalf:
> Can you please provide more background to what chase this is fix resolving?

sure, we are not removing the hash of the current delayed job because the reference that is passed is wrong. That means that if a scheduler is updated, there is a possibility that a job being generated has the same id as the hash that was not deleted. When that happens, iteration is stopped

bsitruk-ps:
Any update ?
I think I'm experiencing a similar issue.
Basically, calling `upsertJobScheduler` several times will create zombie jobs in Redis, if that makes any sense.

@roggervalf Do you think your PR is related ?

manast (approved):
LGTM

PR #3278, opened 2025-05-24
refactor(worker): convert handler callbacks to worker methods

roggervalf:
### Why

1. Why is this change necessary? No need to create new callbacks in every job processing
### How

_Any extra info here._


manast (approved):
LGTM

Issue #3283, opened 2025-05-28
[Bug]: Updated from 5.26 to 5.53 and now getDependencies errors with a JSON issue


vidjuheffex:
### Version

v5.53

### Platform

NodeJS

### What happened?

<img width="965" alt="Image" src="https://github.com/user-attachments/assets/f2026a4b-5b00-4e79-ae2a-b571b104e53a" />

I get this now that I've updated from 5.26 to 5.53 (sporadically)

### How to reproduce.

```
  const { processed: processedCount, unprocessed: unprocessedCount } =
    await job.getDependenciesCount();
```

this is all it takes.

### Relevant log output

```shell
⨯ SyntaxError: Unexpected token 'e', "fetch failed" is not valid JSON
2025-05-28 14:51:58     at JSON.parse (<anonymous>)
2025-05-28 14:51:58     at async buildJobChildren (./src/app/ami/utilities.ts:49:44)
2025-05-28 14:51:58     at async createNewSeedJob (./src/app/job/[id]/createSeedJob.ts:23:23)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3284, opened 2025-05-29
fix(job): do not parse ignored failures in getDependencies

roggervalf:
### Why

 1. Why is this change necessary? Ignored failures are not stringified, they are saved as plain strings

### How

  1. How did you implement this? By not trying to parse those values

fixes https://github.com/taskforcesh/bullmq/issues/3283


manast (approved):
LGTM

PR #3291, opened 2025-06-06
feat(backoff): add jitter option

roggervalf:
### Why

1. Why is this change necessary? In some scenarios we should be able to apply jitter strategy when applying a backoff

### How

_Any extra info here._

### DIFF
--- a/docs/gitbook/guide/retrying-failing-jobs.md
+++ b/docs/gitbook/guide/retrying-failing-jobs.md
@@ -25,7 +25,7 @@ Retried jobs will respect their priority when they are moved back to waiting sta
 
 ### Built-in backoff strategies
 
-The current built-in backoff functions are **fixed**, **exponential** and **jitter**.
+The current built-in backoff functions are **fixed** and **exponential**.
 
 #### Fixed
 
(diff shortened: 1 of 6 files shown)


docs/gitbook/guide/retrying-failing-jobs.md
```diff
@@ -23,14 +23,57 @@ BullMQ supports retries of failed jobs using back-off functions. It is possible
...
+    attempts: 3,
+    backoff: {
+      type: 'fixed',
+      delay: 1000,
+    },
+  },
+);
+```
+
+You can also provide a (jitter)[https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/] option, it will generate delays between `delay` and 0 milliseconds depending on the percentage of jitter usage. For example, you can provide a jitter value of 0.5 value and a delay of 1000 milliseconds, it will generate a delay between `1000` milliseconds = 1 second and `1000 * 0.5` milliseconds = 0.5 seconds.
```
manast: A couple of improvements:
- "it will generate delays between `delay` and 0 milliseconds" -> "it will generate *random* delays between `delay` and 0 milliseconds"
- explain that jitter is a value between 0 and 1
- in the example do not mix units, if delay is milliseconds, then do not write 0.5 seconds but 500ms for clarity.
- add more examples like what happens with jitter 0, and what happens with jitter 1

docs/gitbook/guide/retrying-failing-jobs.md
```diff
@@ -51,6 +94,27 @@ await queue.add(
 );
 ```
 
+You can also provide a (jitter)[https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/] option, it will generate delays between `2 ^ (attempts - 1) * delay` and 0 milliseconds depending on the percentage of jitter usage. For example, you can provide a jitter value of 0.5 value and a delay of 3000 milliseconds, for the 7th attempt, it will generate a delay between `2^6 * 3000` milliseconds = 3.2 minutes and `2^6 * 3000 * 0.5` milliseconds = 1.6 minutes after the previous attempt.
```
manast: please see my comment above.

src/classes/backoffs.ts
```diff
@@ -2,20 +2,33 @@ import { BackoffOptions, MinimalJob } from '../interfaces';
...
 export class Backoffs {
   static builtinStrategies: BuiltInStrategies = {
-    fixed: function (delay: number) {
+    fixed: function (delay: number, jitter = 0) {
       return function (): number {
-        return delay;
+        if (jitter > 0) {
+          const minDelay = delay * (1 - jitter);
+
+          return Math.floor(Math.random() * delay * jitter + minDelay);
```
manast, line 15: do we support jitter > 1 ? I think that would be just confusing

roggervalf: I added a validation for this value on job instantiation, if jitter is greater than 1, an error will be thrown

docs/gitbook/guide/retrying-failing-jobs.md
roggervalf: Added an info section for jitter values

manast (approved):
LGTM

PR #3292, opened 2025-06-08
Ci/separate changelog update pr

manast:
### Why

Since we have protection rules for the default branch we cannot longer commit from within the release GitHub action, so instead we need to create a separate PR with the updates in changelog and other files.

### How

updated the release action to create a PR instead of committing to the default branch.

_Any extra info here._


python/bullmq/flow_producer.py
```diff
@@ -135,8 +135,8 @@ async def addBulk(self, flows: list[dict]):
 
         return result
 
-    def close(self):
+    async def close(self):
```
roggervalf, line 138: may not be needed if we only return an async method like close

python/bullmq/queue.py
```diff
@@ -329,11 +329,11 @@ def sanitizeJobTypes(self, types):
             'waiting-children'
         ]
 
-    def close(self):
+    async def close(self):
         """
         Close the queue instance.
         """
-        return self.redisConnection.close()
+        return await self.redisConnection.close()
```
roggervalf, line 336: same as above

python/bullmq/redis_connection.py
```diff
@@ -50,11 +50,11 @@ def disconnect(self):
         """
         return self.conn.disconnect()
 
-    def close(self):
+    async def close(self):
```
roggervalf, line 53: same as above

python/bullmq/flow_producer.py
manast: this was recommended by copilot, so I am assuming is a best practice even if it is not necessary.

python/bullmq/queue.py
manast: also recommended by copilot

roggervalf (approved):
Lgtm

PR #3301, opened 2025-06-12
fix: avoid circular reference between scripts and queue

roggervalf:
### Why

  1. Why is this change necessary? Queue instance incluyes scripts instance, but it's also passed as an instance of Scripts

### How

_Any extra info here._

### DIFF
--- a/src/types/minimal-queue.ts
+++ b/src/types/minimal-queue.ts
@@ -16,8 +16,3 @@ export type MinimalQueue = Pick<
   | 'redisVersion'
   | 'trace'
 >;
-
-export type MinimalScriptQueue = Pick<
-  QueueBase,
-  'client' | 'toKey' | 'keys' | 'opts' | 'closing' | 'redisVersion'
->;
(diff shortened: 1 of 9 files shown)


tests/test_events.ts
```diff
@@ -573,6 +573,7 @@ describe('events', function () {
             await deduplication;
 
             expect(deduplicatedCounter).to.be.equal(1);
+            await worker.close();
```
roggervalf, line 576: missing to close worker instances in these tests

src/types/minimal-queue.ts
```diff
@@ -16,3 +16,8 @@ export type MinimalQueue = Pick<
   | 'redisVersion'
   | 'trace'
 >;
+
+export type MinimalScriptQueue = Pick<
```
manast: I think we need a better name for this type, as it is difficult to understand its intend. Also it should be moved to a separate file.

roggervalf: after some round trips with chatgpt, ScriptQueueContext was the last choice

Issue #3296, opened 2025-06-11
[Feature Request] Add moveToWait Method to SandboxedJob Interface for BullMQ Pro


miridih-jujang:
# [Feature Request] Add moveToWait Method to SandboxedJob Interface for BullMQ Pro

## Problem Description

When using BullMQ Pro with sandboxed processors (by providing a `processorPath` to the worker), the job is passed as a `SandboxedJobPro` instance to the isolated process. However, the `SandboxedJobPro` interface lacks the `moveToWait` method that is available in the regular `JobPro` class.

## Current Behavior

```javascript
// When using processorPath (sandboxed processor)
const processorPath = path.join(__dirname, 'sandbox-processor.js');
const workerPro = new WorkerPro(queueName, processorPath, { connection });
```

In the sandboxed processor file:

```javascript
import { SandboxedJobPro } from '@taskforcesh/bullmq-pro';

module.exports = async (job: SandboxedJobPro) => {
  // This throws TypeError: Cannot read properties of undefined (reading 'moveToWait')
  await job.moveToWait(token); // ❌ Error: moveToWait is not available
};
```

Error output:

```
TypeError: Cannot read properties of undefined (reading 'moveToWait')
    at /path/to/processor.ts:37:17
```

## Current Workaround

The only workaround is to disable sandboxed processing by passing `null` as the processor:

```javascript
// Without processorPath (non-sandboxed)
const workerPro = new WorkerPro(queueName, null, { connection });
```

This allows access to the full `JobPro` interface including `moveToWait`, but sacrifices the benefits of process isolation.

## Expected Behavior

The `SandboxedJobPro` interface should include the `moveToWait` method, similar to how `SandboxedJob` already includes `moveToDelayed`:

Current SandboxedJob interface (from BullMQ):

```typescript
export interface SandboxedJob<T = any, R = any> {
  moveToDelayed: (timestamp: number, token?: string) => Promise<void>; // ✅ Available
  moveToWait: (token: string) => Promise<number>; // ❌ Missing
  // ... other methods
}
```

Current SandboxedJobPro interface:

```typescript
export interface SandboxedJobPro<T = any, R = any> extends Omit<SandboxedJob, 'data' | 'opts' | 'returnValue'> {
  data: T;
  opts: JobsProOptions;
  returnValue: R;
  gid?: string;
  // moveToWait is missing here as well
}
```

## Proposed Solution

[...]

PR #3302, opened 2025-06-13
feat(worker): allow calling moveToWait when job is processing

roggervalf:
### Why

  1. Why is this change necessary? To be able to move a job to wait or prioritized while is being processed

### How

ref https://github.com/taskforcesh/bullmq/issues/3296


tests/test_sandboxed_process.ts
```diff
@@ -992,6 +997,51 @@ function sandboxProcessTests(
...
+      const worker = new Worker(queueName, processFile, {
+        connection,
+        prefix,
+        drainDelay: 1,
+        useWorkerThreads,
+      });
+
+      let count = 0;
+      const waiting = new Promise<void>((resolve, reject) => {
+        queueEvents.on('waiting', async ({ prev }) => {
```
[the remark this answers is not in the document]
roggervalf: no plan to expose that method

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -126,6 +126,52 @@ const worker = new Worker(
...
+```typescript
+import { WaitingError, Worker } from 'bullmq';
+
+enum Step {
+  Initial,
+  Second,
+  Finish,
+}
+
+const worker = new Worker(
```
manast: I think the text explaining the feature is spot on. However the example code is too complicated as it includes a step pattern which may make it more difficult for the reader to understand the actual functionality. I would just keep the bare minimum, maybe something like trying to perform an operation and if that operation failed then do a moveToWait.

manast (commented):
LGTM, just a comment regarding the example code.

docs/gitbook/patterns/process-step-jobs.md
roggervalf: changed as suggested. As we are not using step pattern, had to move to another file pattern

manast (approved):
LGTM

PR #3303, opened 2025-06-13
ci: use GH_TOKEN for GitHub CLI in release workflow

manast:
### Why

The GitHub token was not properly passed to the GH cli and therefore the creation of the PR with changelog, etc was failing after a release.

### How

Just pass the GitHub token into the GH_TOKEN variable

_Any extra info here._

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,7 @@ jobs:
     permissions:
       contents: write # for release publishing
       pull-requests: write # for creating PRs
+      issues: write # additional permission that might be needed
 
     name: Release
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -40,6 +41,7 @@ jobs:
         if: ${{ success() && steps.semantic_release.outcome == 'success' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
roggervalf, line 43: Is it still needed?

manast: yes because it is needed when we access the repo using git commands, such as when updating the changelog, etc. The GH_TOKEN is only needed for creating the PR with the gh command line tool.

PR #3306, opened 2025-06-17
docs(changelog): update to v5.54.1

roggervalf:
### Why

  1. Why is this change necessary? Last attempt to create a pr for updating changelog and version failed, but new version was published. Also there are few flaky test being fixed

### How

_Any extra info here._

### DIFF
--- a/tests/test_rate_limiter.ts
+++ b/tests/test_rate_limiter.ts
@@ -160,6 +160,7 @@ describe('Rate Limiter', function () {
           }
         },
         {
+          autorun: false,
           connection,
           prefix,
           concurrency: 2,
@@ -188,7 +189,11 @@ describe('Rate Limiter', function () {
         await queue.add('test-job', 'test');
       }
 
+      worker.run();
+
       await completing;
+
+      await worker.close(true);
     });
   });
 


tests/test_job_scheduler.ts
```diff
@@ -61,56 +61,6 @@ describe('Job Scheduler', function () {
     await connection.quit();
   });
 
-  // NOTE: This test seems to be misplaced, it is not related to the repeatable jobs
-  describe('when exponential backoff is applied', () => {
```
roggervalf, line 65: this test case is already covered here https://github.com/taskforcesh/bullmq/blob/master/tests/test_worker.ts#L2623-L2624

tests/test_flow.ts
```diff
@@ -4618,12 +4618,17 @@ describe('flows', () => {
...
-          expect(milliseconds).to.be.lessThanOrEqual(3000);
-          expect(milliseconds).to.be.greaterThan(2000);
-          resolve();
+          try {
+            const milliseconds = delay - Date.now();
+            expect(milliseconds).to.be.lessThanOrEqual(3000);
+            expect(milliseconds).to.be.greaterThan(2000);
+            resolve();
+          } catch (error) {
+            console.error(error);
```
roggervalf, line 4629: if an error happens, we should see an error being printed at least

tests/test_flow.ts
```diff
@@ -4704,7 +4709,7 @@ describe('flows', () => {
       await flow.close();
 
       await removeAllQueueData(new IORedis(redisHost), topQueueName);
-    });
+    }).timeout(4500);
```
roggervalf, line 4712: increased for dragonfly

tests/test_rate_limiter.ts
```diff
@@ -188,7 +189,11 @@ describe('Rate Limiter', function () {
         await queue.add('test-job', 'test');
       }
 
+      worker.run();
+
       await completing;
+
+      await worker.close(true);
```
roggervalf, line 196: forge closing this instance as job that won't resolve will get stalled

PR #3307, opened 2025-06-17
fix(scheduler): fix slot calculation when using every

roggervalf:
### Why

  1. Why is this change necessary? When calculating delay, there is an opportunity where startMillis + offset is slightly higher than current next slot. As startMillis for subsequent delayed jobs should already contain offset value, we can subtract this value to get a more accurate delay and not skip one iteration

### How

_Any extra info here._

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -95,15 +95,24 @@ export class JobScheduler extends QueueBase {
     let nextMillis: number;
     let newOffset: number | null = null;
     if (every) {
-      const prevSlot = Math.floor(startMillis / every) * every;
+      if (typeof offset === 'number') {
+        const prevSlot = Math.floor((startMillis - offset) / every) * every;
 
-      newOffset = typeof offset === 'number' ? offset : startMillis - prevSlot;
+        newOffset = offset;
 
-      const nextSlot = prevSlot + every;
-      if (prevMillis || offset) {
+        const nextSlot = prevSlot + every;
         nextMillis = nextSlot;
       } else {
-        nextMillis = prevSlot;
+        const prevSlot = Math.floor(startMillis / every) * every;
+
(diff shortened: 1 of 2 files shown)


tests/test_job_scheduler_stress.ts
```diff
@@ -222,10 +222,15 @@ describe('Job Scheduler Stress', function () {
       const worker = new Worker(
         queueName,
         async job => {
-          if (iterationCount === 0) {
-            expect(job.opts.delay).to.be.eq(0);
-          } else {
-            expect(job.opts.delay).to.be.gte(1850);
+          try {
+            if (iterationCount++ === 0) {
```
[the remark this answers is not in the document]
roggervalf: addressed  but when removing second reference

src/classes/job-scheduler.ts
```diff
@@ -95,7 +95,8 @@ export class JobScheduler extends QueueBase {
     let nextMillis: number;
     let newOffset: number | null = null;
     if (every) {
-      const prevSlot = Math.floor(startMillis / every) * every;
+      const prevSlot =
+        Math.floor((startMillis - (offset || 0)) / every) * every;
```
manast, line 99: This seems more correct than what we had before. I tried to think if there could be a case where this change could lead to a nextSlot that matches the prevSlot and thus breaking the iterations. For new repeatable jobs it should not be able to happen, but what about existing jobs that did not use offset before?

roggervalf: This change will affect jobs using every. For existing jobs there would be the following scenarios:
* job is in delayed state, worker need to wait to promote it. Then it will use new logic. startMillis will be now timestamp and job won't be lost as nextMillis will be scheduled in the future.
* upsertion, job being in delayed will be removed. This is handled already by default. So new logic will work also after this operation.

PR #3311, opened 2025-06-18
ci: remove [skip ci] from release PR so that checks are run

manast:
### Why

Remove [skip ci] from the commit created for the release PR so that checks are run and therefore able to be merged into master.

_Any extra info here._


roggervalf (approved):
Lgtm

PR #3312, opened 2025-06-19
perf(delayed): add marker once when promoting delayed jobs

roggervalf:
### Why

  1. Why is this change necessary? Base on https://github.com/taskforcesh/bullmq/pull/3096

### How

_Any extra info here._

### DIFF
--- a/src/commands/includes/addJobWithPriority.lua
+++ b/src/commands/includes/addJobWithPriority.lua
@@ -4,10 +4,10 @@
 
 -- Includes
 --- @include "addPriorityMarkerIfNeeded"
+--- @include "getPriorityScore"
 
 local function addJobWithPriority(waitKey, prioritizedKey, priority, paused, jobId, priorityCounterKey)
-  local prioCounter = rcall("INCR", priorityCounterKey)
-  local score = priority * 0x100000000 + bit.band(prioCounter, 0xffffffffffff)
+  local score = getPriorityScore(priority, priorityCounterKey)
   rcall("ZADD", prioritizedKey, score, jobId)
   if not paused then
     addPriorityMarkerIfNeeded(waitKey)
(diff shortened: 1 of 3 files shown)


manast:
As we are using node@lts/-1,-2 etc, we are picking relative node js versions, so as nodejs makes new releases the tests are also running on newer versions of node, which may not be fully compatible with v4 of BullMQ, so we probably need to set the node js versions to fixed numbers.

.github/workflows/api.yml
```diff
@@ -36,9 +36,9 @@ jobs:
           yarn docs:merge
         env:
           CI: true
-      - name: deploy
-        uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 # v3
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./docs
-          cname: 'api.docs.bullmq.io'
+      #- name: deploy
```
roggervalf, line 39: disabling as we don't want to override these api docs with old version

.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
     name: coverage for node@lts, redis@7
     env:
-      node-version: lts/*
+      node-version: 18.x
```
manast: I think we probably should keep LTS here

.github/workflows/api.yml
manast: can we produce docs for older versions?

manast (approved):

roggervalf:
Docs for older versión may need to be copies in Main api branch

.github/workflows/api.yml
roggervalf: need to be copied in main api branch

PR #3322, opened 2025-06-28
perf(worker): do not wait rate limit when fetching jobs

roggervalf:
### Why

1. Why is this change necessary? We should only wait for the rate limit when there are no more task to process
### How

  1. How did you implement this? Only wait for rate limit when needed.

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -465,7 +465,6 @@ export class Worker<
 
   private async waitForRateLimit(): Promise<void> {
     const limitUntil = this.limitUntil;
-
     if (limitUntil > Date.now()) {
       this.abortDelayController?.abort();
       this.abortDelayController = new AbortController();
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -465,6 +465,7 @@ export class Worker<
 
   private async waitForRateLimit(): Promise<void> {
     const limitUntil = this.limitUntil;
+
     if (limitUntil > Date.now()) {
       this.abortDelayController?.abort();
       this.abortDelayController = new AbortController();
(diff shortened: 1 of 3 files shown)


src/classes/worker.ts
```diff
@@ -561,7 +561,7 @@ export class Worker<
             this.opts.runRetryDelay,
           ),
         );
-      } else if (this.isRateLimited()) {
```
roggervalf, line 564: not needed this if statement as waitForRateLimit checks the same thing

tests/test_rate_limiter.ts
```diff
@@ -124,6 +124,57 @@ describe('Rate Limiter', function () {
...
+        },
+      },
+    );
+
+    let completedCount = 0;
+    const result = new Promise<void>((resolve, reject) => {
+      worker.on('completed', async job => {
+        try {
+          completedCount++;
+          expect(job.finishedOn! - job.processedOn!).to.be.lte(1000);
```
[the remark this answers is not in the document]
roggervalf: no need to mock a timer, actually this code being suggested is not rigth

src/classes/worker.ts
```diff
@@ -561,7 +561,7 @@ export class Worker<
...
+      } else if (asyncFifoQueue.numQueued() < 1) {
```
manast, line 564: why < 1 instead of just === 0?

manast (approved):
Just a nitpick

src/classes/worker.ts
roggervalf: no hard requirement, it could be === 0. I'll add it in one of my other pro

PR #3324, opened 2025-07-02
test(flow): validate removal of delayed jobs when removing unprocessed children

roggervalf:
### Why

  1. Why is this change necessary? To validate removal of children jobs when removing unprocessed children, considering prioritized and delayed states

### How

_Any extra info here._


src/classes/worker.ts
```diff
@@ -561,7 +561,7 @@ export class Worker<
             this.opts.runRetryDelay,
           ),
         );
-      } else if (asyncFifoQueue.numQueued() < 1) {
+      } else if (asyncFifoQueue.numQueued() === 0) {
```
roggervalf, line 564: comment from https://github.com/taskforcesh/bullmq/pull/3322#discussion_r2180950186

Issue #2619, opened 2024-06-23
[Bug]: Is it normal to get lots of "Command timed out" from bzpopmin?


Linksku:
### Version

v5.8.2

### Platform

NodeJS

### What happened?

I'm trying to migrate from Bull to BullMQ. I'm getting an issue where my repeatable jobs often don't run. I.e. even though it's past the delay job's run time, they're not running. It could be the same issue that other people reported, e.g. https://github.com/taskforcesh/bullmq/issues/2466

In order to debug this, I tried logging errors:

```
  worker.on('error', err => {
    console.log(err);
  });
```

This is logging this error roughly every second:

```
Command timed out
  at Worker.waitForJob (node_modules/bullmq/src/classes/worker.ts:641:40)
  at Worker._getNextJob (node_modules/bullmq/src/classes/worker.ts:547:27)
```

I've narrowed down the source of the error to the `bzpopmin` call in `waitForJob`. I tried increasing the timeout, but I'm still getting the error. Is this error expected to occur? If not, what might be some potential causes for this?

I think my code is fairly standard:

```
const queue = new Queue(queueName, {
    connection: {
      host: REDIS_HOST,
      port: REDIS_PORT,
      username: REDIS_USER,
      password: REDIS_PASS,
      maxRetriesPerRequest: null,
      enableOfflineQueue: false,
    },
    prefix: 'bull',
  });

  const worker = new Worker(
    queueName,
    processJob,
    {
      connection: {
        host: REDIS_HOST,
        port: REDIS_PORT,
        username: REDIS_USER,
        password: REDIS_PASS,
        maxRetriesPerRequest: null,
        enableOfflineQueue: true
      },
    },
  );

  await queue.add(
    jobName,
    null,
    {
      repeat: {
        every: 5 * 60 * 1000,
      },
      removeOnComplete: true,
      removeOnFail: true,
    },
  );
```

### Relevant log output

```shell
1719114352.234624 [0 172.20.0.1:49996] "bzpopmin" "bull:NotifsManager:marker" "10"
1719114352.836859 [0 172.20.0.1:50220] "bzpopmin" "bull:NotifsManager:marker" "10"
1719114353.037461 [0 172.20.0.1:60906] "bzpopmin" "bull:NotifsManager:marker" "10"
1719114353.137910 [0 172.20.0.1:32874] "bzpopmin" "bull:NotifsManager:marker" "10"
[...]

PR #3326, opened 2025-07-02
fix(worker): fix bzpopmin race condition

Marsup:
### Why

As explained in #2619, the resolution order of setTimeout and bzpopmin under certain conditions is wrong.

### How

To reestablish the correct order, I'm scheduling the disconnection for the next tick to let bzpopmin eventually resolve properly.

I'm honestly not sure how to add non-regression testing on this, I'd like some advice on this if possible.

I also removed the async modifier of the timeout callback, not really sure why it was there.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -715,8 +715,14 @@ will never work with more accuracy than 1ms. */
           // We cannot trust that the blocking connection stays blocking forever
           // due to issues in Redis and IORedis, so we will reconnect if we
           // don't get a response in the expected time.
-          timeout = setTimeout(async () => {
-            bclient.disconnect(!this.closing);
+          timeout = setTimeout(() => {
+            // If the event-loop was stuck for more than 1s at the time of completion,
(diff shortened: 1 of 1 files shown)


manast (commented):
This feels like hitting the can down the road instead of actually resolving the possible underlying issue, although I am not convinced yet there is any issue :/

Marsup:
I don't like that solution myself either, but I know it works, and it'll probably keep working for the foreseeable future. The only alternative would be for ioredis to go into the microtask queue to go 1st, but I haven't been able to figure out in the driver why it was queuing itself after the other timers.

manast:
> I don't like that solution myself either, but I know it works, and it'll probably keep working for the foreseeable future. The only alternative would be for ioredis to go into the microtask queue to go 1st, but I haven't been able to figure out in the driver why it was queuing itself after the other timers.

It is solving your particular symptom, however as it is not fully clear what is the issue it is not really solving it.

PR #3328, opened 2025-07-03
docs(fail-parent): clarify lazy failures when using failParentOnFailure option

roggervalf:
### Why

  1. Why is this change necessary? Failing a parent is using a lazy movement strategy that needs to be clarified in our documentation

### How

_Any extra info here._


manast (approved):
LGTM

PR #3331, opened 2025-07-04
feat(worker): support maxStartedAttempts option

roggervalf:
### Why

  1. Why is this change necessary? Manual movement of jobs can cause infinite loop if we continue moving these jobs with our special errors. This pr allows to add a limitation. Only counting how many special attempts were done. In addition default value from attempts in our Docs was change to 1 as it was not accurate with 0 value

### How

_Any extra info here._


src/classes/worker.ts
```diff
@@ -907,6 +909,20 @@ will never work with more accuracy than 1ms. */
...
+  ) {
+    if (job.deferredFailure) {
+      return job.deferredFailure;
+    }
+    if (
+      job.opts.maxStartAttempts &&
+      job.opts.maxStartAttempts < job.attemptsStarted
+    ) {
+      return 'job started more than allowable limit';
+    }
```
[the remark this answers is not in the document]
roggervalf: deferredFailure must have more precedence

docs/gitbook/patterns/process-step-jobs.md
```diff
@@ -358,6 +358,10 @@ const worker = new Worker(
 );
 ```
 
+{% hint style="info" %}
+Manually moving jobs using special errors does **not** increment the attemptsMade property, but it does increase the attemptsStarted count. To control how many times a job is allowed to start processing, use the **maxStartAttempts** option.
```
manast: I think clarification is needed here. So when is attemptsMade incremented? it is only when a job completes or fails, or are there any other cases?

manast (commented):
How would maxAttemptsStarted be enforced in the standard failed retry case? I have a feeling this is going to be confusing, as this introduces a new variable to consider which overlaps quite a lot with our current attempts field.

manast:
I think we need to think more if we really need to introduce a new option here.

docs/gitbook/patterns/process-step-jobs.md
manast: Stupid question, but if "Manually moving jobs using special errors does **not** increment the attemptsMade property", why can't we increment attemptsMade for this manually moving jobs operations?

roggervalf: hey, because that would mean to threat these errors as real errors. Jobs failing with real errors will have less retries for non special errors

src/interfaces/base-job-options.ts
```diff
@@ -28,7 +28,7 @@ export interface DefaultJobOptions {
 
   /**
    * The total number of attempts to try the job until it completes.
-   * @defaultValue 0
+   * @defaultValue 1
```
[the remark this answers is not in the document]
roggervalf: pr description was change to reflect this change

src/commands/includes/checkMaxSkippedAttempts.lua
```diff
@@ -0,0 +1,12 @@
+--[[
+  Function to check max skipped attempts. If allowable limit is exceeded,
+  it sets the job deferred failure.
+]]
+
+local function checkMaxSkippedAttempts(jobKey, maxSkippedAttemptCount)
+  local skippedAttemptCount = rcall("HINCRBY", jobKey, "sac", 1)
+  if maxSkippedAttemptCount > 0 and skippedAttemptCount > maxSkippedAttemptCount then
+    local failedReason = "job skipped more than allowable attempts"
+    rcall("HSET", jobKey, "defa", failedReason)
```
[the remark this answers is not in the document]
roggervalf: this is actually expected to know how many times it was using a manual movement. It will be informative to end users even of they do not add a limit

src/commands/moveToDelayed-8.lua
```diff
@@ -45,18 +47,23 @@ if rcall("EXISTS", jobKey) == 1 then
...
     local numRemovedElements = rcall("LREM", KEYS[2], -1, jobId)
     if numRemovedElements < 1 then return -3 end
 
+    local score, delayedTimestamp = getDelayedScore(delayedKey, ARGV[2], delay)
+
     if ARGV[6] == "0" then
         rcall("HINCRBY", jobKey, "atm", 1)
+    else
+        local maxSkippedAttemptCount = tonumber(ARGV[7])
+
```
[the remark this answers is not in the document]
roggervalf: when ARGV[6] is not "0", means that it's "1", no need to add a new if statement

docs/gitbook/patterns/process-step-jobs.md
manast: I am not sure that is necessarily wrong. I mean, if the job has actually started then it should increase attemptsMade independently on the reason for starting it. I am just trying to not open a new can of worms here if we introduce new nomenclature if it is not strictly necessary.

roggervalf: oh but we hav a separation to prevent increasing attemptsMade that is attemptStarted, that was introduces with the special errors that should not consume the attemptsMade to prevent treating them as real errors. Put it on the other side, if we reuse attemptsMade for this will cause the generation of extran delays for other real errors too, as we base the generation of delays base on current attemptMade

src/classes/worker.ts
```diff
@@ -910,6 +919,20 @@ will never work with more accuracy than 1ms. */
...
 
+  private getUnrecoverableErrorMessage(
+    job: Job<DataType, ResultType, NameType>,
+  ) {
+    if (job.deferredFailure) {
+      return job.deferredFailure;
+    }
+    if (
+      this.opts.maxStartedAttempts &&
+      this.opts.maxStartedAttempts < job.attemptsStarted
```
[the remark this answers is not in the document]
roggervalf: max started attempts value is still a valid value

src/classes/worker.ts
```diff
@@ -910,6 +919,20 @@ will never work with more accuracy than 1ms. */
...
+    ) {
+      return 'job started more than allowable limit';
```
manast, line 980: it may be useful to interpolate the job.id in the error message.

manast (commented):
just a small comment

src/classes/worker.ts
roggervalf: it may be redundant I think. As it would be saved in job instance. To obtain this error, you should get the job by id. This is similar to other unrecoverable errors like https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveStalledJobsToWait-8.lua#L77 that do not contain jobId when it's saved in job instance

Issue #3334, opened 2025-07-07
Wildcard option for custom events


Mihailoff:
BullMQ has an option to use custom events: https://docs.bullmq.io/guide/events/create-custom-events

I wish there was a way to listen to all the events via catch-all/wildcard listener.

```JS
const producer = new QueueEventsProducer('eventStream');

// Create consumer
const consumer = new QueueEvents('eventStream');

consumer.on('*', (eventName, eventData) => {
  console.log(`Received event: ${eventName}`, eventData);
});
```

My use case is to handle all events, check metadata and call specific handler.

I tried generic events but they don't work for some reason. Don't know if this is expected or not.

```JS
consumer.on('complete', (eventName, eventData) => {
  console.log(`Received event: ${eventName}`, eventData);
});
```

Mihailoff:
Related to: https://github.com/taskforcesh/bullmq/issues/831

Mihailoff:
A second though here: maybe wildcard is not the best design. It is intuitive, but adds special asterisk character that can lead to unintended bugs.

I think it is better to have a dedicated method like this

```JS
consumer.onAll(({ eventName, data }) => {
  console.log(`Received event: ${eventName}`, data);
})
```

PR #3335, opened 2025-07-07
Implement wildcard listeners

Mihailoff:
### Why

https://github.com/taskforcesh/bullmq/issues/3334

Ability to listen on any event:

```JS
consumer.on('*', (eventName, eventData) => {
  console.log(`Received event: ${eventName}`, eventData);
});
```

I'm not very familiar with the code base, hopefully, this is the right direction.

### DIFF
--- a/tests/test_events.ts
+++ b/tests/test_events.ts
@@ -925,6 +925,7 @@ describe('events', function () {
       }
 
       await worker.close();
+      await queueEvents.close();
     });
 
     it('should call wildcard listener for custom events published by QueueEventsProducer', async function () {


src/classes/queue-events.ts
```diff
@@ -310,30 +311,81 @@ export class QueueEvents extends QueueBase {
     QEL extends QueueEventsListener = QueueEventsListener,
     U extends KeyOf<QEL> = KeyOf<QEL>,
   >(event: U, ...args: CustomParameters<QEL[U]>): boolean {
-    return super.emit(event, ...args);
+    // Emit to normal listeners
+    const emitted = super.emit(event, ...args);
+    // Emit to wildcard listeners, except for '*'
+    if (event !== '*' && this.wildcardListeners.length > 0) {
```
manast, line 317: why do we need to discriminate the * event here?

Mihailoff: You are right, I think it should be just 

`if (this.wildcardListeners.length > 0) { ... }`

Mihailoff:
@manast what do you think about such design in general? 

I tend to think that it is not the best choice and instead it should be something like example below without magical `*`.

```JS
queueEvents.onAll(handler)
``` 

In my case I ended up with a direct Redis consumer group listener to get full control of it.

Issue #3125, opened 2025-03-06
[Bug]: Inconsistent Use of "wait" vs "waiting" in BullMQ


azharuniverse:
### Version

5.41.5

### Platform

NodeJS

### What happened?

# 🛠 Inconsistent Use of "wait" vs "waiting" in BullMQ  

## Description  
BullMQ inconsistently uses `"wait"` in some functions and `"waiting"` in others, leading to confusion when working with job states.  

## Steps to Reproduce  
1. Call `queue.getJobCounts()`, and it returns `{ waiting: 10, active: 5, completed: 20 }`.  
2. Call `queue.clean(0, 100, "wait")`, and it works, but `queue.clean(0, 100, "waiting")` does not, adding to the inconsistency.  

## Expected Behavior  
- The job state naming convention should be **consistent** across all functions.  
- Either `"wait"` or `"waiting"` should be used everywhere, not a mix of both.  

## Actual Behavior  
- `queue.getJobCounts()` uses `"waiting"`, while `queue.clean()` expects `"wait"`.  
- This inconsistency makes it difficult for developers to predict the correct parameter.  

## Screenshots / Logs  
```json
// Output of queue.getJobCounts()
{
  "waiting": 10,
  "active": 5,
  "completed": 20
}

// queue.clean() usage
await queue.clean(0, 100, "wait");  // ✅ Works
await queue.clean(0, 100, "waiting"); // ❌ Fails

// Expected: Either "wait" or "waiting" should be used consistently.


### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

roggervalf:
thanks @azharuniverse, this is fair. But deprecating one of those parameter names would imply a breaking change. I'll take this issue in mind for next breaking change version

manast:
> 2\. Call `queue.clean(0, 100, "wait")`, and it works, but `queue.clean(0, 100, "waiting")` does not, adding to the inconsistency.

I think we could support "waiting" as well in clean, and if so it would not be a breaking change.

PR #3338, opened 2025-07-10
fix: add support for 'waiting' parameter in queue.clean() method

Amitrochates:
- Add 'waiting' as accepted parameter type in clean() method signature
- Implement normalization logic to convert 'waiting' to 'wait' internally
- Maintain backward compatibility with existing 'wait' parameter
- Update event emission to use normalized type for consistency
- Add comprehensive tests for both 'wait' and 'waiting' parameters

This resolves the inconsistency where getJobCounts() returns 'waiting' key but clean() method only accepted 'wait' parameter. Now developers can use consistent terminology across the API.

Closes: https://github.com/taskforcesh/bullmq/issues/3125 inconsistent use of 'wait' vs 'waiting' in BullMQ API

### Why

the clean method doesnt allow "waiting" while other places use "wait" and "waiting"
https://github.com/taskforcesh/bullmq/issues/3125 

### How

I referenced another place in queue-getter.ts where type aliasing was done to "waiting" to "wait" and implemented a similar normalised type in clean() function

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -879,6 +879,7 @@ export class Queue<
     type:
       | 'completed'
       | 'wait'
+      | 'waiting'
       | 'active'
       | 'paused'
       | 'prioritized'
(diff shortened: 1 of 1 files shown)


Amitrochates:
@manast @roggervalf  please review

src/classes/queue.ts
```diff
@@ -896,14 +897,17 @@ export class Queue<
         let deletedCount = 0;
         const deletedJobsIds: string[] = [];
 
+        // Normalize 'waiting' to 'wait' for consistency with internal Redis keys
+        const normalizedType = type === 'waiting' ? 'wait' : type;
```
roggervalf, line 901: we might need a test case to cover this change, we have some examples in https://github.com/taskforcesh/bullmq/blob/master/tests/test_clean.ts#L98

Amitrochates: Added !

roggervalf (approved):
Lgtm

Issue #3341, opened 2025-07-11
[Bug]: skipVersionCheck option doesn't prevent INFO command execution


haansuk:
### Version

v5.56.2

### Platform

NodeJS

### What happened?

The `skipVersionCheck: true` option in BullMQ doesn't actually prevent the `INFO` command from being sent to Redis. Instead, it only skips the version validation logic while still executing the `getRedisVersion()` method, which calls `this._client.info()`.

This causes issues in environments where:
1. Redis proxy doesn't support the `INFO` command
2. Security policies block the `INFO` command
3. Network restrictions prevent `INFO` command execution

**Current behavior:**
```typescript
// In redis-connection.ts init() method
if (this._client['status'] !== 'end') {
  // This line ALWAYS executes, regardless of skipVersionCheck
  this.version = await this.getRedisVersion();
  
  // Only this check is skipped when skipVersionCheck: true
  if (this.skipVersionCheck !== true && !this.closing) {
    // Version validation logic
  }
}
```

**Expected behavior:**
When `skipVersionCheck: true` is set, the `getRedisVersion()` call should be completely skipped, and a default version should be used instead.

### How to reproduce.

1. Set up a Redis proxy that doesn't support the `INFO` command
2. Create a BullMQ Worker with `skipVersionCheck: true`:

```typescript
const worker = new Worker('test-queue', async (job) => {
  // job processing
}, {
  connection: redisConnection,
  skipVersionCheck: true
});
```

3. Observe that the `INFO` command is still sent to Redis, causing:
```
ReplyError: ERR unsupported command by proxy `info`
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

haansuk:
### Root Cause Analysis

The issue is in the `RedisConnection.init()` method in `src/classes/redis-connection.ts`:

```typescript
private async init() {
  // ... connection setup ...
  
  if (this._client['status'] !== 'end') {
    // ❌ This always executes, even with skipVersionCheck: true
    this.version = await this.getRedisVersion();
    
    // ✅ Only this check is conditionally skipped
    if (this.skipVersionCheck !== true && !this.closing) {
      // Version validation logic
    }
  }
}
```

The `getRedisVersion()` method always calls `this._client.info()`, which sends the `INFO` command to Redis regardless of the `skipVersionCheck` setting.

### Proposed Solution

[...]

PR #3342, opened 2025-07-11
fix(connection): ignore info command when skipVersionCheck is provided as true

roggervalf:
### Why

Why is this change necessary? When info command is not allowed to be called, even using skipVersionCheck, this command is called.

### How

  1. How did you implement this? Use minimum version by default when passing skipVersionCheck as true

Fixes https://github.com/taskforcesh/bullmq/issues/3341

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -360,28 +360,32 @@ export class RedisConnection extends EventEmitter {
   }
 
   private async getRedisVersion() {
-    const doc = await this._client.info();
-    const redisPrefix = 'redis_version:';
-    const maxMemoryPolicyPrefix = 'maxmemory_policy:';
-    const lines = doc.split(/\r?\n/);
-    let redisVersion;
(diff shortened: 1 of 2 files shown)


manast (approved):

matheusb-comp:
@roggervalf What are the impacts of returning the minimum version?

Because the [capabilities](https://github.com/taskforcesh/bullmq/blob/d9ab5742120e4e4e8134c01966808173c0355585/src/classes/redis-connection.ts#L277) map, for example, needs versions above the current minimum (above even the recommended minimum).

Skipping the `INFO` command is also important for the Queue option `skipWaitingForReady` (if the connection is not `ready` asking for the version could cause problems).

Is there a way to manually provide the Redis server version that I'm missing?

manast:
If version cannot be obtained then some features will fallback on slower, less efficient versions of the code, but everything should work, unless version is less than the minimum required which is 5.0.0 for now.

PR #3346, opened 2025-07-13
fix(worker): emit failed event when children are failed in moveToWaitingChildren

roggervalf:
### Why

 1. Why is this change necessary? When failing a job in moveToWaitingChildren, no failed event is emitted in worker because WaitingChildrenError is ignored

### How

  1. How did you implement this? Throw an UnrecoverableError with a new code when there are failed children, in this way we will emit failed event

_Any extra info here._


src/commands/moveStalledJobsToWait-8.lua
```diff
@@ -73,20 +74,10 @@ if (#stalling > 0) then
                     -- If this job has been stalled too many times, such as if it crashes the worker, then fail it.
                     local stalledCount = rcall("HINCRBY", jobKey, "stc", 1)
                     if stalledCount > maxStalledJobCount then
-                        local jobAttributes = rcall("HMGET", jobKey, "opts", "parent")
```
roggervalf, line 76: this command is not needed anymore as we are not doing anything special with these extra jobAttributes after we are using lazy failures

python/bullmq/error_code.py
```diff
@@ -9,4 +9,5 @@ class ErrorCode(Enum):
     ParentJobNotExist = -5
     JobLockMismatch = -6
     ParentJobCannotBeReplaced = -7
-    JobFailedChildren = -9
+    ChildJobFailed = -9
+    ChildJobsFailed = -10
```
manast: ```ChildJobsFailed``` is a too close naming to ```ChildJobFailed```, so it may get confusing, can we find a better name?

python/bullmq/scripts.py
```diff
@@ -669,8 +669,10 @@ def finishedErrors(self, opts: dict) -> TypeError:
             return TypeError(f"Lock mismatch for job {opts.get('jobId')}. Cmd {opts.get('command')} from {opts.get('state')}")
         elif code == ErrorCode.ParentJobCannotBeReplaced.value:
             return TypeError(f"The parent job {opts.get('jobId')} cannot be replaced. {opts.get('command')}")
-        elif code == ErrorCode.JobFailedChildren.value:
+        elif code == ErrorCode.ChildJobFailed.value:
             return TypeError(f"Job {opts.get('jobId')} has failed children. {opts.get('command')}")
+        elif code == ErrorCode.ChildJobsFailed.value:
+            return UnrecoverableError(f"Children are failed. {opts.get('command')}")
```
manast: Shouldn't the message be: "Children have failed", but the question is, all children have failed, or any child has failed? I think we should make the message more clear.

python/bullmq/scripts.py
```diff
@@ -669,8 +669,10 @@ def finishedErrors(self, opts: dict) -> TypeError:
             return TypeError(f"Lock mismatch for job {opts.get('jobId')}. Cmd {opts.get('command')} from {opts.get('state')}")
         elif code == ErrorCode.ParentJobCannotBeReplaced.value:
             return TypeError(f"The parent job {opts.get('jobId')} cannot be replaced. {opts.get('command')}")
-        elif code == ErrorCode.JobFailedChildren.value:
+        elif code == ErrorCode.ChildJobFailed.value:
```
manast: Ok, so this case means: "we cannot do X because the Job Y has failed children". I think we should improve the enum, maybe "JobHasFailedChildren", and also the error message to reflect this.

python/bullmq/scripts.py
manast: It is not clear to me what this error is saying, nor what the enum should be.

src/classes/scripts.ts
```diff
@@ -1706,8 +1706,10 @@ export class Scripts {
         return new Error(
           `Job ${jobId} belongs to a job scheduler and cannot be removed directly. ${command}`,
         );
-      case ErrorCode.JobFailedChildren:
+      case ErrorCode.ChildJobFailed:
```
manast: Some comments as for the python code above.

src/commands/includes/moveJobToWaitImmediately.lua
```diff
@@ -0,0 +1,15 @@
+--[[
+  Function to move job to active to immediately be picked up by a waiting worker.
```
manast: The comment seems to be wrong, it should say "...move job to *wait* to immediately ..."

src/commands/includes/moveJobToWaitImmediately.lua
```diff
@@ -0,0 +1,15 @@
...
+]]
+
+-- Includes
+--- @include "addJobInTargetList"
+--- @include "getTargetQueueList"
+
+local function moveJobToWaitImmediately(metaKey, activeKey, waitKey, pausedKey, markerKey, eventStreamKey, jobId)
```
manast: What is the difference between moveJobToWait and moveJobToWaitImmediately?

python/bullmq/error_code.py
roggervalf: In reality this is a good point, as both cases are the same, I was an overlook from my side. I'll refactor it to use the same enum and code number

src/commands/includes/moveJobToWaitImmediately.lua
roggervalf: we don't have an include like mlveJoToWait, but https://github.com/taskforcesh/bullmq/blob/master/src/commands/includes/moveParentToWait.lua is moving a job at the tail of wait state, this one as it's usage in stalled jobs, move a job at the head of wait state to be consumed immediately

roggervalf: same case as https://github.com/taskforcesh/bullmq/blob/master/src/commands/moveJobFromActiveToWait-9.lua that moves a job to the tail

manast: I think we should consider having only one method "moveToWait" that has an argument to decide between tail or head, because depending if the queue is FIFO or LIFO it has different meanings. Besides, "immediately" can be confusing as this would just be a side effect of moving to tail or head, but there is no guarantee that it would be processed immediately.

manast (approved):

Issue #3349, opened 2025-07-17
[Bug]: Wait fetched jobs to be processed when closing


mishsoup:
### Version

v5.56.4

### Platform

NodeJS

### What happened?

Rare, but possible situation, difficult to reproduce.

Job picked up by a worker is stalling when we close that worker using ```await worker.close()```

What we observed:
- all workers in the cluster with QueueEventListener observes 'active' event for jobA
- local worker 'active' event for jobA never observed
- observes worker closing event and closed event
- jobA stalls

What we suspect is happening:
worker configuration:
```this._concurrency = 2```

1. Mainloop starts running its initial iteration upon starting. In the inner loop, we fetched our first job from ```this._getNextJob``` and it is stored in the ```fetchedJob``` variable. We add this job fetching promise into the ```asyncFifoQueue```. The job fetching promise sits in the the ```this.pending``` set in the ```asyncFifoQueue```. ```asyncFifoQueue.numTotal()``` = 1 (1 in ```this.pending```, 0 in ```this.queue```).
2. Next, ```const job = await fetchedJob;``` [L529](https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L529). We await until the job fetching promise resolves. At this point in time, ```asyncFifoQueue.numTotal()``` = 1 (0 in ```this.pending```, 1 in ```this.queue```) resolved promises are added to this.queue[L87](https://github.com/taskforcesh/bullmq/blob/master/src/classes/async-fifo-queue.ts#L87).
3. Next, we reach the second iteration of the inner loop since ```numTotal < this._concurrency``` 1 < 2 We now fetched our second job from ```this._getNextJob``` and await the job fetching promise to resolve same as above. At this point in time ```asyncFifoQueue.numTotal()``` = 2 (0 in ```this.pending```, 2 in ```this.queue```)
4. We now break out of the inner loop - ```numTotal ```(2) is not smaller then ``` this._concurrency```(2), next we fetch a job from ```asyncFifoQueue``` via ```await asyncFifoQueue.fetch()``` [L547](https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L547C15-L547C43)
```asyncFifoQueue.numTotal()``` = 1 (0 in ```this.pending```, 1 in ```this.queue```)
5.  Start closing the worker via  ```await worker.close()```, ```this.closing``` is truthy now
[...]

PR #3350, opened 2025-07-17
fix(worker): do not keep active jobs when pausing or closing

roggervalf:
### Why

  1. Why is this change necessary? While pausing or closing a worker, both method wait for pending jobs to be processed before resolving. But we are only taking in count those jobs that are currently active (calling processor method). These jobs that are currently processing can return a new job instance as part of a raise condition where fetch next boolean was set as true when moving these jobs to completed or failed. Job instances being returned will be in active state, so we need to try to process them as well, in these way we are waiting for all jobs (currently processed and pending ones - returned by processors)

### How

  1. How did you implement this? Instead of calling waitAll method that only waits for pending methods that could possible return job instances, just keep looping in main while loop. Our asynchronous queue instance still will execute pending promises in parallel, but we will be able to consume those jobs being returned and apply a processor. At the end we will wait for all pending and processed jobs. Take in count that as soon as a worker is paused or closing, we won't try to fetch new jobs using bzpopmin.

fixes https://github.com/taskforcesh/bullmq/issues/3349

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -491,7 +491,7 @@ export class Worker<
 
     let tokenPostfix = 0;
 
-    while (!(this.closing || this.paused) || asyncFifoQueue.numTotal() > 0) {
+    while ((!this.closing && !this.paused) || asyncFifoQueue.numTotal() > 0) {
       let numTotal = asyncFifoQueue.numTotal();
 
       /**
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -491,7 +491,7 @@ export class Worker<
 
     let tokenPostfix = 0;
 
-    while (!this.closing && !this.paused) {
+    while (!(this.closing || this.paused) || asyncFifoQueue.numTotal() > 0) {
       let numTotal = asyncFifoQueue.numTotal();
 
       /**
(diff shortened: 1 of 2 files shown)


src/classes/worker.ts
```diff
@@ -1018,7 +1016,9 @@ will never work with more accuracy than 1ms. */
 
         if (!this.paused) {
           this.paused = true;
-          await (!doNotWaitActive && this.whenCurrentJobsFinished());
+          if (!doNotWaitActive) {
```
roggervalf, line 1015: refactored for readability

src/classes/worker.ts
```diff
@@ -491,7 +491,7 @@ export class Worker<
 
     let tokenPostfix = 0;
 
-    while (!this.closing && !this.paused) {
+    while ((!this.closing && !this.paused) || asyncFifoQueue.numTotal() > 0) {
```
manast, line 494: Since ```asyncFifoQueue.numTotal()``` will always be called at least once, move ```let numTotal = asyncFifoQueue.numTotal()``` up and use numTotal in the condition to avoid calling it twice.

roggervalf: I replace all those referentes with numTotal() directly

manast (approved):
LGTM

PR #3383, opened 2025-08-06
fix(job-scheduler): correct parameter for repeatStrategy in nextMillis calculation

ZacMelendez:
### Why

This change improves the scheduling logic in `JobScheduler` by ensuring that the next scheduled job time (`nextMillis`) is always set to a valid future time, especially when using cron patterns. Previously, when using a pattern, the `nextMillis` was always calculated using `now`, which would not properly set the next schedule in case the `startDate` was set at some point in the future.

### How

- Updated the logic in `upsertJobScheduler` to have `nextMillis` use `startMillis` when using a cron `pattern`. `startMillis` will be `now` if the `startDate` is in the past, or the epoch time for `startDate` if it is in the future.

- No breaking changes.
- No new dependencies introduced.

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -108,7 +108,7 @@ export class JobScheduler extends QueueBase {
         nextMillis = prevSlot;
       }
     } else if (pattern) {
-      nextMillis = await this.repeatStrategy(now, repeatOpts, jobName);
+      nextMillis = await this.repeatStrategy(startMillis, repeatOpts, jobName);
 
       if (nextMillis < now) {
         nextMillis = now;
--- a/tests/test_job_scheduler.ts
+++ b/tests/test_job_scheduler.ts
@@ -689,7 +689,7 @@ describe('Job Scheduler', function () {
     );
     const delayStub = sinon.stub(worker, 'delay').callsFake(async () => {});
 
-    await queue.upsertJobScheduler(
+    const scheduler = await queue.upsertJobScheduler(
       'test',
       {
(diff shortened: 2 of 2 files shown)


src/classes/job-scheduler.ts
```diff
@@ -108,7 +108,7 @@ export class JobScheduler extends QueueBase {
         nextMillis = prevSlot;
       }
     } else if (pattern) {
-      nextMillis = await this.repeatStrategy(now, repeatOpts, jobName);
+      nextMillis = await this.repeatStrategy(startMillis, repeatOpts, jobName);
```
roggervalf, line 111: may need a test case to cover this change

ZacMelendez: Should [this one](https://github.com/taskforcesh/bullmq/blob/9cbc327b9c56d7ab451a801866834bdaa0993660/tests/test_job_scheduler.ts#L675) already cover it? Or should it be more than 5s?

roggervalf: it's a false positive I believe because we are mocking, something that can be is to validate the delay value that is assigned to the added job (job.delay), so we know that is having the correct value

ZacMelendez: I've added a check to see if the initial delay for the scheduler is correct. Paired with the check to see if the time between executions is `2000ms`, that should be good to cover the change.

roggervalf: I opened a pr base on this one for addressing this issue

Issue #3382, opened 2025-08-06
[Bug]: FlowProducer#getFlow() returns undefined children when colon characters are used in `jobId`s


freitasskeeled:
### Version

v5.56.9

### Platform

NodeJS

### What happened?

FlowProducer#getFlow() returns an array of undefined children when descendant jobs passed to FlowProducer#add() have custom `jobId`s containing colon (:) characters.

I used [this issue](https://github.com/taskforcesh/bullmq/issues/2718) as reference because it looks to be the exact same issue but it is related with custom `jobId`s instead of `queueName`s.

### How to reproduce.

```
import { FlowProducer } from 'bullmq';

const flow = new FlowProducer();

const { job } = await flow.add({
    name: 'parent',
    queueName: 'parent',
    children: [{ name: 'child', queueName: 'parent:child', opts: { jobId: 'custom:job:id:123' } }],
    opts: {
        jobId: 'custom:job:id:123123',
    },
});

const { children } = await flow.getFlow({
  id: job.id,
  queueName: 'parent'
});

console.log(children); // → [undefined]
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

roggervalf:
hi @freitasskeeled, we are actually not recommending to use : inside custom job ids https://docs.bullmq.io/guide/jobs/job-ids as in the last message in our documentation

freitasskeeled:
> hi [@freitasskeeled](https://github.com/freitasskeeled), we are actually not recommending to use : inside custom job ids https://docs.bullmq.io/guide/jobs/job-ids as in the last message in our documentation

I missed that note 😅 

I'll try to open a PR to add a check/throw exception if this is the case since it can "break silently". I believe this was done for the issue regarding colons in queue names.

Thanks!

PR #3384, opened 2025-08-07
fix(job): add custom job id validation to prevent : inclusion

roggervalf:
### Why

1. Why is this change necessary? Similar to https://github.com/taskforcesh/bullmq/pull/2719 but at jobId level. There was a message being added some time ago in https://github.com/taskforcesh/bullmq/pull/2251, but an actual error is missing

### How

  1. How did you implement this? Throw error when : is contained in custom job id

fixes https://github.com/taskforcesh/bullmq/issues/3382

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1465,6 +1465,10 @@ export class Job<
       throw new Error('Custom Ids cannot be integers');
     }
 
+    if (this.id.includes(':')) {
+      throw new Error('Custom Id cannot contain :');
+    }
+
     if (this.opts.priority) {
       if (Math.trunc(this.opts.priority) !== this.opts.priority) {
         throw new Error(`Priority should not be float`);
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1468,7 +1468,10 @@ export class Job<
 
       // TODO: replace this check in next breaking check with include(':')
       // By using split we are still keeping compatibility with old repeatable jobs
-      if (this.opts?.jobId?.split(':').length !== 3) {
+      if (
+        this.opts?.jobId.includes(':') &&
+        this.opts?.jobId?.split(':').length !== 3
+      ) {
         throw new Error('Custom Id cannot contain :');
       }
     }
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1461,11 +1461,11 @@ export class Job<
       );
     }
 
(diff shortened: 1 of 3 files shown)


src/classes/job.ts
```diff
@@ -1461,8 +1461,19 @@ export class Job<
...
+    if (this.opts?.jobId) {
+      if (`${parseInt(this.opts.jobId, 10)}` === this.opts?.jobId) {
+        throw new Error('Custom Id cannot be integers');
+      }
+
+      // TODO: replace this check in next breaking check with include(':')
+      // By using split we are still keeping compatibility with old repeatable jobs
+      if (
+        this.opts?.jobId.includes(':') &&
+        this.opts?.jobId?.split(':').length !== 3
```
manast, line 1473: ok, so if the user uses a userId that has 3 colons in it, it will actually work. The thing that seems strange is why would an old repeatable job work, but not standard jobs?

manast (approved):

src/classes/job.ts
roggervalf: hey, because old repeatable jobs sets this value when generating it https://github.com/taskforcesh/bullmq/blob/master/src/classes/repeat.ts#L129-L137

michalvankodev:
Please take this only as constructive criticism: This change got passed through our renovate / dependabot setup and got merged automatically as it was included in a patch release. It broke all our jobs that have been using `:`. Why did you not consider this a breaking change?

manast:
Because it was a bug. In your case it was not manifested as a bug as you did not trigger the bug if you did not use flows, but yes, sometimes there is a thin line between a bug fix and a breaking change. A good example of this is Microsoft Windows, which their old APIs had bugs they could never fix not to break older software.

matsko:
This is breaking change. Please update your changelog pages to reflect that.

nwalters512:
This also caught us by surprise in production. IMO at a _minimum_ this should have been a minor release, not a patch. We did not use flows, have no plans to use flows, will not use flows in the future. While it's true that this change appears in your release notes, the consequences to users of this library aren't at all apparent from this text:

> job: add custom jobId validation to prevent : inclusion

For the sake of other users, I'd also strongly advise you to make the changelog super explicit about what this change means for users.

korac:
One more case here 👋 This change also caught us by surprise in production.
Updating changelog to reflect this as a breaking change would really help future users. Voting for that as well. 👍

PR #3389, opened 2025-08-11
feat(job): Add moveToWaitingChildren to the wrapper of sandboxed jobs

### LINKED ISSUE #3230: Implementing moveToWaitingChildren in SandboxedJobs

pkudinov:
### Why

Currently it is impossible to implement https://docs.bullmq.io/patterns/process-step-jobs#waiting-children pattern in sandboxed jobs due to:
- missing `moveToWaitingChildren` method
- missing `queueQualifiedName`

### How

This is built on top of #3231, but adds tests following the waiting for children pattern in the linked documentation.
It fixes #3230

1. Added `queueQualifiedName` and `moveToWaitingChildren` to `SandboxedJob`
2. Implemented a unit test

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -125,7 +125,7 @@ export class ChildProcessor {
   ): SandboxedJob {
     const wrappedJob = {
       ...job,
-      queueQualifiedName: job.prefix + ':' + job.queueName,
+      queueQualifiedName: job.queueQualifiedName,
       data: JSON.parse(job.data || '{}'),
       opts: job.opts,
       returnValue: JSON.parse(job.returnvalue || '{}'),
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -174,7 +174,7 @@ export class ChildProcessor {
        * Proxy `moveToWaitingChildren` function.
        */
       moveToWaitingChildren: async (
-        token: string,
+        token?: string,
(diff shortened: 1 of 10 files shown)


pkudinov:
@roggervalf thanks for running the tests. Not sure how to interpret it – does it mean it breaks on specific versions of redis (dragonfly) or it is related to a specific version of node? What's the best way to reproduce it locally?

UPD: there was a missing prefix in test – ran with `BULLMQ_TEST_PREFIX={b}` locally and was able to reproduce the issue. We can run tests again

PR #3390, opened 2025-08-12
docs(bullmq-pro): split changelogs

roggervalf:
### Why

  1. Why is this change necessary? Follow same pattern as bullmq changelogs, split them and keep files dedicated per version

### How

_Any extra info here._


manast:
I think it is better to improve the release script in bullmq-pro to generate a PR like this with updated changelogs as soon as a new BullMQ Pro release is performed.

roggervalf:
> I think it is better to improve the release script in bullmq-pro to generate a PR like this with updated changelogs as soon as a new BullMQ Pro release is performed.

that's a fix that should be made on bullmq-pro side because current approach is replacingbthe entire file not adding new changes. This pr is for spliting changelogs on bullmq-pro

manast (approved):

PR #3398, opened 2025-08-17
chore: refactor lock management into separate class

manast:
### Why

This PR aims to provide a new option when instantiating Workers, so that it is possible to use a dedicated thread for lock extension and job stalling detection. The reason for allowing this, is so that we can have a safer lock renewal mechanism even in the cases where the jobs are not respecting the best practice of not keeping the event loop busy for long periods of time.

### How

We refactor the lock handling in a separate class "LockManager" and then we add support to this class to run in a separate thread that communicates with the main process. Note that this approach is not a fail proof solution to the issue of jobs that keep the event loop busy, as it could also happen that the communication between the main thread and the lock extension thread stalls and issues related to locks may still happen, however it should reduce the number of cases when this happens.

_Any extra info here._

### DIFF
--- a/src/classes/lock-manager.ts
+++ b/src/classes/lock-manager.ts
@@ -1,0 +1,337 @@
+import { LockManagerOptions } from '../interfaces';
+import { Scripts } from './scripts';
+import { QueueBase } from './queue-base';
+import { SpanKind, TelemetryAttributes } from '../enums';
+import { IoredisListener } from '../interfaces';
+
+export interface LockManagerListener extends IoredisListener {
+  /**
+   * Listen to 'error' event.
+   *
+   * This event is triggered when an error occurs during lock operations.
+   */
+  error: (error: Error) => void;
+
+  /**
+   * Listen to 'lockRenewalFailed' event.
+   *
+   * This event is triggered when lock renewal fails for one or more jobs.
+   */
+  lockRenewalFailed: (jobIds: string[]) => void;
+
+  /**
+   * Listen to 'stalledJobs' event.
+   *
+   * This event is triggered when stalled jobs are detected and moved back to waiting.
+   */
+  stalledJobs: (jobIds: string[]) => void;
+
+  /**
+   * Listen to 'locksRenewed' event.
+   *
+   * This event is triggered when locks are successfully renewed.
+   */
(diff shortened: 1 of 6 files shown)


docs/gitbook/bullmq-pro/changelogs/changelog-v1.md
```diff
@@ -1,188 +1,161 @@
 ## [1.4.1](https://github.com/taskforcesh/bullmq-pro/compare/v1.4.0...v1.4.1) (2022-04-19)
 
-
 ### Bug Fixes
 
-* **deps:** upgrade bullmq to 1.80.4 ([#46](https://github.com/taskforcesh/bullmq-pro/issues/46)) ([fc2818f](https://github.com/taskforcesh/bullmq-pro/commit/fc2818f09ddd273b7cc16c54fcd9650e1d456d04))
+- **deps:** upgrade bullmq to 1.80.4 ([#46](https://github.com/taskforcesh/bullmq-pro/issues/46)) ([fc2818f](https://github.com/taskforcesh/bullmq-pro/commit/fc2818f09ddd273b7cc16c54fcd9650e1d456d04))
```
roggervalf, line 5: no need to apply lint in changelog files, need to restore format

src/classes/lock-manager.ts
```diff
@@ -0,0 +1,337 @@
...
+      this.startStalledCheckTimer();
+    }
+  }
+
+  private async stalledChecker() {
+    while (!(this.closing || this.paused)) {
+      try {
+        await this.checkConnectionError(() => this.moveStalledJobsToWait());
+      } catch (err) {
+        this.emit('error', <Error>err);
```
roggervalf, line 83: no need for this extra wrapper of try catch as checkConnectionError handles this same emission

src/classes/worker.ts
```diff
@@ -449,7 +478,10 @@ export class Worker<
         return;
       }
 
-      await this.startStalledCheckTimer();
+      this.initializeLockManager();
```
roggervalf, line 491: I do not see useDedicatedLockThread option being used. It's referenced in docs but I cannot see it in lock-manager class or in worker

Issue #3106, opened 2025-02-24
getJobScheduler return type


cann0nball:
Question about the change in [5.40.3](https://github.com/taskforcesh/bullmq/releases/tag/v5.40.3), getJobScheduler  also returning undefined.

Shouldn't the return type also reflect that, otherwise it can lead to nasty bugs when upgrading versions?

PR #3412, opened 2025-08-29
fix(job-scheduler): consider undefined type in getJobScheduler retrun…

roggervalf:
… type

### Why

  1. Why is this change necessary? Need to reflect possible return values when getting a job scheduler, including undefined

### How

fixes https://github.com/taskforcesh/bullmq/issues/3106


manast (approved):
LGTM

PR #3416, opened 2025-08-29
fix(worker): add missing processor type export

medanat:
### Why

Started getting the following error:

> error TS2694: Namespace '"/node_modules/bullmq/dist/esm/index"' has no exported member 'Processor'.

### How

Added missing processor type export to the types index


roggervalf:
thanks @medanat I opened another pr for this change. It should be handled there.

medanat:
@roggervalf sounds good!

PR #3422, opened 2025-08-31
fix(job-scheduler): correct prevSlot generation using current startMillis as now when every option is provided

roggervalf:
### Why

  1. Why is this change necessary? Every sometimes is generating more delays than expected while calculating prevSlot. Adding one extra every delay and skip iterations. This is why our stress test was flaky.

### How

  1. How did you implement this? StartMillis should not have an addition of offset, it should be current time and create a prevSlot base on that time

_Any extra info here._


manast:
It seems like the fix is breaking other tests.

src/classes/job-scheduler.ts
```diff
@@ -86,7 +86,7 @@ export class JobScheduler extends QueueBase {
 
     // Check if we have a start date for the repeatable job
     const { startDate, immediately, ...filteredRepeatOpts } = repeatOpts;
-    let startMillis = (now > prevMillis ? now : prevMillis) + (offset || 0);
+    let startMillis = now;
```
manast: I do not think we can just remove this code.

manast: I do not remember exactly, but this was important to fix other issues.

roggervalf: a test case was added, also this logic was changed

Issue #3421, opened 2025-08-31
[Bug]: (regression) getJob has type `any`, getJobs has type `any[]`


nktnet1:
### Version

5.58.4

### Platform

NodeJS

### What happened?

Return type of

- `queue.getJob` is `any`
- `queue.getJobs` is `any[]`

I bisected bullmq versions and discovered that it was introduced in:

|||
|-|-|
| Commit | f2495e5ee9ecdb26492da510dc38730718cb28c5 |
| Version | [v5.26.1](https://github.com/taskforcesh/bullmq/releases/tag/v5.26.1) |

i.e. v5.26.0 was the last working version.

### How to reproduce.

Video demo:

https://github.com/user-attachments/assets/4a745177-8b15-40c9-933d-6e7693a33d08

1. Clone this repository: https://github.com/nktnet1/bullmq-abort-demo

1. Upgrade to bullmq v5.26.1 (or later - v5.58.4 is the latest at the time of writing)
   ```
   pnpm i bullmq@5.26.1
   ```
   
1. Find any instance of `getJob` or `getJobs` and hover your mouse over them.

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

manast:
Thanks for the issue, we will look into it shortly.

PR #3423, opened 2025-09-02
fix(queue.ts): preserve Job type inference when no explicit type for JobBase

ntvviktor:
Fixed by simplifying JobBase to always wrap data types in Job instances

### Why

When T = any (default), TypeScript couldn't reliably resolve this to Job<any, any, string>.
```typescript
type JobBase<T, ResultType, NameType extends string> = T extends Job<any, any, any>
  ? T
  : Job<T, ResultType, NameType>;
```
since `any extends Job<any, any, any>` would always true.

### How

Add a check if user T is any, fall back to the Job<any, any, string>

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -90,8 +90,8 @@ export interface QueueListener<JobBase extends Job = Job>
 
 /**
  * IsAny<T> A type helper to determine if a given type `T` is `any`.
- * This works by uing `any` type with the intersection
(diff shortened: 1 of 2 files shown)


ntvviktor:
This is an attempt to solve this issue:
https://github.com/taskforcesh/bullmq/issues/3421

roggervalf (approved):
Lgtm

PR #3426, opened 2025-09-04
fix(events): set prev param as active when calling retryJob script

roggervalf:
### Why

  1. Why is this change necessary? RetryJob script is called when a job is active, so prev value should reflect that and not failed state

### How

_Any extra info here._

### DIFF
--- a/tests/test_worker.ts
+++ b/tests/test_worker.ts
@@ -3290,6 +3290,9 @@ describe('workers', function () {
         worker.on('completed', () => {
           reject(new Error('Failed job was retried more than it should be!'));
         });
+        queueEvents.on('waiting', async ({ prev }) => {
+          expect(prev).to.eql('active');
+        });
         queueEvents.on('retries-exhausted', async ({ jobId, attemptsMade }) => {
           expect(jobId).to.eql(job.id);
           expect(3).to.eql(Number(attemptsMade));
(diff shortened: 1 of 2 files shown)


tests/test_queue.ts
```diff
@@ -52,14 +52,13 @@ describe('queues', function () {
     });
 
     it('should resolve Job<any, any, string> when no generics provided', async function () {
-      const defaultQueue = new Queue(queueName, { prefix, connection });
+      await queue.add('test-job', { foo: 'bar', num: 123 });
```
roggervalf, line 55: just reusing same queue instance that is being provided globally in this file

tests/test_worker.ts
```diff
@@ -3290,6 +3290,9 @@ describe('workers', function () {
         worker.on('completed', () => {
           reject(new Error('Failed job was retried more than it should be!'));
         });
+        queueEvents.on('waiting', async ({ prev }) => {
+          expect(prev).to.eql('active');
```
manast: If this expect throws an error the test will still pass, as the error is not propagated correctly to the test suite handler.

manast: You need to call "reject" on the upper promise if the expectation fails.

manast: Btw, same error on the "retries-exhausted" listener below...

manast (approved):
LGTM

Issue #3424, opened 2025-09-03
[Bug]: Queue clean failing when JobScheduler jobs are in the completed/failed state


tsebas:
### Version

v5.41.5

### Platform

NodeJS

### What happened?

We want to make use of the new jobScheduler functionality, preferred over a custom polling solution and over downgrading our version of bullmq. When running some tests on our solution using the job scheduler, we ran into some issues, namely when trying to remove jobs (`job.remove`) generally, and more so with running `queue.clean`.

We noticed that in order to run `job.remove` successfully, we first had to run `queue.removeJobScheduler`, which worked just fine. However, even here there is some, to me, weird behaviour. It looks like without the call to remove the jobScheduler, all jobs but 1 will be removed successfully (the weird part!); with the last one only successfully removed after removing the jobScheduler (this works!).

We then tried removing a lot of jobs at once with `queue.clean`. Depending on which status the jobs were in, we were successful or not (provided we of course ran the same `queue.removeJobScheduler` function before attempting to clean. Namely, it worked out of the box for the delayed and paused states, but not for the completed and failed. From our investigation, it is simply not possible given the current implementation/integration of the jobScheduler functionality with other general queue functionality.

I looked into the source code and found the following:
* in the commit introducing the functionality, see [here](https://github.com/taskforcesh/bullmq/commit/dd6b6b2263badd8f29db65d1fa6bcdf5a1e9f6e2#diff-18facfda0e83aa8e36173808fa60de92a8665fcbe7fa2e5a4cbf952aca28f1fe), great care is take when it comes to the delayed status
* in another commit also related to jobScheduler and status, see [here](https://github.com/taskforcesh/bullmq/commit/d4cad8402628f1773299c9cf33e6cc6a0e694037#diff-18facfda0e83aa8e36173808fa60de92a8665fcbe7fa2e5a4cbf952aca28f1fe), a comment is even added suggesting the specific variable related to repeat jobs is not needed in case of failed status

I didn't dig more into this, but as far as I can see, specially after modifying the `src/commands/cleanJobsInSet-3.lua` script as to provide the repeatKey as well for the completed/failed status code branch, no consideration has been made to what should happen in those states. The changes can be found below, where the only different from the current implementation is the last argument, KEYS[3], in the documentation presented as `KEYS[3]  repeat key`
```
 else
   rangeEnd = ARGV[2]
   result = cleanSet(KEYS[1], ARGV[1], rangeEnd, ARGV[2], limit,
[...]

PR #3429, opened 2025-09-05
test: add case when cleaning completed jobs and there are job schedulers

roggervalf:
### Why

  1. Why is this change necessary? Validate case of cleaning completed jobs and job scheduler presence .

### How

ref https://github.com/taskforcesh/bullmq/issues/3424

### DIFF
--- a/tests/test_clean.ts
+++ b/tests/test_clean.ts
@@ -77,7 +77,7 @@ describe('Cleaner', () => {
       );
     });
 
-    const addedJobs = await queue.addBulk([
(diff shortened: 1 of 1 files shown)


tests/test_clean.ts
```diff
@@ -77,7 +77,7 @@ describe('Cleaner', () => {
       );
     });
 
-    const addedJobs = await queue.addBulk([
```
roggervalf, line 80: addedJobs is an unused variable

manast (approved):

PR #3439, opened 2025-09-09
fix(scheduler): do not update scheduler data when worker updates

roggervalf:
### Why

  1. Why is this change necessary? No need to update scheduler data when update script is called, only next milliseconds needs to be updated and add delayed job accordingly to scheduler attributes previously added. There is a race condition between updating and adding a scheduler where next delayed job could have old data properties instead of new value

### How

_Any extra info here._


manast:
I have incorporated this test into this PR: https://github.com/taskforcesh/bullmq/pull/3446

PR #3442, opened 2025-09-10
fix(retry): reset attemptsMade and attemptsStarted (python)

roggervalf:
### Why

  1. Why is this change necessary? When retrying a job, attempsMade used for backoff strategy was not reseted or attemptsStarted

### How

_Any extra info here._


roggervalf:
closed as it's expected to retry only once

Issue #3095, opened 2025-02-20
[Bug]: upsertJobSchedule not updating data


FabioNappi:
### Version

v5.41.3

### Platform

NodeJS

### What happened?

Trying to update an existing jobSchedule is not working as expected: the scheduler itself seems to have updated data, but the actual jobs data is not always updated as well.

In my case I execute the following:
```js
await queue.upsertJobScheduler('my-schedule', {
    every: 20000,
}, {
    name: 'my-name',
    data: { key: 'value' },
})
```
The schedule is correctly created and jobs executed as expected. Then I try to update the schedule as follows:
```js
await queue.upsertJobScheduler('my-schedule', {
    every: 20000,
}, {
    name: 'my-name',
    data: { anotherKey: 'another-value' },
})
```
Now the schedule is updated correctly (I'm getting info with `queue.getJobSchedulers()`), but the jobs are executed with "old" data (i.e. `{ key: 'value' }`)

NOTE: this happens from the 2nd update and on. The creation and the first update work fine.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

cann0nball:
I can confirm this behavior. It seems it's trying to update the previous job repeat, that already finished, leaving the next scheduled one untouched.

It seems to work correctly, if I set prevMillis to Date.now(), but that has other side effects, like job not starting immediately, even with every.

This issue, together with [#3063](https://github.com/taskforcesh/bullmq/issues/3063) means there's currently no way to update the existing jobs, or we're doing something wrong?

FabioNappi:
In the end my workaround was to use deprecated APIs: add to the queue a job with the repeat options (insert) and then remove it with `removeRepeatableByKey` and re-add it again with new data (update).

It works as expected and is reliable.

manast:
We are trying to reproduce this issue in a test case and if we can we will make a fix.

marcolubian:
I'm experiencing the same issue and I confirm that the workaround of setting prevMillis to Date.now() doesn't work effectively, it works 2-3 times but then the job schedulation is not upserted/refreshed anymore. 

Even removing and recreating the scheduled job doesn't seem to work (rif. [3063](https://github.com/taskforcesh/bullmq/issues/3063))

enyineer:
Does this also affect upserting a JobScheduler with new RepeatOptions?

[...]

Issue #3430, opened 2025-09-05
[Bug]: Method upsertJobScheduler does not work stably


marcolubian:
### Version

v5.58.5

### Platform

NodeJS

### What happened?

I'm using method **upsertJobScheduler(jobSchedulerId, repeatOpts, jobTemplate)** to update the scheduling of an already existing scheduled job, changing its execution interval using attribute `every` and, for example, changing the scheduling every 8, 4, 9, 2, 3, ... seconds (eg 3 seconds means `every` set to 3000 ms).

In order to do that, I run the script like this

```javascript
await queue.upsertJobScheduler(jobId, jobOptions, scheduleData);
```

where jobId is the identifier (name) of the scheduled job and I put `every` attribute inside the `jobOptions`.

What I'm experiencing is that sometimes the update works as expected, so for example if I ask for a scheduling of 3 seconds, actually the job is run each 3 seconds; but sometimes the change is not taken and even if it seems to pass correctly through the above method, the scheduling remains the previous one (e.g. if it was each 3 seconds and I change it to 8 seconds, it remains each 3 seconds...).

If I go in debug, I see that the related method of file `job-scheduler`

```javascript
async upsertJobScheduler(jobSchedulerId, repeatOpts, jobName, jobData, opts, { override, producerId }) 
```

is called multiple times for each scheduled job:

- it's called with `override` to true when called externally (to actually update it)
- it's called with `override` to false when called internally on the `worker` file on this method:

```javascript
async nextJobFromJobData(jobData, jobId, token) {
```

the problem is that sometimes there is a de-synchronization between the 2 calls and it happens that **upsertJobScheduler** is called from the worker with an old `repeatOpts` and from that moment on it keeps working with the old/wrong options. 

I have not understood well how `nextJobFromJobData` works, but for what I've seen it seems to work with the old copy of the `repeatOpts`: maybe something that remain in memory / in cache ? If so, isn't it possible to let `nextJobFromJobData` to retrieve (re-read) its data from the backend (Redis(DB) instead of the in-memory/cached data?

I've seen many issues around `upsertJobScheduler` topic and some seem very similar to mine, in particular this one that seems not fixed yet (even if marked as closed): https://github.com/taskforcesh/bullmq/issues/3095

Could you please give priority on this topic? For us it's a big problem if we can't count on the reliability of this method. Consider that we've bought the PRO version and need to go in production soon...

### Relevant log output

```shell

```

### Code of Conduct

[...]

PR #3446, opened 2025-09-11
fix(job-scheduler): fix unstable upsert

manast:
### Why
  1. Why is this change necessary?
  Sometimes when upserting a job scheduler, the schedulers new options were not properly reflected on the new jobs.

  2. What problem does it solve or improve?
  Fixes issues #3095 and #3430 

### How

Moving the computation of nextMillis for "every" to the lua scripts that update the job scheduler, to eliminate any race conditions.

_Any extra info here._

### DIFF
--- a/tests/test_job_scheduler.ts
+++ b/tests/test_job_scheduler.ts
@@ -2842,7 +2842,7 @@ describe('Job Scheduler', function () {
 
     it('should allow repeatable jobs to be retried infinitely when stalled', async function () {
       this.clock.restore(); // Use real timers with mocked delay
-      this.timeout(5000); // Reduced timeout since delay will be mocked
+      this.timeout(8000); // Increased timeout for CI environments
 
       const stalledEvents: string[] = [];
       let jobProcessingAttempts = 0;
@@ -2919,15 +2919,28 @@ describe('Job Scheduler', function () {
         },
       );
 
+      // Ensure the worker is ready and starts processing
+      await worker.waitUntilReady();
+
       const jobId = job!.id!;
 
       // Wait for processing and stalling to occur with mocked delays
-      await delay(2000); // Much shorter since delay is mocked - retryIfFailed will be instant
+      // Use a polling approach to wait for the expected conditions
+      let attempts = 0;
+      const maxAttempts = 20; // 20 attempts * 200ms = 4 seconds max
+
+      while (attempts < maxAttempts && mockCallCount === 0) {
+        await delay(200);
+        attempts++;
+      }
+
+      // Give a bit more time for the full retry cycle to complete
+      await delay(1000);
 
       // Verify that the repeatable job attempted scheduler updates (retryIfFailed working)
       expect(mockCallCount).to.be.greaterThan(
-        3,
-        'Should have attempted scheduler update more than retryIfFailed limit',
+        0,
+        `Should have attempted scheduler update at least once, got ${mockCallCount}`,
       );
 
       // Verify that moveToDelayed was called as fallback after retryIfFailed exhausted


src/commands/addJobScheduler-11.lua
```diff
@@ -44,23 +45,41 @@ local eventsKey = KEYS[9]
...
+local prevMillis = rcall("ZSCORE", repeatKey, jobSchedulerId)
+if prevMillis then
+    prevMillis = tonumber(prevMillis)
+end
+local schedulerOpts = cmsgpack.unpack(ARGV[2])
+
+local every = schedulerOpts['every']
+
+local jobOffset = jobOpts['repeat'] and jobOpts['repeat']['offset'] or 0
+local offset = schedulerOpts['offset'] or jobOffset or 0
```
roggervalf, line 78: why we need the jobOffset, both should be the same one. We can use schedulerOpts as the source of truth

src/commands/addJobScheduler-11.lua
```diff
@@ -82,42 +101,52 @@ local function removeJobFromScheduler(prefixKey, delayedKey, prioritizedKey, wai
...
+
+local nextJobKey = schedulerKey .. ":" .. nextMillis
+local nextJobId = "repeat:" .. jobSchedulerId .. ":" .. nextMillis
+
+storeJobScheduler(jobSchedulerId, schedulerKey, repeatKey, nextMillis, schedulerOpts, templateData, templateOpts)
 
 rcall("INCR", KEYS[8])
 
-addJobFromScheduler(nextDelayedJobKey, nextDelayedJobId, ARGV[6], waitKey, pausedKey,
+addJobFromScheduler(nextJobKey, nextJobId, jobOpts, waitKey, pausedKey,
```
roggervalf: how we are making sure that we are not trying to add a duplicate as before?

src/commands/addJobScheduler-11.lua
```diff
@@ -82,42 +101,52 @@ local function removeJobFromScheduler(prefixKey, delayedKey, prioritizedKey, wai
             return true
         end
     end
-    return false
-end
 
-if rcall("EXISTS", nextDelayedJobKey) == 1 then
-    if not removeJobFromScheduler(prefixKey, delayedKey, prioritizedKey, waitKey, pausedKey,
-        nextDelayedJobId, metaKey, eventsKey) then
-        rcall("XADD", eventsKey, "MAXLEN", "~", maxEvents, "*", "event",
```
roggervalf, line 91: this event was added in cases where we are trying to add a duplicated job. i.e this test case https://github.com/taskforcesh/bullmq/blob/master/tests/test_job_scheduler.ts#L267

src/commands/updateJobScheduler-12.lua
```diff
@@ -38,17 +38,41 @@ local jobSchedulerId = ARGV[2]
...
 
 local schedulerKey = repeatKey .. ":" .. jobSchedulerId
+local schedulerAttributes = rcall("HMGET", schedulerKey, "name", "data", "every", "startDate", "offset")
+
+local every = schedulerAttributes[3]
+local now = tonumber(timestamp)
+if every then
+    local startDate = schedulerAttributes[4]
+    local jobOptsOffset = jobOpts['repeat'] and jobOpts['repeat']['offset'] or 0
+    local offset = schedulerAttributes[5] or jobOptsOffset or 0
```
roggervalf: why not only use offset from scheduler as source of truth

tests/test_job_scheduler.ts
```diff
@@ -338,22 +352,17 @@ describe('Job Scheduler', function () {
           await delayedJob!.promote();
           worker.run();
 
-          const duplicating = new Promise<void>(resolve => {
```
roggervalf, line 341: seems like we are stopping to emit duplicated event and description of this test case do not reflect that

tests/test_job_scheduler.ts
```diff
@@ -638,6 +647,7 @@ describe('Job Scheduler', function () {
       key: 'test',
       name: 'test',
       pattern: '*/2 * * * * *',
+      offset: 0,
```
roggervalf, line 659: when scheduler is pattern we can omit to save offset value in Lua

tests/test_job_scheduler_stress.ts
```diff
@@ -216,10 +216,185 @@ describe('Job Scheduler Stress', function () {
...
+    expect(delayedJobs).to.have.length(1);
+
+    // The final job should reflect the last upsert
+    const finalDelayedJob = delayedJobs[0];
+    expect(finalDelayedJob.data.iteration).to.equal(4);
+    expect(finalDelayedJob.name).to.equal('iteration-4');
+
+    // Verify no duplicate events were emitted
+    let duplicateEventCount = 0;
+    queueEvents.on('duplicated', () => {
```
roggervalf, line 414: there are no duplicated events because we are even not trying to emit it?

src/commands/includes/getJobSchedulerEveryNextMillis.lua
```diff
@@ -0,0 +1,27 @@
...
+            nextMillis = tonumber(startDate)
+            nextMillis = nextMillis > now and nextMillis or now
+        else
+            nextMillis = now
+        end
+    else
+        nextMillis = prevMillis + every
+        -- check if we may have missed some iterations
+        if nextMillis < now then
+            nextMillis = math.floor(now / every) * every + every + (offset or 0)
```
roggervalf, line 17: there is another pr that you can use as reference for this logic, maybe add test case from https://github.com/taskforcesh/bullmq/pull/3422 to see if it passes

roggervalf:
I have another pr reference for #3430 in https://github.com/taskforcesh/bullmq/pull/3439, there is a test case to replicate this issue there that can be used here as well

tests/test_job_scheduler_stress.ts
manast: There is still a case where they can be emitted.

src/commands/includes/getJobSchedulerEveryNextMillis.lua
manast: If you look in the PR I already added those tests and they pass :).

tests/test_job_scheduler.ts
manast: For consistency it is now added in all cases.

src/commands/addJobScheduler-11.lua
manast: yes, but we did not store the offset in the scheduler before, so to keep backwards compatibility we need to also check an existing offset in the repeat options of the current job.

manast: I can add a comment for this.

src/commands/addJobScheduler-11.lua
manast: AFAICS it is not possible to add a duplicate anymore. It can in theory only happen when calling update, but not when calling add.

src/commands/addJobScheduler-11.lua
manast: Can you write a test case where this event is needed, I could not find any case with the current code at least.

src/commands/updateJobScheduler-12.lua
manast: answered above.

src/commands/addJobScheduler-11.lua
roggervalf: maybe a todo comment to remove this in next breaking change

tests/test_job_scheduler.ts
roggervalf: still thinking is still valuable to omit it if not used, like if always will be zero for pattern, do not save it. That would save some space in Redis

src/commands/includes/getJobSchedulerEveryNextMillis.lua
roggervalf: looks like it does not pass the same test as in #3422, I added in https://github.com/taskforcesh/bullmq/pull/3447

src/commands/addJobScheduler-11.lua
roggervalf: sure, maybe this one https://github.com/taskforcesh/bullmq/pull/3448

manast: What I mean is a test case where, if not having the duplicate event, something unexpected will happen?

tests/test_job_scheduler.ts
manast: It creates burden in the tests and rest of the logic. As it is only stored in the scheduler, the amount of data is negligible, it would be different if it was stored in the jobs themselves.

src/commands/addJobScheduler-11.lua
roggervalf: event is for notifying that there is a duplicated id being generated, but along side of this event there is a validation of the next delayed job key. Without this validation, we can update an existing job record that is undesirable as it could be already completed or active for example

src/classes/worker.ts
```diff
@@ -813,25 +813,30 @@ will never work with more accuracy than 1ms. */
...
-        } else {
+        } else if (job.opts.repeat) {
           const repeat = await this.repeat;
           await repeat.updateRepeatableJob(job.name, job.data, job.opts, {
             override: false,
           });
         }
+      } catch (err) {
+        throw new Error(
+          `Failed to add repeatable job for next iteration: ${err}`,
```
roggervalf: I would only try catch the repeatable logic and emit an error instead of throwing it something like:
```ts
try {
        // Use new job scheduler if possible
        if (job.repeatJobKey && job.repeatJobKey.split(':').length < 5) {
          const jobScheduler = await this.jobScheduler;
          await jobScheduler.upsertJobScheduler(
          // Most of these arguments are not really needed
            // anymore as we read them from the job scheduler itself
            job.repeatJobKey,
            job.opts.repeat,
            job.name,
            job.data,
            job.opts,
            { override: false, producerId: job.id },
          );
        } else if (job.opts.repeat) {
          const repeat = await this.repeat;
          await repeat.updateRepeatableJob(job.name, job.data, job.opts, {
            override: false,
          });
        }
      } catch (err) {
        this.emit('error', new Error(
          `Failed to add repeatable job for next iteration: ${err}`));
      }
 
```
this way we can still return the previous job that was moved to active as in line 841

roggervalf: also seems like error is only thrown on addition, in update we are emitting the event so we can only try catch the first if block

manast: Hmm, I do not think we can do that, if we fail adding the next repeatable job, by design we must also fail the job itself, because otherwise we cannot guarantee that the iterations will continue.

manast: In other words, we exploit the guarantees of the job itself to give guarantees to the next iteration.

manast: In my new approach this is handled better. No exception is thrown, but we retry the job if possible, if not we let it stall and it will be retried later by the stall checker mechanism.

src/classes/worker.ts
```diff
@@ -813,25 +813,63 @@ will never work with more accuracy than 1ms. */
...
+          // Calculate backoff delay: base delay of 5 seconds, exponentially increasing with attempts
+          const baseDelay = 1000; // 1 second
+          const maxDelay = 300000; // 5 minutes max
+          const attempt = (job.stalledCounter || 0) + 1;
+          const exponentialDelay = Math.min(
+            baseDelay * Math.pow(2, attempt - 1),
+            maxDelay,
+          );
+
+          await job.moveToDelayed(Date.now() + exponentialDelay, job.token);
```
roggervalf, line 863: this movement will increase the attemptsMade. That will consume the retries that users can use on their processes. Why not just use the `retryIfFailed` method.  Also in order to prevent a infinite loop retrying, to pass an extra argument with a limit of retries

manast: ok, I saw we can use "skip attempt" to avoid incrementing the attempts made. Using moveToDelayed is much more robust, as another worker that may be in better condition could take the job. I thought about limiting the number of times a job would be retried, but the problem is that if it consumes all the retries, then there will not be more jobs for a given job scheduler.

manast: Actually it is already disabled:
```ts
  async moveToDelayed(timestamp: number, token?: string): Promise<void> {
    const now = Date.now();
    const delay = timestamp - now;
    const finalDelay = delay > 0 ? delay : 0;
    const movedToDelayed = await this.scripts.moveToDelayed(
      this.id,
      now,
      finalDelay,
      token,
      { skipAttempt: true },
    );
    this.delay = finalDelay;

    return movedToDelayed;
  }
```

roggervalf: what about using moveToWait instead of moveToDelayed

roggervalf: also thinking more on this backoff strategy, it can be better just to move it back to wait because we are setting a base of 1000ms and "every" operations can have values with lower values than 1000ms

roggervalf (approved):
Lgtm

vkouk:
Hello there, after upgrading to this version we noticed that job scheduler started to fail with below error:

```
{"code":-10,"context":"Bootstrap","error":{"code":-10},"level":"error","message":"Cannot create job scheduler iteration - job ID already exists. addJobScheduler","ms":"+223ms","stack":["Error: Cannot create job scheduler iteration - job ID already exists. addJobScheduler\n    at Scripts.finishedErrors (/opt/tipsterio/node_modules/.pnpm/bullmq@5.58.9/node_modules/bullmq/src/classes/scripts.ts:1761:17)\n    at Scripts.addJobScheduler
```

Hence, we had to downgrade. This is how we upsert jobs in our workers side

```
    async onModuleInit() {
        await this.queue.upsertJobScheduler(
            SidelinedPlayerJob.FetchSidelinedPlayers,
            { pattern: "0 0 1 */3 *" },
            {
                opts: {
                    removeOnComplete: 3,
                    removeOnFail: 10,
                    attempts: 3,
                    backoff: {
                        type: "fixed",
                        delay: 1500
                    }
                }
            }
        );
    }
}
```

roggervalf:
hi @vkouk I tried your example without success. It would be helpful to have a reproducible test case so we can take a look. It might be other things that trigger this error. But basically this error should happen if there is an existing job with same id that scheduler is going to produce

vaskouk:
> hi @vkouk I tried your example without success. It would be helpful to have a reproducible test case so we can take a look. It might be other things that trigger this error. But basically this error should happen if there is an existing job with same id that scheduler is going to produce

Hello @roggervalf , sorry for the late reply. Yes this is completely vague issue but pop up after latest update.

vkouk:
> hi @vkouk I tried your example without success. It would be helpful to have a reproducible test case so we can take a look. It might be other things that trigger this error. But basically this error should happen if there is an existing job with same id that scheduler is going to produce

Hello @roggervalf , sorry for the late reply. Yes this is completely vague issue but pop up after latest update.

manast:
@vkouk other than the error log, did it had any other impact? I think this should be harmless in practice, before we emitted a duplicate event but now we produce an error, maybe it is as easy as just going back to emitting the duplicate event.

vkouk:
> @vkouk other than the error log, did it had any other impact? I think this should be harmless in practice, before we emitted a duplicate event but now we produce an error, maybe it is as easy as just going back to emitting the duplicate event.

Not really, we just had to delete the job from redis and restart workers. After that it worked fine.

PR #3449, opened 2025-09-13
build(dev-deps): update dev deps to latest major versions

hainenber:
### Why

Maintain dev deps to latest versions, mostly for code hygiene and to potentially reduce attack surface.

- `sinon`
- `prettier`
- `pretty-quick`

### How

Update in `package.json` and perform corresponding refactor. I ran the tests locally and they all passed.

N/A


manast:
Thanks. However we should configure prettier to keep the same format as before.

hainenber:
More or less, I think the Prettier-related changes in this PR are due to Prettier considers them as defects and formats accordingly. 

For example, this change to nested ternaries statement in [`src/classes/scripts.ts`](https://github.com/taskforcesh/bullmq/pull/3449/files?diff=unified&w=0#diff-7858112f5b8e6ac9e61f03ef5e82b749bbc72c060061ba6b00451d98b97a5a0a) was caused by this [commit](https://github.com/prettier/prettier/pull/9559) in Prettier 3.1. This appears to be non-configurable.

Not sure how I can tinker Prettier to keep old format whilst being on higher version.

tests/test_deduplication.ts
```diff
@@ -586,9 +586,8 @@ describe('deduplication', function () {
             },
           );
 
-          const deduplicationJobId = await queue.getDeduplicationJobId(
-            deduplicationId,
-          );
+          const deduplicationJobId =
+            await queue.getDeduplicationJobId(deduplicationId);
```
roggervalf, line 627: I tried part of these changes and looks like these new lines are not needed https://github.com/taskforcesh/bullmq/pull/3466

hainenber: @roggervalf because `yarn prettier` command doesn't run over files in `tests/**/*.ts`. I wonder why the test files are excluded from prettifier but there must be a historical reason.

roggervalf: oh yeah they are being ignored, I believe this shouldn't be. I'll reformat that script apart

roggervalf (approved):
Lgtm

PR #3455, opened 2025-09-19
ci(deps): fix semantic-release format using conventionalcommits

roggervalf:
### Why

  1. Why is this change necessary? In order to follow our regular semantic-release format as it needs to use conventional-changelog-conventionalcommits preset

### How

_Any extra info here._

### DIFF
--- a/docs/gitbook/SUMMARY.md
+++ b/docs/gitbook/SUMMARY.md
@@ -146,4 +146,6 @@
 ## Python
 
 * [Introduction](python/introduction.md)
-* [Changelog](python/changelog.md)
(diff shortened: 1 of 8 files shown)


docs/gitbook/SUMMARY.md
```diff
@@ -146,4 +146,6 @@
 ## Python
 
 * [Introduction](python/introduction.md)
-* [Changelog](python/changelog.md)
+* [Changelogs](python/changelog.md)
```
roggervalf, line 149: taking the opportunity to split python changelings similar to node

manast (approved):

PR #3458, opened 2025-09-20
fixed typo in queue obliterate jsdoc comment

vehm:
### Why

The typo could have caused confusion, at least at first glance (I had to do a double take while reading it because it tripped me up). The description of `obliterate` is now a bit clearer on first read.

### How

It's a very, very simple change - just a typo fix.


roggervalf (approved):
Lgtm

Issue #2668, opened 2024-07-21
[Bug]: `queue.remove()` doesn't emit `'removed'` event


bitbay:
### Version

v5.8.2

### Platform

NodeJS

### What happened?

Manually removing a job from the queue through `queue.remove(job_id)` doesn't emit the `'removed'` event to it's `QueueEvents`.
This keeps it in cache and can't reuse the same jobid (results in duplicated event).
Using a retrieved job's `job.remove()` emits the event and the jobid can be reused.

### How to reproduce.

Using the same starting point of
```
const broker = new Queue("broker");
const job = await broker.add('job', {}, {
  jobId: 'job_id_a',
  delay: 10000
});

const listener = new QueueEvents('broker');
queueEvents.on('added', ({ jobId }) => {
  console.log('added event');
});
queueEvents.on('removed', ({ jobId }) => {
  console.log('removed event');
});
queueEvents.on('duplicated', ({ jobId }) => {
  console.log('duplicated event');
});
```

Removing through `queue` method:
```
// remove through queue
await broker.remove('job_id_a');  // returns 1 (removed);

// reusing the same job id results in 'duplicated' event
await broker.add('job', {}, {
  jobId: 'job_id_a',
  delay: 10000
});
// console logs 'duplicated'
```

While using job's `.remove()` method
```
const jobRef = await broker.getJob('job_id_a');
await jobRef.remove(); // returns 1 (removed)
// console logs 'removed'
await broker.add('job', {}, {
  jobId: 'job_id_a',
  delay: 10000
});
// console logs 'added'
```

The method in [`queue.ts`#L404](https://github.com/taskforcesh/bullmq/blob/c905d6206aa3d1b0a8f315da220cbc27f6a397c3/src/classes/queue.ts#L404) uses the same script as [`job.ts`#L563](https://github.com/taskforcesh/bullmq/blob/c905d6206aa3d1b0a8f315da220cbc27f6a397c3/src/classes/job.ts#L563) to remove the job, but `job.ts` does emit after removal.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

faizanu94:
Hey Guys, I opened a PR to address this: [#3461](https://github.com/taskforcesh/bullmq/pull/3461)

The change ensures that `queue.remove(jobId)` now emits the `'removed'` event, consistent with `job.remove()`.  
It also includes a test in `test_queue.ts` verifying the behavior.

[...]

PR #3461, opened 2025-09-21
feat(queue): emit 'removed' event when removing jobs via queue.remove()

faizanu94:
### Why
- `queue.remove(jobId)` did not emit the `'removed'` event, unlike `job.remove()`.
- This caused inconsistencies:
  - `QueueEvents` listeners never received a `'removed'` notification.
  - Removed job IDs stayed in cache, making reusing job IDs trigger a `'duplicated'` event instead of `'added'`.
- Fixes [#2668](https://github.com/taskforcesh/bullmq/issues/2668).

### How
- Updated `Queue.remove()` to:
  - Retrieve the job before removal.
  - Call `this.emit('removed', job)` when removal succeeds (`result === 1`).
- Added a new test in `test_queue.ts` to verify:
  - Calling `queue.remove(jobId)` emits the `'removed'` event.
  - The emitted job matches the removed job.
- This mirrors the behavior of `Job.remove()`, keeping API consistency.

### Additional Notes
- Scoped the change to only emit the event when removal actually succeeds.
- No breaking changes: existing functionality remains intact.
- New test ensures regression protection going forward.

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -804,7 +804,14 @@ export class Queue<
           }),
         });
 
-        return await this.scripts.remove(jobId, removeChildren);
+        const job = await this.getJob(jobId);
+        const result = await this.scripts.remove(jobId, removeChildren);
+
+        if (result === 1 && job) {
+          this.emit('removed', job as JobBase<DataType, ResultType, NameType>);
+        }
+
+        return result;
       },
     );
   }
--- a/tests/test_queue.ts
+++ b/tests/test_queue.ts
@@ -451,6 +451,24 @@ describe('queues', function () {
     });
   });
 
+  describe('.remove', () => {
+    it.only('should emit "removed" event when a job is removed via queue.remove()', async () => {
+      await queue.waitUntilReady();
+
+      const job = await queue.add('test-job', { foo: 'bar' });
+
+      const removedPromise = new Promise<Job>(resolve => {
+        queue.on('removed', job => resolve(job));
+      });
+
+      const result = await queue.remove(job.id!);
+      expect(result).to.equal(1);
+
+      const removedJob = await removedPromise;
+      expect(removedJob.id).to.equal(job.id);
+    });
+  });
+
   describe('.removeDeprecatedPriorityKey', () => {
     it('removes old priority key', async () => {
       const client = await queue.client;


tests/test_queue.ts
```diff
@@ -451,6 +451,24 @@ describe('queues', function () {
     });
   });
 
+  describe('.remove', () => {
+    it.only('should emit "removed" event when a job is removed via queue.remove()', async () => {
```
manast: .only

src/classes/queue.ts
```diff
@@ -804,7 +804,14 @@ export class Queue<
           }),
         });
 
-        return await this.scripts.remove(jobId, removeChildren);
+        const job = await this.getJob(jobId);
+        const result = await this.scripts.remove(jobId, removeChildren);
```
manast: Since you have an instance of Job here, why not just call job.remove?

tests/test_queue.ts
faizanu94: resolved

src/classes/queue.ts
faizanu94: Thanks, I deliberately updated the `job.remove()` signature from `Promise<void>` → `Promise<number>` to align with `Queue.remove()` signature:
- 0: job not found
- 1: removed successfully
- throw: locked/unremovable

src/classes/job.ts
```diff
@@ -660,7 +660,7 @@ export class Job<
    *
    * @param opts - Options to remove a job
    */
-  async remove({ removeChildren = true } = {}): Promise<void> {
+  async remove({ removeChildren = true } = {}): Promise<number> {
```
manast: Unfortunately that would be a breaking change...

manast: The solution would be to wrap job.remove in a try catch in queue.remove and then return 0 or 1 depending on the result of the call.

faizanu94: I’ve updated as suggested.

Note: Previously `scripts.remove()` could [throw](https://github.com/taskforcesh/bullmq/blob/master/src/classes/scripts.ts#L535) with `this.finishedErrors(...)` and ultimately `queue.remove()` would also throw.

With the new `job.remove()` + try/catch approach - that case is now flattened into returning 0.

src/classes/queue.ts
```diff
@@ -804,7 +804,18 @@ export class Queue<
           }),
         });
 
-        return await this.scripts.remove(jobId, removeChildren);
```
roggervalf, line 807: removeJob script returns 1 when removal is successful so a roundtrip to get the job and then call remove can be avoided

faizanu94: yes this was actually my [first commit](https://github.com/taskforcesh/bullmq/pull/3461/commits/5f3853de5e304a569ba2fbd33c52423d2e8546b6) using scripts.remove() and emitting when result === 1.

I later switched to job.remove() per @manast suggestion to leverage the job instance (job.remove).

I’m fine reverting to the lighter scripts.remove() path if that’s the preferred direction.

faizanu94: Hi folks, circling back to align on the approach.

My recommendation is to revert back to the [first commit](https://github.com/taskforcesh/bullmq/pull/3461/commits/5f3853de5e304a569ba2fbd33c52423d2e8546b6) using getJob (to get job instance), scripts.remove() (for removal) and then emitting event when result === 1.

If we drop getJob entirely, we avoid roundtrips but lose the ability to emit a proper Job instance in the event. That’s the main tradeoff here:
- Lighter path: just scripts.remove() and this.emit('removed') → no roundtrip, but no Job instance.
- Balanced path (my first commit): getJob + scripts.remove() → still efficient, but we emit a real Job instance.
- Current path: getJob + job.remove() → consistent but adds 1 more extra Redis op.

Open to feedbacks!

roggervalf: hi @faizanu94 we chose the lighter path but also passing the job id as a parameter for that event as part of this [pr](https://github.com/taskforcesh/bullmq/pull/3492)

roggervalf:
closing this pr as it's covered by https://github.com/taskforcesh/bullmq/pull/3492

PR #3472, opened 2025-09-26
feat(deduplication): support replace option in single mode

roggervalf:
### Why

  1. Why is this change necessary? Support replace option in single mode

### How

_Any extra info here._

### DIFF
--- a/src/commands/includes/deduplicateJob.lua
+++ b/src/commands/includes/deduplicateJob.lua
@@ -32,8 +32,8 @@ local function deduplicateJob(deduplicationOpts, jobId, delayedKey, deduplicatio
             if ttl and ttl > 0 then
                 local currentDebounceJobId = rcall('GET', deduplicationKey)
                 if currentDebounceJobId then
-                    local isRemoved = removeDelayedJob(delayedKey, deduplicationKey, eventsKey, maxEvents, currentDebounceJobId,
-                        jobId, deduplicationId, prefix)
+                    local isRemoved = removeDelayedJob(delayedKey, deduplicationKey, eventsKey, maxEvents,
+                        currentDebounceJobId, jobId, deduplicationId, prefix)
                     if isRemoved then
                         if deduplicationOpts['extend'] then
                             rcall('SET', deduplicationKey, jobId, 'PX', ttl)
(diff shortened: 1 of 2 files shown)


src/commands/includes/deduplicateJob.lua
```diff
@@ -58,6 +79,7 @@ local function deduplicateJob(deduplicationOpts, jobId, delayedKey, deduplicatio
 
             if deduplicationKeyExists then
                 local currentDebounceJobId = rcall('GET', deduplicationKey)
+                -- TODO remove debounced event in next breaking change
```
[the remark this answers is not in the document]
roggervalf: too verbose, a single line should be sufficient

src/commands/includes/deduplicateJob.lua
```diff
@@ -3,39 +3,71 @@
...
+
+local function removeDelayedJob(delayedKey, deduplicationKey, eventsKey, maxEvents, currentDeduplicatedJobId,
+    jobId, deduplicationId, prefix)
+    if rcall("ZREM", delayedKey, currentDeduplicatedJobId) > 0 then
+        removeJobKeys(prefix .. currentDeduplicatedJobId)
+        rcall("XADD", eventsKey, "*", "event", "removed", "jobId", currentDeduplicatedJobId,
+            "prev", "delayed")
+
+        -- TODO remove debounced event in next breaking change
+        rcall("XADD", eventsKey, "MAXLEN", "~", maxEvents, "*", "event", "debounced", "jobId",
```
manast, line 15: I think we should keep a list of all the changes that are breaking so that we can decide for the next major version which ones to incorporate or delete.

src/commands/includes/deduplicateJob.lua
```diff
@@ -3,39 +3,71 @@
...
-            local currentDebounceJobId = rcall('GET', deduplicationKey)
-            if currentDebounceJobId then
-                if rcall("ZREM", delayedKey, currentDebounceJobId) > 0 then
-                    removeJobKeys(prefix .. currentDebounceJobId)
-                    rcall("XADD", eventsKey, "*", "event", "removed", "jobId", currentDebounceJobId,
-                        "prev", "delayed")
-
-                    if deduplicationOpts['extend'] then
-                        rcall('SET', deduplicationKey, jobId, 'PX', ttl)
+        if deduplicationOpts['replace'] then
```
manast, line 31: This if statement is quite long, maybe there are opportunities for refactoring here...

roggervalf: I may need to think more about it, this is why I created the removeDelayedJob helper too. I'll add a TODO in this line

src/commands/includes/deduplicateJob.lua
roggervalf: usually I identify it by the TODO statements that I always add

PR #3476, opened 2025-09-27
feat(sandbox): support mjs files

roggervalf:
### Why

### How

_Any extra info here._


monolithed:
@roggervalf, thanks for the PR! Could you explain why it wasn’t possible to add the `*.mts` extension?
`*.ts` is already in the list, so it seems like backward compatibility shouldn’t be broken,  am I mistaken?

PR #3482, opened 2025-10-04
fix(worker): do not retry processor when connection errors happen

roggervalf:
### Why

  1. Why is this change necessary? retryIfFailed method was applied to the whole processJob method, this one includes the call of job processor plus our handler of completed and failed. While we want to retry this operation for our scripts when there is a connection error, we should not re-execute the processor operation, but the ones that we manage

### How

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -556,15 +556,11 @@ export class Worker<
       if (job) {
         const token = job.token;
         asyncFifoQueue.add(
-          this.retryIfFailed<void | Job<DataType, ResultType, NameType>>(
-            () =>
-              this.processJob(
-                <Job<DataType, ResultType, NameType>>job,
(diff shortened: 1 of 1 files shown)


src/classes/worker.ts
```diff
@@ -1327,13 +1344,10 @@ will never work with more accuracy than 1ms. */
       } catch (err) {
         if (isNotConnectionError(err as Error)) {
           this.emit('error', <Error>err);
-          if (delayInMs) {
-            await this.delay(delayInMs);
```
roggervalf, line 1331: need to remove this delay as tests were timing out when errors were different than connection errors, i.e when missing lock on stalled jobs. Instead we should throw error as soon as possible

manast (approved):
LGTM

PR #3503, opened 2025-10-17
Ci/remove npm token env

manast:
### Why

For OIDC to activate it seems like we cannot have the NPM_TOKEN as it would interfere in the publishing process.

### How

_Any extra info here._


roggervalf (approved):
Lgtm let's try it

PR #3504, opened 2025-10-18
fix(worker): only emit error when moveToActive failed

roggervalf:
### Why

  1. Why is this change necessary? When an error is thrown by _getNextJob, it stops the worker to continue its loop. Instead we should only emit the error and return undefined.

### How

_Any extra info here._


manast (commented):
looks good, but do we need to introduce an option ```onlyEmitError``` that will always be true?

roggervalf:
> looks good, but do we need to introduce an option `onlyEmitError` that will always be true?

For this method call yes, because if an error is throw it will break our loop, similar to https://github.com/taskforcesh/bullmq/blob/master/src%2Fclasses%2Fworker.ts#L971

manast (approved):

PR #3513, opened 2025-10-23
feat(queue): support getMeta method

roggervalf:
### Why

  1. Why is this change necessary? In order to get all queue configurations at once

### How

_Any extra info here._

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -215,7 +215,7 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       ...rest
     } = config;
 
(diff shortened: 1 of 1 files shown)


tests/test_telemetry_interface.ts
```diff
@@ -44,7 +44,7 @@ describe('Telemetry', () => {
   class MockContextManager<Context = any> implements ContextManager<Context> {
     private activeContext: Context = {} as Context;
 
-    with<A extends(...args: any[]) => any>(
+    with<A extends (...args: any[]) => any>(
```
manast: There is an unexpected space here

PR #3522, opened 2025-10-29
test(stalled): consider retry case

roggervalf:
### Why

  1. Why is this change necessary? Make sure to cover case when retrying a job that failed because all stalled retries were consumed

### How

_Any extra info here._

### DIFF
--- a/tests/test_stalled_jobs.ts
+++ b/tests/test_stalled_jobs.ts
@@ -173,11 +173,11 @@ describe('stalled jobs', function () {
 
       await worker.waitUntilReady();
 
-      const jobs = await Promise.all([
-        queue.add('test', { bar: 'baz' }, { removeOnFail: true }),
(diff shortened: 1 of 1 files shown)


tests/test_stalled_jobs.ts
```diff
@@ -143,81 +143,6 @@ describe('stalled jobs', function () {
     await worker.close();
   });
 
-  it('fail stalled jobs that stall more than allowable stalled limit', async function () {
```
roggervalf, line 146: this test case is basically the same as next test case, I'm removing it as considering a duplicate

manast, line 146: I think it is actually useful to have a test with this title, as in, we are explicitly testing just the case that we fail stalled jobs that stall more than the given limit, but we may be able to write a much simpler test just to test this specific feature.

roggervalf: I moved this test and refactor another so we can test different things

Issue #3500, opened 2025-10-15
[Bug]: upsertJobSchedule does not apply change on existing scheduled job


dangtran47:
### Version

5.61.0

### Platform

NodeJS

### What happened?

upsertJobScheduler does not update existing job

### How to reproduce.

```    
  await dailyArchiveManager.getQueue().upsertJobScheduler(
        'daily-archive-recurring',
        {
          pattern: '*/1 * * * *', 
          offset: 0,
          tz: 'America/New_York',
        },
        {
          opts: {
            attempts: 3,
            removeOnFail: true,
            repeatJobKey: 'daily-archive-recurring',
          },
          data: {
            timestamp: new Date().toISOString(),
            triggeredBy: 'schedule',
            // targetDate will be calculated in the processor
          },
          name: 'daily-archive-task',
        },
      );
```
Then change the pattern.

```    
  await dailyArchiveManager.getQueue().upsertJobScheduler(
        'daily-archive-recurring',
        {
          pattern: '* 8 * * *', // <<<<<<<<<<
          offset: 0,
          tz: 'America/New_York',
        },
        {
          opts: {
            attempts: 3,
            removeOnFail: true,
            repeatJobKey: 'daily-archive-recurring',
          },
          data: {
            timestamp: new Date().toISOString(),
            triggeredBy: 'schedule',
            // targetDate will be calculated in the processor
          },
          name: 'daily-archive-task',
        },
      );
```
I can see that the data is updated through the Bull dashboard, but the schedule still isn't updated until the existing task is run.

<img width="1285" height="541" alt="Image" src="https://github.com/user-attachments/assets/6a94d65d-6fdd-4cf0-a7cd-9e78f630ff76" />

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3524, opened 2025-10-29
fix: upsertJobScheduler does not apply change on existing scheduled job

manast:
### Why

To fix issue #3500

### How

Making sure we update a job scheduler even if there is a job id collision when calling upsert.

_Any extra info here._

### DIFF
--- a/tests/test_job_scheduler.ts
+++ b/tests/test_job_scheduler.ts
@@ -3049,20 +3049,20 @@ describe('Job Scheduler', function () {
       const stalledEvents: string[] = [];
       let jobProcessingAttempts = 0;
       let mockCallCount = 0;
-      let moveToDelayedCallCount = 0;
 
       const worker = new Worker(
         queueName,
-        async job => {
+        async () => {
           jobProcessingAttempts++;
           return { result: 'completed' };
         },
         {
+          autorun: false,
           connection,
           prefix,
           maxStalledCount: 2, // Regular jobs would fail after 2 stalls
           lockDuration: 1000, // 1 second lock
-          stalledInterval: 500, // Check for stalled jobs every 500ms
+          stalledInterval: 100, // Check for stalled jobs every 500ms
         },
       );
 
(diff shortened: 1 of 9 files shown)


src/commands/addJobScheduler-11.lua
```diff
@@ -150,7 +154,7 @@ if rcall("EXISTS", jobKey) == 1 then
         end
     else
         -- For 'pattern' case: return error code
```
roggervalf: this comment should be moved when the -10 error code is being thrown

tests/test_job_scheduler.ts
```diff
@@ -2549,6 +2550,55 @@ describe('Job Scheduler', function () {
...
+        key,
+        {
+          pattern: '10 * * * * *', // At 10 seconds past every minute
+        },
+        { name: 'test1', data: { foo: 'bar' } },
+      );
+
+      let delayedJobs = await queue.getDelayed();
+      expect(delayedJobs).to.have.length(1);
+      const firstJobId = delayedJobs[0].id;
```
roggervalf: seems like this variable is not used, it can be removed

tests/test_job_scheduler.ts
```diff
@@ -2549,6 +2550,55 @@ describe('Job Scheduler', function () {
...
+      const firstDelay = delayedJobs[0].delay;
+
+      // The first job should be scheduled for :10 seconds
+      expect(firstDelay).to.be.greaterThan(0);
```
roggervalf: it can be changed to:
`expect(firstDelay).to.be.equal(10000);`

tests/test_job_scheduler.ts
```diff
@@ -2549,6 +2550,55 @@ describe('Job Scheduler', function () {
...
+      delayedJobs = await queue.getDelayed();
+      expect(delayedJobs).to.have.length(1);
+
+      // Verify the delayed job was updated
+      expect(delayedJobs[0].name).to.be.equal('test2');
+      expect(delayedJobs[0].data).to.deep.equal({ foo: 'baz' });
+
+      // The new delay should be different (for :30 instead of :10)
+      const secondDelay = delayedJobs[0].delay;
+      expect(secondDelay).to.not.equal(firstDelay);
```
roggervalf, line 2594: after this validation
`expect(secondDelay).to.be.equal(30000);`
this expectation can be added to make sure that delay was properly updated with right value

manast: Fixed in commit [#8b57e1c](https://github.com/taskforcesh/bullmq/pull/3524/commits/8b57e1c82c79983d3448af7f3597902f49ffa336)

tests/test_job_scheduler.ts
manast: Fixed in commit [#8b57e1c](https://github.com/taskforcesh/bullmq/pull/3524/commits/8b57e1c82c79983d3448af7f3597902f49ffa336)

tests/test_job_scheduler.ts
manast: fixed in commit https://github.com/taskforcesh/bullmq/pull/3524/commits/b3a571455e9af9d8500a8f571b0326744136edd9

roggervalf (approved):
Lgtm

PR #3532, opened 2025-11-03
(Python) Add cooldown for connection error

julianbuettner:
### Why
When starting a BullMQ worker with a bad connection URL, the connect fails, prints an error and tries connecting immediately again. This generates a non-stopping flood of error logs until the connection is available.

I think it is better to have a small cooldown (one second) after a connection error.

It is highly likely you encountered this problem, if your already worked with the python bullmq.

### How
I only added a small sleep(1) for connection errors in the `getNextJob()` method.

It should be obvious, that the change has the desired effect and changes nothing else of the application logic.

Please let me know if you want a minimal reproduction demonstration.

### DIFF
--- a/docs/gitbook/bullmq-pro/changelog.md
+++ b/docs/gitbook/bullmq-pro/changelog.md
@@ -1,3 +1,23 @@
+## [7.40.1](https://github.com/taskforcesh/bullmq-pro/compare/v7.40.0...v7.40.1) (2025-11-15)
+
+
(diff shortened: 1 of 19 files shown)


roggervalf:
hi @julianbuettner pls try v2.16.2 I added a fix for this, similar to what we did in node version

julianbuettner:
Thank you!

-------- Original Message --------

PR #3537, opened 2025-11-05
fix(connection): consider error code when checking connection error

roggervalf:
### Why

1. Why is this change necessary? Ioredis changes their connection error format, this pr adds that format validation and keep the old format as well to prevent a breaking change. There is this vulnerability https://osv.dev/vulnerability/GHSA-29xp-372q-xqph that is addressed by this pr. I also removed some unnecessary resolutions.
commitizen package is only a helper to create commits following semantic-release format. Since it's not being updated since 1 year ago and not commonly used by us. It can be removed
  
### How

_Any extra info here._

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bullmq",
-  "version": "5.63.0",
+  "version": "5.63.1",
   "description": "Queue for messages and jobs based on Redis",
   "homepage": "https://bullmq.io/",
   "main": "./dist/cjs/index.js",
--- a/docs/gitbook/changelog.md
+++ b/docs/gitbook/changelog.md
@@ -1,3 +1,10 @@
+## [5.63.1](https://github.com/taskforcesh/bullmq/compare/v5.63.0...v5.63.1) (2025-11-13)
+
+
+### Bug Fixes
+
+* **job-scheduler:** changing every in upsert resets the iterations ([#3551](https://github.com/taskforcesh/bullmq/issues/3551)) ([b4c7c65](https://github.com/taskforcesh/bullmq/commit/b4c7c6579b430b53d135b7a21d20d01d14c1814e))
+
 # [5.63.0](https://github.com/taskforcesh/bullmq/compare/v5.62.2...v5.63.0) (2025-10-31)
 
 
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -119,11 +117,6 @@
     "typedoc": "^0.28.2",
     "typescript": "^5.9.2"
   },
-  "resolutions": {
-    "@semantic-release/commit-analyzer/conventional-changelog-angular": "8.1.0",
```
roggervalf, line 123: conventional-changelog-angular is being updated now when recreating lock file

package.json
```diff
@@ -63,8 +62,8 @@
     "uuid": "^11.1.0"
   },
   "devDependencies": {
-    "@commitlint/cli": "^17.0.3",
-    "@commitlint/config-conventional": "^17.0.3",
+    "@commitlint/cli": "^20.1.0",
+    "@commitlint/config-conventional": "^20.0.0",
```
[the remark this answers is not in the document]
roggervalf: updated to update tar sub dependency

tests/test_connection.ts
```diff
@@ -227,9 +227,11 @@ describe('connection', () => {
         },
       });
 
-      await expect(queue.waitUntilReady()).to.be.eventually.rejectedWith(
-        'connect ECONNREFUSED 127.0.0.1:9000',
-      );
+      try {
+        await queue.waitUntilReady();
+      } catch (error) {
+        expect(error.code).to.be.equal('ECONNREFUSED');
```
roggervalf, line 233: error format changed in ioredis, message is empty but code is filled with ECONNREFUSED

manast:
Before we can merge this PR we must be 100% sure this does not involve a breaking change, as users may be using a version of IORedis that is not compatible with this new one and passing that connection as an option to Queue, Worker, etc.

roggervalf:
> Before we can merge this PR we must be 100% sure this does not involve a breaking change, as users may be using a version of IORedis that is not compatible with this new one and passing that connection as an option to Queue, Worker, etc.

I believe we should be good as now all our tests pass. Hopefully this was detected now and we are having this issue now because we use a caret in our ioredis dependency, so for new users that install bullmq, they will use the most updated ioredis version
Screenshot 2025-11-11 at 8 17 44 p m

manast (approved):

PR #3543, opened 2025-11-07
refactor(worker): refactor lock renewal into separate class

manast:
### Why

We are planning to move the lock renewal to a separate thread, so the first step is to refactor the code outside of the worker class.

### How

Creating a new separate LockManager class.

### DIFF
--- a/docs/gitbook/changelog.md
+++ b/docs/gitbook/changelog.md
@@ -1,3 +1,10 @@
+## [5.63.1](https://github.com/taskforcesh/bullmq/compare/v5.63.0...v5.63.1) (2025-11-13)
+
+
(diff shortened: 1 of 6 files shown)


src/classes/worker.ts
```diff
@@ -461,6 +496,10 @@ export class Worker<
 
       await this.startStalledCheckTimer();
```
roggervalf, line 497: is this method not supported by new class already?

manast: No, the new class only handles locks, we may refactor to a separate class the stalled checker in a similar fashion.

roggervalf (approved):
Lgtm

Issue #3539, opened 2025-11-05
[Bug]: Method upsertJobScheduler does not trigger the new scheduling immediately


marcolubian:
### Version

v5.63.0

### Platform

NodeJS

### What happened?

I'm using version **v7.40.0** of BullMQ PRO, that is using version **v5.63.0** of bullmq.

When I update an existing scheduled job using `queue.upsertJobScheduler` and change the "every" interval (for example, from "every 10 seconds" to "every 2 seconds"), the new schedule is applied, but the job only starts running with the new interval after the previous schedule time has elapsed. In other words, the job does not reschedule immediately after the update, but only after the original schedule interval has finished.

How can I force the new scheduling to be taken into account immediately? I thought that this issue could be related to https://github.com/taskforcesh/bullmq/issues/3500 but it seems fixed from v5.62.2...

What am I doing wrong? What is the right way to achieve what I need?

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

manast:
Looking into this now.

manast:
After some time anylizing this issue, one question to be answered is what would be the actual correct behaviour. If you upsert a job scheduler with a new "every" setting, when should the next job be scheduled? should it be processed 2 seconds from the moment the new job scheduler is executed, or right at the moment it is executed?
Currently when you add a job scheduler for the first time (with every setting), the first iteration will be as soon as possible and then it will repeat according to the every setting. But if you are updating an existing job scheduler, what would be the desired behaviour?

marcolubian:
In my opinion, when we're updating a job already scheduled with `upsertJobScheduler`:
- by default you could leave the behavior as is: the job scheduling is actually changed, but before executing the job again it will wait the "old" scheduling once (after 2 seconds) and then will start the new scheduling (each 10 seconds)
- it could be useful to add a couple of flags/options to the `upsertJobScheduler` method
  - one to force the **immediate** execution of the job and then apply immediately the new scheduling: in this case a job originally scheduled each 2 seconds, when changing interval to 10 seconds the job would be executed once NOW and then each 10 seconds from NOW
  - one to just apply immediately the new scheduling, starting from NOW: in this case a job originally scheduled each 2 seconds, when changing interval to 10 seconds the job would NOT be executed NOW but correctly after each 10 seconds from NOW

[...]

PR #3551, opened 2025-11-10
fix(job-scheduler): changing every in upsert resets the iterations

manast:
### Why

When updating the every setting in a job scheduler we must reset the iterations, instead of the faulty current behaviour where one old iteration is kept and then the next one will be the one actually updated.

### How

we reset the iteration if the every setting is changed when calling upsertJobScheduler.

Fixes issue #3539

### DIFF
--- a/src/commands/addJobScheduler-11.lua
+++ b/src/commands/addJobScheduler-11.lua
@@ -81,7 +81,7 @@ if every then
     -- if we changed the 'every' value we need to reset millis to nil
     local millis = prevMillis
     if prevMillis then
-        local prevEvery = rcall("HGET", schedulerKey, "every")
+        local prevEvery = tonumber(rcall("HGET", schedulerKey, "every"))
         if prevEvery ~= every then
             millis = nil
             updatedEvery = true


package.json
```diff
@@ -122,7 +122,8 @@
   "resolutions": {
     "@semantic-release/commit-analyzer/conventional-changelog-angular": "8.1.0",
     "@semantic-release/release-notes-generator/conventional-changelog-angular": "8.1.0",
-    "commitizen/inquirer/external-editor/tmp": "0.2.4"
+    "commitizen/inquirer/external-editor/tmp": "0.2.4",
+    "tar": ">=7.5.2"
```
roggervalf, line 126: instead of adding a resolution, this can be fixed by updating our dependencies as in https://github.com/taskforcesh/bullmq/pull/3537

manast: Sometimes it is possible, but it in general it is quite time consuming to find the relevant dependencies to upgrade and if the dependency upgrade is not available you just waste a bunch of time. Maybe we can use resolutions and then once every month or something like that, spend some time to see if we can upgrade some deps and get rid of resolutions.

PR #3564, opened 2025-11-18
feat(worker): add job cancellation

manast:
### Why

Sometimes it is useful to be able to cancel a job. This PR provides a mechanism based on AbortController to cancel jobs, however it leaves the actual duty of cancelling the job to the job itself. This is the only sensible way to accomplish cancellation cleanly in NodeJS.

### How

We pass an abort signal to every job that cares about cancellation, and the job can listen to this signal for cancellations.

_Any extra info here._

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -424,6 +424,10 @@ export class QueueEvents extends QueueBase {
     if (!this.closing) {
       this.closing = (async () => {
         try {
+          // As the connection has been wrongly markes as "shared" by QueueBase,
+          // we need to forcibly close it here. We should fix QueueBase to avoid this in the future.
+          const client = await this.client;
+          client.disconnect();
           await this.connection.close(this.blocking);
         } finally {
           this.closed = true;


src/classes/lock-manager.ts
```diff
@@ -205,4 +218,39 @@ export class LockManager {
...
+    }
+    return false;
+  }
+
+  /**
+   * Cancels all tracked jobs by aborting their signals.
+   * @param reason - Optional reason for the cancellation
+   */
+  cancelAllJobs(reason?: string): void {
+    for (const [jobId, tracked] of this.trackedJobs.entries()) {
```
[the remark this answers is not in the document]
roggervalf: same as above

tests/test_job_cancellation.ts
```diff
@@ -0,0 +1,1037 @@
...
+                clearInterval(interval);
+                resolve({ done: true });
+              }
+            }, 10);
+
+            // Clean up interval if aborted
+            signal?.addEventListener('abort', () => clearInterval(interval));
+          });
+        },
+        { connection, prefix },
```
roggervalf, line 76: I would add autorun false, and worker.run() after setting the active listener

tests/test_job_cancellation.ts
```diff
@@ -0,0 +1,1037 @@
...
+        worker.on('active', () => resolve());
+      });
+
+      await delay(50);
+
+      // Cancel the job
+      worker.cancelJob(job.id!);
+
+      // Wait for the job to fail
+      await new Promise<void>(resolve => {
```
roggervalf, line 95: events can be declared before doing cancellation and await it here. Trying to prevent flaky tests where events are produced very fast before we can attach listeners

tests/test_job_cancellation.ts
```diff
@@ -0,0 +1,1037 @@
...
+        },
+        { connection, prefix },
+      );
+
+      await worker.waitUntilReady();
+
+      const job = await queue.add('test', { foo: 'bar' });
+
+      // Wait for job to start processing
+      await new Promise<void>(resolve => {
```
roggervalf: same as above

src/classes/queue-events.ts
```diff
@@ -417,9 +417,13 @@ export class QueueEvents extends QueueBase {
...
    * @returns
    */
-  close(): Promise<void> {
+  async close(): Promise<void> {
     if (!this.closing) {
-      this.closing = this.disconnect();
+      this.closing = (async () => {
+        await this.disconnect();
+        await this.connection.close();
+        this.closed = true;
```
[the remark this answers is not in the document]
manast: @copilot not sure just swallowing an error here is a good approach.

tests/test_job_cancellation_advanced.ts
```diff
@@ -0,0 +1,1166 @@
...
+      const job = await queue.add('test', { foo: 'bar' });
+
+      // Cancel immediately after adding (race condition test)
+      let wasCancelled = false;
+      const racePromise = new Promise<void>(resolve => {
+        worker.on('active', async () => {
+          worker.cancelJob(job.id!);
+          wasCancelled = true;
+          resolve();
+        });
```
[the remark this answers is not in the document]
manast: @copilot As this is a test, I do not think we need to care about it.

tests/test_job_cancellation.ts
manast: I see what is your point, however autorun would not improve this situation, you still cannot put anything after the promise that is waiting for active as it would never resolve in that case. But I will change it to be correct.

tests/test_job_cancellation.ts
manast: what I mean is that you cannot just await before actually calling run on the worker as the promise would never resolve.

roggervalf (approved):
Lgtm

mt-bt:
This is a great addition, thanks @manast!

I noticed this implementation still requires access to the `Worker` instance to deliver the signal. Are there any plans to make it possible to reliably deliver this cancellation signal directly from the `Queue` class?"

PR #3575, opened 2025-11-23
feat(job): support removeDeduplicationKey method

roggervalf:
### Why

  1. Why is this change necessary? To support removal of deduplication key from job instance, considering to remove it when job is the cause of deduplication

### How

_Any extra info here._

### DIFF
--- a/src/classes/index.ts
+++ b/src/classes/index.ts
@@ -9,6 +9,7 @@ export * from './job';
 export * from './job-scheduler';
 // export * from './main'; this file must not be exported
 // export * from './main-worker'; this file must not be exported
+export * from './lock-manager';
 export * from './queue-base';
 export * from './queue-events';
 export * from './queue-events-producer';


src/classes/index.ts
```diff
@@ -9,6 +9,7 @@ export * from './job';
 export * from './job-scheduler';
 // export * from './main'; this file must not be exported
 // export * from './main-worker'; this file must not be exported
+export * from './lock-manager';
```
roggervalf, line 12: missed to be reused in extension in pro version

docs/gitbook/guide/jobs/deduplication.md
```diff
@@ -119,14 +119,21 @@ const jobId = await myQueue.getDeduplicationJobId('customValue');
...
 ## Remove Deduplication Key
 
-If you need to stop deduplication before ttl finishes or before finishing a job, you can call the **removeDeduplicationKey** method.
+If you need to stop deduplication before ttl finishes or before finishing a job, you can call the **queue.removeDeduplicationKey** method.
 
 ```typescript
 await myQueue.removeDeduplicationKey('customValue');
 ```
 
+Or if you want to stop deduplication only if a specific job is the one that caused the deduplication
```
manast, line 128: I don not understand this text, what case is it exactly?

manast (approved):

docs/gitbook/guide/jobs/deduplication.md
roggervalf: when it's being used in processor as in the dedup pattern. While job is still active and need to remove that deduplication key making sure that this is the job that produced that dedup key

PR #3579, opened 2025-11-26
fix(stalled): prevent lock errors while job is not longer in active state

roggervalf:
### Why

  1. Why is this change necessary? when job is completed, job instance is removed from lock manager right after, but there is a possibility that extendLocks is called between job completion and job untracking. This is a race condition where "could not renew lock for job" error will be thrown. 

### How

  1. How did you implement this? Apart of lock key we have stalled set to validate if a job is still active. If it's not longer in stalled set we can say that we don't need to renew lock from this specific worker instance

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -1031,6 +1031,8 @@ will never work with more accuracy than 1ms. */
           );
           return failed;
         } finally {
+          this.lockManager.untrackJob(job.id);
+
           span?.setAttributes({
             [TelemetryAttributes.JobFinishedTimestamp]: Date.now(),
             [TelemetryAttributes.JobProcessedTimestamp]: processedOn,


src/commands/extendLocks-1.lua
```diff
@@ -37,7 +37,9 @@ for i = 1, jobCount, 1 do
             table.insert(failedJobs, jobId)
         end
     else
-        table.insert(failedJobs, jobId)
+        if rcall("SISMEMBER", stalledKey, jobId) == 1 then
+            table.insert(failedJobs, jobId)
+        end
```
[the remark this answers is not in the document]
roggervalf: added a test case

src/classes/worker.ts
```diff
@@ -359,6 +353,19 @@ export class Worker<
     );
   }
 
+  /**
+   * Creates and configures the lock manager for processing jobs.
+   * This method can be overridden in subclasses to customize lock manager behavior.
+   */
+  protected createLockManager() {
```
roggervalf, line 360: this pr also includes some refactors for lock manager extension in pro version

src/commands/extendLocks-1.lua
```diff
@@ -37,7 +37,9 @@ for i = 1, jobCount, 1 do
             table.insert(failedJobs, jobId)
         end
     else
-        table.insert(failedJobs, jobId)
+        if rcall("SISMEMBER", stalledKey, jobId) == 1 then
```
manast: I am not sure this solution works properly actually, as it is possible that stalledKey does not yet have the jobId of the job that is being completed if for instance the stalled checker has not been called between the job's addition to active and its completion. In other words we are just moving the edge case from one place to another, potentially creating a new one.

src/classes/worker.ts
```diff
@@ -1000,20 +1017,20 @@ will never work with more accuracy than 1ms. */
...
                 token,
                 fetchNextCallback,
                 span,
-              ),
+              );
+            },
             { delayInMs: this.opts.runRetryDelay, span, onlyEmitError: true },
           );
           return failed;
         } finally {
```
manast, line 1033: Maybe we have to keep this untrack here as well in case there are exceptions, no?

PR #3582, opened 2025-11-26
Fix duration units in retrying failing jobs documentation

yawhide:
### Why

Fix typo in retry documentation

### How

n/a

n/a


manast:
It is not really the formatting as far as I can see, but the units from minutes to milliseconds.

PR #3588, opened 2025-12-02
feat: add initial version of BullMQ [elixir]

manast:
### Why

Elixir is growing in popularity and being based on Erlang's VM it has some very interesting features for building large, fault tolerant and distributed systems, where a queue library like BullMQ could be really useful and fill a gap between easiness of use and performance compared with other options available in the platform.

### How

The Elixir library is a hex package that can be installed as any other elixir package using mix. It implements all APIs available in the NodeJS version and uses the same LUA scripts, so we could gain the same level of robustness if not better as we utilise the facilities provided by the BEAM VM, such as light processes and supervisors.r 

_Any extra info here._

### DIFF
--- a/elixir/lib/bullmq/flow_producer.ex
+++ b/elixir/lib/bullmq/flow_producer.ex
@@ -127,6 +127,7 @@ defmodule BullMQ.FlowProducer do
           {:ok, results} ->
             # Check for errors in results
             errors = Enum.filter(results, &match?({:error, _}, &1))
+
             if Enum.empty?(errors) do
               # Extract job IDs from results and populate the job tree
               job_ids = Enum.map(results, fn {:ok, id} -> to_string(id) end)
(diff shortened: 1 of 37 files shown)


elixir/CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
+# Changelog
```
roggervalf, line 1: semantic release similar to python can be used here, not a blocker but it could be added later

elixir/guides/deduplication.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Modes
+
+BullMQ supports three deduplication modes:
+
+### Simple Mode
+
+In Simple Mode, deduplication lasts until the job completes or fails. This is useful for long-running jobs that should not be duplicated while in progress.
+
+```elixir
```
roggervalf, line 13: there are tab sections in our documentation, may no need to duplicate same information but only add a new tab for elixir examples

manast: This guide will appear in hex.pm independently of our current main guide, so we need to repeat the information for now at least.

PR #3594, opened 2025-12-09
test: enhance local development [python]

roggervalf:
### Why

  1. Why is this change necessary? In order to execute python tests in local with pytest that is being used in our ci pipeline. Also this change contains a Union typing in order to be compatible with older versions of python

### How

Base on https://github.com/taskforcesh/bullmq/pull/2558


python/tests/worker_test.py
```diff
@@ -185,7 +185,7 @@ async def process(job: Job, token: str):
         self.assertEqual(failedJob.id, job.id)
         self.assertEqual(failedJob.attemptsMade, 1)
         self.assertEqual(failedJob.data, data)
-        self.assertEqual(failedJob.failedReason, f'"{failedReason}"')
+        assert failedReason in failedJob.failedReason
```
roggervalf: older version of python do not add nan suffix in this error message

PR #3623, opened 2025-12-19
fix(telemetry): send most updated attemptsMade value when finishing

roggervalf:
### Why

  1. Why is this change necessary? AttempstMade value is having the previous value before getting updated when a processor finished its execution

### How

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -949,7 +949,6 @@ will never work with more accuracy than 1ms. */
           [TelemetryAttributes.WorkerName]: this.opts.name,
           [TelemetryAttributes.JobId]: job.id,
           [TelemetryAttributes.JobName]: job.name,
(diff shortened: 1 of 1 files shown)


src/classes/worker.ts
```diff
@@ -1118,12 +1123,15 @@ will never work with more accuracy than 1ms. */
 
       this.emit('failed', job, err, 'active');
 
-      // Note: result can be undefined if moveToFailed fails (e.g., lock was lost)
-      if (result) {
-        span?.addEvent('job failed', {
-          [TelemetryAttributes.JobFailedReason]: err.message,
-        });
+      span?.addEvent('job failed', {
```
roggervalf, line 1126: this event should be emitted independently of the moveToFinished response as if it's undefined 0 that only means that there are no more jobs to process. Also this should match previous failed event emission as in line 1124

manast (approved):
LGTM

PR #3657, opened 2025-12-30
feat(queue): support obliterate method [elixir]

roggervalf:
### Why

1. Why is this change necessary? support obliterate method in elixir

### How

_Any extra info here._

### DIFF
--- a/docs/gitbook/guide/queues/removing-jobs.md
+++ b/docs/gitbook/guide/queues/removing-jobs.md
@@ -125,7 +125,7 @@ asyncio.run(main())
 alias BullMQ.Queue
 
 # Force obliteration even with active jobs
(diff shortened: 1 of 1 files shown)


src/commands/obliterate-2.lua
```diff
@@ -102,11 +102,12 @@ end
 if(maxCount > 0) then
   rcall("DEL",
     baseKey .. 'events',
-    baseKey .. 'delay', 
+    baseKey .. 'delay',
     baseKey .. 'stalled-check',
     baseKey .. 'stalled',
     baseKey .. 'id',
     baseKey .. 'pc',
+    baseKey .. 'marker',
```
roggervalf, line 110: all jobs in waiting or delayed states are removed, marker can be removed as well as there is no reason to wake up a worker in an emptied queue

Issue #3340, opened 2025-07-10
[Bug]: getWorkers does not work with Redis Cluster


DamienVicet:
### Version

v5.53.2

### Platform

NodeJS

### What happened?

The queue getter `getWorkers` does not work with Redis Cluster.

### How to reproduce.

I made a fork to test this : https://github.com/DamienVicet/bullmq/tree/master

I launched a Redis Cluster in Docker with

```
docker compose -f docker-compose-cluster.yml up
```

Then I launched getWorkers tests :

```
yarn test './tests/test_getters.ts' -g '.getWorkers'
```

The method returns always no results, after digging a bit, in https://github.com/taskforcesh/bullmq/blob/b4d3a5b36df3d383a3e43b7aef540d7401fc14df/src/classes/worker.ts#L319, `connectionName` is not passed to the options with Redis Cluster because the `Cluster` class of `ioredis` accepts it under `redisOptions`. The other problem is that `CLIENT LIST` is used to get the list of the workers, but with a cluster, this command runs only on one node, and I am not sure this is the right node that is used.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

danielpires:
is there any news on this issue? I'm having the same problem

micdabic:
+1

manast:
This may not be trivial to implement based on the current functionality, we may need to use a different approach for this to work.

danielpires:
Created a [PR](https://github.com/taskforcesh/bullmq/pull/3660) in an attempt to fix this issue, lmk what you think

PR #3660, opened 2026-01-03
Fix worker connection name on cluster #3340

danielpires:
### Why

This is an attempt to fix the issue https://github.com/taskforcesh/bullmq/issues/3340
When the bullmq is connected to a redis cluster the connection option change and in order to assure that the connection to redis uses the proper name it needs to have the redis connections options fixed. And it also aims to fix the client list functions when the its connected to a redis cluster

### How

1. When creating new worker, check if the redis client is a cluster before duplicating the connection so it's possible to pass different connection options
2. When listing the connections to fetch all worker connections, verify is redis client is a cluster, if so, get all connections from all nodes, the node with the most connections with the same name should be the node where all the connections where redirected so that should be the only node cluster to take into consideration for those workers connections

- Tests aren't prepared to run with a redis cluster, had do manually changed it to run the tests but didn't commit those changes
- Tests some times become flaky

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -19,7 +19,7 @@ jobs:
           # check out all branches
           fetch-depth: 0
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Setup Node.js
-        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
         with:
           node-version: '22.14.0'
           cache: 'yarn'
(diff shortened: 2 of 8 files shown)


src/classes/queue-getters.ts
```diff
@@ -5,7 +5,8 @@ import { QueueBase } from './queue-base';
 import { Job } from './job';
 import { clientCommandMessageReg, QUEUE_EVENT_SUFFIX } from '../utils';
 import { JobState, JobType } from '../types';
-import { JobJsonRaw, Metrics, QueueMeta } from '../interfaces';
+import { JobJsonRaw, Metrics, QueueMeta, RedisClient } from '../interfaces';
+import { Cluster } from 'ioredis';
```
danielpires: I don't think this should import Cluster directly, what do you think?

roggervalf: looks like this is just used as type. You can import only the type as:
```ts
import type { Cluster } from 'ioredis';
```

danielpires: you're right, I've fixed accordingly

manast:
@danielpires thank you for this PR. Looking into it now.

php/package.json
```diff
@@ -27,6 +27,12 @@
       "@commitlint/config-conventional"
     ]
   },
+  "resolutions": {
+    "diff": "8.0.3",
```
roggervalf, line 31: all these resolutions are not needed, only undici. I'm addressing these vulnerabilities in https://github.com/taskforcesh/bullmq/pull/3728

manast: lets merge this PR and then clean it up in your PR.

mouadhbb:
Hello, I believe this issue is related to this PR
https://github.com/taskforcesh/bullmq/issues/3758
Thank you in advance

PR #3669, opened 2026-01-07
fix(deps): upgrade ioredis to v5.9.0 to prevent incompatible typing

roggervalf:
### Why

### How

_Any extra info here._


roggervalf:
addressed in https://github.com/taskforcesh/bullmq/pull/3676

PR #3673, opened 2026-01-09
feat(queue): support drain method [python]

roggervalf:
### Why

  1. Why is this change necessary? To support drain method in python port. It also adds example codes in our docs

### How

_Any extra info here._


manast (approved):
LGTM

Issue #3667, opened 2026-01-06
[Bug]: Incompatibility with ioredis version 5.9


Diluka:
### Version

v5.66.4

### Platform

NodeJS

### What happened?

BullMQ is currently incompatible with ioredis version 5.9. Attempting to use BullMQ with this version of ioredis results in the following TypeScript error:

```
Type 'Redis' is not assignable to type 'ConnectionOptions'.
  Type 'Redis' is not assignable to type 'ClusterOptions | Redis | Cluster'.
    Type 'import("e:/Sources/bullmq-demo/node_modules/ioredis/built/Redis").default' is not assignable to type 'import("e:/Sources/bullmq-demo/node_modules/bullmq/node_modules/ioredis/built/Redis").default'.
      Types of property 'options' are incompatible.
        Type 'import("e:/Sources/bullmq-demo/node_modules/ioredis/built/redis/RedisOptions").RedisOptions' is not assignable to type 'import("e:/Sources/bullmq-demo/node_modules/bullmq/node_modules/ioredis/built/redis/RedisOptions").RedisOptions'.
          Type 'RedisOptions' is not assignable to type 'CommonRedisOptions'.
            Types of property 'Connector' are incompatible.
              Type 'import("e:/Sources/bullmq-demo/node_modules/ioredis/built/connectors/ConnectorConstructor").default | undefined' is not assignable to type 'import("e:/Sources/bullmq-demo/node_modules/bullmq/node_modules/ioredis/built/connectors/ConnectorConstructor").default | undefined'.
                Type 'import("e:/Sources/bullmq-demo/node_modules/ioredis/built/connectors/ConnectorConstructor").default' is not assignable to type 'import("e:/Sources/bullmq-demo/node_modules/bullmq/node_modules/ioredis/built/connectors/ConnectorConstructor").default'.
                  Type 'import("e:/Sources/bullmq-demo/node_modules/ioredis/built/connectors/AbstractConnector").default' is not assignable to type 'import("e:/Sources/bullmq-demo/node_modules/bullmq/node_modules/ioredis/built/connectors/AbstractConnector").default'.
                    Property 'connecting' is protected but type 'AbstractConnector' is not a class derived from 'AbstractConnector'.ts(2322)
queue-options.d.ts(16, 5): The required type comes from property "connection" in type "QueueOptions"
(property) QueueBaseOptions.connection: ConnectionOptions
Options for connecting to a Redis instance.
```

[...]

PR #3682, opened 2026-01-12
chore(deps): ioredis move to peerDependencies and set range version

mdoi2:
### Why

Currently, IORedis is included in the dependencies section of package.json, and its version is pinned.
There are two cons to this.

- As with the recent issue (#3667), you need to update package.json every time IORedis is updated.
- When users install IORedis and pass the IORedis instance as a connection to BullMQ, version mismatches may occur.

### How

- Move the current `“ioredis”: “5.9.1”` setting from `dependencies` to `devDependencies`.
- Add `“ioredis”: “^5.9.1”` to `peerDependencies`

🚨 __Caution__
If you have only installed `bullmq` and not `ioredis`, you will need to install `ioredis`.
I think it should be announced as a breaking change.

__P.S.__
I believe this change will benefit both maintainers and users.
I kindly ask for your consideration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,12 +56,14 @@
   },
   "dependencies": {
     "cron-parser": "4.9.0",
-    "ioredis": "5.9.1",
     "msgpackr": "1.11.5",
     "node-abort-controller": "3.1.1",
     "semver": "7.7.3",
     "tslib": "2.8.1",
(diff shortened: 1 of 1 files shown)


mdoi2:
I forgot to include a signature in the commit, so I'm closing it for now.

PR #3683, opened 2026-01-12
ioredis move to peerDependencies

mdoi2:
### Why

Currently, IORedis is included in the dependencies section of package.json, and its version is pinned.
There are tree cons to this.

- As with the recent issue (#3667), you need to update package.json every time IORedis is updated.
- You cannot freely choose the version of IORedis.
- When users install IORedis and pass the IORedis instance as a connection to BullMQ, version mismatches may occur.

### How

- Move the current `“ioredis”: “5.9.1”` setting from `dependencies` to `devDependencies`.
- Add `“ioredis”: “^5.9.1”` to `peerDependencies`

🚨 __Caution__
If you have only installed `bullmq` and not `ioredis`, you will need to install `ioredis`.
I think it should be announced as a breaking change.

__P.S.__
I believe this change will benefit both maintainers and users.
I kindly ask for your consideration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,12 +56,14 @@
   },
   "dependencies": {
     "cron-parser": "4.9.0",
-    "ioredis": "5.9.1",
     "msgpackr": "1.11.5",
     "node-abort-controller": "3.1.1",
     "semver": "7.7.3",
     "tslib": "2.8.1",
     "uuid": "11.1.0"
+  },
+  "peerDependencies": {
+    "ioredis": "^5.9.1"
   },
   "devDependencies": {
     "@commitlint/cli": "20.3.1",
@@ -97,6 +99,7 @@
     "eslint-plugin-tsdoc": "0.5.0",
     "fast-glob": "3.3.3",
     "husky": "8.0.3",
+    "ioredis": "5.9.1",
     "lint-staged": "16.2.7",
     "madge": "8.0.0",
     "minimatch": "9.0.5",


roggervalf:
hi @mdoi2, this will be a breaking change. We will consider this in v6. This is still pending to be done as we would like to include other extra changes on v6

mfulton26:
> hi @mdoi2, this will be a breaking change. We will consider this in v6. This is still pending to be done as we would like to include other extra changes on v6

I donn't know the timeline for v6 but unless it is going to be very soon what do you think of changing the existing dependency from a pinned version to a range? e.g. `"^5.9.1"` would not be breaking change at all (leaving it as a dependency for now and not moving it until v6). This would enable developers to deduplicate their package manager lock files (e.g. `pnpm dedupe`) and have only one version of "ioredis" in their dependency tree as long as it matches the specified range (e.g. pulling in "5.9.1", "5.9.2", etc.).

mdoi2:
This issue was resolved in BullMQ v6.

mfulton26:
> This issue was resolved in BullMQ v6.

Great news. Thank you. Are there no breaking changes in v6? I don't see any listed in https://github.com/taskforcesh/bullmq/releases/tag/v6.0.0. Why the major version bump if no breaking changes?

mdoi2:
The details of the breaking change are described in [this commit](https://github.com/taskforcesh/bullmq/commit/e1f86effc5bddda5b70d890866d9a490555e41ff).

It appears that ioredis has been moved to peerDependencies (optional) as a result of the abstraction of low-level drivers (ioredis, node-redis, Bun).
https://github.com/taskforcesh/bullmq/blob/v6.0.0/package.json

Issue #3703, opened 2026-01-16
[Bug]: Running with sandboxed processes on forked processes with the --watch flag prevents jobs being started


lewnelson:
### Version

v5.66.5

### Platform

NodeJS

### What happened?

The issue is inside [Child.initChild](https://github.com/taskforcesh/bullmq/blob/master/src/classes/child.ts#L177-L188) there is a message handler to handle messages sent from the child process to the parent process. This expects the first message to belong to BullMQ, however when `--watch` is enabled the child process sends the first message and continues to send messages. Once a message has been received regardless of origin the event handler is detached and the process stalls with no log output.

# Potential solutions

## Quick fix

Exiting early if `msg.cmd` is not one of the expected enum values. This implicitly determines a message to belong to BullMQ if the message has a `cmd` property that corresponds to the expected enum value.

```TypeScript
      const onMessageHandler = (msg: any) => {
        if (!Object.values(ParentCommand).includes(msg.cmd)) {
          return;
        }

        if (msg.cmd === ParentCommand.InitCompleted) {
          resolve();
        } else if (msg.cmd === ParentCommand.InitFailed) {
          const err = new Error();
          err.stack = msg.err.stack;
          err.message = msg.err.message;
          reject(err);
        }
        this.off('message', onMessageHandler);
        this.off('close', onCloseHandler);
      };
```

## Explicit messages

Exiting early if the `msg` is not a BullMQ message. This would involve a similar approach to the quick fix, but adding a constant identifier to all messages to determine if a message is a BullMQ message or not.

### How to reproduce.

I have a demo repository which demonstrates what happens.

1. Clone https://github.com/lewnelson/bullmq-sandbox
2. Checkout the `bug/watch-mode-child-processes` branch
3. Run the following:
   1. If using [mise](https://mise.jdx.dev/) run mise install, otherwise setup `node`
   2. `npm install`
   3. `npm start` in one terminal
   4. `npm run worker-child-process-watch` in another terminal
   5. Open the browser and go to http://localhost:3020
   6. Trigger a job
   7. Note that in the worker terminal the job is picked up, but never started
   8. Repeat with using either `worker-child-process`, `worker-threads`, or `worker-threads-watch` to start the worker process and note that the job is executed as expected

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3704, opened 2026-01-16
fix(sandboxed-processes): ensure non-bullmq messages are ignored

lewnelson:
### Why

Fixes issue #3703 and introduces a test to cover regressions.

### How

Ignores messages without a valid `cmd` property when running sandboxed processors as forked child processes.

_Any extra info here._

### DIFF
--- a/tests/fixtures/fixture_main_non_bullmq_messages.js
+++ b/tests/fixtures/fixture_main_non_bullmq_messages.js
@@ -1,0 +1,65 @@
+/**
+ * Custom main file that sends non-BullMQ messages during initialization
+ * to test that the child process properly ignores them.
+ */
+'use strict';
+
+const { ChildProcessor } = require('../../dist/cjs/classes/child-processor');
+const { ParentCommand, ChildCommand } = require('../../dist/cjs/enums');
+const { errorToJSON, toString } = require('../../dist/cjs/utils');
+
+const send = async msg => {
+  if (process.send) {
+    return new Promise((resolve, reject) => {
+      process.send(msg, err => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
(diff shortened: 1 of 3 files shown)


src/classes/child.ts
```diff
@@ -175,6 +175,10 @@ export class Child extends EventEmitter {
   private async initChild() {
     const onComplete = new Promise<void>((resolve, reject) => {
       const onMessageHandler = (msg: any) => {
+        if (!Object.values(ParentCommand).includes(msg.cmd)) {
+          return;
```
roggervalf, line 179: you can do:
```js
resolve();
return;
```
in order to finalize the Promise execution

lewnelson: We don't want to resolve at this point as we need to keep the listener open. This is essentially saying to "drop non-bullmq messages" i.e. ignore this and wait for the next message. Once we receive a message that has a valid cmd (implicit check for BullMQ message) we can handle it accordingly.

roggervalf: now I can see what you meant. Looks good to me. Pls resolve the merge conflict when you get some time

lewnelson: @roggervalf resolved the conflict now

src/classes/child.ts
```diff
@@ -175,6 +175,10 @@ export class Child extends EventEmitter {
   private async initChild() {
     const onComplete = new Promise<void>((resolve, reject) => {
       const onMessageHandler = (msg: any) => {
+        if (!Object.values(ParentCommand).includes(msg.cmd)) {
```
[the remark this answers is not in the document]
roggervalf: if all values are numeric, this check should be fine

roggervalf (approved):

Issue #3531, opened 2025-11-02
Add safeguards for invalid job methods in the process function


manast:
**Is your feature request related to a problem? Please describe.**
Currently, it is possible to call invalid job methods inside the process function of a worker, for example:

```ts
const worker = new Worker(
      queueName,
      async job => {
        await job.moveToDelayed();
      }, { ... }
```

This call will be performed successfully but later when the job actually completes we will end with strange errors like  
```Error: Missing lock for job [jobId]... moveToDelayed```

**Describe the solution you'd like**
We should implement safeguards so that the invalid job methods directly throw an exception that will early in development notify the user that the method is invalid and thus it cannot be used inside the process function.

**Additional context**
For most methods we have special errors that can be thrown from the process function to achieve similar results in a correct way, for example: https://docs.bullmq.io/patterns/process-step-jobs#delaying

PR #3706, opened 2026-01-17
feat(job): add safeguards for invalid job methods in process function

smileeunsong:
Refs: #3531

### Why

When calling `job.moveToDelayed()` or `job.moveToWaitingChildren()` without passing the token inside a processor, the job state changes but the processor returns normally. Then the worker tries to call `moveToCompleted`, resulting in confusing errors:

- `Missing lock for job [jobId]. moveToFinished`
- `Job [jobId] is not in the active state. moveToFinished`

These errors are difficult to debug because they don't indicate the actual cause.

### How

Added a safeguard check at the beginning of `moveToDelayed` and `moveToWaitingChildren` methods:

```typescript
if (this.token && !token) {
  throw new Error(`Missing token for job ${this.id}. moveToDelayed`);
}
```

When `job.token` is set (meaning the job is being processed), calling these methods without passing the token will throw a clear error immediately.

**Questions for maintainers:**

1. Is this the correct approach for #3531? I used `this.token` to detect if the job is currently being processed.

2. Are there other methods that need similar safeguards? (e.g., `moveToWait`, `moveToCompleted`, `moveToFailed`)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,13 +100,13 @@ jobs:
           else
             echo "No changes detected after semantic-release"
           fi
-      - run: yarn copy:lua:python
       # Python Release
       - name: Setup Python
         uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
         with:
           python-version: '3.10.19'
           cache: 'pip'
+      - run: yarn copy:lua:python
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


roggervalf:
hi @smileeunsong, thanks for your pr. In my opinion this can be considered a breaking change as we have a logic that when token is not present in those methods token is considered as '0', this is an special value in our scripts to avoid validating the token and continue the job movement. We will consider to remove that behavior as it's confusing and force users to pass always the token to those methods. I'll keep this pr in mind for v6. For now pls refer to our documentation https://docs.bullmq.io/patterns/process-step-jobs#delaying as the right way to call these methods within a processor is explained there.¿

smileeunsong:
Thanks for the clarification! I understand that this would be a breaking change since token='0' is used to skip validation intentionally.

I'll keep this PR open for v6 reference, or feel free to close it if you prefer. Let me know if there's anything else I can help with for v6.

Issue #3692, opened 2026-01-13
[Feature]: improve telemetry spans with Job ID and Job Name


alexbaileyuk:
**Is your feature request related to a problem? Please describe.**
When my application is instrumented with telemetry traces and I search for spans from failed jobs then I would like to be able to quickly identify the job ID which has failed.

In the example below, you can see that the job 'complete' span has no job ID. Neither does the failed span. Therefore, I cannot identify which job it was for:

<img width="1562" height="800" alt="Image" src="https://github.com/user-attachments/assets/2ea393e3-cf28-4b46-9ec2-134ed952ebd4" />

However, if you compare this to the 'process' span, we can clearly see that the job ID is included:

<img width="1567" height="778" alt="Image" src="https://github.com/user-attachments/assets/501be571-f901-4065-93b5-2c47ebaa747a" />

By including the job ID as an attribute in 'failed' spans (also possibly 'complete' spans) I will be able to quickly identify the failed job ID in other traces, logs, and in my failed job queue.

**Describe the solution you'd like**
Add the `bullmq.job.id` attribute to the job spans for 'complete' and 'failed'. I think these are in https://github.com/taskforcesh/bullmq/blob/8a174a2d0c59810a2af67a3a8b94c832e7189042/src/classes/job.ts#L721C22-L721C28 and https://github.com/taskforcesh/bullmq/blob/8a174a2d0c59810a2af67a3a8b94c832e7189042/src/classes/job.ts#L811 respectively.

I think something like this would do:

```typescript
        span?.setAttributes({
          [TelemetryAttributes.JobName]: name,
          [TelemetryAttributes.JobId]: job.id,
        });
```

**Describe alternatives you've considered**
Implementing my own tracing but that's less ideal.

**Additional context**
I can probably submit a PR later this week if it helps.

roggervalf:
available since v5.67.0

PR #3707, opened 2026-01-17
Feat/improved telemetry spans

alexbaileyuk:
### Why

1. Telemtry spans are not consistent which makes it more difficult to correlate and analyze observability traces.
2. It allows users to correlate open telemetry traces. Specifically aggregating metrics based on job name or job ID.
3. https://github.com/taskforcesh/bullmq/issues/3692

### How

Referring to similar examples in `src/classes/flow-producer.ts`, I added a reuseable method to the `Job` class which adds relevant attributes to spans. I updated two methods in `Job` which now add the expected attributes.

There was a small block (3 lines) of unused code which I removed while I was working in the class.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -722,13 +722,8 @@ export class Job<
       SpanKind.INTERNAL,
       'complete',
       this.queue.name,
-      async (span, dstPropagationMedatadata) => {
+      async span => {
         this.setSpanJobAttributes(span);
-
-        let tm;
-        if (!this.opts?.telemetry?.omitContext && dstPropagationMedatadata) {
-          tm = dstPropagationMedatadata;
-        }
 
         await this.queue.waitUntilReady();
 


src/classes/job.ts
```diff
@@ -722,11 +722,8 @@ export class Job<
       SpanKind.INTERNAL,
       'complete',
       this.queue.name,
-      async (span, dstPropagationMedatadata) => {
-        let tm;
-        if (!this.opts?.telemetry?.omitContext && dstPropagationMedatadata) {
-          tm = dstPropagationMedatadata;
```
roggervalf, line 728: I don't think that this logic was intended to be removing in this pr

alexbaileyuk: Hey Roggervalf

Thanks for the review. I found this code unused and removed it in commit `6b2171b`. I'd be happy to move that to a different PR or leave it there for safety if we would rather do that?

roggervalf: hey @alexbaileyuk, I see what you mean. In that case it should be fine to remove it

PR #3746, opened 2026-01-30
feat: add datatype for version differentiation

roggervalf:
### Why

  1. Why is this change necessary? We are supporting different databases where they handle their own versioning. We must differentiate redis db from other dbs version when checking some bullmq capabilities. This pr introduces a new attribute called databaseType, it extracts database type from info command

### How

_Any extra info here._


docker-compose.yml
```diff
@@ -5,3 +5,12 @@ services:
     container_name: cache
     ports:
       - 6379:6379
+  dragonfly:
```
roggervalf, line 8: added this image for local testing

.github/workflows/test.yml
```diff
@@ -125,7 +125,7 @@ jobs:
 
     services:
       dragonflydb:
-        image: docker.dragonflydb.io/dragonflydb/dragonfly:v1.24.0
+        image: docker.dragonflydb.io/dragonflydb/dragonfly:v1.28.0
```
manast: Shouldn't we take latest instead?

roggervalf: there is still failures in our tests for new versions

roggervalf: tests are working now

manast: I mean the tag latest so that it always uses the latest.

roggervalf: I think that it was added by renovate, it pinned all versions and creates new prs to upgrade versions. Also I prefer this way ensuring reproducibility, predictability, and stability across different tests and over time. That way if a new version is no longer compatible, unrelated prs won't be blocked but that will be detected on a different renovate pr.

manast: It has cons and pros. One con is that a new version of DragonflyDB may include a regression that breaks BullMQ and it will pass unnoticed until we upgrade our CI scripts.

manast: so in other words, pinning a version makes our life easier but increases the risk for breaking deployments for users...

manast: I prefer getting red tests here than on a production deployment :)

roggervalf: ok I'll update it to latest

bamse16:
This may be broken for TypeScript projects.

I'm getting

```
node_modules/bullmq/dist/esm/utils/index.d.ts:6:30 - error TS2307: Cannot find module 'src/types' or its corresponding type declarations.
6 import { DatabaseType } from 'src/types';
```

when trying to build the project, using `bullmq@v5.68.0`.

When inspecting `.../types/DatabaseType` is empty.

Issue #3758, opened 2026-02-03
[Bug]: ClusterAllFailedError: NOAUTH Authentication required after upgrading from 5.67.0 to 5.67.1+


mouadhbb:
### Version

v5.67.1+

### Platform

NodeJS

### What happened?

After upgrading from BullMQ 5.67.0 to 5.67.1 or 5.67.2, I'm getting a ClusterAllFailedError with NOAUTH 

```
Authentication required when connecting to a Redis Cluster with authentication.
ClusterAllFailedError: Failed to refresh slots cache.
    at tryNode (/node_modules/ioredis/built/cluster/index.js:335:31)
    ...lastNodeError: ReplyError: NOAUTH Authentication required.
      at parseError (/node_modules/redis-parser/lib/parser.js:179:12)
      at parseType (/node_modules/redis-parser/lib/parser.js:302:14) {
    command: { name: 'cluster', args: [Array] }
  }
```

### How to reproduce.

Set up a Redis Cluster with authentication (username/password)
Create a Queue with the following configuration:

```
const Redis = require('ioredis');

const redisInstance = new Redis.Cluster(clusterHosts, {
  redisOptions: {
    username: 'myuser',
    password: 'mypassword',
  }
});

const myQueue = new Queue('queueName', {
  connection: redisInstance
});
```

### Relevant log output

```shell
Connecting to Redis Cluster with hosts: [
  { host: '192.168.1.52', port: 7001 },
  { host: '192.168.1.52', port: 7002 },
  { host: '192.168.1.52', port: 7003 }
]
Using Redis options: {
  redisOptions: { username: 'redisuser', password: 'redisuserpassword' }
}
🚀 Connected to Redis Cluster
Connected to port 8010
🟢 [redis] connect
Database connected successfully !!
✅ [redis] ready (slots cache OK)
ClusterAllFailedError: Failed to refresh slots cache.
    at tryNode (/Users/benbrahammouadh/workspace/tic-nova/geo-nova/geo-nova-node/node_modules/ioredis/built/cluster/index.js:335:31)
    at /Users/benbrahammouadh/workspace/tic-nova/geo-nova/geo-nova-node/node_modules/ioredis/built/cluster/index.js:352:21
    at /Users/benbrahammouadh/workspace/tic-nova/geo-nova/geo-nova-node/node_modules/ioredis/built/cluster/index.js:721:24
[...]

PR #3759, opened 2026-02-03
fix(connection): merge redisOptions when duplicating cluster

roggervalf:
### Why

  1. Why is this change necessary? When authentication parameters are passed, they were ignored in duplicate method

### How

  1. How did you implement this? Spread original options when duplicating a cluster

fix https://github.com/taskforcesh/bullmq/issues/3758

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -1,4 +1,4 @@
-import { Cluster } from 'ioredis';
+import type { Cluster } from 'ioredis';
 import { JobProgress } from '../types';
 import {
   IoredisListener,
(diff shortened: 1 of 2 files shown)


tests/cluster.test.ts
```diff
@@ -700,3 +700,529 @@ describe('Cluster support', () => {
...
+      console.warn(
+        'Redis cluster not available, skipping cluster integration tests:',
+        (err as Error).message,
+      );
+    }
+  }, 10000);
+
+  beforeEach(async ctx => {
+    if (!clusterAvailable) {
+      ctx.skip();
```
roggervalf, line 762: for now these tests are running in local if cluster is built

mouadhbb:
Thanks for your PR

manast (approved):
LGTM

Issue #3774, opened 2026-02-10
[Bug]: Internal error events from Repeat / JobScheduler are swallowed


next-n:
### Version

v5.67.3

### Platform

NodeJS

### What happened?

Error forwarding from Repeat/JobScheduler is effectively a no-op in both Queue and Worker.
In queue.ts
this._repeat.on('error', e => this.emit.bind(this, e)); (line: 250)
this._jobScheduler.on('error', e => this.emit.bind(this, e)); (line:263)

In worker.ts
this._repeat.on('error', e => this.emit.bind(this, e));(line: 510)
this._jobScheduler.on('error', e => this.emit.bind(this, e));(line:524)
 
These use e => this.emit.bind(this, e), which returns a function but never emits. Result: parent queue.on('error') / worker.on('error') will miss those child errors.

---

### What happened?

Error forwarding from **Repeat** / **JobScheduler** is broken (no-op) in both **Queue** and **Worker**.

In **queue.ts**:

* `this._repeat.on('error', e => this.emit.bind(this, e));` (line ~250)
* `this._jobScheduler.on('error', e => this.emit.bind(this, e));` (line ~263)

In **worker.ts**:

* `this._repeat.on('error', e => this.emit.bind(this, e));` (line ~510)
* `this._jobScheduler.on('error', e => this.emit.bind(this, e));` (line ~524)

These handlers return a bound function but **never call `emit`**. EventEmitter ignores listener return values, so the parent **never re-emits** the error. As a result, `queue.on('error')` / `worker.on('error')` will miss errors emitted by Repeat/JobScheduler.

### Expected behavior

Errors emitted by Repeat/JobScheduler should propagate to the parent emitter, so users can catch them via:

```js
queue.on('error', handler)
worker.on('error', handler)
```

### Suggested fix

Replace with one of the following

```js
child.on('error', err => this.emit('error', err));
```

or

```js
child.on('error', this.emit.bind(this, 'error'));
```

Happy to open PR after confirmation from maintainers.

### How to reproduce.

create new test file tests/error-forwarding.test.ts and run "yarn vitest run tests/error-forwarding.test.ts" at the repo root. Following code is for tests/error-forwarding.test.ts.
```js
import { describe, it, expect, vi } from 'vitest';
import { Queue } from '../src';

describe('Repeat/JobScheduler error forwarding', () => {
  it('Queue should forward Repeat "error" to queue.on("error")', async () => {
    const queue = new Queue('test-forward-repeat');

[...]

PR #3775, opened 2026-02-10
fix(queue): forward error from repeat or job-scheduler instances

next-n:
Fixes https://github.com/taskforcesh/bullmq/issues/3774
### Why

Internal `'error'` events emitted by `Repeat` and `JobScheduler` were not propagated to their parent `Queue` / `Worker`.
The existing listeners returned a bound function (`this.emit.bind(...)`) instead of calling `emit`, so the error was silently swallowed and `queue.on('error')` / `worker.on('error')` never fired.

This breaks observability for internal failures (e.g. Redis or scheduler errors).

---

### How

* Fixed error forwarding by explicitly emitting the `'error'` event from `Repeat` and `JobScheduler` to their parent emitters.
* Replaced `e => this.emit.bind(this, e)` with `e => this.emit('error', e)` in both `Queue` and `Worker`.
* Added a regression test that reproduces the issue and verifies correct propagation.

---

### Additional Notes

* Added tests fail on `main` and pass after this fix.
* No behavior change for job-level failures; this only affects internal error propagation.
* No public API changes.

### DIFF
--- a/tests/error-forwarding.test.ts
+++ b/tests/error-forwarding.test.ts
@@ -1,12 +1,29 @@
-import { describe, it, expect, vi } from 'vitest';
+import { default as IORedis } from 'ioredis';
+import { v4 } from 'uuid';
+import { describe, it, expect, vi, beforeAll, afterAll } from 'vitest';
 import { Queue } from '../src';
+import { removeAllQueueData } from '../src/utils';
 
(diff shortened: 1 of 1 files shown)


next-n:
I updated with new test for this, can someone approve the CI test?

roggervalf (approved):
Lgtm

PR #3784, opened 2026-02-11
feat: set BullMQ driver info for Redis connections

vchomakov:
### Why

This PR adds support for Redis driver identification via the `CLIENT SETINFO` command.

When BullMQ creates a Redis client internally, it automatically identifies itself to Redis. This helps with debugging and monitoring Redis connections.

As recommended in the [Redis CLIENT SETINFO documentation](https://redis.io/docs/latest/commands/client-setinfo/), libraries should identify themselves to help with debugging and monitoring.

### How

**TypeScript (ioredis):**
- Sets `clientInfoTag` to `bullmq_v{version}` (e.g., `bullmq_v5.67.3`)
- Applied to `RedisConnection`, `Worker`, and `QueueEvents`

**Python (redis-py):**
- Uses `DriverInfo` API (sets both lib name and version)
- Applied to `RedisConnection`

Key points:
- Users can override with a custom value
- Only applies when BullMQ creates the client (not for custom clients passed via `connection` option)

Example Redis `CLIENT LIST` output:
```
lib-name=ioredis(bullmq_v5.67.3)
```

### Additional Notes

- Added tests for both TypeScript and Python implementations


manast:
Thank you for the PR. As a side note, workers must identify themselves so that the getWorkers api works as expected. We have just not implemented a simular identification pattern for the other connections.

manast:
ah, sorry, I was mixing SETINFO with SETNAME. We use SETNAME for the workers. One thing that needs to be considered is that this is a quite new API and thus will not work on Redis 6.2+ which is our baseline version for BullMQ.

manast:
Some references:
https://redis.io/docs/latest/commands/client-setinfo/

manast:
@copilot can you update the documentation to cover this new feature?

vchomakov:
> ah, sorry, I was mixing SETINFO with SETNAME. We use SETNAME for the workers. One thing that needs to be considered is that this is a quite new API and thus will not work on Redis 6.2+ which is our baseline version for BullMQ.

Thanks for the review, @manast !
`redis-py` handles this gracefully - when it sends `CLIENT SETINFO` during connection, it silently ignores failures on older Redis servers (like our 6.2 baseline).

This is by design per [Redis docs](https://redis.io/docs/latest/commands/client-setinfo/): *"Client libraries are expected to pipeline this command after authentication and ignore failures since they could be connected to an older version."*

vchomakov:
@manast I'd appreciate an approval for running the workflows. Thanks!

PR #3794, opened 2026-02-14
chore(deps): address markdown-it vulnerability

roggervalf:
### Why

  1. Why is this change necessary? Reported by OSV

### How

_Any extra info here._


tests/error-forwarding.test.ts
```diff
@@ -1,67 +0,0 @@
-import { default as IORedis } from 'ioredis';
```
roggervalf, line 1: no need for a single file for these tests, they can be to their corresponding scope tests

manast (approved):
it seems like there are unrelated code in this PR besides the vulnerability fix, please double check this is intended.

roggervalf:
> it seems like there are unrelated code in this PR besides the vulnerability fix, please double check this is intended.

Yes those were intended as my comment above

Issue #3796, opened 2026-02-14
[Bug]: Fix for TODO:  Remove legacy fallback entirely in the future.


next-n:
### Version

5.69.1

### Platform

NodeJS

### What happened?

There is an existing TODO in src/classes/job-scheduler.ts:325 to eventually remove legacy parsing (keyToData).

Until that cleanup happens, the current legacy fallback should still work correctly. 

The minimal fix is to treat empty Redis hash data as missing data, so fallback is reached instead of returning malformed scheduler objects.

### How to reproduce.

There is a TODO right there at job-scheduler.ts (line 325).

I want to propose a fix for current behavior, not implementation of that TODO.

Implementing the TODO now would be a larger behavior change.  My propsed fix is to

1. Keep TODO as-is (not remove legacy fallback yet).

2. Fix condition in transformSchedulerData:
- from: if (jobData)
- to: if (jobData && Object.keys(jobData).length > 0)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

next-n:
PR incoming

PR #3797, opened 2026-02-14
fix(job-scheduler): handle empty scheduler hash fallback

next-n:
Fixes https://github.com/taskforcesh/bullmq/issues/3796

### Why

I make this change to check when having empty hash data (Object.keys(jobData).length > 0), so it does not build malformed object. 
It then goes to legacy fallback (keyToData) when applicable, and returns parsed fields like name, id, pattern, etc.

We now treat empty scheduler hash data as missing key data(same handling), instead of treating it as valid and returning malformed scheduler objects.

this change specifically prevents the name: undefined bad output for empty hash data.

### How

Implemented safe fix
Changed condition:
- from: if (jobData)
- to: if (jobData && Object.keys(jobData).length > 0)
- 
This keeps existing TODO/legacy behavior intact and only fixes current bad behavior.

I also added regression test
job_scheduler.test.ts
new test: when listing legacy schedulers without hash data

- Explanation

It seeds a legacy scheduler key in repeat ZSET without hash and asserts getJobSchedulers() returns correctly parsed legacy fields.
- Verified behavior
Before fix: new test failed (name: undefined).
After fix: new test passed.

### DIFF
--- a/tests/job_scheduler.test.ts
+++ b/tests/job_scheduler.test.ts
@@ -2255,6 +2255,29 @@ describe('Job Scheduler', () => {
     });
   });
 
+  describe('when listing legacy schedulers without hash data', () => {
+    it('should parse scheduler fields from legacy key format', async () => {
+      const client = await queue.client;
+      const next = Date.now() + ONE_MINUTE;
+      const legacyKey = 'legacy-name:legacy-id:::*/5 * * * * *';
+
+      await client.zadd(queue.toKey('repeat'), next, legacyKey);
+
+      const schedulers = await queue.getJobSchedulers();
+
+      expect(schedulers).toHaveLength(1);
+      expect(schedulers[0]).toEqual({
+        key: legacyKey,
+        name: 'legacy-name',
+        id: 'legacy-id',
+        endDate: null,
+        tz: null,
+        pattern: '*/5 * * * * *',
+        next,
+      });
+    });
+  });
+
   describe('when repeatable job fails', () => {
     it('should continue repeating', async () => {
       const date = new Date('2017-02-07T15:24:00.000Z');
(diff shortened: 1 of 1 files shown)


tests/job_scheduler.test.ts
```diff
@@ -3909,3 +3909,46 @@ describe('Job Scheduler', () => {
...
+        endDate: null,
+        tz: null,
+        pattern: '*/5 * * * * *',
+        next,
+      });
+    } finally {
+      await queue.close();
+      await removeAllQueueData(new IORedis(redisHost), queueName);
+    }
+  });
```
roggervalf:
```suggestion
  it('falls back to legacy key parsing when scheduler hash data is missing', async () => {
      const next = Date.now() + ONE_MINUTE;
      const legacyKey = 'legacy-name:legacy-id:::*/5 * * * * *';

      const client = await queue.client;
      await client.zadd(queue.toKey('repeat'), next, legacyKey);

      const schedulers = await queue.getJobSchedulers();

      expect(schedulers).toHaveLength(1);
      expect(schedulers[0]).toEqual({
        key: legacyKey,
        name: 'legacy-name',
        id: 'legacy-id',
        endDate: null,
        tz: null,
        pattern: '*/5 * * * * *',
        next,
      });
    });
```

roggervalf: no need to create a new queue or connection, those are defined at the top of this file

next-n: I Refactored that test to reuse existing Job Scheduler suite fixtures

roggervalf: looks good to me. Thank you

roggervalf (approved):
Lgtm

PR #3811, opened 2026-02-18
feat(otel): support gauge metric when counting jobs by state

roggervalf:
### Why

  1. Why is this change necessary? Queue size can be emitted as a metric. In this matter we can use gauge metric

### How

_Any extra info here._

### DIFF
--- a/docs/gitbook/guide/telemetry/metrics.md
+++ b/docs/gitbook/guide/telemetry/metrics.md
@@ -83,14 +83,26 @@ Gauges are recorded when calling `recordJobCountsMetric()`. The `bullmq.queue.jo
 
 ## Metric Attributes
 
-All metrics include the following attributes for filtering and grouping:
+Different metrics include different attributes for filtering and grouping:
 
-| Attribute                 | Description                                                                        |
-| ------------------------- | ---------------------------------------------------------------------------------- |
-| `bullmq.queue.name`       | Name of the queue                                                                  |
-| `bullmq.job.name`         | Name of the job                                                                    |
-| `bullmq.job.status`       | Status of the job (completed, failed, delayed, retried, waiting, waiting-children) |
-| `bullmq.queue.jobs.state` | Job state for gauge metrics (waiting, active, completed, failed, etc.)             |
+### Common Attributes (all metrics)
+
+| Attribute           | Description       |
+| ------------------- | ----------------- |
+| `bullmq.queue.name` | Name of the queue |
+
+### Job Metric Attributes (counters and histograms only)
+
+| Attribute           | Description                                                                        |
+| ------------------- | ---------------------------------------------------------------------------------- |
+| `bullmq.job.name`   | Name of the job                                                                    |
+| `bullmq.job.status` | Status of the job (completed, failed, delayed, retried, waiting, waiting-children) |
+
+### Gauge Attributes (`bullmq.queue.jobs` only)
+
+| Attribute                 | Description                                                            |
+| ------------------------- | ---------------------------------------------------------------------- |
+| `bullmq.queue.jobs.state` | Job state for gauge metrics (waiting, active, completed, failed, etc.) |
 
 ## Configuration Options
 


docs/gitbook/guide/telemetry/metrics.md
```diff
@@ -156,10 +179,14 @@ const worker = new Worker(
     telemetry,
   },
 );
+
+// When calling count() or getJobCounts(), gauge metrics are recorded
+const waitingCount = await queue.count();
```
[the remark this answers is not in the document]
roggervalf: In this case should be fine as count method returns all waiting statuses count

src/classes/queue-getters.ts
```diff
@@ -182,6 +183,20 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       counts[currentTypes[index]] = res || 0;
     });
 
+    const meter = this.opts.telemetry?.meter;
```
manast, line 199: I find this a code smell. This is a getter, but we are creating a gauge inside it in a side-effect kind of way. It should be refactored into a separate function that does this job, properly named and documented.

roggervalf: this was refactored to be included in a function apart

docs/gitbook/guide/telemetry/metrics.md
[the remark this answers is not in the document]
manast: I think copilot has a point here, "queued" is better than "waiting" as "queued" would imply all jobs that are still in some status other than completed or failed.

docs/gitbook/guide/telemetry/metrics.md
```diff
@@ -156,10 +179,14 @@ const worker = new Worker(
     telemetry,
   },
 );
+
+// When calling count() or getJobCounts(), gauge metrics are recorded
```
manast: "gauge metrics are recorded" is an incorrect comment now that we need to explicitly record the gauge metrics, no?

src/classes/queue-getters.ts
```diff
@@ -182,9 +183,33 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       counts[currentTypes[index]] = res || 0;
     });
 
+    this.recordJobCountsMetric(counts);
```
manast: now the code to record the count metrics is in one function, but it should not be called from a getter. A getter should return information and not have side effects that write data to some external component such as Otel.

src/classes/queue-getters.ts
```diff
@@ -182,9 +183,33 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
...
     return counts;
   }
 
+  /**
+   * Records job counts as gauge metrics for telemetry purposes.
+   * Each job state count is recorded with the queue name and state as attributes.
+   *
+   * @param counts - An object mapping job states to their counts
+   */
+  private recordJobCountsMetric(counts: { [index: string]: number }): void {
```
manast: I think this should be a public function and the user should call it whenever it wants to record the metrics. So basically it should be the reverse of what we have, instead of getJobCounts calling recordJobCountesMetric, getJobCounts should be called from within recordJobCountsMetric

manast (changes requested):

docs/gitbook/guide/telemetry/metrics.md
roggervalf: do you mean when doing the inverse process as below?

src/classes/queue-getters.ts
roggervalf: this was addressed in last changes

src/classes/queue-getters.ts
```diff
@@ -185,6 +186,32 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
...
+  async recordJobCountsMetric(...types: JobType[]): Promise<{
+    [index: string]: number;
+  }> {
+    const counts = await this.getJobCounts(...types);
+    const meter = this.opts.telemetry?.meter;
+    if (meter && typeof (meter as any).createGauge === 'function') {
+      const gauge = meter.createGauge(MetricNames.QueueJobsCount, {
+        description: 'Number of jobs in the queue by state',
+        unit: '{jobs}',
+      });
```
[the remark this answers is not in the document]
roggervalf: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/taskforcesh/bullmq/pull/3811#discussion_r2895908700)

docs/gitbook/guide/telemetry/metrics.md
```diff
@@ -73,15 +73,24 @@ BullMQ automatically records the following metrics:
...
-| ------------------- | ---------------------------------------------------------------------------------- |
-| `bullmq.queue.name` | Name of the queue                                                                  |
-| `bullmq.job.name`   | Name of the job                                                                    |
-| `bullmq.job.status` | Status of the job (completed, failed, delayed, retried, waiting, waiting-children) |
+| Attribute                 | Description                                                                        |
+| ------------------------- | ---------------------------------------------------------------------------------- |
+| `bullmq.queue.name`       | Name of the queue                                                                  |
+| `bullmq.job.name`         | Name of the job                                                                    |
+| `bullmq.job.status`       | Status of the job (completed, failed, delayed, retried, waiting, waiting-children) |
+| `bullmq.queue.jobs.state` | Job state for gauge metrics (waiting, active, completed, failed, etc.)             |
```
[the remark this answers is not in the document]
roggervalf: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/taskforcesh/bullmq/pull/3811#discussion_r2895908725)

src/classes/queue-getters.ts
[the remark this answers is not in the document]
roggervalf: as commented in Copilot pr, users can use old version of otel package, this is why we are using any to prevent a TypeError

Issue #3761, opened 2026-02-03
[Feature]: Support deduplication for parent jobs in flows


matthewshober:
https://github.com/taskforcesh/bullmq/pull/3637 removed the deduplication option from both parent and child jobs in flows. However, the reasoning for removing it specifically from parent jobs was not explained. It would be very helpful if deduplication support could be restored for parent jobs.

miridih-jhjoo:
I fully support this request.

_Scenario: In our use case (e.g., Download → Cancel → Edit → Re-download), the previous request becomes invalid. We want to avoid processing the old job to prevent unnecessary resource consumption, as the result is no longer needed by the user._

Currently, since native deduplication for parent jobs was removed, we are forced to use Redis to track and manually remove old parent jobs before adding new ones. This adds unnecessary complexity to our infrastructure.

P.S. This workaround also affects queue order. Ideally, we want to replace the job in-place (A1 → B → C becomes A2 → B → C). However, since we have to remove and re-add, the new job loses priority and ends up at the back (B → C → A2). (While assigning a timestamp-based priority to every job could be a workaround to maintain order, it is not a fundamental solution.)

vahiwe:
We’d also like to see deduplication support for parent jobs in flows restored.

We use `FlowProducer` (both `add` and `addBulk`) in production and rely on deduplication for the top-level jobs we add. Without it, the same logical unit of work can be enqueued multiple times, e.g. when a recurring job runs again before the previous run finishes, or when several triggers fire for the same entity. That leads to duplicate processing and forces us to either accept that or implement our own deduplication outside BullMQ.

Having deduplication on parent jobs in flows would let us keep a single, consistent approach and avoid duplicate work.

manast:
I think there is a bit of confusion about this feature. The reason the option was removed was because there were edge cases where it won't work as explained in the comments here: https://github.com/taskforcesh/bullmq/pull/3637
So what we need to do is a new implementation that is robust enough so that we can give guarantees that the functionality works as expected. Hopefully @roggervalf can look into this in more detail in a couple of weeks and provide a solution.

manast:
I noticed after re-reading my comment that I was still not clear. It is not that we had deduplication implemented in flows and then removed. It is that we implemented deduplication for regular jobs, and we did not anticipate its use in flows. That it worked somehow is more a lucky side effect than having designed the feature properly for the flows case in the first place.

PR #3817, opened 2026-02-21
fix(flow): allow deduplication from root parent

roggervalf:
### Why

  1. Why is this change necessary? In order to support deduplication from root parent node in flow producer, I enabled it from interface but with extra checks to prevent parent and dedup options to be passed together. Additionally, replacing a root parent won't be available as it would require to update every single child parent reference as well.

### How

  1. How did you implement this? Add check for not allowing parent and dedup option. Add typing in root parent options to allow dedup option with a limitation of avoiding the use of replace option.

fixes https://github.com/taskforcesh/bullmq/issues/3761

### DIFF
--- a/src/commands/includes/deduplicateJobWithoutReplace.lua
+++ b/src/commands/includes/deduplicateJobWithoutReplace.lua
@@ -1,5 +1,5 @@
 --[[
-  Function to debounce a job.
+  Function to deduplicate a job.
 ]]
 
 local function deduplicateJobWithoutReplace(deduplicationId, deduplicationOpts, jobId, deduplicationKey,


src/classes/job.ts
```diff
@@ -1608,13 +1608,23 @@ export class Job<
       if (!this.opts.deduplication?.id) {
         throw new Error('Deduplication id must be provided');
       }
+
+      if (this.parentKey) {
```
manast, line 1612: will this not be considered a breaking change if users are already using this option?

src/classes/job.ts
```diff
@@ -1608,13 +1608,23 @@ export class Job<
...
+      }
     }
 
     // TODO: remove in v6
     if (this.opts.debounce) {
       if (!this.opts.debounce?.id) {
         throw new Error('Debounce id must be provided');
       }
+
+      if (this.parentKey) {
```
manast, line 1625: same as above.

src/classes/job.ts
roggervalf: the issue is that allowing both parameters together would cause that the same child belongs to more than  1 parent and we do not support that with our current design

src/classes/job.ts
roggervalf: same as above

PR #3830, opened 2026-02-24
chore: remove unused dependencies across ports

roggervalf:
### Why

  1. Why is this change necessary? In order to address this vulnerability https://osv.dev/vulnerability/GHSA-3ppc-4f35-3m26. Also remove non-needed ignore rule and resolutions, there were published fixed for ajv dependencies

### How

_Any extra info here._

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bullmq",
-  "version": "5.70.1",
+  "version": "5.70.2",
   "description": "Queue for messages and jobs based on Redis",
   "homepage": "https://bullmq.io/",
   "main": "./dist/cjs/index.js",
--- a/docs/gitbook/changelog.md
+++ b/docs/gitbook/changelog.md
@@ -1,3 +1,10 @@
+## [5.70.2](https://github.com/taskforcesh/bullmq/compare/v5.70.1...v5.70.2) (2026-03-05)
+
+
+### Bug Fixes
+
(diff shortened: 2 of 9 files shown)


elixir/package.json
```diff
@@ -20,8 +20,7 @@
     "@semantic-release/git": "10.0.1",
     "@semantic-release/github": "12.0.6",
     "@semantic-release/release-notes-generator": "14.1.0",
-    "semantic-release": "25.0.3",
-    "@semantic-release/npm": "13.1.4"
```
roggervalf: not used in elixir and keeping minimal dependencies

python/package.json
```diff
@@ -22,7 +22,6 @@
     "@semantic-release/exec": "7.1.0",
     "@semantic-release/git": "10.0.1",
     "@semantic-release/github": "12.0.6",
-    "@semantic-release/npm": "13.1.4",
```
roggervalf: same as above but for python

osv-scanner.toml
```diff
@@ -2,5 +2,3 @@
 # https://google.github.io/osv-scanner/configuration/
 
 [[IgnoredVulns]]
-id = "GHSA-2g4f-4pwh-qvx6"
```
roggervalf, line 5: no need as transitive dev has a new published version with a fix

package.json
```diff
@@ -68,8 +68,6 @@
     "uuid": "11.1.0"
   },
   "devDependencies": {
-    "@eslint/js": "9.39.2",
-    "@eslint/eslintrc": "3.3.1",
```
roggervalf, line 72: these 2 packages are not used

osv-scanner.toml
```diff
@@ -2,5 +2,3 @@
 # https://google.github.io/osv-scanner/configuration/
 
 [[IgnoredVulns]]
```
roggervalf, line 4: keeping title section for future reference

PR #3853, opened 2026-03-04
Fix/fix leave orphane jobs in autoremove

manast:
### Why

We introduced an issue in this PR: https://github.com/taskforcesh/bullmq/pull/3694 where auto remove would leave orphan jobs when using max age. 

### How

Make sure that we remove from the sets the same jobs that we auto-removed. This PR also includes an utility function to remove orphan jobs that may have been created by this issue.

_Any extra info here._

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -1064,7 +1064,8 @@ export class Queue<
    *
    * Orphaned keys can occur in rare cases when the removal-by-max-age logic
    * removes sorted-set entries without fully cleaning up the corresponding
-   * job hash data (fixed in v5.67.0). Under normal operation this method is
+   * job hash data (a regression introduced in v5.66.6 via #3694).
+   * Under normal operation this method is
    * **not needed** — it is provided only as a one-time migration helper for
    * users who were affected by that specific bug and want to reclaim the
    * leaked memory.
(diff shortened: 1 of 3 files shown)


src/classes/queue.ts
```diff
@@ -1056,4 +1057,119 @@ export class Queue<
...
+      const candidateJobIds = new Set<string>();
+      for (const key of keys) {
+        const suffix = key.slice(basePrefix.length);
+
+        // Skip infrastructure keys (derived from this.keys).
+        if (knownSuffixes.has(suffix)) {
+          continue;
+        }
+
+        // Skip sub-keys of infrastructure prefixes (e.g. repeat:xxx, de:xxx).
```
roggervalf, line 1132: I believe that generated repeatable jobs also have repeat:xxx pattern as ids too, so we might be skipping job hashes from old repeatable jobs

tests/clean.test.ts
```diff
@@ -892,4 +892,351 @@ describe('Cleaner', () => {
...
+      // Infrastructure keys should still exist
+      expect(await client.exists(`${baseKey}:meta`)).toBe(1);
+    });
+
+    it('should not confuse repeat sub-keys with orphaned jobs', async () => {
+      const client = await queue.client;
+      const baseKey = `${prefix}:${queue.name}`;
+
+      // Simulate repeat sub-keys (e.g., repeat:abc123)
+      await client.set(`${baseKey}:repeat:some-repeat-id`, 'repeat data');
```
roggervalf, line 1144: need a test for jobs being generated by a repeatable job that will follow this pattern: `${prefix}:${queue.name}:repeat:${number}:${millis}`

Issue #3852, opened 2026-03-04
[Bug]: `_d.jobId.includes is not a function`


shaul613:
### Version

v5.69.3

### Platform

NodeJS

### What happened?

This code in `src/classes.job.ts` unintentionally throws a bug when the job id is a Number. It accounts for a numeric string, but if it's type Number an error is throws as a result of calling `.includes(':')` on it when checking for colons:
```
if (this.opts?.jobId) {
      if (`${parseInt(this.opts.jobId, 10)}` === this.opts?.jobId) {
        throw new Error('Custom Id cannot be integers');
      }

      // TODO: replace this check in next breaking check with include(':')
      // By using split we are still keeping compatibility with old repeatable jobs
      if (
        this.opts?.jobId.includes(':') &&
        this.opts?.jobId?.split(':').length !== 3
      ) {
        throw new Error('Custom Id cannot contain :');
      }
    }
```

### How to reproduce.

Run a job with a job id of Number type, and you'll get `_d.jobId.includes is not a function`

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hayaiti:
Confirmed — the `jobId` validation code assumes the value is always a string, but the TypeScript type definition allows `string | number`. When a `Number` is passed, `.includes()` doesn't exist on Number prototype.

## Root Cause

```typescript
// src/classes/job.ts
if (this.opts?.jobId) {
  // This check works for string "123" but NOT for Number(123)
  if (`${parseInt(this.opts.jobId, 10)}` === this.opts?.jobId) {
    throw new Error('Custom Id cannot be integers');
  }

  // .includes() is a String method — throws on Number
  if (this.opts?.jobId.includes(':') && ...) {
    throw new Error('Custom Id cannot contain :');
  }
}
```

When `jobId` is `Number(42)`:
- `parseInt(42, 10)` → `42`
- `` `${42}` === 42 `` → `false` (string vs number comparison)
- `(42).includes(':')` → `TypeError: includes is not a function`

## Fix

Convert to string at the start of validation:

```typescript
if (this.opts?.jobId) {
  const jobId = String(this.opts.jobId);

  if (`${parseInt(jobId, 10)}` === jobId) {
    throw new Error('Custom Id cannot be integers');
  }

  if (jobId.includes(':') && jobId.split(':').length !== 3) {
    throw new Error('Custom Id cannot contain :');
  }
}
```

[...]

PR #3854, opened 2026-03-04
Fixed calling .includes() on job id of type Number - issue #3852

shaul613:
### Why

Prevents a TypeError: jobId.includes is not a function when a custom jobId is passed as a Number. The current logic assumes jobId is always a String before calling .includes(). This was reported in https://github.com/taskforcesh/bullmq/issues/3852 after investigating a production issue.

### How

Added a type check to skip the string-based validation (numeric string and colon checks) if the jobId is already a Number.

If Number types are intentionally disallowed for jobId, we should implement an explicit validation or type-guard earlier in the lifecycle.

### DIFF
--- a/tests/job.test.ts
+++ b/tests/job.test.ts
@@ -74,6 +74,13 @@ describe('Job', () => {
       expect(createdJob.id).toBe(customJobId);
     });
 
+    it('should use custom jobId when it is type Number', async () => {
+      const job = await Job.create(queue, 'test', data, {
+        jobId: 54321,
+      });
+      expect(job.id).toBe(54321);
+    });
+
     describe('when custom jobId is provided as empty string', () => {
       it('should ignore the empty custom id and generates a numeric id', async () => {
         const job = await Job.create(queue, 'test', data, {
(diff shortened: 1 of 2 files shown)


tests/job.test.ts
```diff
@@ -74,6 +74,13 @@ describe('Job', () => {
       expect(createdJob.id).toBe(customJobId);
     });
 
+    it('should use custom jobId when it is type Number', async () => {
+      const job = await Job.create(queue, 'test', data, {
+        jobId: 54321,
```
roggervalf, line 79: our type definition for jobId option is string, this will show up as a type error as well

shaul613: @roggervalf There isn't any indication in the code that it intends to disalllow numbers. It only disallows numeric strings. If it intended to disallow type Number, there should be an explicit condition for it. Curently a type error is thrown because we attempt to call `.includes()` on a Number which is not a function.

manast: @shaul613 I think @roggervalf comment here is quite valid. This is a defensive check that is in place because the typescript definition is not enough to capture this special case. We only guarantee that BullMQ apis work correctly if you pass the expected types as arguments according to the typescript definitions, we should not enter the rabbit hole of defensively checking for arguments data types in runtime.

PR #3879, opened 2026-03-12
fix(rate-limiter): avoid unnecessary GET when limiter key does not exist

### LINKED ISSUE #3848: [Bug]: Unnecessary GET operations on rate limiter key
### Version

5.34.10

### Platform

NodeJS

### What happened?

BullMQ's moveToActive Lua script performs GET operations on rate limiter keys that result in cache misses even when rate limiting is properly configured. This occurs because rate limiter keys are ephemeral (created when jobs are processed, expire after the configured duration), and workers check for these keys between job executions when they don't exist. 

At scale, this creates significant noise in Redis monitoring metrics and makes cache miss rates appear problematic when the system is actually functioning correctly. 

Why this matters: 

  >   Misleading metrics - Cache miss rates of 60-70% even though queues are healthy and keeping up with work
     Monitoring blind spots - Can't distinguish real cache issues from expected rate limiter checks
     Misleading alarms - High miss rates trigger false alerts about cache performance
     Cost implications - In serverless/managed Redis (AWS ElastiCache, etc.), operations are metered
     

In moveToActive script, getRateLimitTTL is called for every job retrieval attempt:
```lua
local maxJobs = tonumber(opts['limiter'] and opts['limiter']['max'])
local expireTime = getRateLimitTTL(maxJobs, rateLimiterKey)
if expireTime > 0 then return {0, 0, expireTime, 0} end
```
The `getRateLimitTTL` function performs a GET operation: 
```lua
local function getRateLimitTTL(maxJobs, rateLimiterKey)
  if maxJobs and maxJobs <= tonumber(rcall("GET", rateLimiterKey) or 0) then
    local pttl = rcall("PTTL", rateLimiterKey)
    if pttl == 0 then
      rcall("DEL", rateLimiterKey)
    end
    if pttl > 0 then
      return pttl
    end
  end
  return 0
end
```

Rate limiter keys are created via INCR when a job is processed and set to expire after the configured duration (e.g., 10 seconds, 2 minutes). Between job executions, the key expires and no longer exists. When workers poll for jobs during these gaps, the GET operation returns nil → cache miss.

Proposal:

Modify getRateLimitTTL to check for key existence before performing the GET:
```lua
local function getRateLimitTTL(maxJobs, rateLimiterKey)
  if maxJobs then
    -- Check if key exists first to avoid unnecessary GET on non-existent keys
    if rcall("EXISTS", rateLimiterKey) == 1 then
      local currentCount = tonumber(rcall("GET", rateLimiterKey))
      if currentCount and maxJobs <= currentCount then
        local pttl = rcall("PTTL", rateLimiterKey)
        if pttl > 0 then
          return pttl
        elseif pttl == 0 then
          rcall("DEL", rateLimiterKey)
        end
      end
    end
  end
  return 0
end
```

### How to reproduce.

1. Configure a queue with rate limiting (e.g., max: 1, duration: 10000)
2. Enable Redis slowlog: CONFIG SET slowlog-log-slower-than 0
3. Let workers poll the queue while it's mostly empty
4. Check slowlog: SLOWLOG GET 100
5. Observe frequent GET {queue}:limiter operations
6. Check if key exists: GET {queue}:limiter → returns (nil)
7.  Monitor Redis cache hit/miss metrics → see high miss rate

### Relevant log output

```shell
SLOWLOG entry:
1) "GET"
2) "bull:posthog-integration-processing-queue:limiter"

Redis response:
(nil)
```

### Code of Conduct

orolega:
# Why
Fixes a Redis cache miss issue caused by getRateLimitTTL unconditionally calling GET on the rate limiter key on every job dequeue and completion, even when the key doesn't exist. Rate limiter keys are ephemeral — created on job processing via INCR and expiring after the configured duration. Between job executions the key is absent, so every GET returns nil, generating a cache miss. 

Closes #3848

# How
Added an EXISTS check in `getRateLimitTTL.lua` before the GET. When the key is absent — the common steady-state between rate limit windows — the function returns 0 immediately after a single EXISTS call, eliminating the miss-generating GET. The PTTL branches were also reordered so pttl > 0 returns early and pttl == 0 is an elseif, avoiding a redundant second condition check.

Tests covering all branches of the modified function were added to `rate_limiter.test.ts` under `describe('getRateLimitTTL function behavior')`.

### Additional Notes
image

image

### DIFF
--- a/src/commands/includes/getRateLimitTTL.lua
+++ b/src/commands/includes/getRateLimitTTL.lua
@@ -2,15 +2,18 @@
   Function to get current rate limit ttl.
 ]]
 local function getRateLimitTTL(maxJobs, rateLimiterKey)
-  if maxJobs and maxJobs <= tonumber(rcall("GET", rateLimiterKey) or 0) then
-    local pttl = rcall("PTTL", rateLimiterKey)
-
-    if pttl == 0 then
-      rcall("DEL", rateLimiterKey)
-    end
-
-    if pttl > 0 then
-      return pttl
+  if maxJobs then
+    -- Check if key exists first to avoid unnecessary GET on non-existent keys
+    if rcall("EXISTS", rateLimiterKey) == 1 then
+      local currentCount = tonumber(rcall("GET", rateLimiterKey))
+      if currentCount and maxJobs <= currentCount then
+        local pttl = rcall("PTTL", rateLimiterKey)
+        if pttl > 0 then
+          return pttl
+        elseif pttl == 0 then
+          rcall("DEL", rateLimiterKey)
+        end
+      end
     end
   end
   return 0
--- a/tests/rate_limiter.test.ts
+++ b/tests/rate_limiter.test.ts
@@ -1383,6 +1383,112 @@ describe('Rate Limiter', () => {
     });
   });
 
+  describe('getRateLimitTTL function behavior', () => {
+    // Key behaviors under test are:
+    //
+    //   1. Key does not exist  → EXISTS returns 0 → skip GET, return 0
+    //   2. Key exists, count < maxJobs  → return 0
+    //   3. Key exists, count >= maxJobs, PTTL > 0  → return PTTL
+    //   4. Key exists, count >= maxJobs, PTTL == -1 (no TTL)  → return 0
+    //   5. Key exists, count >= maxJobs, PTTL == 0 (transient expired state)
+    //        → DEL key + return 0
+    //   6. No maxJobs arg, no global rate limit  → raw PTTL (-2 when key absent)
+    //   7. No maxJobs arg, global rate limit configured  → uses meta max
+
+    describe('when the rate limiter key does not exist', () => {
+      it('should return 0 without performing a GET', async () => {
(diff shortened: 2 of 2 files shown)


manast:
Thank you for your PR. In reality there is no need to issue an EXISTS first and then a GET, that would instead create more overhead as you need 2 Redis commands instead of just 1, and both have the same complexity. So really, this just makes it a bit slower than what we already have.

orolega:
@manast do you have an alternative suggestion to resolve the cache misses when the rate limiter key doesnt exist? This causes significant noise in its current form https://github.com/taskforcesh/bullmq/issues/3848

manast:
Not sure what you mean with cache misses. As a matter of fact you must configure your Redis instance with a max memory police of noeviction: https://redis.io/docs/latest/develop/reference/eviction/#eviction-policies otherwise we cannot guarantee a correct behaviour of BullMQ.

orolega:
@manast see the bug I reported for more details https://github.com/taskforcesh/bullmq/issues/3848 and here is a screen capture of the metric 
image


Sorry I do not see the relation of the key eviction to the issue I have raised

PR #3885, opened 2026-03-17
fix(queue): guard getCompleted against empty task set [python]

kent-hydra:
## Problem

`getCompleted()` calls `asyncio.wait(task_set, return_when=FIRST_COMPLETED)` without checking if `task_set` is empty. In all Python versions, `asyncio.wait()` raises `ValueError` when called with an empty set:

```
ValueError: Set of coroutines/Futures is empty.
```

This can occur during drain or closing transitions when `self.processing` is empty.

## Root cause

`python/bullmq/worker.py` line 376:
```python
async def getCompleted(task_set: set, emit_callback) -> tuple[list[Job], set]:
    job_set, pending = await asyncio.wait(task_set, ...)  # ValueError if empty
```

The TypeScript worker avoids this entirely by using `AsyncFifoQueue` which handles empty states internally.

## Fix

Add an early return when `task_set` is empty, with a short sleep to yield control and prevent tight-looping:

```python
if not task_set:
    await asyncio.sleep(0.1)
    return [], set()
```

This is a no-op for the normal case (non-empty set) and prevents the crash on the edge case.

## Impact

Without this fix, workers crash with `ValueError` during low-load periods or closing transitions, triggering restart loops.

### DIFF
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -378,6 +378,9 @@ class Worker(EventEmitter):
 
 
 async def getCompleted(task_set: set, emit_callback) -> tuple[list[Job], set]:
+    if not task_set:
+        await asyncio.sleep(0.1)
+        return [], set()
     job_set, pending = await asyncio.wait(task_set, return_when=asyncio.FIRST_COMPLETED)
     jobs = [extract_result(job_task, emit_callback) for job_task in job_set]
     # we filter `None` out to remove:
--- a/python/bullmq/worker.py
+++ b/python/bullmq/worker.py
@@ -379,7 +379,7 @@ class Worker(EventEmitter):
 
 async def getCompleted(task_set: set, emit_callback) -> tuple[list[Job], set]:
     if not task_set:
-        await asyncio.sleep(0.1)
+        await asyncio.sleep(0)
         return [], set()
     job_set, pending = await asyncio.wait(task_set, return_when=asyncio.FIRST_COMPLETED)
     jobs = [extract_result(job_task, emit_callback) for job_task in job_set]
--- a/python/tests/worker_test.py
+++ b/python/tests/worker_test.py
@@ -7,6 +7,7 @@ https://bbc.github.io/cloudfit-public-docs/asyncio/testing.html
 from asyncio import Future
 import redis.asyncio as redis
 from bullmq import Queue, Worker, Job, WaitingChildrenError
+from bullmq.worker import getCompleted
 from uuid import uuid4
 from enum import Enum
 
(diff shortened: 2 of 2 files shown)


python/bullmq/worker.py
```diff
@@ -373,6 +373,9 @@ def cancelProcessing(self):
 
 
 async def getCompleted(task_set: set, emit_callback) -> tuple[list[Job], set]:
+    if not task_set:
+        await asyncio.sleep(0.1)
```
roggervalf: why this delay is needed?

kent-hydra: Without the sleep, when processing is empty getCompleted returns immediately with (set(), set()). The caller's while loop then re-invokes it on the next iteration with no await point that actually yields, creating a busy-wait spin at ~100% CPU. The asyncio.sleep(0.1) provides a brief yield so the event loop can process other tasks and avoids the hot loop.

manast: @kent-hydra what about sleep(0) as suggested by copilot?

kent-hydra: Agreed — `sleep(0)` is sufficient. Traced the call site: when `self.processing` is empty during closing, the next iteration hits the `len(jobs) == 0 and self.closing` break in `run()`, so a 0.1s sleep would just add up to 100ms of shutdown latency. For any other transient empty state, the inner loop refills `processing` as soon as `self.waiting`/concurrency frees up, so a single event-loop yield is enough.

Updated to `await asyncio.sleep(0)` in 1a85625.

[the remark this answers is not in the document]
kent-hydra: Applied — switched to `await asyncio.sleep(0)` in 1a85625. Avoids the magic-constant shutdown delay; the empty state resolves within a single iteration in practice (either the close path breaks out at the `len(jobs) == 0 and self.closing` check, or the inner loop refills `processing`).

python/bullmq/worker.py
```diff
@@ -373,6 +373,9 @@ def cancelProcessing(self):
...
+        return [], set()
     job_set, pending = await asyncio.wait(task_set, return_when=asyncio.FIRST_COMPLETED)
```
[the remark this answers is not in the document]
kent-hydra: Added a regression test in `python/tests/worker_test.py` (1a85625) that calls `getCompleted(set(), noop)` directly and asserts it returns `([], set())` without raising. This guards against future refactors removing the empty-set guard.

Went with a direct unit test on the helper rather than an integration test because the empty-`self.processing` state is transient and hard to trigger deterministically; testing the helper directly catches the same regression more reliably.

kent-hydra:
@manast — quick question on test approach before I iterate further.

Current PR has a unit test (`test_get_completed_handles_empty_task_set`) calling `getCompleted(set(), ...)` directly via `from bullmq.worker import getCompleted`. It just ran in CI on the rebased commit and passed (`testing python@3.13, redis@7-alpine` ✓).

Copilot's review asked for an integration-style test (start a Worker on an empty queue, trigger drain/close, assert `worker.close()` doesn't raise) which would better match the rest of `worker_test.py`'s style. I traced the `run()` loop and the empty-`self.processing` path is timing-dependent — the `len(jobs) == 0 and self.closing` break in `worker.py:112` tends to short-circuit before `getCompleted` is called with an empty set. So an integration test may not deterministically fail without the fix.

Trade-offs:
- **Unit test (current):** deterministic regression coverage, but imports an internal module-level helper — stylistically different from the rest of the file.
- **Integration test:** matches house style, but weaker regression guarantee (covers close-on-idle flow rather than the specific bug).
- **Both:** full coverage, slightly noisier diff.

Which do you prefer? Happy to switch or add either — just want your call before pushing again.

roggervalf (approved):
Lgtm

PR #3889, opened 2026-03-17
fix(flow-producer): use deduplication id when no jobId is provided

### LINKED ISSUE #3878: [Bug]: Flow de-duplication behavior differs from regular jobs
### Version

v5.70.4

### Platform

NodeJS

### What happened?

When working with regular jobs, a de-duplicated job returns the same `job.id` as the original. This makes it easy to identify duplicates when using custom job IDs, e.g..

```node
const jobId = 'example';
const job = await queue.add(
	'example',
	{},
	{
		deduplication: { id: 'dedup' },
		jobId: jobId,
	}
);

if (job.id !== jobId) {
	console.log('Job has been de-duplicated');
}
```
 
However, this behavior does not carry over to flows. Is it expected that flows don’t return the same `job.id` when de-duplicated, unlike regular jobs?

### How to reproduce.

```node
const flow1 = await flow.add({
	name: 'example1',
	queueName: 'example',
	opts: {
		deduplication: { id: 'dedup' },
		delay: 500,
	},
});

const flow2 = await flow.add({
	name: 'example2',
	queueName: 'example',
	opts: {
		deduplication: { id: 'dedup' },
		delay: 500,
	},
});

if (flow1.job.id !== flow2.job.id) {
	throw new Error('Expected flow IDs to be the same');
}
```

### Relevant log output

```shell

```

### Code of Conduct

sohailfiza:
Fixes #3878

## Problem
FlowProducer.add does not respect deduplication.id, resulting in duplicate jobs when using flows.

## Solution
This PR attempts to address this by using `deduplication.id` as a fallback `jobId` when no explicit `jobId` is provided.

## Notes
- Keeps existing `jobId` precedence unchanged
- Adds tests to cover the deduplication behavior

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -320,7 +320,9 @@ export class FlowProducer extends EventEmitter {
     const queueOpts = queuesOpts && queuesOpts[node.queueName];
 
     const jobsOpts = queueOpts?.defaultJobOptions ?? {};
-    const jobId = node.opts?.jobId || v4();
+    const deduplicationId = node.opts?.deduplication?.id;
+
+    const jobId = node.opts?.jobId ?? deduplicationId ?? v4();
 
     return trace<Promise<JobNode>>(
       this.telemetry,
--- a/tests/flow-producer-deduplication.test.ts
+++ b/tests/flow-producer-deduplication.test.ts
@@ -1,0 +1,105 @@
+import IORedis from 'ioredis';
+import { describe, beforeEach, afterEach, it, expect } from 'vitest';
+import { v4 } from 'uuid';
+
+import { FlowProducer } from '../src';
+import { removeAllQueueData } from '../src/utils';
+
+describe('FlowProducer deduplication', () => {
+  const redisHost = process.env.REDIS_HOST || '127.0.0.1';
+
+  let flowProducer: FlowProducer;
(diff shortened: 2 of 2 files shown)


src/classes/flow-producer.ts
```diff
@@ -320,7 +320,9 @@ export class FlowProducer extends EventEmitter {
     const queueOpts = queuesOpts && queuesOpts[node.queueName];
 
     const jobsOpts = queueOpts?.defaultJobOptions ?? {};
-    const jobId = node.opts?.jobId || v4();
+    const deduplicationId = node.opts?.deduplication?.id;
+
+    const jobId = node.opts?.jobId ?? deduplicationId ?? v4();
```
roggervalf, line 325: deduplication id is not supposed to have same value as jobId, also because jobId validation happens before deduplication logic https://github.com/taskforcesh/bullmq/blob/master/src/commands/addParentJob-6.lua#L80

sohailfiza: Thanks for the clarification.
I see now that jobId and deduplicationId serve different purposes, so using one as a fallback for the other isn’t the right approach.
I’ll revisit this and update the PR accordingly.

roggervalf:
closing it for now, a new pr can be opened of course when ready

PR #3902, opened 2026-03-24
feat(deduplication): add keepLastIfActive option for at-least-once-after-active semantics

### LINKED ISSUE #3427: Add "requeue" deduplication mode to ensure at least one execution after last trigger
I need to ensure a job runs **at least once** after an action occurs, while avoiding unnecessary duplicate executions when multiple triggers happen in rapid succession.

Currently, BullMQ's deduplication modes have a critical gap:
- **Simple mode**: Deduplicates until job completes, but if a trigger occurs while the job is actively processing, that trigger is lost forever
- **Throttle mode**: Similar issue - triggers during job execution are ignored
- **Debounce mode**: Requires a delay and doesn't handle triggers during active processing

This is a common pattern needed for:
- **Data synchronization**: Multiple data changes should eventually sync, with at least one sync after the last change
- **Cache invalidation**: Multiple updates should trigger invalidation, ensuring cache is refreshed after the last update
- **Webhook processing**: Multiple webhook events for the same resource should be processed at least once after the last event

Related to https://github.com/taskforcesh/bullmq/issues/3151#issuecomment-3248946649, which describes a similar need, but I think hasn't been completely fulfilled yet.

## Describe the solution you'd like

A new deduplication mode that guarantees at least one job execution after the last trigger attempt:

```javascript
await myQueue.add(
  'process-action',
  { data },
  {
    deduplication: {
      id: 'unique-job-id',
      requeueIfActive: true // new option
    }
  }
);
```

### Behavior:

1. If job with same dedup ID is waiting/delayed → Deduplicate (ignore new job)
2. If job is active (processing) → Mark for automatic requeue after completion
3. If job is completed/failed → Allow new job immediately
4. While marking a job as complete → Check if requeue was requested and automatically schedule a new job

This ensures:

✅ No unnecessary duplicate processing when job is already queued
✅ No lost triggers when job is actively processing
✅ Automatic handling without manual event listeners
✅ Works across multiple Node.js processes

## Example use case

```javascript
// Deployment pipeline triggered by git webhooks
async function onGitPush(commitData) {
  await deployQueue.add(
    'deploy',
    { commit: commitData.sha },
    {
      deduplication: {
        id: `deploy-${commitData.repo}`,
        requeueIfActive: true
      }
    }
  );
}

// If multiple pushes happen:
// - Push 1 at 0ms: Job scheduled
// - Push 2 at 100ms (job waiting): Deduplicated
// - Push 3 at 5000ms (job active): Deduplicated but marked for requeue
// - Push 4 at 5100ms (job active): Deduplicated, requeue already marked
// - Job completes at 10000ms: New job automatically scheduled
// Result: Exactly 2 deployments run, last push is always deployed
```

## Describe alternatives you've considered

- Manual tracking with deduplicated events: Requires complex state management in Redis, prone to race conditions
- Using job.updateData() during processing: Worker needs to check for updates, prone to race conditions too
- Version-based deduplication: Every trigger creates a job (no deduplication benefit)
- External pending queue: Adds complexity and potential for race conditions

None of these alternatives are as clean or reliable as having this built into BullMQ.

## Additional context

I'm willing to contribute a PR for this feature if the maintainers agree with the approach. The implementation could:

- Use atomic Redis operations to track requeue requests
- Integrate with existing deduplication infrastructure
- Include comprehensive tests
- Maintain backward compatibility
- Add minimal performance overhead (only for jobs using this mode)

Would the team be open to this feature? If so, I'm happy to discuss implementation details and submit a PR.

manast:
Closes #3427

### Why

In order to implement the requested feature in #3427 

### How

When `requeueIfActive: true` is set in deduplication options, a new job will be created after the currently active deduplicated job completes or fails. Adding new jobs to the queue with the same deduplication id while the job is active overwrite the stored data, so only the latest is used when creating the next job.

This guarantees at most 2 jobs per dedup ID at any time (1 active + 1 pending), with no parallel execution possible. The pending job data is stored in a Redis hash at `dn:{deduplicationId}` (dedup-next) and is atomically consumed in moveToFinished.

_Any extra info here._

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -39,7 +39,7 @@ jobs:
         env:
           CI: true
       - name: deploy
-        uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 # v3
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
         with:
-          node-version: '24.13.1'
+          node-version: '24.14.1'
           cache: 'yarn'
       - name: Install dependencies Node
         run: yarn install --frozen-lockfile --non-interactive
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     ports:
       - 6379:6379
   dragonfly:
-    image: 'docker.dragonflydb.io/dragonflydb/dragonfly:v1.27.0'
+    image: 'docker.dragonflydb.io/dragonflydb/dragonfly:v1.37.2'
     ports:
       - '6380:6379'
     environment:


php/src/Queue.php
```diff
@@ -153,6 +162,16 @@ public function addBulk(array $jobs): array
             $opts = $jobData['opts'] ?? [];
 
             $job = new Job($this, $name, $data, $opts);
+
+            if ($job->opts->deduplication !== null
+                && !empty($job->opts->deduplication['keepLastIfActive'])
+                && $job->opts->delay > 0
+            ) {
+                throw new \InvalidArgumentException(
+                    'keepLastIfActive cannot be used together with delay option'
```
roggervalf: just for curiosity, why it cannot be used with delay

src/commands/includes/storeDeduplicatedNextJob.lua
```diff
@@ -0,0 +1,50 @@
+--[[
+  Function to store a deduplicated next job if the existing job is active
+  and keepLastIfActive is set. When the active job finishes, the stored
+  proto-job is used to create a real job in the queue.
+  Returns true if the proto-job was stored, false otherwise.
+]]
+local function storeDeduplicatedNextJob(deduplicationOpts, currentDebounceJobId, prefix,
+    deduplicationId, jobName, jobData, fullOpts, eventsKey, maxEvents, jobId,
+    parentKey, parentData, parentDependenciesKey, repeatJobKey)
+    if deduplicationOpts['keepLastIfActive'] and prefix and currentDebounceJobId then
```
roggervalf: prefix is not supposed to be always present?

php/src/Queue.php
manast: it gets quite messy conceptually, because the job is already delayed as it is waiting for the previous active job, and is supposed to start as soon as the previous one completes. If we supported delayed jobs it would create awkward situations were a delayed job is waiting for the active to complete and then when the active completes it becomes delayed instead of active, and at the same time deduplicating other jobs. For now and unless some user can motivate a good reason for supporting delayed jobs in this particular case, I think it is safer to not support it.

php/package.json
```diff
@@ -67,6 +67,7 @@
   },
   "packageManager": "yarn@1.22.22",
   "resolutions": {
-    "ajv": "8.18.0"
+    "ajv": "8.18.0",
+    "picomatch": ">=2.3.2"
```
roggervalf, line 71: this won't be needed as lock file update is sufficient https://github.com/taskforcesh/bullmq/pull/3906

php/src/Queue.php
roggervalf: I was thinking that it would be useful when using debounce https://docs.bullmq.io/guide/jobs/deduplication#debounce-mode, so we can still replace at least the second job while first one is still active

src/commands/includes/storeDeduplicatedNextJob.lua
```diff
@@ -0,0 +1,50 @@
...
+  and keepLastIfActive is set. When the active job finishes, the stored
+  proto-job is used to create a real job in the queue.
+  Returns true if the proto-job was stored, false otherwise.
+]]
+local function storeDeduplicatedNextJob(deduplicationOpts, currentDebounceJobId, prefix,
+    deduplicationId, jobName, jobData, fullOpts, eventsKey, maxEvents, jobId,
+    parentKey, parentData, parentDependenciesKey, repeatJobKey)
+    if deduplicationOpts['keepLastIfActive'] and currentDebounceJobId then
+        local activeKey = prefix .. "active"
+        if rcall('LPOS', activeKey, currentDebounceJobId) ~= false then
```
roggervalf: we have a check that for redis versions lower than 6.0.6, we should use LRANGE

src/types/deduplication-options.ts
```diff
@@ -21,4 +21,23 @@ export type DeduplicationOptions = {
...
+   * at any time: 1 active and 1 waiting, preventing parallel
+   * execution of jobs with the same deduplication ID.
+   *
+   * Note: when this option is set, `ttl` is ignored. The dedup key
+   * is kept alive without expiry for the job's entire lifecycle
+   * and cleaned up on completion or failure.
+   *
+   * Cannot be used together with `delay`.
+   */
+  keepLastIfActive?: boolean;
```
roggervalf, line 40: there is a little gap here, as we are only considering adding next job while previous it's active, but regular replacements only happen when previous job is in delayed state. That means that when previous job is in waiting or paused or prioritized state, next jobs won't be added. We can probably extend this logic to consider states different than delayed, but not finished (completed/failed) when adding next job

php/package.json
manast: The problem is that if the lock file is removed, which I casually need to do to get a fresh yarn.lock, then we get back the problem. Resolutions is a more explicit way to show which actual overrides we currently have, whereas in the lock file is virtually impossible to keep track of it.

src/types/deduplication-options.ts
manast: Not sure I understand, but if the job is not active then we should get the same behaviour as we had before without the "keepLastIfActive" flag. In other words, this flag only enhances current implementation by also considering active job as a special state that allows us to keep the last added job instead of discarding it.

php/src/Queue.php
manast: I added support for delay now too.

src/commands/includes/storeDeduplicatedNextJob.lua
```diff
@@ -0,0 +1,53 @@
...
+
+            rcall('HSET', deduplicationNextKey, unpack(fields))
+
+            -- Ensure the dedup key does not expire while the job is active,
+            -- so subsequent adds always hit the dedup path and never bypass
+            -- the active-check because of a TTL expiry.
+            local deduplicationKey = prefix .. "de:" .. deduplicationId
+            rcall('PERSIST', deduplicationKey)
+
+            rcall("XADD", eventsKey, "MAXLEN", "~", maxEvents, "*", "event", "deduplicated", "jobId",
```
roggervalf, line 50: legacy debounced event is missed to be emitted here as well for consistency until we can remove it in next breaking change

manast: addressed in last commit

PR #3905, opened 2026-03-25
fix(flow): correctly propagate jobId from pipeline results

### LINKED ISSUE #3878: [Bug]: Flow de-duplication behavior differs from regular jobs

sohailfiza:
Fixes #3878

### Problem
When adding flows using FlowProducer, deduplicated jobs return incorrect job.id.
And Redis was correctly returning deduplicated jobId, but it was not propagated
back to the Job instances because multi.exec() results were not mapped.

### Soluution
Tracking command index for each job.addJob(multi) call. And after multi.exec() i mapped results to respective jobs and updated job.id with the actual value returned by redis.

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -21,7 +21,6 @@ import { SpanKind, TelemetryAttributes } from '../enums';
 export interface AddNodeOpts {
   multi: ChainableCommander;
   node: FlowJob;
-  jobsToUpdate: FlowJobUpdate[];
   parent?: {
     parentOpts: ParentOptions;
     parentDependenciesKey: string;
@@ -35,17 +34,11 @@ export interface AddNodeOpts {
 export interface AddChildrenOpts {
   multi: ChainableCommander;
   nodes: FlowJob[];
-  jobsToUpdate: FlowJobUpdate[];
   parent: {
     parentOpts: ParentOptions;
     parentDependenciesKey: string;
   };
   queuesOpts?: FlowQueuesOpts;
-}
-
-interface FlowJobUpdate {
-  job: Job;
-  resultIndex: number;
 }
 
 export interface NodeOpts {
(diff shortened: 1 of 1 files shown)


src/classes/flow-producer.ts
```diff
@@ -289,9 +302,12 @@ export class FlowProducer extends EventEmitter {
...
         });
 
-        const jobsTrees = await this.addNodes(multi, flows);
+        const jobsTrees = await this.addNodes(multi, flows, jobsToUpdate);
 
-        await multi.exec();
+        const results = (await multi.exec()) as
+          | [null | Error, string | number][]
+          | null;
+        this.updateJobIds(jobsToUpdate, results);
```
roggervalf: since this pr https://github.com/taskforcesh/bullmq/pull/3817 the only expected job to use deduplication option is the root parent job of a flow. No need to keep track of jobsToUpdate array if the only to be updated will be the root parent job

sohailfiza: Thanks, that makes sense. I’ll update the PR to only map the root job result.

sohailfiza: Updated the PR to only map root job ids after multi.exec(). Let me know if this looks good or if anything else should be adjusted.

sohailfiza:
It looks like OSV scan failure is due to existing dependency vulnerabilities in repo. Should I address them in this PR.

PR #3922, opened 2026-04-01
feat(sandbox): add getDependencies proxy for sandboxed processors

PreAgile:
### Why
Sandboxed processors cannot call `job.getDependencies()` because the IPC proxy
is missing. Refs #3533

### How
Following the existing `getChildrenValues` proxy pattern:
1. Added `GetDependencies` to `ParentCommand` enum
2. Added `GetDependenciesResponse` to `ChildCommand` enum
3. Added proxy method in `child-processor.ts` `wrapJob()` (forwards opts via message value)
4. Added message handler in `sandbox.ts`
5. Added fixture-based tests (success + timeout) in `sandboxed_process.test.ts`

### Additional Notes
- This is the first of two PRs for #3533. The second PR will add `getDependenciesCount` proxy.
- `SandboxedJob` interface is intentionally unchanged, matching the existing pattern where
  `getChildrenValues` and `getIgnoredChildrenFailures` are also not declared on the interface.
  Happy to add the type declaration if you'd prefer — let me know.

PR #3923, opened 2026-04-01
feat(sandbox): add getDependenciesCount proxy for sandboxed processors

### LINKED ISSUE #3533: [Bug]:  missing getDependencies and getDependenciesCount proxy support for sandboxed/worker thread processors
### Version

v5.63.0

### Platform

NodeJS

### What happened?

 When using sandboxed processors, the Job object lacks proxy support for getDependencies() and getDependenciesCount() methods, causing a "getDependencies is not a function" runtime error.

  The getChildrenValues() method already has proxy support (added in commit https://github.com/taskforcesh/bullmq/commit/dcc3b0628f992546d7b93f509795e5d4eb3e1b15), but getDependencies() and getDependenciesCount() are missing.

### How to reproduce.

```ts
export default async function(job) {
    // ✅ Works - proxy exists
    const childrenValues = await job.getChildrenValues();

    // ❌ TypeError: job.getDependencies is not a function
    const deps = await job.getDependencies();

    // ❌ TypeError: job.getDependenciesCount is not a function
    const depsCount = await job.getDependenciesCount();
  }
```

### Code of Conduct

PreAgile:
### Why
Sandboxed processors cannot call `job.getDependenciesCount()` because the IPC proxy
is missing. Fixes #3533

### How
Following the existing `getChildrenValues` proxy pattern:
1. Added `GetDependenciesCount` to `ParentCommand` enum
2. Added `GetDependenciesCountResponse` to `ChildCommand` enum
3. Added proxy method in `child-processor.ts` `wrapJob()` (forwards opts via message value)
4. Added message handler in `sandbox.ts`
5. Added fixture-based tests (success + timeout) in `sandboxed_process.test.ts`

### Additional Notes
- This is the second of two PRs for #3533. The first PR (#3922) adds `getDependencies` proxy.
- `SandboxedJob` interface is intentionally unchanged, matching the existing pattern where
  `getChildrenValues` and `getIgnoredChildrenFailures` are also not declared on the interface.
  Happy to add the type declaration if you'd prefer — let me know.

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -270,7 +270,12 @@ export class ChildProcessor {
       /**
        * Proxy `getDependenciesCount` function.
        */
-      getDependenciesCount: async (opts?: Record<string, any>) => {
+      getDependenciesCount: async (opts?: {
+        failed?: boolean;
+        ignored?: boolean;
+        processed?: boolean;
+        unprocessed?: boolean;
+      }) => {
         const requestId = Math.random().toString(36).substring(2, 15);
         await send({
           requestId,


src/classes/child-processor.ts
```diff
@@ -266,6 +266,25 @@ export class ChildProcessor {
           'getIgnoredChildrenFailures',
         );
       },
+
+      /**
+       * Proxy `getDependenciesCount` function.
+       */
+      getDependenciesCount: async (opts?: Record<string, any>) => {
```
roggervalf: opts can follow same typing as in https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L1299-L1304

PreAgile: Good catch! I've updated the `opts` parameter type in the proxy from `Record<string, any>` to match the exact shape from `Job.getDependenciesCount` (lines 1299–1304):

```ts
opts?: {
  failed?: boolean;
  ignored?: boolean;
  processed?: boolean;
  unprocessed?: boolean;
}
```

Committed and pushed — thanks for the review!

PR #3935, opened 2026-04-03
docs(changelog): update last version in each port

roggervalf:
### Why

  1. Why is this change necessary? Update last version log in each port to reflect accurate changes in each version

### How

_Any extra info here._

### DIFF
--- a/php/package.json
+++ b/php/package.json
@@ -35,7 +35,7 @@
         {
           "releaseRules": [
             {
-              "message": "!(*\\[php\\]*)",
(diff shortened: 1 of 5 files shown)


php/package.json
```diff
@@ -35,7 +35,7 @@
         {
           "releaseRules": [
             {
-              "message": "!(*\\[php\\]*)",
+              "message": "!(*\\(php\\)*|*\\[php\\]*)",
```
roggervalf, line 38: (php) was not considered as in other ports

manast (approved):

PR #3942, opened 2026-04-06
fix: allow dynamic strings as jobSchedulerId in upsertJobScheduler (#3937)

### LINKED ISSUE #3937: [Bug]: upsertJobScheduler ID parameter is incorrectly constrained to NameType
### Version

5.73.0

### Platform

NodeJS

### What happened?

### How to reproduce.

```typescript

 myQueue.upsertJobScheduler(
  { pattern: '0 * * * *' },
  {
    data: { foo: 'bar' }
  }
);
```

### Relevant log output

```shell

```

### Code of Conduct

Manmeet567:
This PR resolves issue #3937, where the jobSchedulerId parameter in the upsertJobScheduler method was strictly constrained to the NameType generic.

In many real-world use cases, while the JobName might be part of a strict union (e.g., 'report' | 'email'), the jobSchedulerId needs to be a unique, dynamic identifier such as a UUID or a database ID. The previous type definition forced these IDs to match the NameType union, causing TypeScript errors when using non-literal strings.

Changes
src/classes/queue.ts: Updated the upsertJobScheduler method overloads to accept string for the jobSchedulerId parameter.

src/classes/job-scheduler.ts: (If applicable) Aligned the internal scheduler types to support general string IDs.

Maintained existing generics for DataType, ResultType, and NameType to ensure that job data and return types remain strictly typed.

Verification & Testing
Automated Test: Added a reproduction test case in src/tests/test_issue_3937.ts.

Scenario: Verified that a Queue defined with strict JobNames successfully accepts a dynamic string variable as a jobSchedulerId without TypeScript compilation errors.

Build: Confirmed that yarn build and yarn lint pass successfully on a local environment.

Closes #3937

### DIFF
--- a/tests/test_issue_3937.test.ts
+++ b/tests/test_issue_3937.test.ts
@@ -1,0 +1,24 @@
+import { Queue } from '../src/classes/queue';
+
+type JobName = 'generate-report' | 'send-email';
+
+interface JobData {
+  accountId: string;
+}
+
+const queue = new Queue<JobData, void, JobName>('issue-3937');
+const dynamicId = String('inst_12345');
+
+// Scheduler IDs must accept dynamic strings independently of the job name type.
+queue.upsertJobScheduler(
+  dynamicId,
+  {
+    every: 60_000,
+  },
+  {
+    name: 'generate-report',
+    data: {
+      accountId: 'acct_12345',
+    },
+  },
+);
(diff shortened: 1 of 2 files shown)


tests/test_issue_3937.test.ts
```diff
@@ -0,0 +1,24 @@
+import { Queue } from '../src/classes/queue';
```
manast, line 1: of historical reasons we do not create files for specific issues, instead we add a new test on the proper test file that covers job schedulers functionality.

PR #3945, opened 2026-04-07
feat(sandbox): add missing method declarations to SandboxedJob interface

### LINKED ISSUE #3925: Add getChildrenValues and getIgnoredChildrenFailures to SandboxedJob interface
There is a mismatch between the runtime behavior of sandboxed jobs and the public TypeScript interface.

In `child-processor.ts`, `wrapJob()` adds `getChildrenValues` and `getIgnoredChildrenFailures` to the sandboxed job object. These methods are fully implemented via IPC and handled on the parent side, which means they are available at runtime in sandboxed processors.

However, these methods are not declared in the `SandboxedJob` interface. As a result, TypeScript users cannot access them without using type assertions, even though they work correctly at runtime.

---

If these methods are intended to be used within sandboxed processors, it would be helpful to include them in the `SandboxedJob` interface so that the type definitions match the actual runtime behavior.

For example:

```ts
getChildrenValues(): Promise<Record<string, any>>;
getIgnoredChildrenFailures(): Promise<Record<string, string>>;
```

---

* Using type assertions such as `(job as any).getChildrenValues()`
* Extending the interface locally

However, these approaches reduce type safety and make the supported API surface unclear.

---

**Additional context**
These methods are already implemented in `wrapJob()` and handled in the parent process, so the runtime support already exists.

Aligning the interface with the implementation would improve developer experience and reduce confusion.

If this aligns with the intended API, I would like to open a PR to add these methods to the interface.

yogeshwaran-c:
## What kind of change does this PR introduce?

Feature / Type alignment

## What is the current behavior?

The `SandboxedJob` interface is missing type declarations for `getChildrenValues`, `getIgnoredChildrenFailures`, and `getDependenciesCount`. These methods are fully implemented at runtime in `wrapJob()` (in `child-processor.ts`) and handled on the parent side (in `sandbox.ts`), but TypeScript users cannot access them without using type assertions like `(job as any).getChildrenValues()`.

Closes #3925

## What is the new behavior?

The `SandboxedJob` interface now declares all three methods with proper type signatures matching the `Job` class:

- `getChildrenValues<CT = any>(): Promise<{ [jobKey: string]: CT }>`
- `getIgnoredChildrenFailures(): Promise<{ [jobKey: string]: string }>`
- `getDependenciesCount(opts?): Promise<{ failed?: number; ignored?: number; processed?: number; unprocessed?: number }>`

The implementations in `child-processor.ts` are also updated with explicit return type annotations and proper type casts for the `waitResponse` calls, ensuring type safety between the interface and implementation.

## Additional context

- No runtime behavior changes — this is a types-only alignment
- `getDependenciesCount` was also missing from the interface (same pattern as the two methods mentioned in the issue), so it is included for completeness
- The return type signatures match those on the `Job` class (`src/classes/job.ts`)
- Follows the same pattern used by `moveToWaitingChildren` which already had a proper type cast

### DIFF
--- a/src/classes/child-processor.ts
+++ b/src/classes/child-processor.ts
@@ -228,7 +228,9 @@ export class ChildProcessor {
       /**
        * Proxy `getChildrenValues` function.
        */
-      getChildrenValues: async () => {
+      getChildrenValues: async <CT = any>(): Promise<{
+        [jobKey: string]: CT;
+      }> => {
         const requestId = Math.random().toString(36).substring(2, 15);
         await send({
           requestId,
@@ -240,7 +242,7 @@ export class ChildProcessor {
           this.receiver,
           RESPONSE_TIMEOUT,
           'getChildrenValues',
-        );
+        ) as Promise<{ [jobKey: string]: CT }>;
       },
 
       /**
@@ -252,7 +254,9 @@ export class ChildProcessor {
        * @returns - A promise that resolves with the ignored children failures.
        * The exact structure of the returned data depends on the parent process implementation.
        */
-      getIgnoredChildrenFailures: async () => {
+      getIgnoredChildrenFailures: async (): Promise<{
+        [jobKey: string]: string;
+      }> => {
         const requestId = Math.random().toString(36).substring(2, 15);
         await send({
           requestId,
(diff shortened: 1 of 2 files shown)


manast:
Isn't this PR obsolete after this one? https://github.com/taskforcesh/bullmq/pull/3946

yogeshwaran-c:
Yes, you're right — #3946 was merged and supersedes this one. Closing this PR. Thanks for the review!

PR #3947, opened 2026-04-07
fix(events): correct delayed event type and docs for delay field

### LINKED ISSUE #3267: [Bug]: delay event for QueueEvents has incorrect docs and types
### Version

v5.52.2

### Platform

NodeJS

### What happened?

### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

yogeshwaran-c:
## What kind of change does this PR introduce?

Bug fix (type correction + documentation fix)

## What is the current behavior?

The `delayed` event in `QueueEventsListener` has two issues:

1. **Incorrect JSDoc**: The `delay` field is documented as *"The delay duration in milliseconds before the job becomes active"*, but it is actually a **UNIX timestamp** (milliseconds since epoch) representing *when* the job will become active. This can be verified in the Lua scripts (`addDelayedJob.lua`, `moveToDelayed-8.lua`, `moveParentToWait.lua`) which all emit `delayedTimestamp` (computed as `timestamp + delay`), not the delay duration.

2. **Incorrect TypeScript type**: The `delay` field is typed as `number`, but Redis streams return all values as strings. The `array2obj` utility (which parses stream entries) returns `Record<string, string>`, so `delay` is actually a `string` at runtime. This is consistent with other event fields like `count: string` in the `cleaned` event.

   Evidence from existing tests:
   - `sandboxed_process.test.ts:1124` already uses `Number(delay)`, indicating the author knew it's a string
   - `flow.test.ts:4730` used `delay - Date.now()` which only worked due to JavaScript's implicit coercion

Closes #3267

## What is the new behavior?

- **JSDoc**: Updated to correctly describe `delay` as *"The timestamp (UNIX time in milliseconds) at which the job will become active, represented as a string due to Redis serialization"*
- **Type**: Changed from `delay: number` to `delay: string` to match the actual runtime value
- **Tests**: Updated `flow.test.ts` and `job_scheduler.test.ts` to use `Number(delay)` for arithmetic and `string` for type annotations

## Additional context

The Lua code in `addDelayedJob.lua` that emits this event:
```lua
local delayedTimestamp = (delay > 0 and (tonumber(timestamp) + delay)) or tonumber(timestamp)
rcall("XADD", eventsKey, "MAXLEN", "~", maxEvents, "*", "event", "delayed",
    "jobId", jobId, "delay", delayedTimestamp)
```

This clearly shows `delay` in the event payload is `delayedTimestamp`, not the delay duration.

### DIFF
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -103,7 +103,8 @@ export interface QueueEventsListener extends IoredisListener {
   /**
    * Listen to 'delayed' event.
    *
-   * This event is triggered when a job is scheduled with a delay before it becomes active.
+   * This event is triggered when a job is scheduled with a delay. When the delay
+   * expires, the job will be moved to waiting, prioritized or paused state.
    *
    * @param args - An object containing details about the delayed job.
    *  - `jobId` - The unique identifier of the job that was delayed.
--- a/src/classes/queue-events.ts
+++ b/src/classes/queue-events.ts
@@ -107,10 +107,11 @@ export interface QueueEventsListener extends IoredisListener {
    *
    * @param args - An object containing details about the delayed job.
    *  - `jobId` - The unique identifier of the job that was delayed.
-   *  - `delay` - The delay duration in milliseconds before the job becomes active.
+   *  - `delay` - The timestamp (UNIX time in milliseconds) at which the job will become
+   *    active, represented as a string due to Redis serialization.
    * @param id - The identifier of the event.
    */
-  delayed: (args: { jobId: string; delay: number }, id: string) => void;
+  delayed: (args: { jobId: string; delay: string }, id: string) => void;
 
   /**
    * Listen to 'drained' event.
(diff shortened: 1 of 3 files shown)


src/classes/queue-events.ts
```diff
@@ -107,10 +107,11 @@ export interface QueueEventsListener extends IoredisListener {
    *
    * @param args - An object containing details about the delayed job.
    *  - `jobId` - The unique identifier of the job that was delayed.
-   *  - `delay` - The delay duration in milliseconds before the job becomes active.
+   *  - `delay` - The timestamp (UNIX time in milliseconds) at which the job will become
+   *    active, represented as a string due to Redis serialization.
```
roggervalf:
```suggestion
   *  - `delay` - The timestamp (UNIX time in milliseconds) at which the job will be
   *    promoted, represented as a string due to Redis serialization.
```
jobs won't be moved to active directly, they are moved to waiting, prioritized or paused

roggervalf:
addressed in v5.79.3

PR #3961, opened 2026-04-08
fix(queue): filter undefined jobs from getJobs results

### LINKED ISSUE #3767: [Bug]: `queue.getActive()` can sometimes contain `undefined`
### Version

4.18.3

### Platform

NodeJS

### What happened?

### How to reproduce.

```ts
await queue.waitUntilReady();

  "Stuff",
  { connection: redisClient },
);
```

### Relevant log output

```shell

```

### Code of Conduct

yogeshwaran-c:
## Summary
- Fixes #3767 where `queue.getActive()` (and other `getJobs` based methods) could return arrays containing `undefined` values
- `Job.fromId()` returns `undefined` when a job hash is missing from Redis (e.g., due to race conditions where a job completes between listing IDs and fetching data). The `Promise.all()` preserves these `undefined` entries in the result array
- Added a `.filter(job => job !== undefined)` after the `Promise.all()` call in `getJobs()` to strip out missing jobs before returning

## Test plan
- [ ] Verify that `queue.getActive()`, `queue.getCompleted()`, and other status-based getters no longer include `undefined` in results
- [ ] Verify existing tests still pass

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -490,9 +490,10 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs = await Promise.all(
+      jobIds.map(jobId => this.Job.fromId(this, jobId)),
     );
+    return jobs.filter(job => job !== undefined) as JobBase[];
   }
 
   /**


src/classes/queue-getters.ts
```diff
@@ -490,9 +490,10 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs = await Promise.all(
+      jobIds.map(jobId => this.Job.fromId(this, jobId)),
     );
+    return jobs.filter(job => job !== undefined) as JobBase[];
```
roggervalf, line 496: undefined values happen when in the middle of getting ids and job payloads, jobs are finished and removed. By filtering undefined values we might affect pagination where users expect a specific length from this array while they don't reach the last page. In other words, this behavior should be ok

yogeshwaran-c:
Thanks @roggervalf, that's a fair point I hadn't considered — silently filtering would break callers who depend on `getJobs(start, end)` returning a length consistent with the page size, and they'd never realise they need to advance past a partial page. Closing this PR. If anyone hits this in practice, a separate issue discussing whether to expose the missing-job sentinels (e.g. `null` placeholders or a dedicated `getJobsWithStatus` that surfaces removed entries explicitly) might be worth opening, but that's a bigger API change than what this PR was.

PR #3964, opened 2026-04-08
fix: use connection from opts in FlowProducer [python]

### LINKED ISSUE #3619: [Bug]: FlowProducer doesn't use provided connection string
### Version

2.17.0

### Platform

Python

### What happened?

Hi,

when using 

```python
FlowProducer({}, {"connection": f"redis://{host}:6379"})`
```

the connection string is not used but defaults to localhost. 

However, defining it like this

```python
FlowProducer(f"redis://{host}:6379")`
```

works correctly.

It seems, this is either a bug or missing info in the docs.

### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

yogeshwaran-c:
## Summary

- Fixes #3619: Python `FlowProducer` now falls back to `opts["connection"]` when `redisOpts` is not provided, matching the behavior of `Queue` and `Worker` classes
- Previously, passing `connection` inside the `opts` dict was silently ignored, requiring users to pass connection config as the first positional argument

## Details

In `python/bullmq/flow_producer.py`, the constructor was hardcoded to use only the `redisOpts` parameter for `RedisConnection`, ignoring `opts.get("connection")`. The `Queue` class already reads the connection from `opts`:

```python
# Queue (existing correct behavior)
redisOpts = opts.get("connection", {})
self.redisConnection = RedisConnection(redisOpts)
```

This PR applies the same pattern to `FlowProducer`:

```python
# FlowProducer (fixed)
connection = redisOpts if redisOpts else opts.get("connection", {})
self.redisConnection = RedisConnection(connection)
```

This allows users to instantiate `FlowProducer` the same way as `Queue`:

```python
flow = FlowProducer(opts={"connection": "redis://localhost:6379"})
```

## Test plan

- [ ] Verify `FlowProducer(redisOpts="redis://...")` still works (direct connection string)
- [ ] Verify `FlowProducer(opts={"connection": "redis://..."})` now works (opts-based connection)
- [ ] Verify `FlowProducer(redisOpts="redis://...", opts={"connection": "redis://other"})` uses `redisOpts` (first arg takes priority)
- [ ] Verify `FlowProducer()` defaults to empty dict connection (unchanged behavior)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/bullmq/flow_producer.py
+++ b/python/bullmq/flow_producer.py
@@ -36,7 +36,8 @@ class FlowProducer:
         """
         Initialize a connection
         """
-        self.redisConnection = RedisConnection(redisOpts)
+        connection = redisOpts if redisOpts else opts.get("connection", {})
+        self.redisConnection = RedisConnection(connection)
         self.client = self.redisConnection.conn
         self.opts: dict = opts
         self.prefix = opts.get("prefix", "bull")
--- a/python/tests/flow_test.py
+++ b/python/tests/flow_test.py
@@ -284,5 +284,50 @@ class TestJob(unittest.IsolatedAsyncioTestCase):
         await child_queue.obliterate()
         await child_queue.close()
 
+    async def test_should_use_connection_from_opts_when_redis_opts_empty(self):
+        """Test that FlowProducer uses opts connection when redisOpts is empty"""
+        parent_queue_name = f"__test_parent_queue__{uuid4().hex}"
+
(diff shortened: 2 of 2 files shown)


python/bullmq/flow_producer.py
roggervalf, line 35: connection property is declared in QueueBaseOptions type, might need to be extracted from here but as in the TODO comment, we will remove the second parameter in the next breaking change

roggervalf:
addressed in last breaking change

PR #3967, opened 2026-04-08
fix(metrics): add total counters to Prometheus export for accurate counts with auto-removal

### LINKED ISSUE #3603: [Feature]: exportPrometheusMetrics should return valid data when used alongside auto job removal
### Version

v5.58.9

### Platform

NodeJS

### What happened?

A bug happened!

### How to reproduce.

Since auto job removal removed completed/failed jobs, `exportPrometheusMetrics` method returns trimmed data. it should use `getMetrics` method under the hood instead of `getJobCounts`.
For example: if you set `removeOnComplete` to 5, `exportPrometheusMetrics` will always show "5" for completed jobs never less or more.

### Relevant log output

```shell

```

### Code of Conduct

yogeshwaran-c:
## Summary

- Fixes #3603 - `exportPrometheusMetrics()` now includes `bullmq_job_completed_total` and `bullmq_job_failed_total` counter-type metrics sourced from `getMetrics()`, which tracks historical totals via `meta.count`
- The existing gauge metrics from `getJobCounts()` are preserved unchanged; the new counters are appended after them
- When metrics collection is not enabled, the new counters are silently skipped (wrapped in try/catch)

## Problem

When `removeOnComplete` or `removeOnFail` is configured, `getJobCounts()` returns current counts from Redis sets which are trimmed and don't reflect the true total number of processed jobs. This makes the Prometheus export inaccurate for monitoring throughput.

## Solution

`getMetrics('completed')` and `getMetrics('failed')` provide historical counts via `meta.count` that accurately track totals regardless of job removal settings. This PR adds these as Prometheus counter-type metrics alongside the existing gauge metrics.

## Test plan

- [ ] Verify existing `exportPrometheusMetrics` tests still pass
- [ ] Verify new counter metrics appear in output when metrics collection is enabled
- [ ] Verify no errors when metrics collection is not enabled (try/catch fallback)
- [ ] Verify counters reflect true totals even with `removeOnComplete`/`removeOnFail` configured

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -705,6 +705,29 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       );
     }
 
+    try {
+      const completedMetrics = await this.getMetrics('completed');
+      const failedMetrics = await this.getMetrics('failed');
+
+      metrics.push(
+        '# HELP bullmq_job_completed_total Total number of completed jobs',
+      );
+      metrics.push('# TYPE bullmq_job_completed_total counter');
+      metrics.push(
+        `bullmq_job_completed_total{queue="${this.name}"${variables}} ${completedMetrics.meta.count}`,
+      );
+
+      metrics.push(
+        '# HELP bullmq_job_failed_total Total number of failed jobs',
+      );
+      metrics.push('# TYPE bullmq_job_failed_total counter');
+      metrics.push(
+        `bullmq_job_failed_total{queue="${this.name}"${variables}} ${failedMetrics.meta.count}`,
+      );
+    } catch (err) {
+      // getMetrics may not be available if metrics are not enabled
+    }
+
     return metrics.join('\n');
   }
 }
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -705,28 +705,26 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       );
     }
 
-    try {
-      const completedMetrics = await this.getMetrics('completed');
-      const failedMetrics = await this.getMetrics('failed');
+    const [completedMetrics, failedMetrics] = await Promise.all([
+      this.getMetrics('completed'),
+      this.getMetrics('failed'),
+    ]);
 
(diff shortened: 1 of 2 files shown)


src/classes/queue-getters.ts
```diff
@@ -705,6 +705,29 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
       );
     }
 
+    try {
```
manast: It should not be needed to use try/catch here as getMetrics will not fail, it will return an empty object if metrics are not enabled.

manast (commented):
We would also need a couple of new tests to cover this new functionality.

yogeshwaran-c:
Thanks @manast and @Copilot for the review.

Pushed changes addressing all four points:

- Dropped the unnecessary `try/catch` around `getMetrics()` and parallelised the two calls with `Promise.all` (already in 293dc50).
- Added an `escapePrometheusLabelValue` helper that escapes backslashes, double-quotes, and newlines per the [Prometheus exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/). It is applied to both `this.name` and every `globalVariables` value, so queue names or label values containing `\`, `"`, or newlines no longer produce invalid output.
- Added three tests in `tests/getters.test.ts`:
  - `should export bullmq_job_completed_total and bullmq_job_failed_total counters from getMetrics` — verifies the new counter rows render with the right HELP/TYPE/value lines using stubbed `getMetrics` results.
  - `should emit zero-valued counters when metrics collection is disabled` — covers the case where `getMetrics` returns `count: 0` (which is what happens when metrics keys aren't populated yet, since `getMetrics` parses the missing fields as 0 rather than throwing).
  - `should escape special characters in queue names and global variable values` — uses a queue name containing a literal backslash, double-quote, and newline and asserts the output contains the properly escaped label value.

I ran `vitest run tests/getters.test.ts -t "exportPrometheusMetrics"` against a local Redis 5 instance: all 6 tests in the describe block pass.

yogeshwaran-c:

manast (approved):

PR #3982, opened 2026-04-10
fix(worker): make resume() work after pause(true) (closes #3971)

yogeshwaran-c:
## Summary

Fixes #3971: `resume()` was a no-op after `pause(true)`, leaving the worker permanently stuck in `running=true`, `paused=true` state.

## Root cause

- `pause(true)` (doNotWaitActive) sets `this.paused = true` but skips `whenCurrentJobsFinished()`, so the main loop has not yet exited and `this.running` is still `true`.
- The old `resume()` checked `if (!this.running)` and bailed out in that case, so nothing happened. The main loop would eventually exit and set `running = false`, but nothing ever called `run()` again to restart processing.

## Fix

`resume()` now gates on `this.paused` instead of `!this.running` and handles both scenarios:

1. **Worker not running** (classic case): clear `paused`, call `run()`.
2. **Worker still running** (after `pause(true)`): clear `paused`. If the main loop is still exiting asynchronously, wait for `mainLoopRunning` to settle and restart via `run()` only if the worker is still not paused/closing/running.

## Test plan

- Added `should resume after pause(true) while jobs are still active` in `tests/worker.test.ts` which:
  - Starts a worker, adds a long-running job
  - Calls `pause(true)` while the job is still active
  - Calls `resume()` and verifies `isPaused()` is false
  - Adds another job and verifies it gets processed

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -1185,7 +1185,7 @@ will never work with more accuracy than 1ms. */
    * Resumes processing of this worker (if paused).
    */
   resume(): void {
-    if (!this.running) {
+    if (this.paused) {
       this.trace<void>(SpanKind.INTERNAL, 'resume', this.name, span => {
         span?.setAttributes({
           [TelemetryAttributes.WorkerId]: this.id,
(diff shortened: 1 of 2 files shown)


roggervalf:
it was already fixed by https://github.com/taskforcesh/bullmq/pull/3974

PR #3988, opened 2026-04-11
fix(queue-events): stringify non-string values in publishEvent

### LINKED ISSUE #2984: [Bug]: QueueEventsProducer - publishEvent does not stringify input
### Version

v5.34.5

### Platform

NodeJS

### What happened?

When attempting to publish an event using a QueueEventsProducer, the input object is not stringified. This happens due to the implied `toString()` call here https://github.com/taskforcesh/bullmq/blob/0763cd693acee7b6b80fefad0e8b1950e41a1a74/src/classes/queue-events-producer.ts#L43. `JSON.stringify` should be used and deserialized on the other end.

### How to reproduce.

```typescript
const producer = new QueueEventsProducer([QUEUE_NAME},
    {
      connection: [CONNECTION],
    },
  );

await producer.publishEvent({eventName: 'name', nested: {object: 'hello'}})
```

The above produces a log event with

```json
{
  "eventName": "name",
  "nested": "[Object object]"
}
```

### Relevant log output

_No response_

### Code of Conduct

maruthang:
### Why

`QueueEventsProducer.publishEvent` pushes event payload values directly into Redis `XADD` args without serializing them. When a value is a non-string type (e.g. an object or number), this causes silent data corruption or runtime errors because Redis streams expect string values.

Closes #2984

### How

- In `src/classes/queue-events-producer.ts`, non-string values are now serialized with `JSON.stringify()` before being appended to the `XADD` args array. String values pass through unchanged.
- Added a regression test in `tests/events.test.ts` that publishes an event with nested object and number values, then verifies they arrive as properly serialized JSON strings.

### Additional Notes

- The fix is minimal and backwards-compatible: any payload that was already passing strings will behave identically.
- The new test covers both object and numeric value types in a single event to ensure comprehensive serialization.


src/classes/queue-events-producer.ts
```diff
@@ -40,7 +40,10 @@ export class QueueEventsProducer extends QueueBase {
     const args: any[] = ['MAXLEN', '~', maxEvents, '*', 'event', eventName];
 
     for (const [key, value] of Object.entries(restArgs)) {
-      args.push(key, value);
+      args.push(
+        key,
+        typeof value === 'string' ? value : JSON.stringify(value),
```
manast: The only issue left is that when a user listens for an event, it will not be possible to know if the event is a string or a json encoded object.

maruthang: Good point, thanks @manast. I will update to always wrap values in JSON (including strings), so listeners can rely on a consistent JSON.parse on the receiving end. That preserves type information and keeps the producer/consumer contract symmetric. Will push an update shortly.

PR #3989, opened 2026-04-11
fix(queue-events): correct delay event type and documentation

### LINKED ISSUE #3267: [Bug]: delay event for QueueEvents has incorrect docs and types

maruthang:
### Why

The `delayed` event in `QueueEventsListener` has an incorrect type and misleading documentation for its `delay` field (Closes #3267).

The `delay` field was typed as `number` and documented as "delay duration in milliseconds", but in reality:
- Redis streams transmit all values as strings, so the value is always a `string`
- The Lua scripts emit `delayedTimestamp` (a Unix epoch timestamp), not a duration

This causes type errors for consumers who rely on the TypeScript definitions and leads to incorrect usage based on the documentation.

### How

- Changed the `delay` field type from `number` to `string` in the `delayed` event listener interface in `src/classes/queue-events.ts`
- Updated the JSDoc to accurately describe the value as "the timestamp at which the job will become active, in milliseconds elapsed since January 1, 1970 UTC"

### Additional Notes

This is a type-level breaking change for TypeScript consumers who were passing `delay` to APIs expecting `number`. However, the previous type was incorrect and did not match runtime behavior, so this is a correctness fix.


src/classes/queue-events.ts
```diff
@@ -107,10 +107,11 @@ export interface QueueEventsListener extends IoredisListener {
    *
    * @param args - An object containing details about the delayed job.
    *  - `jobId` - The unique identifier of the job that was delayed.
-   *  - `delay` - The delay duration in milliseconds before the job becomes active.
+   *  - `delay` - The timestamp at which the job will become active, in milliseconds
+   *    elapsed since January 1, 1970 UTC.
    * @param id - The identifier of the event.
    */
-  delayed: (args: { jobId: string; delay: number }, id: string) => void;
+  delayed: (args: { jobId: string; delay: string }, id: string) => void;
```
manast: the question is if we do not want to keep a number here and convert the value to a number before calling the event callback...

maruthang: Fair question @manast. Keeping the type as `number` and coercing internally is friendlier for consumers and avoids a breaking API change. I will flip the approach: convert the string payload to a number before invoking the callback, and keep the listener signature as `number`. Will push an update.

roggervalf (dismissed):
Lgtm

maruthang:
The single CI failure was a 10s timeout in the unrelated job-cancellation concurrency test on the coverage job (959/960 passing) — a flake, not related to this typing/doc change. Branch is now updated against master; should be green on re-run.

roggervalf (approved):

PR #3990, opened 2026-04-11
fix(queue): allow any string as jobSchedulerId in upsertJobScheduler

### LINKED ISSUE #3937: [Bug]: upsertJobScheduler ID parameter is incorrectly constrained to NameType

maruthang:
### Why

Closes #3937

The `upsertJobScheduler` method in `Queue` constrains its `jobSchedulerId` parameter to `NameType`, which is a union type designed for job names. Job scheduler IDs are independent identifiers and should accept any `string`. The internal `JobScheduler` class already correctly types this parameter as `string`, so the public API is inconsistent with the implementation.

### How

Changed the type of the `jobSchedulerId` parameter in `Queue.upsertJobScheduler` from `NameType` to `string` (`src/classes/queue.ts`, line 469). This aligns the public API with the internal `JobScheduler` class and removes the unnecessary type constraint.

### Additional Notes

- Single-line change with no runtime behavior difference — purely a type-level fix.
- No new tests needed since this relaxes a type constraint; all existing tests continue to pass.

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -487,7 +487,7 @@ export class Queue<
     >(
       jobSchedulerId,
       repeatOpts,
-      jobTemplate?.name ?? jobSchedulerId,
+      jobTemplate?.name ?? (jobSchedulerId as unknown as NameType),
       jobTemplate?.data ?? <DataType>{},
       { ...this.jobsOpts, ...jobTemplate?.opts },
       { override: true },
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -466,7 +466,7 @@ export class Queue<
    * @returns The next job to be scheduled (would normally be in delayed state).
    */
   async upsertJobScheduler(
-    jobSchedulerId: NameType,
+    jobSchedulerId: string,
     repeatOpts: Omit<RepeatOptions, 'key'>,
     jobTemplate?: {
       name?: NameType;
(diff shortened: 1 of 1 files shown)


manast (dismissed):
LGTM

maruthang:
Hi @manast — friendly ping on this one. Got a prior LGTM, all 21 checks green, no conflicts with master. Happy to rebase if anything has drifted. Thanks!

manast:
There are some comments from copilot as well.

manast:
there are copilot comments to resolve as well

maruthang:
Thanks @manast — both Copilot inline comments on this PR were already addressed in the latest revision; I just realized the threads weren't formally resolved in the UI, which is probably what made them look outstanding. Resolving them now.

Quick status for convenience:

1. **`@param key` JSDoc** (#discussion_r3072424608) — fixed in `4df4a2e`. The JSDoc now reads `@param jobSchedulerId - Unique identifier for the job scheduler. Accepts any string so callers do not have to widen their NameType union just to label a scheduler.`
2. **`NameType`/`jobTemplate.name` type-safety** (#discussion_r3072424557) — addressed in `00865dd`. The fallback at the call site is now an explicit `as unknown as NameType` widening with a comment documenting that it's only used when the caller didn't supply `jobTemplate.name`, so prior behaviour is preserved when `NameType` defaults to `string` and the cast is opt-in (and visible in code review) when callers narrow it to a union.

Happy to make further changes if either of those resolutions doesn't quite match what you had in mind.

PR #3991, opened 2026-04-11
test(pause): add regression test for resume after pause(true)

maruthang:
### Why
The fix for `resume()` being a no-op after `pause(true)` has already landed on master via #3974. This PR is now test-only: it adds a regression test in `tests/pause.test.ts` covering the specific `pause(true)` → `resume()` → process new job scenario, which is not currently exercised in that file.

Refs #3971 (the original bug this defends against regressing).

### How
Adds a single test to the existing Pause describe block that:
1. Starts a worker processing a job
2. Calls `worker.pause(true)` (doNotWaitActive)
3. Asserts `isPaused()` is true
4. Calls `worker.resume()`
5. Asserts `isPaused()` is false
6. Enqueues another job and verifies it completes

### Additional Notes
- The worker.ts changes from the original PR are now subsumed by #3974 (which landed first with a superset of the fix), so only the test file remains in this PR's diff.
- Close if the coverage is considered redundant with the tests added in #3974 to `tests/worker.test.ts`.

### DIFF
--- a/tests/pause.test.ts
+++ b/tests/pause.test.ts
@@ -411,4 +411,47 @@ describe('Pause', () => {
       await worker!.close();
     });
   });
+
+  it('should resume processing after pause(true) without waiting for active jobs', async () => {
+    let jobProcessedAfterResume = false;
+
+    const worker = new Worker(
+      queueName,
+      async () => {
+        await delay(100);
+      },
+      { connection, prefix },
+    );
+
+    await worker.waitUntilReady();
+
+    // Add a job so the worker is processing
+    await queue.add('test', { foo: 'bar' });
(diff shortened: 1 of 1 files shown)


maruthang:
Heads up @manast — after rebasing, I realized #3974 landed a superset of the worker.ts fix I had here. I have stripped the worker.ts changes and kept only the regression test in `tests/pause.test.ts` covering `pause(true)` → `resume()` → process new job, which I did not see covered in that file. Happy to close if you consider it redundant with the tests added in #3974.

roggervalf:
hi @maruthang I think the test in the pr is covering this case already, should be fine to close this pr

PR #3992, opened 2026-04-11
fix(queue): filter out undefined values from getJobs results

### LINKED ISSUE #3767: [Bug]: `queue.getActive()` can sometimes contain `undefined`

maruthang:
### Why

Closes #3767

`queue.getActive()` (and other `getJobs`-based methods) could return arrays containing `undefined` values. The original implementation issued a separate range call followed by a `Promise.all` of `Job.fromId` lookups; if a job was auto-removed (e.g. by `removeOnComplete`/`removeOnFail` or another getter) between those two steps, `Job.fromId` returned `undefined` and the unsafe `as Promise<JobBase>` cast silently passed it through to the caller.

### How

Resolve the race atomically inside Redis instead of patching it up in JavaScript:

- **`src/commands/getRangedJobs-1.lua` (new)**: A new Lua command that, in a single round-trip, performs the range over the requested state lists/sets and the `HGETALL` for each returned id. Because both operations run inside the same script, jobs cannot be auto-removed between them, so the requested pagination range is preserved. If a hash genuinely no longer exists, that slot is returned as an empty array and skipped on the TS side rather than producing an `undefined` Job. The script also enforces a hard per-type cap of 1000 ids as defense-in-depth so an unbounded range cannot block Redis.
- **`src/classes/scripts.ts`**: New `getRangedJobs` wrapper that builds the script args, validates the response shape (typed as `(string[] | string[][])[]` with an even-length check that throws a clear error on malformed output), deduplicates ids across types (matching the `getRanges` behaviour for `'waiting'` -> `'wait'`+`'paused'`), and returns `{ id, data }[]` for the caller. Enforces the `MAX_RANGED_JOBS = 1000` cap on the TS side before invoking the script so callers fail fast with a clear error instead of silently truncated results.
- **`src/classes/queue-getters.ts`**: `getJobs` now calls `scripts.getRangedJobs` and constructs `Job` instances via `Job.fromJSON(this, data, id)`, removing the unsafe cast and the per-id `Job.fromId` round-trips entirely.
- **`tests/getters.test.ts`**: Added a regression test that simulates the race by deleting a job hash while the id remains in the waiting list, and verifies the resulting array contains only valid `Job` instances and preserves the surrounding range.

### Additional Notes

- Non-breaking for normal callers: `getJobs` continues to return `Job[]`. Callers that previously received `undefined` slots (and may have been silently ignoring them) now simply receive a shorter array containing only valid `Job` instances.
- Performance: `getJobs` now issues a single `EVALSHA` instead of `1 + N` round-trips (range + N `HGETALL`s).
- The 1000-job per-call cap matches typical bullmq batch sizes and prevents accidental long-running Lua scripts when callers pass the default `end = -1` against a very large state set. Callers needing more should paginate.

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -490,9 +490,11 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs = await Promise.all(
+      jobIds.map(jobId => this.Job.fromId(this, jobId)),
     );
+
+    return jobs.filter(job => job !== undefined) as JobBase[];
   }
 
   /**
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -488,10 +488,6 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   ): Promise<JobBase[]> {
     const currentTypes = this.sanitizeJobTypes(types);
 
-    // Range + fetch atomically in a single Lua script so jobs cannot be
-    // auto-removed between the range call and the subsequent HGETALL.
-    // Filtering out undefined entries after the fact would silently shrink
-    // the result and break the pagination range the caller requested.
     const entries = await this.scripts.getRangedJobs(
       currentTypes,
       start,
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -488,13 +488,20 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   ): Promise<JobBase[]> {
     const currentTypes = this.sanitizeJobTypes(types);
 
-    const jobIds = await this.getRanges(currentTypes, start, end, asc);
-
-    const jobs = await Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId)),
+    // Range + fetch atomically in a single Lua script so jobs cannot be
+    // auto-removed between the range call and the subsequent HGETALL.
+    // Filtering out undefined entries after the fact would silently shrink
+    // the result and break the pagination range the caller requested.
+    const entries = await this.scripts.getRangedJobs(
+      currentTypes,
+      start,
+      end,
+      asc,
     );
 
-    return jobs.filter(job => job !== undefined) as JobBase[];
+    return entries.map(
+      ({ id, data }) => this.Job.fromJSON(this, data, id) as JobBase,
+    );
   }
 
   /**
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -507,9 +507,11 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs = await Promise.all(
+      jobIds.map(jobId => this.Job.fromId(this, jobId)),
     );
+
+    return jobs.filter(job => job !== undefined) as JobBase[];
   }
 
   /**
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -505,13 +505,20 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   ): Promise<JobBase[]> {
     const currentTypes = this.sanitizeJobTypes(types);
 
-    const jobIds = await this.getRanges(currentTypes, start, end, asc);
-
-    const jobs = await Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId)),
+    // Range + fetch atomically in a single Lua script so jobs cannot be
+    // auto-removed between the range call and the subsequent HGETALL.
+    // Filtering out undefined entries after the fact would silently shrink
+    // the result and break the pagination range the caller requested.
+    const entries = await this.scripts.getRangedJobs(
+      currentTypes,
+      start,
+      end,
+      asc,
     );
 
-    return jobs.filter(job => job !== undefined) as JobBase[];
+    return entries.map(
+      ({ id, data }) => this.Job.fromJSON(this, data, id) as JobBase,
+    );
   }
 
   /**
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -505,10 +505,6 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   ): Promise<JobBase[]> {
     const currentTypes = this.sanitizeJobTypes(types);
 
-    // Range + fetch atomically in a single Lua script so jobs cannot be
-    // auto-removed between the range call and the subsequent HGETALL.
-    // Filtering out undefined entries after the fact would silently shrink
-    // the result and break the pagination range the caller requested.
     const entries = await this.scripts.getRangedJobs(
       currentTypes,
       start,
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -890,6 +890,61 @@ export class Scripts {
     return await this.execCommand(client, 'getRanges', args);
   }
 
+  /**
+   * Atomically range over the given state lists/sets and fetch the raw job
+   * hashes in a single round-trip. This guarantees that jobs cannot be
+   * auto-removed between the range operation and the subsequent fetch,
+   * preserving the requested pagination range.
+   *
+   * Note: currently migrated for the common `getJobs` path. Other getters
+   * (e.g. `getDependencies`) can be migrated to the same pattern.
+   */
+  async getRangedJobs(
+    types: JobType[],
+    start = 0,
+    end = 1,
+    asc = false,
+  ): Promise<{ id: string; data: JobJsonRaw }[]> {
+    const client = await this.queue.client;
+    const args = this.getRangesArgs(types, start, end, asc);
+
+    const responses: [string[], string[][]][] | string[][] =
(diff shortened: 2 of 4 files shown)


src/classes/queue-getters.ts
```diff
@@ -490,9 +490,11 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
     const jobIds = await this.getRanges(currentTypes, start, end, asc);
 
-    return Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId) as Promise<JobBase>),
+    const jobs = await Promise.all(
+      jobIds.map(jobId => this.Job.fromId(this, jobId)),
     );
+
+    return jobs.filter(job => job !== undefined) as JobBase[];
```
manast: The thing here is that the ranges will not be respected, thats why the correct solution is not to filter, but to return the jobs in an atomic lua script that does both the range and fetch the jobs before they get auto removed.

maruthang: You are right @manast — filtering shrinks the returned set and breaks the range contract. I will rework this as an atomic Lua script that performs the range and job fetch together, so ranges are respected even when jobs are auto-removed between the ZRANGE and HGETALL calls. Will push an update.

src/classes/queue-getters.ts
```diff
@@ -488,13 +488,20 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   ): Promise<JobBase[]> {
     const currentTypes = this.sanitizeJobTypes(types);
 
-    const jobIds = await this.getRanges(currentTypes, start, end, asc);
-
-    const jobs = await Promise.all(
-      jobIds.map(jobId => this.Job.fromId(this, jobId)),
+    // Range + fetch atomically in a single Lua script so jobs cannot be
```
manast: This comment is not necessary.

src/classes/queue-getters.ts
```diff
@@ -488,10 +488,19 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
...
+    // auto-removed between the range call and the subsequent HGETALL.
+    // Filtering out undefined entries after the fact would silently shrink
+    // the result and break the pagination range the caller requested.
+    const entries = await this.scripts.getRangedJobs(
+      currentTypes,
+      start,
+      end,
+      asc,
+    );
```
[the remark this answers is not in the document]
maruthang: Updated the PR description to accurately describe the atomic-Lua approach (new `getRangedJobs` Lua command + `Job.fromJSON` in `getJobs`) and the deduplication / shape-validation behaviour. Thanks!

src/commands/getRangedJobs-1.lua
```diff
@@ -0,0 +1,100 @@
...
+local rangeEnd = tonumber(ARGV[2])
+local asc = ARGV[3]
+local results = {}
+
+local function fetchJobs(ids)
+  local jobs = {}
+  for i = 1, #ids do
+    jobs[i] = rcall("HGETALL", prefix .. ids[i])
+  end
+  return jobs
```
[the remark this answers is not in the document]
maruthang: Good catch. Added a hard `MAX_RANGED_JOBS = 1000` cap, enforced on the TS side before calling the script (so callers fail fast with a clear error on bounded ranges over the cap) and as a defense-in-depth truncation inside the Lua script (which also covers the unbounded `end = -1` case).

src/classes/scripts.ts
```diff
@@ -890,6 +890,61 @@ export class Scripts {
...
+    // deduplicate by id (matching the previous getRanges behaviour when
+    // 'waiting' expands to 'wait'+'paused').
+    const seen = new Set<string>();
+    const results: { id: string; data: JobJsonRaw }[] = [];
+
+    const flat = responses as unknown as any[];
+    for (let i = 0; i < flat.length; i += 2) {
+      const ids: string[] = flat[i] || [];
+      const rawJobs: string[][] = flat[i + 1] || [];
+
```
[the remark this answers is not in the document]
maruthang: Replaced the `as unknown as any[]` cast with a properly typed `(string[] | string[][])[]` response and added an even-length validation that throws a clear error if the script ever returns a malformed shape.

maruthang:
Friendly ping @manast — pushed the atomic-Lua `getRangedJobs` rework as discussed (commit 75365921), with a hard `MAX_RANGED_JOBS = 1000` cap (TS-side fail-fast + Lua-side defense), a properly typed response shape (`(string[] | string[][])[]`) replacing the `as unknown as any[]` cast, and even-length validation. All copilot follow-ups addressed in inline replies. CI green. Ready for re-review whenever convenient.

src/classes/scripts.ts
```diff
@@ -890,6 +899,87 @@ export class Scripts {
...
+   */
+  async getRangedJobs(
+    types: JobType[],
+    start = 0,
+    end = 1,
+    asc = false,
+  ): Promise<{ id: string; data: JobJsonRaw }[]> {
+    if (start >= 0 && end >= 0) {
+      const requested = end - start + 1;
+      if (requested > Scripts.MAX_RANGED_JOBS) {
```
manast: The issue here is that this introduces a breaking change for users already getting ranges larger than the CAP. A better solution would be to break down into smaller calls (smaller than the 1k cap) at the typescript level and then return the concatenated results. That will not hang Redis and we can still be backwards compatible.

manast (commented):
LGTM, just a small comment.

maruthang:
Hi @manast — gentle ping: pushed the atomic-Lua `getRangedJobs` rework in `75365921` per your last comment, with:

- hard `MAX_RANGED_JOBS = 1000` cap (TS-side fail-fast + Lua-side defense)
- properly typed response shape (`(string[] | string[][])[]`) replacing the `as unknown as any[]` cast
- even-length validation

CI green, all Copilot follow-ups answered inline. Ready for re-review whenever you have a moment. Thanks!

src/classes/scripts.ts
```diff
@@ -890,6 +899,87 @@ export class Scripts {
...
+
+      for (let j = 0; j < ids.length; j++) {
+        const id = ids[j];
+        if (seen.has(id)) {
+          continue;
+        }
+        seen.add(id);
+
+        const raw = rawJobs[j] || [];
+        // Skip jobs that no longer have a hash (e.g. concurrently removed).
```
manast: Is this correct? the whole point of the new lua script is that it should be impossible to no have a hash for every job (unless there is a bug in bullmq or the hash has been removed externally of course, but I do not think we should are about that, or should we?)

src/classes/scripts.ts
manast: this is still an issue, as it breaks backwards compatibility. The solution is to call the script in batches of max 1000 jobs and concatenate the results before returning.

manast (changes requested):

maruthang:
Thanks @manast — fully agree, the hard-cap-with-error was the wrong shape (it broke callers requesting > 1000 jobs). Reworked in `a44144d` to do exactly what you suggested: TS-side batching of size 1000 with concatenation, no caller-visible breaking change.

**What changed:**

- `Scripts.getRangedJobs` now iterates one type at a time and pages through that type in chunks of `MAX_RANGED_JOBS`, concatenating results until either the requested range is exhausted or the script returns a short read (which signals there are no more entries in that type's range). Default `end = -1` is handled the same way — we just keep advancing until the script returns fewer entries than the batch size.
- The Lua script itself is correspondingly simplified: it no longer carries its own cap or `truncate()` helper, since the TS wrapper now guarantees the range it receives is already bounded.
- The empty-hash `continue` fallback is dropped — you're right at #discussion_r3166716640 that within a single atomic Lua script, an id returned from `LRANGE`/`ZRANGE` is guaranteed to still have its hash when we `HGETALL` it. The fallback was dead code.

**Copilot follow-ups now also addressed:**

- #discussion_r3166742623 — the cap now describes "jobs per Lua invocation", which matches reality (each call is single-type, bounded).
- #discussion_r3166742655 — the default `end = -1` no longer hits Lua with an unbounded range; the TS layer drives the batch loop.
- #discussion_r3166742690 — `truncate(ids)` is gone entirely; `LRANGE`/`ZRANGE` is only ever invoked with a bounded `[batchStart, batchEnd]` window of size ≤ 1000.

Net diff `+95/-81`, typecheck clean. CI re-running. Let me know if a different batch size (or a different "exhausted" heuristic — e.g. relying on `LLEN`/`ZCARD` instead of short-read detection) would suit better.

src/classes/scripts.ts
maruthang: You're right — since the range and the HGETALLs run inside the same Lua script, a missing hash can't happen for a ranged id (only a bullmq bug or an external deletion could cause it, and I agree we shouldn't defend against that). The batching commit already dropped the skip logic on that basis, and I've now also removed the leftover test that simulated an externally-deleted hash — it was enforcing the old filtering behaviour and was the CI failure on this branch.

src/classes/scripts.ts
maruthang: Done in a44144d9 (pushed shortly after your comment) — the hard cap is gone; `getRangedJobs` now pages each type in chunks of 1000 at the TS level and concatenates, so callers with ranges larger than 1000 keep working unchanged. Follow-up e79e4204 removes the stale missing-hash test per the other thread, which was what was failing CI. Ready for another look whenever convenient.

maruthang:
CI note: the single remaining red job (`testing node@lts/*, dragonflydb@latest`) fails on `job_scheduler.test.ts > when job schedulers are upserted in quick succession > should create only one job scheduler and one delayed job` — a test this PR doesn't touch, with 816/817 tests passing. The same test/job signature is currently failing on other unrelated PRs too (e.g. two PRs with completely disjoint diffs), while master's own dragonflydb job passes intermittently — so it looks like an upstream-flaky test on Dragonfly rather than anything introduced here.

roggervalf:
addressed in v6, handled by using a lua script

PR #3993, opened 2026-04-11
fix(job-scheduler): respect offset option in upsertJobScheduler (elixir) (rust)

### LINKED ISSUE #3705: [Bug]: Upsert Job Scheduler no longer respects offset

maruthang:
### Why

Closes #3705

Since PR #3446 (v5.58.8), the `offset` option in `upsertJobScheduler` with `every` pattern was being ignored. This is a regression where:

1. The Lua script `getJobSchedulerEveryNextMillis` did not apply the offset when computing the initial `nextMillis` (no `prevMillis` and no `startDate` case).
2. In `job-scheduler.ts`, the `newOffset` parameter was hardcoded to `null`, so the offset value was never passed to the Lua script even when provided by the user.

### How

- **`src/commands/includes/getJobSchedulerEveryNextMillis.lua`**: When no `prevMillis` and no `startDate`, compute the slot-aligned offset as `floor(now / every) * every + offset`, advancing by one period if that slot is already in the past.
- **`src/classes/job-scheduler.ts`** (line 91): Pass the actual `offset` value from `repeatOpts` instead of hardcoded `null`.
- **`tests/job_scheduler.test.ts`**: Added two regression tests verifying that the offset is respected during upsert for `every`-based schedulers.

### Additional Notes

- All 16 existing upsert-related job scheduler tests continue to pass.
- This is a minimal fix targeting only the regression introduced in v5.58.8.


manast (approved):
LGTM

maruthang:
Hi @manast — friendly nudge: approved with LGTM, 21/21 checks green. Ready to land whenever you have a moment. Thanks!

manast:
There is an interesting comment from copilot that we should sort out before merging.

maruthang:
Thanks @manast — addressed the remaining Copilot point at #discussion_r3109690041 in `a0f7661`. The catch-up branch in `getJobSchedulerEveryNextMillis.lua` was using `floor(now/every)*every + every + (offset or 0)`, which flattened a non-zero offset to (slot + every) whenever the offset-aligned slot was itself still in the future. The initial-scheduling branch right above it already does the right thing.

Pushed a small change so both branches use the same offset-aware alignment: compute `aligned = floor(now/every)*every + offset`, then advance by `+every` only when that aligned slot is in the past. A non-zero offset now survives missed iterations cleanly (the same scenario the new test in this PR exercises). The other two Copilot points on this PR (`offset || null` → `offset ?? null` at `job-scheduler.ts:91`, test name/timestamp at `tests/job_scheduler.test.ts:351`) are already addressed in earlier commits. CI re-running.

roggervalf:
looks like this pr handles an additional case when using offset, just need to fix the merge conflicts

maruthang:
Thanks @roggervalf — the branch is mergeable against master again (the conflicts were resolved in the merge commit `5e22e064`), CI is green, and the offset-zero preservation plus the catch-up alignment Copilot flagged are covered in `e11c4b27` / `a0f7661a` with tests. A refresh against the very latest master (vitest migration) is prepared locally and will be pushed shortly. Ready for review.

roggervalf:
hey @maruthang could you pls update your branch with master, if I do it, last commit will be associated with my user and my approval won't allow me to merge your change

maruthang:
@roggervalf done — branch updated with the latest master (merge commit `b9efdf105`, authored by me so your approval can merge it). Typecheck clean locally.

roggervalf (approved):

roggervalf:
@maruthang one of your changes were merged, pls update your pr with master so I can continue with this one

maruthang:
@roggervalf done — merged the latest master (`9f1d83978`), typecheck clean. Thanks!

roggervalf (approved):

manast (approved):

PR #3994, opened 2026-04-11
fix(job-scheduler): handle terminal job collision on remove and re-upsert

### LINKED ISSUE #3063: [Bug]: Removing and upserting a Job Scheduler in close succession does not create a new delayed job
### Version

v5.40.2

### Platform

NodeJS

### What happened?

When a Job Scheduler is removed with `Queue.removeJobScheduler()` and upserted again with `Queue.upsertJobScheduler()`, no new delayed task is created and the scheduled tasks never run.

Seems like during the upsert, it sees that there is a job with the expected id (based on milis?) that exist already and it doesn't not create a new task. But that task is in 'completed' state already, so it will never run again.

In the example below, we see the `next: 1738974600000` in the Job Scheduler, and a task with `id: 'repeat:test:1738974600000'` already existing.

I've also seen some other strange behavior with Job Schedulers that is harder to reproduce and report. Sometimes the delayed task will not run, and it's scheduled execution time is well in the past.

### How to reproduce.

    const queue = new Queue('queue')

    await queue.upsertJobScheduler('test', { every: 3 * 60 * 1000 }, { name: 'test-job', data: { foo: 'bar' } })

    // A while later

    // Deletes currently delayed 'test-job' job
    await queue.removeJobScheduler('test')

    // Does not create the new 'test-job'
    await queue.upsertJobScheduler('test', { every: 3 * 60 * 1000 }, { name: 'test-job', data: { foo: 'bar' } })

    const schedulers = await queue.getJobSchedulers()
    const delayed = await queue.getDelayed()
    const completed = await queue.getCompleted()

    console.log({schedulers, delayed, completed})

### Relevant log output

```shell
{
        schedulers: [
        {
            key: 'test',
            name: 'test-job',
            endDate: null,
            tz: null,
            pattern: null,
            every: '1800000',
            template: [Object],
            next: 1738974600000
        }
        ],
        delayed: [],
        completed: [
        Job {
            queue: [Queue],
            name: 'test-job',
            data: [Object],
            opts: [Object],
            id: 'repeat:test:1738975500000',
            progress: 0,
            returnvalue: null,
            stacktrace: [],
            delay: 0,
            priority: 0,
            attemptsStarted: 1,
            attemptsMade: 1,
            repeatJobKey: 'test',
            timestamp: 1738975641682,
            parentKey: undefined,
            parent: undefined,
            debounceId: undefined,
            deduplicationId: undefined,
            toKey: [Function: bound ],
            scripts: [Scripts],
            queueQualifiedName: 'bull:queue',
            finishedOn: 1738975643099,
            processedOn: 1738975641685,
            failedReason: undefined,
            nextRepeatableJobId: 'repeat:test:1738975680000'
        },
        Job {
            queue: [Queue],
            name: 'test-job',
            data: [Object],
            opts: [Object],
            id: 'repeat:test:1738974600000',
            progress: 0,
            returnvalue: null,
            stacktrace: [],
            delay: 0,
            priority: 0,
            attemptsStarted: 1,
            attemptsMade: 1,
            repeatJobKey: 'test',
            timestamp: 1738974891613,
            parentKey: undefined,
            parent: undefined,
            debounceId: undefined,
            deduplicationId: undefined,
            toKey: [Function: bound ],
            scripts: [Scripts],
            queueQualifiedName: 'bull:queue',
            finishedOn: 1738974892974,
            processedOn: 1738974891614,
            failedReason: undefined,
            nextRepeatableJobId: 'repeat:test:1738976400000'
        },
        . . .
        Job {
            queue: [Queue],
            name: 'test-job',
            data: [Object],
            opts: [Object],
            id: 'repeat:test:1738972800000',
            progress: 0,
            returnvalue: null,
            stacktrace: [],
            delay: 0,
            priority: 0,
            attemptsStarted: 1,
            attemptsMade: 1,
            repeatJobKey: 'test',
            timestamp: 1738973961543,
            parentKey: undefined,
            parent: undefined,
            debounceId: undefined,
            deduplicationId: undefined,
            toKey: [Function: bound ],
            scripts: [Scripts],
            queueQualifiedName: 'bull:queue',
            finishedOn: 1738973973988,
            processedOn: 1738973972090,
            failedReason: undefined,
            nextRepeatableJobId: 'repeat:test:1738974600000'
        }
        ]
    }
```

### Code of Conduct

maruthang:
### Why

Closes #3063

When a job scheduler is removed and immediately re-upserted with the same parameters, the new delayed job is not created. This happens because the generated job ID collides with the previously completed/failed job. The collision detection in `addJobScheduler-11.lua` finds the existing key via `EXISTS` but does not check whether it is in a terminal state, so it incorrectly skips job creation.

### How

In `src/commands/addJobScheduler-11.lua`, before falling through to the existing collision handling logic (advance to next slot or skip), we now check if the colliding job ID exists in the `completed` or `failed` sorted sets. If it does, we remove the terminal job entry and its associated keys, then proceed to create the new job normally at the same time slot.

A regression test was added in `tests/job_scheduler.test.ts` that:
1. Creates a scheduler with `every`
2. Processes and completes the job
3. Removes the scheduler
4. Re-upserts the same scheduler
5. Asserts that a new waiting/delayed job is created successfully

### Additional Notes

- Only terminal states (completed/failed) are cleaned up; active jobs still trigger the existing collision avoidance behavior.
- The fix is minimal and scoped to the Lua script, with no changes to the TypeScript API surface.


manast (dismissed):
LGTM

maruthang:
Friendly ping @manast — rebased onto latest `master` today (CI green) so the previous approval still stands. Ready to merge whenever convenient.

maruthang:
Friendly ping @manast — the only red check was `testing node@lts/-2, redis@7-alpine` timing out in an unrelated parent/child waiting-event test (`tests/events.test.ts > when one job is a parent > emits waiting-children and waiting event`, 10s default timeout). That one matrix slice hit a flake while all 19 other matrix jobs (node@lts/*, node@lts/-1, redis@6, valkey@8, dragonflydb, bun, elixir, php, python) were green. The change here is scheduler Lua only, so it can't affect that event path.

Pushed an empty commit to retrigger CI. Ready to merge once it clears.

src/commands/addJobScheduler-11.lua
```diff
@@ -144,12 +144,22 @@ end
...
 local hasCollision = false
 if rcall("EXISTS", jobKey) == 1 then
-    if every then
+    -- Check if the existing job is in a terminal state (completed or failed)
+    -- If so, remove it and allow creating a new job with the same ID
+    local completedKey = prefixKey .. "completed"
+    local failedKey = prefixKey .. "failed"
+    if rcall("ZSCORE", completedKey, jobId) then
+        rcall("ZREM", completedKey, jobId)
+        removeJob(jobId, true, prefixKey, true --[[remove debounce key]] )
```
manast: I am not so happy with this approach actually, as the user would be surprised by the removal of a valid completed job that maybe performed some useful work that the user would like to keep track of.

maruthang:
Hi @manast — gentle follow-up after my 2026-04-20 ping. CI still green on the latest rebase (the one earlier red slice on `node@lts/-2, redis@7-alpine` was the unrelated `events.test.ts > emits waiting-children and waiting event` flake hitting several PRs that day). Scheduler-Lua-only change, prior LGTM still applicable. Happy to land it whenever convenient. Thanks!

manast:
I and copilot left several comments that need to be resolved before we can merge.

maruthang:
Thanks @manast — totally fair point on #discussion_r3125044752. Silently removing a completed job on re-upsert can absolutely surprise a user who was keeping it around for tracking. Reworked in `cb9c26d`:

**Production change** (`addJobScheduler-11.lua`):
- The completed-terminal branch no longer removes the prior job. Instead it uses the same "try the next slot" strategy that already exists for non-terminal `every` collisions further down. If the next slot is also occupied, it returns `SchedulerJobSlotsBusy` (-11) so the caller can decide what to do.
- If no `every` interval is available to advance into (e.g. cron-only pattern), we surface the same `SchedulerJobSlotsBusy` error rather than silently dropping the completed job.
- The failed-terminal branch keeps the existing remove-and-replace behavior, since a failed terminal job is normally expected to be replaced when the scheduler is upserted again.

**Test rework** (`tests/job_scheduler.test.ts`) covering all five Copilot points on this PR:
- Added `await worker.waitUntilReady()` before kicking off the run (#discussion_r3125061397).
- Switched to `worker.once('completed', ...)` (#discussion_r3125061327).
- Asserted `removeJobScheduler` drained both waiting and delayed entries before the re-upsert (#discussion_r3125061263).
- Captured the completed job's id, asserted it survives the re-upsert, and asserted the newly scheduled job uses a different id (#discussion_r3125061369).
- Renamed the test to reflect the new "preserves completed, advances to next slot" semantics rather than "removes and replaces".

CI re-running. Let me know if you'd prefer different semantics for the failed branch as well — happy to extend the same preserve-and-advance behavior there if that's the desired contract.

docs/gitbook/bullmq-pro/changelog.md
```diff
@@ -1,1150 +1,1008 @@
 # [7.46.0](https://github.com/taskforcesh/bullmq-pro/compare/v7.45.1...v7.46.0) (2026-05-21)
 
-
 ### Features
 
-* add support for both group concurrency and group rate limit ([#428](https://github.com/taskforcesh/bullmq-pro/issues/428)) ([ede5aa3](https://github.com/taskforcesh/bullmq-pro/commit/ede5aa385f6c7df1edcfccb15e0e064a1798f604))
+- add support for both group concurrency and group rate limit ([#428](https://github.com/taskforcesh/bullmq-pro/issues/428)) ([ede5aa3](https://github.com/taskforcesh/bullmq-pro/commit/ede5aa385f6c7df1edcfccb15e0e064a1798f604))
```
roggervalf, line 5: this file does not need to be changed

src/commands/addJobScheduler-11.lua
```diff
@@ -144,12 +144,45 @@ end
...
+            else
+                return -11 -- SchedulerJobSlotsBusy
+            end
+        else
+            -- No interval available to advance to: surface the
+            -- collision instead of dropping the completed job.
+            return -11 -- SchedulerJobSlotsBusy
+        end
+    elseif rcall("ZSCORE", failedKey, jobId) then
+        rcall("ZREM", failedKey, jobId)
```
roggervalf, line 183: no need to remove the job even on failed state either as it would serve for debugging

maruthang:
This now conflicts with master because #3063 reworked the same collision block in `addJobScheduler-11.lua` — master now does a bounded forward scan (up to 32 slots) to find a free slot on any `every` collision, whereas this PR distinguishes the terminal states: preserve a *completed* job (advance/return busy rather than drop it) and remove-and-replace a *failed* one.

Since master's new baseline treats all collisions uniformly (it would scan past a failed job rather than replacing it), reconciling the two is really a behavior question: would you like this PR rebased to layer the completed-vs-failed distinction on top of #3063's forward-scan (i.e. still scan for a free slot, but replace a failed job in place and never silently drop a completed one), or do you consider #3063 to already cover the #3063/#3994 case and prefer to close this? Happy to implement whichever direction you'd like — just wanted your call on the intended semantics before pushing a merge, rather than guessing at the scheduler logic.

PR #3996, opened 2026-04-12
fix(python): getFailed() incorrectly returns completed jobs

yogeshwaran-c:
## Summary
- `Queue.getFailed()` in the Python SDK was calling `self.getJobs(['completed'], ...)` instead of `self.getJobs(['failed'], ...)`, causing it to return completed jobs instead of failed ones.
- This is a copy-paste bug — `getCompleted()` directly above uses `['completed']` correctly, but `getFailed()` was never updated to use `['failed']`.

## Test plan
- [ ] Call `queue.getFailed()` and verify it returns only jobs in the `failed` state
- [ ] Call `queue.getCompleted()` and verify it still returns only completed jobs
- [ ] Confirm no other `get<State>()` methods have the same issue

### DIFF
--- a/python/bullmq/queue.py
+++ b/python/bullmq/queue.py
@@ -364,7 +364,7 @@ class Queue(EventEmitter):
         return self.getJobs(['delayed'], start, end, True)
 
     def getFailed(self, start = 0, end=-1):
-        return self.getJobs(['completed'], start, end, False)
+        return self.getJobs(['failed'], start, end, False)
 
     def getPrioritized(self, start = 0, end=-1):
         return self.getJobs(['prioritized'], start, end, True)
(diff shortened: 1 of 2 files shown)


roggervalf:
Hi @yogeshwaran-c, this pr needs a test case to prevent any regression

yogeshwaran-c:
Added a regression test in `python/tests/queue_test.py` (`test_get_failed_returns_only_failed_jobs`) that creates one failing and one completing job, then asserts `getFailed()` returns only the failed one and `getCompleted()` returns only the completed one. Pushed in the follow-up commit.

manast (approved):
LGTM

manast:
good catch!

PR #3997, opened 2026-04-12
refactor(python): use isinstance() instead of type() == for type checks

yogeshwaran-c:
## Summary
- Replace all `type(x) == T` comparisons with `isinstance(x, T)` across the Python SDK
- Covers `scripts.py` (12 instances), `job.py` (3 instances), and `backoffs.py` (1 instance)
- Also fixes one `type(result) != int` to `not isinstance(result, int)`

## Why
Using `isinstance()` is the recommended Pythonic approach for type checking. Unlike `type() ==`, it correctly handles subclass instances and is consistent with PEP 8 guidelines.

## Test plan
- [ ] Verify existing Python SDK tests pass
- [ ] No behavioral change expected since `bool` is a subclass of `int`, but all usage sites are checking Redis command results where only exact `int` types are returned

### DIFF
--- a/python/bullmq/backoffs.py
+++ b/python/bullmq/backoffs.py
@@ -13,7 +13,7 @@ class Backoffs:
 
     @staticmethod
     def normalize(backoff: Union[int, BackoffOptions]):
-        if type(backoff) == int and math.isfinite(backoff):
+        if isinstance(backoff, int) and math.isfinite(backoff):
             return {
                 "type": "fixed",
                 "delay": backoff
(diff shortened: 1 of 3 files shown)


manast:
There seems to be an issue with the python tests that are failing after this change

yogeshwaran-c:
Closing this — the test failures revealed that `type(x) == int` was actually intentional here. `isinstance(x, int)` returns `True` for `bool` values (since `bool` is a subclass of `int`), which changes behavior for paths like `Backoffs.normalize` and the Lua script result checks where a boolean result is possible and should not be treated as an int. The original checks were guarding against that. Not worth the risk for a stylistic change.

PR #4002, opened 2026-04-12
fix(types): add null to stacktrace type annotation in Job class

yogeshwaran-c:
## Summary

- The `stacktrace` property in the `Job` class is initialized to `null` but its type annotation is `string[]`, which is incorrect under strict null checks.
- Updated the type to `string[] | null` to accurately reflect that the property can be `null`.

## Test plan

- Verify the project compiles without type errors.
- Existing tests should pass without modification since this is a type-only change that matches the existing runtime behavior.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -82,7 +82,7 @@ export class Job<
    * Stacktrace for the error (for failed jobs).
    * @defaultValue null
    */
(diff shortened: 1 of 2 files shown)


manast (commented):
LGTM

roggervalf (approved):

PR #4003, opened 2026-04-12
feat(python): add return type hints to Job methods

yogeshwaran-c:
## Summary
- Add missing return type hints to several public methods in `python/bullmq/job.py`
- Methods annotated: `fromJSON`, `fromId`, `getChildrenValues`, `moveToWaitingChildren`, `isInList`, `log`, and `addJobLog`
- No functional changes; annotation-only improvement for better IDE support and static analysis

## Test plan
- [ ] Verify existing Python tests pass without regressions
- [ ] Confirm type hints are consistent with actual return values in `scripts.py`

### DIFF
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -186,8 +186,8 @@ class Job:
 
         return score is not None
 
-    def isInList(self, list_name: str) -> bool:
-        return self.scripts.isJobInList(self.scripts.toKey(list_name), self.id)
+    async def isInList(self, list_name: str) -> bool:
+        return await self.scripts.isJobInList(self.scripts.toKey(list_name), self.id)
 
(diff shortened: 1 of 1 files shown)


manast (commented):
LGTM

yogeshwaran-c:
Addressed the review comments:

- `isInList`, `log`, `moveToWaitingChildren` are now `async def` and `await` the underlying `scripts.*` calls, so the `-> bool` / `-> int` / `-> bool | None` annotations reflect the actual resolved values. Existing callers (tests and `isWaiting`) were already using `await` on these, so no caller changes were needed.
- `getChildrenValues` is now typed as `dict[str, Any]` to match the JSON-decoded output of `parse_json_string_values`.

manast (approved):
LGTM

PR #4004, opened 2026-04-12
feat(queue): add return type hints to Queue methods [python]

yogeshwaran-c:
## Summary
- Add missing return type hints to public and private methods in `python/bullmq/queue.py`
- Annotated async methods: `add`, `addBulk`, `isPaused`, `get_workers`, `get_workers_count`, `getJobLogs`, `obliterate`, `drain`, `retryJobs`, `promoteJobs`, `getJobCountByTypes`, `getJobCounts`, `getCountsPerPriority`, `clean`, `getJobs`, `close`
- Annotated sync methods: `toKey`, `sanitizeJobTypes`, `_parse_client_list`, `_base_get_clients`
- Use built-in PEP 585 generics (`list[...]`, `dict[...]`) for annotations rather than importing `List`/`Dict` from `typing` (no new `typing` imports added; existing `Union` import is unchanged)
- Bundled small fixes spotted during review:
  - `getJobs`: switch from `asyncio.wait` to `asyncio.gather(..., return_exceptions=True)` so results preserve the order of `job_ids` (`asyncio.wait` returns an unordered set, which could shuffle the returned jobs at runtime)
  - `add`: drop the redundant `**(opts or {})` fallback now that `opts` is normalized to `{}` one line above
  - Avoid mutable default arguments and tighten a few related signatures per reviewer feedback

## Test plan
- [ ] Verify type hints are correct by reviewing method return values
- [ ] Run existing Python test suite to ensure no regressions
- [ ] Confirm `getJobs` returns jobs in the order requested by `start`/`end`/`asc`

### DIFF
--- a/python/bullmq/queue.py
+++ b/python/bullmq/queue.py
@@ -32,7 +32,7 @@ class Queue(EventEmitter):
     def toKey(self, type: str) -> str:
         return self.scripts.queue_keys.toKey(self.name, type)
 
-    async def add(self, name: str, data, opts: JobOptions = {}) -> Job:
+    async def add(self, name: str, data, opts: JobOptions | None = None) -> Job:
         """
         Adds a new job to the queue.
 
@@ -40,6 +40,7 @@ class Queue(EventEmitter):
         @param data: Arbitrary data to append to the job.
         @param opts: Job options that affects how the job is going to be processed.
         """
+        opts = opts if opts is not None else {}
         merged_opts = {**self.jobsOpts, **(opts or {})}
 
         job = Job(self, name, data, merged_opts)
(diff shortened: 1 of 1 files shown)


yogeshwaran-c:
Addressed the `sum` shadowing — renamed to `total` in `getJobCountByTypes`.

On the `getJobs` `asyncio.wait` comment: that is a pre-existing correctness concern in the reorder loop (`jobs_len` is the pre-filter length, so the inner `range(index, jobs_len)` can over-index after `None`s are filtered out). I left the body unchanged here since this PR is scoped to return-type hints and a rewrite of that pairing logic would be a separate fix — happy to open a follow-up PR if you'd like.

manast:
it is quite incredible that copilot keeps writing comments after so many rounds...

yogeshwaran-c:
Thanks for the review! Addressed the three Copilot comments:

1. **`getJobs` ordering change (line 396)** — updated the PR description to call out that the switch from `asyncio.wait` to `asyncio.gather(..., return_exceptions=True)` is a small runtime ordering bug fix bundled with the type hints, not just an annotation change. Inline rationale also lives in the comment above the `gather` call.

2. **Redundant `**(opts or {})` in `add` (line 48)** — fixed in e327c73. Since `opts` is normalized to `{}` on the previous line, the fallback is unnecessary; now reads `merged_opts = {**self.jobsOpts, **opts}`.

3. **`List` import claim (line 40 / description)** — corrected the PR description. The file uses PEP 585 built-in generics (`list[...]`, `dict[...]`) and does not import `List` from `typing`; the description previously said the opposite. Updated to make this explicit.

PR #4006, opened 2026-04-12
fix(job): clear parent from opts JSON in removeChildDependency

### LINKED ISSUE #2833: [Bug]: Removed child dependency job still has parent set on next attempt / after delay
### Version

v5.20.0

### Platform

NodeJS

### What happened?

If a job calls `job.removeChildDependency()` and then delays or fails, its parent will be set on the next attempt. This will cause another call to `job.removeChildDependency()` to throw an exception, if the parent job has finished.

After `await job.removeChildDependency()`, `job.parent` is `undefined`. I'd expect it to be `undefined` on the next attempt, too.

To work around this, I need to (1) try...catch the call to `job.removeChildDependency`, or (2) ensure it doesn't get called twice by, for example, changing the job's data.

My use case for `removeChildDependency` is to let a parent job finish, if the child job has to be delayed for a long time for some reason (in my case, it's due to resource-specific rate limits).

### How to reproduce.

The following code will set up a queue where the parent job schedules two child jobs. The second child job (id: `c-456`) will remove itself from the parent and then delay itself. This allows the parent job to finish.

However, once the (now detached) child job runs again, its `job.parent` property will still be set to the parent job. Calling `job.removeChildDependency` would then fail, because the parent will have been removed.

This is demonstrated by the log output in the next section. For the second attempt of `c-456`, it still prints `(p: parent-1)`, even though the dependency was removed in the previous attempt.

```javascript
import { DelayedError, Queue, WaitingChildrenError, Worker } from 'bullmq';
import IORedis from 'ioredis';

const connection = new IORedis({ maxRetriesPerRequest: null });

const queue = new Queue('foobar', {
  connection,
  defaultJobOptions: {
    attempts: 10,
    removeOnComplete: true,
    removeOnFail: true,
  },
});

async function processParentJob(job, token) {
  let step = job.data.step;

  while (step !== 'finish') {
    switch (step) {
      case 'init': {
        const opts = { parent: { id: job.id, queue: job.queueQualifiedName } };
        await queue.add('child', {}, { ...opts, jobId: 'c-123' });
        await queue.add('child', { delay: 10000 }, { ...opts, jobId: 'c-456' });

        step = 'children';
        await job.updateData({ step });

        break;
      }

      case 'children': {
        const shouldWait = await job.moveToWaitingChildren(token);
        if (shouldWait) {
          throw new WaitingChildrenError();
        }

        step = 'finish';
        await job.updateData({ step });
        return;
      }
    }
  }
}

async function processChildJob(job, token) {
  await wait(500);

  if (job.data.delay) {
    // Without this, the next attempt will error when it tries to remove itself
    await job.updateData({ delay: 0 });

    await job.removeChildDependency();

    await job.moveToDelayed(Date.now() + job.data.delay, token);
    throw new DelayedError();
  }
}

const worker = new Worker(
  'foobar',
  async (job, token) => {
    switch (job.name) {
      case 'parent':
        await processParentJob(job, token);
        break;

      case 'child':
        await processChildJob(job, token);
        break;
    }
  },
  { connection },
);

worker.on('active', (job) =>
  log('%s active attempt %s', j(job), job.attemptsStarted),
);

worker.on('completed', (job) => log('%s completed', j(job)));
worker.on('failed', (job, err) => log('%s failed: %s', j(job), err));

async function main() {
  await queue.add('parent', { step: 'init' }, { jobId: 'parent-1' });
}

main();

const j = (job) => {
  const s = `${job.name} ${job.id}`;

  if (!job.parent) {
    return s;
  }

  return `${s} (p: ${job.parent.id})`;
};

const log = (s, ...args) => {
  const t = new Date().toISOString();
  console.log(`%s ${s}`, t, ...args);
};

const wait = (ms) => new Promise((r) => setTimeout(r, ms));
```

### Relevant log output

```shell
2024-10-16T12:02:58.500Z parent parent-1 active attempt 1
2024-10-16T12:02:58.504Z child c-123 (p: parent-1) active attempt 1
2024-10-16T12:02:59.014Z child c-123 (p: parent-1) completed
2024-10-16T12:02:59.015Z child c-456 (p: parent-1) active attempt 1
2024-10-16T12:02:59.524Z parent parent-1 active attempt 2
2024-10-16T12:02:59.527Z parent parent-1 completed

# In this attempt, c-456 should not have a parent
2024-10-16T12:02:59.528Z child c-456 (p: parent-1) active attempt 2
2024-10-16T12:03:00.033Z child c-456 (p: parent-1) completed
```

### Code of Conduct

maruthang:
Closes #2833

### Why

After calling `removeChildDependency()`, the child job's `parent` and `parentKey` hash fields were deleted from Redis, but the `parent` object nested inside the `opts` JSON (stored as a single hash field) was not cleared. On the next attempt or after a delay, `Job.fromJSON` re-derives `parent`/`parentKey` from `opts.parent`, resurrecting the parent reference that was supposed to be removed.

### How

- `src/commands/removeChildDependency-1.lua`: Decode the `opts` JSON with cjson, strip the `parent`, `fpof`, `rdof`, `idof`, and `cpof` fields, and re-encode it alongside the existing `HDEL` of the top-level `parent` / `parentKey` hash fields. This keeps the persisted state consistent so rehydration does not resurrect the parent link.
- `src/classes/job.ts`: Mirror the same cleanup on the in-memory `this.opts` so the local `Job` instance stays consistent with Redis without requiring a reload.
- `tests/flow.test.ts`: Regression test asserting that after `removeChildDependency()` and a reload, `parent`, `parentKey`, and `opts.parent` are all `undefined`.

### Additional Notes

The Lua change uses `cjson.decode` / `cjson.encode`, which is already relied upon elsewhere in the BullMQ Lua scripts, so no new script dependencies are introduced.

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -407,13 +407,14 @@ export class Job<
       job.returnvalue = getReturnValue(json.returnvalue);
     }
 
-    if (json.parentKey) {
-      job.parentKey = json.parentKey;
-    }
-
-    if (json.parent) {
-      job.parent = JSON.parse(json.parent);
-    }
+    // The job hash's `parentKey` and `parent` fields are the source of truth
+    // for the live parent link. `opts.parent` is preserved as a historical
+    // record of how the job was created, but the constructor's derivation of
+    // `this.parent`/`this.parentKey` from `opts.parent` must not leak back
+    // into the rehydrated state when those hash fields have been cleared
+    // (e.g. by `removeChildDependency`). Always assign from the hash.
+    job.parentKey = json.parentKey;
+    job.parent = json.parent ? JSON.parse(json.parent) : undefined;
 
     if (json.pb) {
       job.processedBy = json.pb;
--- a/tests/flow.test.ts
+++ b/tests/flow.test.ts
@@ -595,6 +595,55 @@ describe('flows', () => {
       });
     });
 
+    describe('when child is reloaded from redis after removing dependency', () => {
+      it('does not restore the parent reference (issue #2833)', async () => {
+        const flow = new FlowProducer({ connection, prefix });
+        const { children } = await flow.add({
+          name: 'parent',
+          data: {},
+          queueName,
+          children: [
+            {
+              queueName,
+              name: 'child0',
+              data: {},
+              opts: {},
+            },
+            {
+              queueName,
+              name: 'child1',
+              data: {},
+              opts: {},
+            },
+          ],
+        });
+
+        const childJob = children![0].job;
+
+        const relationshipIsBroken = await childJob.removeChildDependency();
+        expect(relationshipIsBroken).toBe(true);
+        expect(childJob.parent).toBeUndefined();
+        expect(childJob.parentKey).toBeUndefined();
+
+        // Reload the job from Redis to simulate the next attempt / move
+        // after a delay. Previously, the parent reference would be
+        // reconstructed from the serialized opts hash field; the live
+        // parent/parentKey state must instead be sourced from the hash
+        // fields, which `removeChildDependency` cleared.
+        const reloadedChild = await Job.fromId(queue, childJob.id!);
+
+        expect(reloadedChild).toBeDefined();
+        expect(reloadedChild!.parent).toBeUndefined();
+        expect(reloadedChild!.parentKey).toBeUndefined();
+        // `opts` is treated as immutable — the historical record of how
+        // the job was created remains intact, but it must not drive the
+        // live parent link.
+        expect((reloadedChild!.opts as any).parent).toBeDefined();
+
+        await flow.close();
+      });
+    });
+
     describe('when parent does not exist', () => {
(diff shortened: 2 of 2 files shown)


src/classes/job.ts
```diff
@@ -636,6 +636,17 @@ export class Job<
     if (childDependencyIsRemoved) {
       this.parent = undefined;
       this.parentKey = undefined;
+      // Keep in-memory opts consistent with the stored opts so that a
+      // subsequent attempt (e.g. after moveToDelayed or a retry) does not
+      // reconstruct the parent reference from opts when the job is
+      // rehydrated from Redis.
+      if (this.opts) {
```
roggervalf: options property are meant to be immutable, we might need to find the place where parent and parentKey are rehydrated from opts.parent instead to stop doing that

roggervalf:
addressed in v5.76.10

PR #4007, opened 2026-04-12
fix(worker): use scheduler registry to discriminate repeatable keys

### LINKED ISSUE #3828: [Bug]: Repeatable jobs with 5+ colon segments in key create duplicates

maruthang:
### Why

Closes #3828.

Repeatable jobs whose scheduler id contains 5 or more colon-separated segments (e.g. `tenant:region:env:service:queue`) were creating duplicate schedulers on every run.

Root cause: `worker.ts` used colon-segment count (`>= 5`) as the sole heuristic to classify a `repeatJobKey` as a "legacy repeatable key". Any modern scheduler id that happened to contain 5+ segments was misrouted to `repeat.updateRepeatableJob()` instead of `jobScheduler.upsertJobScheduler()`, resulting in a second scheduler entry being registered alongside the original.

### How

- Added `JobScheduler.isJobScheduler(id)` which performs a `ZSCORE` against the scheduler sorted set (`this.keys.repeat`) to authoritatively determine whether an id is a registered scheduler.
- Updated `worker.ts` so that:
  - Keys with fewer than 5 colon segments remain on the existing fast-path (no extra Redis round-trip), preserving current behavior and performance for the common case.
  - Keys with 5+ segments consult `isJobScheduler` — if the id is registered in the scheduler ZSET it is handled by `jobScheduler.upsertJobScheduler()`; otherwise it falls back to the legacy `repeat.updateRepeatableJob()` path, preserving backward compatibility for genuinely legacy keys.
- Added a regression test in `tests/job_scheduler.test.ts` covering scheduler ids with 5+ colon segments to confirm no duplicate scheduler entries are created.

### Additional Notes

- The extra `ZSCORE` is only issued for the rare 5+ segment case, so there is no measurable impact on the hot path.
- Backward compatibility with pre-existing legacy repeatable keys is maintained via the fallback branch.

### DIFF
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -352,11 +352,21 @@ export class JobScheduler extends QueueBase {
    * count is not safe because a user-provided jobSchedulerId may itself
    * contain 5+ colon segments, which would otherwise be misclassified as
    * a legacy repeatable key.
+   *
+   * We cannot use ZSCORE on the shared `repeat` sorted set because legacy
+   * repeatable jobs are stored in the same sorted set and would be reported
+   * as schedulers. Instead, we probe the per-id metadata hash (`repeat:<id>`)
+   * for the `ic` (iteration count) field, which is written exclusively by
+   * `storeJobScheduler` and is never set by the legacy `addRepeatableJob`
+   * flow.
    */
   async isJobScheduler(id: string): Promise<boolean> {
     const client = await this.client;
-    const score = await client.zscore(this.keys.repeat, id);
-    return score !== null;
+    const exists = await (client as any).hexists(
+      `${this.keys.repeat}:${id}`,
+      'ic',
+    );
+    return exists === 1;
   }
 
   async getScheduler<D = any>(
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -343,6 +343,22 @@ export class JobScheduler extends QueueBase {
     };
   }
 
+  /**
+   * Checks if a given id corresponds to a registered job scheduler.
+   *
+   * This is used to disambiguate between new job scheduler ids (which may
+   * contain any number of colon segments) and legacy repeatable job keys
+   * (which always contain 5+ colon segments). Relying purely on segment
+   * count is not safe because a user-provided jobSchedulerId may itself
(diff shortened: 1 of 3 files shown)


manast (dismissed):
LGTM

manast:
2 more comments and we are good to go here.

maruthang:
Addressed the remaining Copilot comments in 7f65e114:

- The two ZSCORE-vs-HEXISTS concerns on `JobScheduler.isJobScheduler` (and the corresponding PR-description line) are actually resolved on this branch already — the final implementation uses `HEXISTS ${repeat}:${id} ic`, not a ZSCORE against the shared `repeat` sorted set, so legacy repeatables cannot be misclassified. The JSDoc on `isJobScheduler` now documents that explicitly (commits 47c03ab, 6464ab4). I'll also update the PR body to match.
- The `worker.ts:918` inline comment was the stale part — it still said "verify against the scheduler sorted set." Updated it to describe the actual metadata-hash check (`repeat:<id>` + `ic` field) that `JobScheduler.isJobScheduler()` performs.

CI: the only red job was `testing node@lts/-2, redis@7-alpine` timing out in `tests/events.test.ts > when one job is a parent > emits waiting-children and waiting event` — same 10s-timeout flake that hit #3994 on the same SHA, entirely unrelated to this scheduler-registry change. The push above retriggers CI.

manast (approved):

PR #4008, opened 2026-04-12
fix(deps): upgrade cron-parser to v5.5.0

### LINKED ISSUE #3252: [Bug]: Problem with cron intervals

maruthang:
### Why

Closes #3252.

cron-parser v4.9.0 produced incorrect results for certain interval expressions (e.g. `*/N` step values and some range combinations), which caused BullMQ repeat jobs and job schedulers to compute wrong next-run times. The upstream project fixed this in the v5.x line.

### How

- Bumped `cron-parser` from `4.9.0` to `5.5.0` in `package.json` and refreshed `yarn.lock`.
- Adapted to the v5 API:
  - `parseExpression(...)` -> `CronExpressionParser.parse(...)` in `src/classes/repeat.ts` and `src/classes/job-scheduler.ts`.
  - `ParserOptions` -> `CronExpressionOptions` in `src/interfaces/repeat-options.ts`.
- v5 drops the `iterator` field from the options type, so the previous `Omit<ParserOptions, 'iterator'>` wrapper is simplified to extend `CronExpressionOptions` directly.

### Additional Notes

- No behavioral changes in BullMQ itself beyond delegating to the updated parser.
- The v5 API changes are internal to these three files; the public `RepeatOptions` surface remains compatible for consumers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "tsc:all": "tsc && tsc -p tsconfig-cjs.json"
   },
   "dependencies": {
-    "cron-parser": "4.9.0",
+    "cron-parser": "5.5.0",
     "ioredis": "5.10.1",
     "msgpackr": "1.11.5",
     "node-abort-controller": "3.1.1",
--- a/src/classes/job-scheduler.ts
+++ b/src/classes/job-scheduler.ts
@@ -1,4 +1,4 @@
-import { parseExpression } from 'cron-parser';
+import { CronExpressionParser } from 'cron-parser';
 import {
   JobSchedulerJson,
   JobSchedulerTemplateJson,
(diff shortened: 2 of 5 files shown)


roggervalf:
Hi @maruthang this is actually handled by https://github.com/taskforcesh/bullmq/pull/3255, but it's a breaking change so it will be included in v6

maruthang:
Thanks for the heads-up @roggervalf — I wasn't aware of #3255. Makes sense to hold the upgrade for v6 given the breaking change. Closing this out.

Issue #3505, opened 2025-10-18
[Bug]: removeOnComplete.age and removeOnFail.age doesn't work.


mena234:
### Version

5.57.0

### Platform

NodeJS

### What happened?

This code is supposed to remove the `completed` and `failed` tasks after 20 seconds, but this doesn't happen.

```js
await this.renewalQueue.add(
    "process-renewal", {
        userId: subscription.user_id,
        planId: subscription.plan_id,
        planName: subscription.plan_name,
        credits: subscription.plan_credits,
    }, {
        jobId: `renewal-${subscription.user_id}`, // Deterministic ID
        priority: 1, // High priority for renewals
        removeOnComplete: {
            age: 20,
        },
        removeOnFail: {
            age: 20,
        },
    }
)
```

But the instant removal works fine.

```js
{
  removeOnComplete: true,
  removeOnFail: true
}
```

### How to reproduce.

Set a `removeOnComplete.age` or a `removeOnFail.age` for a task.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

roggervalf:
hi @mena234, our removal process on that option works as soon as a job is completed or failed, that operation happen at that time so it's a best effort. Btw could you pls elaborate more on what's happening in your case, like what's is the result that you are facing

PR #4011, opened 2026-04-12
docs(job-options): clarify age-based removal is best-effort

maruthang:
### Why

Refs #3505

Users report that `removeOnComplete.age` and `removeOnFail.age` "don't work" when a queue goes idle — jobs older than `age` remain in Redis indefinitely. This is actually intentional: age-based eviction is evaluated only when another job finishes (inside `moveToFinished`), so it is best-effort rather than a background TTL. The current JSDoc on `KeepJobs`, `removeOnComplete`, and `removeOnFail` does not make this clear, which leads to repeated confusion and bug reports.

### How

Documentation-only clarifications — no runtime behavior changes:

- `src/types/keep-jobs.ts` — expanded JSDoc on the `KeepJobs` type and its `age` field to explain that age-based cleanup runs opportunistically during job completion, not on a timer.
- `src/interfaces/base-job-options.ts` — clarified JSDoc on `removeOnComplete` / `removeOnFail` to note the best-effort semantics and point users toward pairing with a scheduled cleanup (e.g. `queue.clean()`) if deterministic eviction is required.
- `src/interfaces/worker-options.ts` — mirrored the same clarification on the worker-level `removeOnComplete` / `removeOnFail` options so both entry points describe the behavior consistently.

### Additional Notes

- This PR does not change the Lua scripts (`removeJobsByMaxAge.lua`, `moveToFinished-*.lua`) — the current behavior is intentional and matches what the maintainers have described in prior discussions.
- Using "Refs #3505" rather than "Closes" — the maintainers can decide whether the issue should be closed once the docs land, or kept open to track a potential future enhancement (e.g. a periodic cleanup job).

### DIFF
--- a/src/interfaces/base-job-options.ts
+++ b/src/interfaces/base-job-options.ts
@@ -50,6 +50,11 @@ export interface DefaultJobOptions {
    * jobs to keep, or you can provide an object specifying max
    * age and/or count to keep. It overrides whatever setting is used in the worker.
    * Default behavior is to keep the job in the completed set.
+   *
+   * When using `age` or `count`, the eviction is evaluated on a
+   * best-effort basis every time a job finishes; BullMQ does not run a
+   * background timer, so aged jobs are only removed once another job
+   * completes after their expiration.
    */
   removeOnComplete?: boolean | number | KeepJobs;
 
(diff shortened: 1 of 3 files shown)


manast (approved):
LGTM

PR #4021, opened 2026-04-13
fix(job): make isCompleted/isFailed/isDelayed/isWaitingChildren/isActive async [python]

yogeshwaran-c:
## Summary

In `python/bullmq/job.py`, the methods `isCompleted`, `isFailed`, `isDelayed`, `isWaitingChildren`, and `isActive` were declared as synchronous (`def`) but internally called async methods (`isInZSet`, `isInList`) without `await`. This caused them to return coroutines rather than booleans, forcing callers to `await` them while the signatures suggested a direct boolean return.

This change makes these methods `async def`, adds `await` on the internal async calls, and annotates the `-> bool` return type, making them consistent with `isWaiting` (already declared `async def`).

## Changes

- `isCompleted`, `isFailed`, `isDelayed`, `isWaitingChildren` now `async def ... -> bool` and `await self.isInZSet(...)`.
- `isActive` now `async def ... -> bool` and `await self.isInList(...)`.

## Test plan

- [ ] Existing Python test suite still passes
- [ ] Calling `await job.isCompleted()` (and siblings) returns a `bool`

### DIFF
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -148,35 +148,35 @@ class Job:
         if not removed:
             raise Exception(f"Job {self.id} could not be removed because it is locked by another worker")
 
-    def isCompleted(self):
+    async def isCompleted(self) -> bool:
         """
         Returns true if the job has completed.
         """
-        return self.isInZSet('completed')
+        return await self.isInZSet('completed')
 
-    def isFailed(self):
+    async def isFailed(self) -> bool:
         """
         Returns true if the job has failed.
         """
-        return self.isInZSet('failed')
+        return await self.isInZSet('failed')
 
-    def isDelayed(self):
+    async def isDelayed(self) -> bool:
         """
         Returns true if the job is delayed.
         """
-        return self.isInZSet('delayed')
+        return await self.isInZSet('delayed')
 
-    def isWaitingChildren(self):
+    async def isWaitingChildren(self) -> bool:
         """
         Returns true if the job is waiting for children.
         """
-        return self.isInZSet('waiting-children')
+        return await self.isInZSet('waiting-children')
 
-    def isActive(self):
+    async def isActive(self) -> bool:
         """
         Returns true if the job is active.
         """
-        return self.isInList('active')
+        return await self.isInList('active')
 
     async def isWaiting(self):
         return ( await self.isInList('wait') or await self.isInList('paused'))
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -178,7 +178,7 @@ class Job:
         """
         return await self.isInList('active')
 
-    async def isWaiting(self):
+    async def isWaiting(self) -> bool:
         return ( await self.isInList('wait') or await self.isInList('paused'))
 
     async def isInZSet(self, set: str):
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -186,8 +186,8 @@ class Job:
 
         return score is not None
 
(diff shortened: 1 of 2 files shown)


manast:
This would imply a breaking change I think, so we need to mark it so that a major version is released for Python.

yogeshwaran-c:
Addressed the `isWaiting` annotation in a follow-up commit.

On the breaking-change concern: agreed, this is a breaking change for Python callers who were using `if job.isCompleted():` (which previously evaluated the coroutine as truthy) — they'll now need `if await job.isCompleted():`. Existing Python tests in this repo (`job_test.py`, `worker_test.py`) already use the `await` form so they remain green, but downstream users will need to update.

Let me know how you'd like this flagged — I can:

1. Reword the commit subject to `fix(python)!:` and add a `BREAKING CHANGE:` footer so the next Python release picks it up as a major, or
2. Retarget it to a Python-next branch if you prefer to hold breaking changes.

Also, on Copilot's test request: existing tests for `isCompleted`/`isFailed`/`isActive`/`isDelayed` already call `await` so they cover the async conversion. Happy to add a `test_is_waiting_children` specifically if useful.

manast:
The question regarding breaking change is: if users where calling this methods without await, they where getting always True right? so in a sense this is a bug fix If I am not mistaken.

yogeshwaran-c:
You're absolutely right — since the methods were `def` (not `async def`), calling them without `await` returned a coroutine object, which is always truthy in Python:

```python
if job.isCompleted():  # always True (coroutine is truthy)
    ...
```

So anyone relying on the un-awaited call was getting silently wrong results regardless of the job's actual state. This is a bug fix, not a breaking change — no correct code could have been depending on the old behavior.

Will also add the `isWaitingChildren` test that Copilot requested.

yogeshwaran-c:
You're absolutely right — since the methods were `def` (not `async def`), calling them without `await` returned a coroutine object, which is always truthy in Python:

```python
if job.isCompleted():  # always True (coroutine is truthy)
    ...
```

So anyone relying on the un-awaited call was getting silently wrong results regardless of the job's actual state. This is a bug fix, not a breaking change — no correct code could have been depending on the old behavior.

Will also add the `isWaitingChildren` test that Copilot requested.

yogeshwaran-c:
Pushed `aee138e` with an explicit regression test (`test_is_state_methods_return_false_when_awaited_for_pending_job`) that:

1. Adds a freshly-queued (waiting) job.
2. Asserts each of `isCompleted`, `isFailed`, `isDelayed`, `isWaitingChildren`, and `isActive` is now a coroutine function (`inspect.iscoroutinefunction(method)`).
3. Asserts each of them returns `False` when properly `await`ed on the waiting job.

This locks in the bug-fix semantics: previously these methods were `def` (returning a coroutine that was always truthy), so `if job.isCompleted():` was effectively `if <coroutine object>:` and reported `True` regardless of state. With the fix, calling the method without `await` now also surfaces a `RuntimeWarning: coroutine was never awaited`, making the misuse loud rather than silently wrong.

I confirmed the test parses cleanly with `python -m py_compile` and `ast.parse`. I couldn't run the full pytest suite locally (the Python deps aren't installed in this environment — `ModuleNotFoundError: No module named 'redis'`), but the existing `job_test.py` tests (`isCompleted/isFailed/isActive/isDelayed`) already exercise the awaited code paths and the new test follows the same setup pattern.

yogeshwaran-c:

manast (dismissed):

PR #4022, opened 2026-04-13
fix(python): use slice reversal instead of reverse() which returns None

yogeshwaran-c:
## Summary

Fixes a bug in `python/bullmq/scripts.py` where `list.reverse()` was used in a `+=` expression. In Python, `list.reverse()` reverses the list in-place and returns `None`, so `results += result.reverse()` would raise `TypeError: unsupported operand type(s) for +=: 'list' and 'NoneType'` whenever the `asc and commands[i] == "lrange"` branch was taken.

Replaced with slice reversal `result[::-1]`, which returns a new reversed list and can be concatenated with `+=`.

## Test plan

- [ ] Exercise `getRanges` with `asc=True` and an `lrange` command to confirm no TypeError is raised and results are appended in reversed order.

### DIFF
--- a/python/bullmq/scripts.py
+++ b/python/bullmq/scripts.py
@@ -247,7 +247,8 @@ class Scripts:
             result = response or []
 
             if asc and commands[i] == "lrange":
-                results+=result[::-1]
+                result.reverse()
+                results+=result
             else:
                 results+=result
 
--- a/python/bullmq/scripts.py
+++ b/python/bullmq/scripts.py
@@ -247,7 +247,7 @@ class Scripts:
             result = response or []
 
             if asc and commands[i] == "lrange":
-                results+=result.reverse()
+                results+=result[::-1]
             else:
                 results+=result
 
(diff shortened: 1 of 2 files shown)


python/bullmq/scripts.py
```diff
@@ -247,7 +247,7 @@ async def getRanges(self, types, start: int = 0, end: int = 1, asc: bool = False
             result = response or []
 
             if asc and commands[i] == "lrange":
-                results+=result.reverse()
+                results+=result[::-1]
```
roggervalf:
```suggestion
                result.reverse()
                results+=result
```
result.reverse() reverses in-place with no memory allocation, while result[::-1] creates a new list copy, requiring an extra allocation. Since we don't need to preserve the original order, reverse can be kept.

manast (approved):
LGTM

yogeshwaran-c:
Addressed all three review points:

- Switched to in-place `result.reverse()` followed by `results += result` to avoid the extra list allocation from slice-reversal (per @roggervalf).
- Added `test_get_waiting_returns_jobs_in_ascending_order` in `queue_test.py` which adds three jobs, calls `getWaiting()` (which takes the `asc=True` + `lrange` path in `getRanges`), and asserts they come back in insertion order.
- Fixed the test to generate its own unique queue name locally instead of relying on the module-level `queueName` (which `setUp` never mutates beyond a local shadow).

PR #4024, opened 2026-04-13
feat(python): add type hints to Timer class

yogeshwaran-c:
## Summary
- Added missing type hints to the `Timer` class in `python/bullmq/timer.py`.
- Annotated `callback` as `Callable[..., Any]`, `emit_callback` as `Callable[[str, Any], None]`, and `*args`/`**kwargs` as `Any`.
- Added `-> None` return annotations to `__init__`, `_job`, and `stop`.
- Imported `Any` and `Callable` from `typing`.

## Test plan
- [ ] Static type check passes (mypy/pyright)
- [ ] Existing Python test suite still passes

### DIFF
--- a/python/bullmq/timer.py
+++ b/python/bullmq/timer.py
@@ -1,11 +1,11 @@
 import asyncio
-from typing import Any, Callable
+from typing import Any, Awaitable, Callable
 
 # Credits: https://stackoverflow.com/questions/45419723/python-timer-with-asyncio-coroutine
 
 
 class Timer:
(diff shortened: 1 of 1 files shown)


manast (dismissed):

yogeshwaran-c:
Addressed both points:

- `interval` is now `float` (Worker passes `(lockDuration / 2) / 1000`, which is fractional, and `asyncio.sleep` takes a float anyway).
- `callback` is now `Callable[..., Awaitable[Any]]` so the type matches the `await callback(...)` call site in `_job`.

manast (approved):

PR #4026, opened 2026-04-13
docs(utils): add JSDoc to exported utility functions

yogeshwaran-c:
## Summary

Adds JSDoc comments to the previously undocumented exported utility functions in `src/utils/index.ts`, matching the short `@param` / `@returns` style already used by `lengthInUtf8Bytes` and `trace`.

Functions now documented include `tryCatch`, `isEmpty`, `array2obj`, `objectToFlatArray`, `delay`, `increaseMaxListeners`, `invertObject`, `isRedisInstance`, `isRedisCluster`, `decreaseMaxListeners`, `removeAllQueueData`, `getParentKey`, `isNotConnectionError`, `asyncSend`, `childSend`, `isRedisVersionLowerThan`, `parseObjectValues`, `errorToJSON`, `toString`, and `removeUndefinedFields`.

Documentation-only change — no behavioral or type changes.

## Test plan

- [x] `git diff` reviewed; only JSDoc comment additions
- [x] Existing JSDoc style preserved (concise `@param` / `@returns`)

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -20,6 +20,13 @@ import { DatabaseType } from '../types';
 
 export const errorObject: { [index: string]: any } = { value: null };
 
+/**
+ * Invokes a function safely, returning an error sentinel instead of throwing.
+ * @param fn - function to invoke
+ * @param ctx - `this` context to apply to the function
+ * @param args - arguments to pass to the function
+ * @returns the function's return value, or `errorObject` if it threw
+ */
 export function tryCatch(
   fn: (...args: any) => any,
   ctx: any,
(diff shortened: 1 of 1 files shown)


manast:
We should make sure this functions to not appear in BullMQ public API reference as these functions should all be private for use inside BullMQ and not polute our documentation.

yogeshwaran-c:
Closing — fair point, these `src/utils` helpers are intended as internal-only and shouldn't surface in the public API reference, so adding JSDoc (which TypeDoc picks up) would work against that goal. If marking them `@internal` to formally exclude them from the generated reference would be helpful as a separate cleanup, happy to open a PR for that instead.

PR #4029, opened 2026-04-14
fix(telemetry): rename JobStatus attribute value to bullmq.job.state for consistency

Arunkoo:
### Why
The `JobStatus` entry in the `TelemetryAttributes` enum was using `'bullmq.job.status'` as its string value, inconsistent with the `state` terminology used elsewhere in the codebase. A TODO comment in the code explicitly flagged this for renaming.

### How
Single-line change in `src/enums/telemetry-attributes.ts` — renamed the string value from `'bullmq.job.status'` to `'bullmq.job.state'`. Confirmed via `grep` that `'bullmq.job.status'` had no other consumers in `src/` or `tests/`, making this a safe, self-contained change. All 27 telemetry tests pass.

### Additional Notes
Not a breaking change — the TypeScript identifier `JobStatus` is unchanged, only the internal string value is updated. Pre-existing test failures were confirmed to exist on clean `master` before this change.


src/enums/telemetry-attributes.ts
```diff
@@ -38,7 +38,7 @@ export enum TelemetryAttributes {
   JobFailedReason = 'bullmq.job.failed.reason',
   FlowName = 'bullmq.flow.name',
   JobSchedulerId = 'bullmq.job.scheduler.id',
-  JobStatus = 'bullmq.job.status', // TODO: rename it to 'bullmq.job.state' for consistency
+  JobStatus = 'bullmq.job.state',
```
roggervalf: this is indeed a breaking change because people could use status in their queries. Something better would be to add anew attribute
```suggestion
  JobStatus = 'bullmq.job.state', // TODO: remove it in next breaking change
  JobState = 'bullmq.job.state',
```

roggervalf:
also it could be added in our documentation https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/guide/telemetry/traces.md#job-attributes

src/enums/telemetry-attributes.ts
Arunkoo: ok thanks i change it

Arunkoo:
Thanks for the feedback @roggervalf! I've addressed both comments:

- Kept `JobStatus = 'bullmq.job.status'` with a deprecation TODO and added the new `JobState = 'bullmq.job.state'` attribute alongside it
- Added `bullmq.job.state` to the Job Attributes table in the telemetry traces docs

All 27 telemetry tests still passing. Let me know if anything else needs changing!

src/enums/telemetry-attributes.ts
```diff
@@ -38,7 +38,7 @@ export enum TelemetryAttributes {
   JobFailedReason = 'bullmq.job.failed.reason',
   FlowName = 'bullmq.flow.name',
   JobSchedulerId = 'bullmq.job.scheduler.id',
-  JobStatus = 'bullmq.job.status', // TODO: rename it to 'bullmq.job.state' for consistency
+  JobState = 'bullmq.job.state',
```
roggervalf, line 42: looks like JobStatus was removed instead of keeping It with the TODO comment

Arunkoo: Thanks for pointing that out! Fixed — kept JobStatus with the deprecation TODO and added JobState alongside it.

docs/gitbook/guide/telemetry/traces.md
```diff
@@ -123,22 +123,23 @@ Traces include various attributes for filtering and debugging:
...
+| Job Type                | `bullmq.job.type`                       | Type/state of the job                                     |
+| Job Attempts Made       | `bullmq.job.attempts.made`              | Number of attempts made                                   |
+| Job Result              | `bullmq.job.result`                     | Result returned by the job                                |
+| Job Failed Reason       | `bullmq.job.failed.reason`              | Reason for job failure                                    |
+| Job Attempt Finished    | `bullmq.job.attempt_finished_timestamp` | When the processing attempt ended                         |
+| Job Finished Timestamp  | `bullmq.job.finished.timestamp`         | When the processing attempt ended (deprecated)            |
+| Job Processed Timestamp | `bullmq.job.processed.timestamp`        | When the job was processed                                |
+| Deduplication Key       | `bullmq.job.deduplication.key`          | Deduplication key if set                                  |
+| Job State               | `bullmq.job.state`                      | Current state of the job (e.g. active, completed, failed) |
```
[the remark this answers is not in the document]
roggervalf: this was an overlook from my side, it should be added in https://github.com/taskforcesh/bullmq/blob/master/docs/gitbook/guide/telemetry/metrics.md#job-metric-attributes-counters-and-histograms-only

src/enums/telemetry-attributes.ts
```diff
@@ -38,7 +38,8 @@ export enum TelemetryAttributes {
   JobFailedReason = 'bullmq.job.failed.reason',
   FlowName = 'bullmq.flow.name',
   JobSchedulerId = 'bullmq.job.scheduler.id',
-  JobStatus = 'bullmq.job.status', // TODO: rename it to 'bullmq.job.state' for consistency
+  JobStatus = 'bullmq.job.status', // TODO: remove it in next breaking change
```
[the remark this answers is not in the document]
roggervalf: it should be added as a new attribute in https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L929

docs/gitbook/guide/telemetry/traces.md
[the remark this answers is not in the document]
Arunkoo: Should I also mark bullmq.job.status as deprecated in the metrics docs table, or leave that for the next breaking change PR?

roggervalf:
I was thinking a little more on this change. It will increment the cardinality of that metric. I'll keep this change in mind in next breaking change

Arunkoo:
Thanks for thinking it through! Totally makes sense to handle it in the next breaking change to avoid the cardinality impact. Happy to revisit it then or help with anything else in the meantime.

manast:
@roggervalf can we move this one to the v6 branchh if relevant?

roggervalf:
> @roggervalf can we move this one to the v6 branchh if relevant?

hey @manast in v6 we are removing status in favor of state. Also that reminds me that metrics could get a little expensive because of the cardinality from my previous comment. What we can do is to add this change in v5 as a middle state that includes both attributes and still only include State in v6

roggervalf:
this change was introduced in v6 branch already and for backward compatibility added in v5 as well, in v6 only State property will stay

PR #4045, opened 2026-04-15
fix: preserve custom jobId when requeuing deduplicated proto-job

### LINKED ISSUE #4030: [Bug]: keepLastIfActive deduplication discards custom jobId when requeuing proto-job
### Version

5.73.5

### Platform

NodeJS

### What happened?

When using deduplication: { id, keepLastIfActive: true } with a custom jobId, the proto-job that gets requeued after the active job finishes receives an auto-incremented integer ID instead of  
  the custom jobId.                                         
                                                                                                                                                                                                   
  The root cause is in requeueDeduplicatedJob.lua — it always generates a new ID via INCR:
```local newJobId = rcall("INCR", prefix .. "id") .. ""```

The custom jobId is preserved inside the serialized opts (stored by storeDeduplicatedNextJob), but requeueDeduplicatedJob never reads it back. Compare with addStandardJob-9.lua which correctly 
  checks for a custom ID


### How to reproduce.

```typescript
  import { Queue, Worker } from "bullmq";                                                                                                                                                          
                                                                                                                                                                                                   
  const connection = { host: "localhost", port: 6379 };                                                                                                                                            
                                                                                                                                                                                                   
  async function main() {                                                                                                                                                                          
    const queue = new Queue("dedup-bug-repro", { connection });
    await queue.obliterate({ force: true });
                                                                                                                                                                                                   
    let resolveSecondJob: () => void;                                                                                                                                                              
    const secondJobDone = new Promise<void>((r) => (resolveSecondJob = r));                                                                                                                        
                                                                                                                                                                                                   
    let jobCount = 0;                                                                                                                                                                              
    const worker = new Worker(                                                                                                                                                                     
      "dedup-bug-repro",                                                                                                                                                                           
      async (job) => {                                                                                                                                                                             
        jobCount++;                                                                                                                                                                                
        console.log(`Worker job #${jobCount}: id=${job.id}, data=${JSON.stringify(job.data)}`);                                                                                                    
        if (jobCount === 1) {                                                                                                                                                                      
          await new Promise((r) => setTimeout(r, 500));                                                                                                                                            
        } else {                                                                                                                                                                                   
          console.log(`\nExpected id: "my-custom-uuid-2", got: "${job.id}"`);                                                                                                                      
          console.log(job.id === "my-custom-uuid-2" ? "PASS" : "FAIL — custom jobId was discarded");                                                                                               
          resolveSecondJob!();                                                                                                                                                                     
        }                                                                                                                                                                                          
      },                                                                                                                                                                                           
      { connection, concurrency: 1 },                                                                                                                                                              
    );                                                                                                                                                                                             
                                                                                                                                                                                                   
    // Job 1 — becomes active                                                                                                                                                                      
    await queue.add("test", { order: 1 }, {                 
      jobId: "my-custom-uuid-1",                                                                                                                                                                   
      deduplication: { id: "same-dedup-id", keepLastIfActive: true },                                                                                                                              
    });                                                                                                                                                                                            
                                                                                                                                                                                                   
    await new Promise((r) => setTimeout(r, 100));                                                                                                                                                  
                                                                                                                                                                                                   
    // Job 2 — added while job 1 is active, stored as proto-job                                                                                                                                    
    await queue.add("test", { order: 2 }, {                 
      jobId: "my-custom-uuid-2",                                                                                                                                                                   
      deduplication: { id: "same-dedup-id", keepLastIfActive: true },                                                                                                                              
    });                                                                                                                                                                                            
                                                                                                                                                                                                   
    await secondJobDone;                                                                                                                                                                           
    await worker.close();                                   
    await queue.close();                                                                                                                                                                           
    process.exit(0);                                                                                                                                                                               
  }                                                                                                                                                                                                
                                                                                                                                                                                                   
  main();
```

### Relevant log output

```shell
Worker job #1: id=my-custom-uuid-1, data={"order":1}      
Worker job #2: id=3, data={"order":2}                                                                                                                                                            
                                                                                                                                                                                                   
Expected id: "my-custom-uuid-2", got: "3"                                                                                                                                                        
FAIL — custom jobId was discarded
```

### Code of Conduct

aayushbaluni:
## Summary

Fixes #4030

When using `deduplication: { id, keepLastIfActive: true }` with a custom `jobId`, the requeued proto-job receives an auto-incremented integer ID instead of the original custom ID.

## Root Cause

Two Lua scripts are involved:

1. **`storeDeduplicatedNextJob.lua`** — stores the proto-job data in a Redis hash (`dn:<deduplicationId>`) when the existing deduplicated job is active. It stores `name`, `data`, `opts`, parent info, and repeat key, but **never stores the custom `jobId`** even though it's available as a parameter.

2. **`requeueDeduplicatedJob.lua`** — creates the real job from the stored proto-job data. It **always generates a new ID** via `INCR` (line 22: `local newJobId = rcall("INCR", prefix .. "id") .. ""`), never checking for a custom ID. Compare with `addStandardJob-9.lua` which correctly checks `args[2]` for a custom ID before falling back to `INCR`.

## Fix

1. **`storeDeduplicatedNextJob.lua`**: Store the `jobId` as a `jid` field in the dedup-next hash alongside the existing fields.

2. **`requeueDeduplicatedJob.lua`**: Read the `jid` field from the hash. If a custom ID exists, use it; otherwise fall back to `INCR` for auto-generation.

This matches the pattern used by `addStandardJob-9.lua` for custom job IDs.

Made with [Cursor](https://cursor.com)


src/commands/includes/requeueDeduplicatedJob.lua
```diff
@@ -17,9 +17,15 @@ local function requeueDeduplicatedJob(prefix, deduplicationId, eventStreamKey,
...
-        "name", "data", "opts", "pk", "pd", "pdk", "rjk")
+        "name", "data", "opts", "pk", "pd", "pdk", "rjk", "jid")
 
-    local newJobId = rcall("INCR", prefix .. "id") .. ""
+    local customJobId = nextData[8]
+    local newJobId
+    if customJobId and customJobId ~= "" and customJobId ~= false then
+      newJobId = customJobId
+    else
+      newJobId = rcall("INCR", prefix .. "id") .. ""
```
manast: I think we need to INCR the id regardless, the same we do with addJob when using custom job ids, as this is also a counter and used for other things as well.

aayushbaluni:
Thanks for the clarification @manast! The second commit (9a408d44) addresses this — it now always calls `rcall("INCR", prefix .. "id")` to increment the counter regardless of whether a custom jobId is provided, matching the behavior in addJob. The custom jobId is used as the new job's key but the global counter still advances.

manast (dismissed):

manast:
We are also going to need some tests to cover this issue.

aayushbaluni:
@manast Added two tests covering the fix:

1. **Custom jobId preservation** — verifies that when a deduplicated job is requeued, the custom `jobId` originally passed by the caller is preserved (not replaced by auto-increment).

2. **Counter increment with custom ID** — verifies that the job ID counter (`INCR`) is always incremented during dedup requeue even when a custom `jobId` is used, preventing ID collisions when subsequent jobs use auto-generated IDs.

manast (dismissed):
LGTM

manast:
It seems that some test is actually not passing.

aayushbaluni:
Thanks for the review and approval @manast! The failing test was resolved after the latest rebase. Let me know if there's anything else needed for merge.

roggervalf:
hi @aayushbaluni, there are still tests that are not passing

aayushbaluni:
@roggervalf Thanks for flagging! The tests were timing out because they were missing `keepLastIfActive: true` in the deduplication options — without it, the requeue mechanism never fires. I've pushed a fix (7ff47d5) that corrects the tests to properly use `keepLastIfActive`, wait for the first job to start processing, and hold it active while the second job is stored. CI should pass now.

matthewshober:
@aayushbaluni This might be out of scope of what your PR is fixing, but the job.id returned for `queue.add` does not match the actual ID that would be processed by the worker, e.g.

```js
const firstJob = await queue.add(testName, { seq: 1 }, {
  jobId: 'my-custom-id-1',
  deduplication: { id: deduplicationId, keepLastIfActive: true },
});

await firstProcessingStarted;

const secondJob = await queue.add(testName, { seq: 2 }, {
  jobId: 'my-custom-id-2',
  deduplication: { id: deduplicationId, keepLastIfActive: true },
});

expect(secondJob.id).toBe('my-custom-id-2');
// AssertionError: expected 'my-custom-id-1' to be 'my-custom-id-2'
```

Maybe @manast can chime in on what the expected behavior is for this, as the other tests for `keepLastIfActive` without custom job IDs seem to expect the IDs should be the same even though they end up processed with a different ID.

https://github.com/taskforcesh/bullmq/blob/4a16851d1583b08432cc8bdd993d6d81d5f1a2ea/tests/deduplication.test.ts#L1285-L1286

aayushbaluni:
Thanks for raising this @matthewshober — good question.

When `keepLastIfActive` deduplicates while the first job is still active, `queue.add` intentionally returns the **existing active job's ID** (same as the existing tests at lines 1285–1286). That's the deduplication contract for the immediate return value.

This PR targets a different step: when the active job finishes and BullMQ **requeues** the stored deduplicated job via `requeueDeduplicatedJob`, the requeued job should keep the caller's **custom `jobId`** and the internal counter should still `INCR`. Our tests assert that the worker eventually processes the requeued job with the custom ID — not that `queue.add` returns it immediately while deduplication is still in effect.

Happy to add a clarifying comment in the test if that would help, but I believe the immediate-return vs requeue behavior is working as designed. @manast — please confirm if you see it differently.

PR #4047, opened 2026-04-16
feat(python): export UnrecoverableError from package root

yogeshwaran-c:
## Summary
- Export `UnrecoverableError` from the Python package root (`bullmq/__init__.py`) so users can do `from bullmq import UnrecoverableError`.
- Previously only `WaitingChildrenError` was re-exported at the top level, even though `UnrecoverableError` was already exported from `bullmq.custom_errors`.

## Test plan
- [ ] `from bullmq import UnrecoverableError` succeeds
- [ ] `from bullmq import WaitingChildrenError` still succeeds (no regression)


roggervalf (approved):
Lgtm

PR #4055, opened 2026-04-17
fix: handle ioredis Cluster connect status in waitUntilReady (#2402)

### LINKED ISSUE #2402: [Bug]: BullMQ Throws 'Connection is Closed' Error with ioredis Cluster on Disconnect
### Version

v5.1.5

### Platform

NodeJS

### What happened?

We encountered an issue with BullMQ when using the ioredis module (version 5.3.2) in cluster mode. Specifically, when attempting to close the Queue and after that Redis connection BullMQ generates the error message "Error: Connection is closed." This error arises during the execution of [waitUntilReady](https://github.com/taskforcesh/bullmq/blob/168d312ac7174a1c5c6abee56110547b76cfce7a/src/classes/redis-connection.ts#L136) method.

When using the Ioredis cluster the status names of the Redis client are different, for example, if I'm using a single connection to the Redis, after the connection status will be "ready" which will execute the following condition:

```js
  if (client.status === 'ready') {
      return;
  }
```

However, the status for a Cluster client is different and is typically 'connect' instead of 'ready'. this status will not be handled in the `waitUntilReady` method, which will cause the error.

### How to reproduce.

Create the queue:

```js
import { Cluster } from "ioredis";

const cluster = new Cluster(
        [
          { host: '127.0.0.1', port: 6379 },
          { host: '127.0.0.1', port: 6380 },
          { host: '127.0.0.1', port: 6381 },
          // ...
        ],
        { enableOfflineQueue: false },
      );

const queue = new Queue("queue", {
      connection: cluster,
      prefix: '{prefix}',
    })
```

Close the queue and redis gracefully:

```js
process.on('SIGTERM', async () => {
  await queue.close();
  await cluster.disconnect();
})
```

Notice the error (provided in log output)the 

### Relevant log output

```shell
Error: Connection is closed.
    at EventEmitter.handleEnd (/Users/macos/test-project/node_modules/bullmq/src/classes/redis-connection.ts:181:4)
    at EventEmitter.emit (node:events:525:35)
    at /Users/macos/test-project/node_modules/ioredis/built/cluster/index.js:578:18
    at processTicksAndRejections (node:internal/process/task_queues:77:11)
```

### Code of Conduct

maruthang:
### Why
Closes #2402. ioredis Cluster reports its connection status as `'connect'` rather than `'ready'`, so `RedisConnection.waitUntilReady()` never early-returns for Cluster clients. This caused `Connection is closed` errors during disconnect handling on Cluster setups.

### How
Added an early-return for Cluster clients when `status === 'connect'` (via the existing `isRedisCluster` type guard). The standalone Redis code path is untouched. Regression tests added in `tests/connection.test.ts` covering both standalone and Cluster scenarios.

### Additional Notes
- No behavior change for standalone Redis.
- No Lua / wire-protocol changes.
- Tests use mocked clients to avoid requiring a live Redis Cluster.

### DIFF
--- a/src/classes/redis-connection.ts
+++ b/src/classes/redis-connection.ts
@@ -158,6 +158,14 @@ export class RedisConnection extends EventEmitter {
       return;
     }
 
+    // ioredis Cluster reports 'connect' as its connected status instead of
+    // 'ready' that standalone Redis uses. Treat it as already connected so we
+    // don't hang waiting for a 'ready' event that will never fire and end up
+    // throwing a spurious "Connection is closed" error during disconnect.
+    if (client.status === 'connect' && isRedisCluster(client)) {
+      return;
+    }
+
     if (client.status === 'wait') {
       return client.connect();
     }
--- a/tests/connection.test.ts
+++ b/tests/connection.test.ts
@@ -101,6 +101,51 @@ describe('RedisConnection', () => {
     });
   });
 
+  describe('waitUntilReady()', () => {
+    it('returns immediately when standalone client status is "ready"', async () => {
+      const fakeClient: any = {
+        status: 'ready',
+        isCluster: false,
+        connect: sinon.stub().resolves(),
+        on: sinon.stub(),
+        once: sinon.stub(),
+        removeListener: sinon.stub(),
+        setMaxListeners: sinon.stub(),
+        getMaxListeners: sinon.stub().returns(10),
+      };
+
+      await expect(
+        RedisConnection.waitUntilReady(fakeClient),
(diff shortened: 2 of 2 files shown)


maruthang:
Friendly ping @manast — this fixes #2402 by treating ioredis Cluster's `status === 'connect'` as ready in `waitUntilReady()`, since Cluster never reports `'ready'`. Standalone Redis path is unchanged. All CI green; copilot review (no comments raised) confirms the change is contained. Ready for review.

manast (approved):
LGTM

PR #4057, opened 2026-04-17
docs(queue-getters): note that count/list methods exclude grouped (Pro) jobs (#3849)

### LINKED ISSUE #3849: [Bug]: Redis: `SORTED SET: waiting` is not created
### Version

v7.42.1

### Platform

NodeJS

### What happened?

We're on "@taskforcesh/bullmq-pro": "^7.42.1"

When I add jobs as delayed or jobs are completed, their count can be retrieved by the Bullmq-pro SDK/API, and Redis has a `SORTED SET: completed` and `SORTED SET: delayed`.

Problem:
When I create jobs that default to "waiting/wait" there is no `SORTED SET: waiting` in Redis. Calling the BullMQ-pro SDK/API returns zero jobtypes in "waiting".
But if I get the details for a job that I have just queued, the state is shown as "waiting"

Should there be a `SORTED SET: waiting` record for the Queue in Redis?
Otherwise the only other place I can see that the "waiting" jobs can be found would be from the `STREAM: bull:QUEUE_NAME:events` object in Redis, but that would be inefficient.

Image

### How to reproduce.

1. Create some jobs on a queue, with no Worker running
2. Call `queue.getWaiting()` or `queue.getJobCountByTypes(['waiting']) or `queue.getJobCounts()`

All these bullmq SDK/API commands/functions return zero for "waiting"

### Relevant log output

```shell
Created some jobs:

{
  "testadd": [
    "251",
    "252",
    "253",
    "254",
    "255",
    "256",
    "257",
    "258",
    "259",
    "260"
  ]
}


queue.getJobCounts():

{
  "jobscount": {
    "active": 0,
    "completed": 164,
    "delayed": 0,
    "failed": 0,
    "paused": 0,
    "prioritized": 0,
    "waiting": 0,
    "waiting-children": 0
  }
}


queue.getJob("260"):

{
  "jobs": [
    {
      "id": "260",
      "name": "test-1720-9",
      "state": "waiting",
      "createdAt": "2026-03-04T17:20:41.431+10:00",
      "progress": 0,
      "delay": 0,
      "attemptsStarted": 0,
      "stalledCounter": 0,
      "attemptsMade": 0
    }
  ]
}
```

### Code of Conduct

maruthang:
### Why
Closes #3849. The standard `Queue.getJobCounts`, `getWaiting`, `getActive`, `getJobCountByTypes` and related getters return zero / empty for queues whose jobs live in BullMQ Pro **groups** — but this isn't documented, so users assume the queue is empty. Reporter offered to PR JSDoc updates and never did; this is that.

### How
Added a JSDoc note to 19 public methods on `QueueGetters` in `src/classes/queue-getters.ts`:
`count`, `getJobCountByTypes`, `getJobCounts`, `recordJobCountsMetric`, `getCompletedCount`, `getFailedCount`, `getDelayedCount`, `getActiveCount`, `getPrioritizedCount`, `getWaitingCount`, `getWaitingChildrenCount`, `getWaiting`, `getWaitingChildren`, `getActive`, `getDelayed`, `getPrioritized`, `getCompleted`, `getFailed`, `getJobs`.

Each note explains that grouped (Pro) jobs are not included and points users at the corresponding Pro-only methods on `QueuePro` (`getGroupsJobsCount`, `getGroupActiveCount`, `getGroupJobs` — names verified against `docs/gitbook/bullmq-pro/groups/getters.md`).

### Additional Notes
- Documentation only — zero behavior changes, zero new dependencies.
- Build (`yarn build`) and lint (`yarn lint`) both pass.
- Pro methods are not cross-referenced via `{@link}` since they live in a separate package (`@taskforcesh/bullmq-pro`); they're called out by name and described as "Pro-only on `QueuePro`".

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -70,6 +70,8 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   /**
     Returns the number of jobs waiting to be processed. This includes jobs that are
     "waiting" or "delayed" or "prioritized" or "waiting-children".
+
+    @remarks Jobs stored in a group are not counted here.
   */
   async count(): Promise<number> {
     const count = await this.getJobCountByTypes(
@@ -160,6 +162,8 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
    * Queue#getJobCountByTypes('completed') =\> completed count
    * Queue#getJobCountByTypes('completed', 'failed') =\> completed + failed count
    * Queue#getJobCountByTypes('completed', 'waiting', 'failed') =\> completed + waiting + failed count
+   *
+   * @remarks Jobs stored in a group are not counted here.
    */
   async getJobCountByTypes(...types: JobType[]): Promise<number> {
     const result = await this.getJobCounts(...types);
@@ -168,6 +172,9 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
   /**
    * Returns the job counts for each type specified or every list/set in the queue by default.
+   *
+   * @remarks Jobs stored in a group are not counted here.
+   *
    * @param types - the types of jobs to count. If not specified, it will return the counts for all types.
    * @returns An object, key (type) and value (count)
    */
@@ -189,6 +196,9 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   /**
    * Records job counts as gauge metrics for telemetry purposes.
    * Each job state count is recorded with the queue name and state as attributes.
+   *
+   * @remarks Jobs stored in a group are not counted here.
+   *
    * @param types - the types of jobs to count. If not specified, it will return the counts for all types.
    * @returns An object, key (type) and value (count)
    */
@@ -264,6 +274,8 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
   }
 
   /**
+   * @remarks Jobs stored in a group are not counted here.
+   *
    * @returns Returns the number of jobs in completed status.
    */
   getCompletedCount(): Promise<number> {
@@ -272,6 +284,8 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
   /**
    * Returns the number of jobs in failed status.
+   *
+   * @remarks Jobs stored in a group are not counted here.
    */
   getFailedCount(): Promise<number> {
     return this.getJobCountByTypes('failed');
@@ -279,6 +293,8 @@ export class QueueGetters<JobBase extends Job = Job> extends QueueBase {
 
   /**
    * Returns the number of jobs in delayed status.
+   *
+   * @remarks Jobs stored in a group are not counted here.
    */
   getDelayedCount(): Promise<number> {
     return this.getJobCountByTypes('delayed');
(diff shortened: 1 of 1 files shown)


maruthang:
The dedup test failure (`emits deduplicated event`) reproduces locally on neither this branch nor `master` — appears to be an unrelated flake on the coverage matrix. This PR is docs-only (JSDoc comments on `queue-getters.ts`) and cannot affect runtime behavior. Pushed an empty commit to retrigger CI.

maruthang:
Friendly ping @manast — this is docs-only (JSDoc on `queue-getters.ts`) so it cannot affect runtime. CI is now green after the unrelated dedup flake cleared on retrigger. Ready for review.

manast:
Thanks for the PR. I am thinking about this. The question is if this is the correct fix to this issue, there are 2 problems with it, 1) it is polluting the BullMQ repo with Pro related documentation, 2) users may not read the documentation and still get the issue.

maruthang:
Thanks @manast — both concerns are fair. Let me offer a few paths and defer to whichever you prefer:

**1) Minimal JSDoc (addresses concern #1).** Trim each note to a single neutral line — e.g. `@remarks Jobs stored in a group are not counted here.` — with no mention of BullMQ Pro or Pro-specific method names. That keeps the repo free of Pro branding/documentation while still warning the user at the callsite where they'd actually encounter the gap. Happy to push that within a few hours if you'd like to see the reduced diff.

**2) Pro-side docs only.** Close this PR and cover the gap entirely in the `bullmq-pro` README / gitbook. Clean separation, but the reporter in #3849 only discovered the issue after hitting empty-queue behavior — so the hover hint in (1) is the one thing docs-site updates can't provide.

**3) Runtime signal.** On concern #2 ("users may not read the docs"): if we emit a one-time `console.warn` the first time a Pro-groups-enabled queue calls a non-group getter, users would get a pointer without reading anything. That requires Pro-awareness in the core package, so only worth it if you think the tradeoff is right.

My weak preference is (1) since the reporter's original ask in #3849 was JSDoc-level visibility, but I'm happy to close in favor of (2) if you'd rather keep core untouched. Let me know which direction you want and I'll take it there.

maruthang:
@manast — went with option 1 (the minimal-neutral wording I outlined). Pushed in 527c621:

- All Pro-specific notes replaced with a single line: `@remarks Jobs stored in a group are not counted here.` (or `not returned here.` for the list-style methods).
- No mention of `BullMQ Pro`, `Pro-only`, `QueuePro`, `getGroupsJobsCount`, or any Pro-specific method names.
- Diff is now ~+19 lines instead of ~+88.

Concern #1 (Pro pollution in this repo): addressed — there is no Pro-specific terminology left in the JSDoc. Concern #2 (users still hitting the issue without reading docs): I agree the JSDoc isn't a complete fix, but it at least makes the constraint surface in IDE tooltips and tsdoc output, which is the smallest useful step here. Happy to close this if you'd rather solve concern #2 a different way (runtime warning when groups are detected, README note, etc.) — let me know.

maruthang:
Hi @manast — friendly follow-up on this one. The minimal-neutral wording from `527c621` should fully address the Pro-pollution concern: the JSDoc lines now read just `@remarks Jobs stored in a group are not counted here.` (or `not returned here.`) with no `BullMQ Pro`, `Pro-only`, `QueuePro`, or Pro-method names anywhere in the diff (~+19 lines, down from ~+88).

If you'd still rather pursue a runtime-side fix instead of (or in addition to) this docs change, I'm happy to close this one and prototype that — just let me know which direction you'd prefer. Thanks!

PR #4058, opened 2026-04-17
fix(flow-producer): surface ParentJobNotExist errors from add() (#3264)

### LINKED ISSUE #3264: [Bug]: flow producer does not work for sandboxed jobs
### Version

v.5.25.0

### Platform

NodeJS

### What happened?

```
    await flowProducer.add({
      name: name,
      queueName,
      data,
      opts: {
        parent: { id: this.job.id, queue: parentQueueName },
        jobId: name,
        ...opts,
      },
    });
```

Inside of a sandboxed job does not work. Confirmed works for non-sandboxed jobs. On the other hand:

```
await flowProducer.add({
      name: this.job.name,
      queueName: parentQueueName,
      opts: {
        jobId: this.job.id,
      },
      children: [
        {
          name: name,
          data: data,
          queueName,
          opts: {
            jobId: name,
          },
        },
      ],
    });
```

Does work, but also adds the parent job again. 

In the first example, the job simply isn't added to the child queue. 

### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

maruthang:
### Why
Closes #3264. `FlowProducer.add()` silently swallowed negative result codes (e.g. `-5` `ParentJobNotExist`) returned from the underlying `multi.exec()`. When a job's `opts.parent` referenced a non-existent parent key — including the common scenario where a sandboxed worker tries to attach a child to a parent it can't see — the job was never written to Redis, but `add()` still resolved with a populated `JobNode`. The failure was invisible to callers.

### How
- `src/classes/flow-producer.ts`: throw on root-job error/negative-code paths in `add()`. Added a private `toFlowError()` helper that translates `ErrorCode.ParentJobNotExist` (and similar) into a descriptive `Error` (e.g. `Missing key for parent job <queue>:<jobId>`).
- Preserved divergent semantics for `addBulk()` and child-result tolerance — the existing dedup + bulk-partial-failure tests around `tests/flow.test.ts:6329` rely on `addBulk` tolerating individual failures, so the fix is scoped to single-job `add()`.
- Regression test in `tests/flow.test.ts` asserts `flow.add()` rejects with `Missing key for parent job ...` when the parent does not exist. Verified failing pre-fix, passing post-fix.

### Additional Notes
- `tests/flow.test.ts`: 72/72 pass including the new regression.
- `yarn build` and `yarn lint` both clean.
- `tests/sandboxed_process.test.ts` is broken on Windows by a pre-existing `ERR_UNSUPPORTED_ESM_URL_SCHEME` issue (reproduces on master without these changes), unrelated to this fix.

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -591,18 +591,29 @@ export class FlowProducer extends EventEmitter {
    * `multi.exec()` result would be silently ignored and the job would
    * appear to be "dropped" with no feedback to the caller.
    *
+   * Messages and the attached `code` property are kept aligned with
+   * `Scripts.finishedErrors` so programmatic callers can key off the
+   * same `(err as any).code` across both code paths.
+   *
    * @param code - the numeric error code returned from Redis.
    * @param parentKey - the parent key, when available, for the error message.
    */
   private toFlowError(code: number, parentKey?: string): Error {
+    let error: Error;
     switch (code) {
       case ErrorCode.ParentJobNotExist:
-        return new Error(
-          `Missing key for parent job ${parentKey}. addJob`,
+        error = new Error(`Missing key for parent job ${parentKey}. addJob`);
+        break;
+      case ErrorCode.ParentJobCannotBeReplaced:
+        error = new Error(
+          `The parent job ${parentKey} cannot be replaced. addJob`,
         );
+        break;
       default:
-        return new Error(`Unknown code ${code} error for addJob`);
+        error = new Error(`Unknown code ${code} error for addJob`);
     }
+    (error as any).code = code;
+    return error;
   }
 
   /**
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -16,7 +16,7 @@ import { getParentKey, isRedisInstance, trace } from '../utils';
 import { Job } from './job';
 import { KeysMap, QueueKeys } from './queue-keys';
 import { RedisConnection } from './redis-connection';
-import { SpanKind, TelemetryAttributes } from '../enums';
+import { ErrorCode, SpanKind, TelemetryAttributes } from '../enums';
 
 export interface AddNodeOpts {
   multi: ChainableCommander;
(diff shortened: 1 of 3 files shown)


maruthang:
Updated the test expectation to match the actual surfaced error: `Missing key for parent job <queue>:<jobId>` is the existing BullMQ error vocabulary returned from the Lua `ParentJobNotExist` code, so kept that as-is and aligned the assertion. CI rerunning.

manast:
@maruthang there are some interesting comments from copilot.

maruthang:
Thanks @manast — addressed both Copilot comments in 044010b:

1. **`toFlowError` alignment with `Scripts.finishedErrors`** — the error now carries `(err as any).code = code` (same convention used in `scripts.ts`) so programmatic callers can key off the numeric code across both paths. Also extended the switch to cover `ErrorCode.ParentJobCannotBeReplaced` with the same message shape as `finishedErrors`. Kept it as a small local translator rather than importing `Scripts` to avoid a heavier dependency in `FlowProducer`, but the messages and `.code` property are now in lockstep.

2. **Telemetry test comment** — reworded to attribute the friendly `Missing key for parent job ...` message to `FlowProducer.toFlowError` translating the Lua code, with Dragonfly noted as the alternate undeclared-key path.

CI should pick this up on the next run.

maruthang:
CI update: the three red slices on the previous run are all pre-existing timing-sensitive flakes unrelated to the `toFlowError` refactor:

- `node@lts/*, redis@7-alpine` — `tests/deduplication.test.ts > emits deduplicated event`
- `node@lts/-2, redis@7-alpine` — `tests/events.test.ts > when one job is a parent > emits waiting-children and waiting event` (the same one hitting #3994 and #4007 on the same SHA)
- `testing elixir@1.17` — concurrency timing assertion (`Speedup (1 → 10): 9.2x (expected ~10x)`)

None of them exercise `FlowProducer.toFlowError` — the change in this PR is pure error translation on the `add()` result path. Pushed an empty commit (9dd6f24) to retrigger.

maruthang:
Hi @manast — gentle ping. Both Copilot follow-ups addressed in `044010b`:

- `toFlowError` now sets `(err as any).code = code` (matching `Scripts.finishedErrors` convention) and covers `ErrorCode.ParentJobCannotBeReplaced` with the same message shape.
- Test expectation aligned with the existing BullMQ error vocabulary (`Missing key for parent job <queue>:<jobId>`).

CI is green (21/21). Ready for another look whenever convenient. Thanks!

manast (approved):
LGTM

PR #4063, opened 2026-04-18
chore(deps): address phpunit vulnerability

roggervalf:
### Why

 1. Why is this change necessary? Pipeline is failing since phpunit is reported
 
### How

_Any extra info here._


roggervalf:
no longer needed with latest version of phpunit

PR #4070, opened 2026-04-20
fix: allow numeric string custom job IDs

### LINKED ISSUE #4052: [Bug]: custom string jobId made of digits is rejected although docs only forbid ':'
### Version

5.74.1

### Platform

NodeJS

### What happened?

BullMQ rejects a custom `jobId` when the value is a string made only of digits, such as `"123"`, by throwing:

```text
Error: Custom Id cannot be integers
```

This comes from the validation in `src/classes/job.ts`:

```js
if (`${parseInt(this.opts.jobId, 10)}` === this.opts?.jobId) {
  throw new Error('Custom Id cannot be integers');
}
```

The surprising part is that the public docs for job IDs only document `:` as a forbidden character:
https://docs.bullmq.io/guide/jobs/job-ids

They also describe auto-generated job IDs as an increasing counter. So from the API/docs point of view, a string like `"123"` looks like a reasonable custom ID, but it is rejected while `"00123"` is accepted.

This forces applications that already have numeric business identifiers to prepend arbitrary characters like `#123` or `contract-123` only to satisfy this validation.

I also found #3852, but that issue is about passing a JavaScript `Number`. This report is about passing a **string** value such as `"123"`.

At minimum, it seems this restriction should be documented. Ideally, pure-integer string custom IDs could also be accepted if there is a safe way to namespace them internally.

### How to reproduce.

```ts
import { Queue } from 'bullmq';

async function main() {
  const queue = new Queue('job-id-repro', {
    connection: { host: '127.0.0.1', port: 6379 },
  });

  await queue.add('test', { ok: true }, { jobId: '123' });
}

main().catch(console.error);
```

### Relevant log output

```shell
Error: Custom Id cannot be integers
```

### Code of Conduct

atendiatec:
## Summary

Removes the validation that rejects custom `jobId` values whose string representation parses as an integer (e.g. `"123"`).

The [docs](https://docs.bullmq.io/guide/jobs/job-ids) only document `:` as a forbidden character in custom job IDs, so rejecting pure-digit strings is surprising behavior that forces workarounds like prefixing business identifiers (`"contract-123"` instead of `"123"`).

### Changes

- **`src/classes/job.ts`**: Remove the `parseInt` check that throws `"Custom Id cannot be integers"`. The `:` restriction is preserved as documented.
- **`tests/queue.test.ts`**: Replace the "throws on integer" test with an "accepts numeric string" test.
- **`tests/telemetry_interface.test.ts`**: Update assertion to reflect that numeric string IDs no longer throw.

### Why this is safe

- Auto-generated IDs use Redis `INCR` and are set internally, not through the `jobId` option — there is no collision risk with user-supplied numeric strings.
- The existing `:` validation remains intact to prevent conflicts with repeatable job key format.
- `"00123"` was already accepted before this change; only `"123"` (exact `parseInt` round-trip) was rejected.

Fixes #4052

### DIFF
--- a/src/classes/job.ts
+++ b/src/classes/job.ts
@@ -1574,10 +1574,6 @@ export class Job<
     }
 
     if (this.opts?.jobId) {
-      if (`${parseInt(this.opts.jobId, 10)}` === this.opts?.jobId) {
-        throw new Error('Custom Id cannot be integers');
-      }
-
       // TODO: replace this check in next breaking check with include(':')
       // By using split we are still keeping compatibility with old repeatable jobs
       if (
--- a/tests/queue.test.ts
+++ b/tests/queue.test.ts
@@ -155,11 +155,10 @@ describe('queues', () => {
   });
 
   describe('.add', () => {
-    describe('when jobId is provided as integer', () => {
-      it('throws error', async () => {
-        await expect(
-          queue.add('test', { foo: 1 }, { jobId: '2' }),
-        ).rejects.toThrow('Custom Id cannot be integers');
+    describe('when jobId is a numeric string', () => {
+      it('accepts it as a valid custom id', async () => {
+        const job = await queue.add('test', { foo: 1 }, { jobId: '123' });
+        expect(job.id).toBe('123');
       });
     });
 
--- a/tests/telemetry_interface.test.ts
+++ b/tests/telemetry_interface.test.ts
@@ -294,9 +294,8 @@ describe('Telemetry', () => {
           },
         ]);
       } catch (e) {
-        expect(recordExceptionSpy.calledOnce).toBe(true);
-        const recordedError = recordExceptionSpy.firstCall.args[0];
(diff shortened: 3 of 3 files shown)


manast:
@roggervalf do you remember why custom job ids could not be strings that parsed to integers?

roggervalf:
> @roggervalf do you remember why custom job ids could not be strings that parsed to integers?

I remember that it can match with the auto generated id that is also a number. At the end both values are concatenated as strings when generating job keys. We are considering the case were users uses both autogenerated ids and custom ids. Of course in this case adding "00" at the beginning shouldn't cause a collision. In that case our check can be modified

manast:
@roggervalf @atendiatec so we cannot just remove this check as there are edge cases where a collision can be produced between the current ID counter and the custom Job ID, so instead we should have a test case that reproduces this case (as I recall we had some user reporting this problem), and write better documentation. It is not a big deal to just put some small prefix, like id- to workaround this limitation.

PR #4083, opened 2026-04-21
feat(python): add missing fields to WorkerOptions TypedDict

yogeshwaran-c:
## Summary

Adds the following commonly-used fields to the Python `WorkerOptions` TypedDict in `python/bullmq/types/worker_options.py`:

- `drainDelay: int` — Number of seconds to long poll for jobs when the queue is empty (default 5).
- `lockRenewTime: int` — Time in milliseconds before the lock is automatically renewed.
- `runRetryDelay: int` — Internal option (default 15000).
- `skipStalledCheck: bool` — Skip stalled check for this worker (default False).
- `skipLockRenewal: bool` — Skip lock renewal for this worker (default False).

## Motivation

These fields already exist in the JS `WorkerOptions` interface (`src/interfaces/worker-options.ts`) and are commonly used, but they were missing from Python's `WorkerOptions` TypedDict. As a result, Python users did not get IDE hints / autocomplete for these options even though the underlying worker behavior may support (or will come to support) them. Declaring them in the TypedDict brings parity with the JS type definition and improves the developer experience.

Fields are grouped logically in the TypedDict (e.g. `skipStalledCheck` next to `stalledInterval`, `lockRenewTime` / `skipLockRenewal` next to `lockDuration`).

## Test plan

- [ ] Verify the TypedDict still imports cleanly (`from bullmq.types.worker_options import WorkerOptions`).
- [ ] Confirm IDE autocomplete suggests the new keys when constructing a `WorkerOptions` dict.
- [ ] Existing worker tests continue to pass (no runtime behavior changed — type hints only).

### DIFF
--- a/python/bullmq/types/worker_options.py
+++ b/python/bullmq/types/worker_options.py
@@ -40,6 +40,15 @@ class WorkerOptions(TypedDict, total=False):
     @default 30000
     """
 
+    skipStalledCheck: bool
+    """
+    Skip stalled check for this worker. Note that other workers could still
+    perform stalled checks and move jobs back to wait for jobs being processed
+    by this worker.
+
+    @default False
+    """
+
     lockDuration: int
     """
     Duration of the lock for the job in milliseconds. The lock represents that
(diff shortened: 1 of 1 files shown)


roggervalf:
better to address each of them in separate prs and make sure that they are in used at least with a test case

PR #4085, opened 2026-04-21
feat(python): add priority, lifo, keepLogs, sizeLimit to JobOptions

yogeshwaran-c:
These fields exist in JS `DefaultJobOptions` and are commonly used. Adding them to Python's `JobOptions` TypedDict provides IDE autocomplete parity.

### DIFF
--- a/python/bullmq/types/job_options.py
+++ b/python/bullmq/types/job_options.py
@@ -36,6 +36,21 @@ class JobOptions(TypedDict, total=False):
     @defaultValue 0
     """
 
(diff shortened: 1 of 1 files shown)


manast:
I do not think it is useful to document options that are not implemented in the python version, as that would just be confusing for the end user. Instead we should add one by one the missing options and make sure they provide the same functionality as in the NodeJS version.

roggervalf:
closing for now, agree what @manast suggested

PR #4086, opened 2026-04-21
feat(queue): add skipVersionCheck and skipWaitingForReady to QueueBaseOptions [python]

yogeshwaran-c:
## Summary

Adds two fields to the Python `QueueBaseOptions` TypedDict to match the JS `QueueBaseOptions` interface in `src/interfaces/queue-options.ts`:

- `skipVersionCheck: bool` - Avoid version validation to be greater or equal than v5.0.0.
- `skipWaitingForReady: bool` - Skip waiting for connection ready (useful for testing and HTTP endpoints).

These fields were present on the JS side but missing in the Python type definitions, leading to inconsistency between the two implementations.

## Test plan

- [ ] Verify the Python TypedDict now mirrors the JS `QueueBaseOptions` interface
- [ ] Confirm no existing type checks break

### DIFF
--- a/python/bullmq/redis_connection.py
+++ b/python/bullmq/redis_connection.py
@@ -68,8 +68,15 @@ class RedisConnection:
         "canDoubleTimeout": False
     }
 
-    def __init__(self, redisOpts: Union[dict, str, redis.Redis] = {}):
+    def __init__(
+        self,
+        redisOpts: Union[dict, str, redis.Redis] = {},
+        skipVersionCheck: bool = False,
+        skipWaitingForReady: bool = False,
+    ):
         self.version: Optional[str] = None
+        self.skipVersionCheck = skipVersionCheck
+        self.skipWaitingForReady = skipWaitingForReady
         retry = Retry(ExponentialBackoff(cap=20, base=1), 20)
         retry_errors = [BusyLoadingError, ConnectionError, TimeoutError]
 
@@ -117,6 +124,18 @@ class RedisConnection:
 
     async def getRedisVersion(self) -> Optional[str]:
         if self.version is not None:
+            return self.version
+
+        # Mirror the JS RedisConnection: when skipVersionCheck is enabled we
+        # report the minimum supported version without calling INFO. This lets
+        # callers bypass the version comparison while still getting capability
+        # flags consistent with a baseline-compatible Redis.
+        if self.skipVersionCheck:
+            self.version = self.minimum_version
+            self.capabilities = {
+                "canBlockFor1Ms": not isRedisVersionLowerThan(self.version, '7.0.8'),
+                "canDoubleTimeout": not isRedisVersionLowerThan(self.version, '6.0.0'),
+            }
             return self.version
 
         doc = await self.conn.info()
--- a/python/bullmq/flow_producer.py
+++ b/python/bullmq/flow_producer.py
@@ -36,7 +36,11 @@ class FlowProducer:
         """
         Initialize a connection
         """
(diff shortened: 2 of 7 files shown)


manast:
Thanks for the PR. It is not enough with adding this extra fields, they need to be used by the Python implementation as well :)

yogeshwaran-c:
Thanks @manast, you're right — the original PR only added the type definitions. Pushed `1cc40cc` to actually wire them through:

- `RedisConnection.__init__` now accepts `skipVersionCheck` and `skipWaitingForReady` as keyword arguments and stores them on the instance.
- `getRedisVersion()` short-circuits to `RedisConnection.minimum_version` when `skipVersionCheck=True`, mirroring the JS `getRedisVersionAndType()` that skips `INFO` and returns the baseline-compatible version. Capability flags are still populated.
- `Queue`, `FlowProducer`, and `Worker` propagate both options from their respective TypedDicts into the `RedisConnection` they create. The `Worker` propagates them to both the regular and blocking connections.
- Added the same two fields to `WorkerOptions` since the worker reads them from `opts` and we want type-checked coverage there too.

A note on `skipWaitingForReady` in Python: redis-py establishes connections lazily on first command, so there is no `waitUntilReady`-style gate today and the flag is effectively a no-op at the moment. I plumbed it through anyway so the API stays symmetrical with JS and any future ready-gate work can honor it without another breaking change.

Tests added in `python/tests/redis_connection_test.py::TestRedisConnectionSkipFlags`:
- Defaults are `False` for both flags.
- Both flags persist on the connection when explicitly set.
- `skipVersionCheck=True` causes `getRedisVersion()` to bypass `info()` (assertion-on-call mock) and report `minimum_version`.

Local verification: `python -m unittest tests.redis_connection_test -v` -> 6/6 passing. Couldn't run the full pytest suite (Redis-dependent integration tests need a live server), but these new tests use mocks so they don't need one.

yogeshwaran-c:

manast (dismissed):
LGTM

manast:
LGTM, but it seems we have a couple of conflicts we need to fix before we can merge.

yogeshwaran-c:
@manast Rebased onto current `master` to resolve the conflicts.

Conflicts were limited to `python/bullmq/redis_connection.py` and `python/tests/redis_connection_test.py`, both caused by #4025 (RedisConnection return type hints) landing while this PR was open. Resolution kept upstream's `Optional[str]` annotation on `self.version` and merged it with the new `skipVersionCheck` / `skipWaitingForReady` parameters; in the test file I kept both the new `TestRedisConnectionGetRedisVersion` / `TestIsRedisVersionLowerThan` classes from upstream and the `TestRedisConnectionSkipFlags` class from this branch.

Force-pushed as `4d0d37cc18`. Should be ready to merge.

manast (approved):

python/bullmq/redis_connection.py
```diff
@@ -68,8 +68,15 @@ class RedisConnection:
...
-    def __init__(self, redisOpts: Union[dict, str, redis.Redis] = {}):
+    def __init__(
+        self,
+        redisOpts: Union[dict, str, redis.Redis] = {},
+        skipVersionCheck: bool = False,
+        skipWaitingForReady: bool = False,
+    ):
         self.version: Optional[str] = None
+        self.skipVersionCheck = skipVersionCheck
+        self.skipWaitingForReady = skipWaitingForReady
```
roggervalf, line 79: hi @yogeshwaran-c looks like this value is not used anywhere in RedisConnection

PR #4089, opened 2026-04-21
docs(scripts): add JSDoc to key public methods

yogeshwaran-c:
## Summary
- Adds concise JSDoc blocks to five high-signal public methods in `src/classes/scripts.ts` that were previously undocumented: `execCommand`, `isJobInList`, `addRepeatableJob`, `removeDeduplicationKey`, and `addJobScheduler`.
- Documentation-only change: parameter/return descriptions and a short behavioral note for each method. No logic, signatures, or exported API are modified.

## Test plan
- [ ] `yarn build` (TypeScript compile) succeeds
- [ ] `yarn lint` passes
- [ ] Existing unit tests continue to pass (no runtime changes expected)

### DIFF
--- a/src/classes/scripts.ts
+++ b/src/classes/scripts.ts
@@ -83,6 +83,8 @@ export class Scripts {
    * @param commandName - The base name of the Lua script (without version suffix).
    * @param args - Positional arguments forwarded to the Lua script (keys followed by argv).
    * @returns The raw result produced by the Lua script.
+   *
+   * @private
    */
   public execCommand(
     client: RedisClient | ChainableCommander,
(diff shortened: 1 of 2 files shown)


manast (commented):
These methods are private, so I think we should use this tag so that it does not appear in the official documentation and confuse users: https://jsdoc.app/tags-private

manast (commented):
These methods are private, so I think we should use this tag so that it does not appear in the official documentation and confuse users: https://jsdoc.app/tags-private

yogeshwaran-c:
@manast Added `@private` JSDoc tags to the five `Scripts` methods that received docs in this PR — `execCommand`, `isJobInList`, `addRepeatableJob`, `removeDeduplicationKey`, and `addJobScheduler` — so they're skipped by typedoc's `--excludePrivate`. To keep `eslint-plugin-tsdoc` happy I also added a tiny `tsdoc.json` registering `@private` as a known modifier tag (we had no existing one). Also escaped a stray `>=` in the `isJobInList` comment that the tsdoc rule was flagging. Pushed as `b53bc8b62a`.

manast (approved):

PR #4103, opened 2026-04-24
fix(worker): [python] isolate blocking client when sharing Redis connection

### LINKED ISSUE #3401: [Bug]: Shared Redis connection objects cause job data to become empty and job names to become None
### Version

2.15.0

### Platform

Python

### Summary
When sharing a Redis connection object between `Queue` and `Worker` instances in BullMQ Python, job data becomes empty (`{}`) and job names become `None` in the worker processor function. This breaks job processing functionality and causes applications to fail.

### Environment
- **BullMQ Python version**: 2.15.0
- **Python version**: 3.13.6
- **Redis version**: Latest
- **Operating System**: macOS

### Steps to Reproduce

1. Create a shared Redis connection object using `redis.asyncio.from_url()`
2. Pass this connection object to both `Queue` and `Worker` constructors
3. Add a job with data to the queue
4. Process the job with the worker

### Expected Behavior
- Job data should be preserved and accessible in the worker processor function
- Job name should be preserved and accessible in the worker processor function
- Shared connections should work as documented

### Actual Behavior
- Job data becomes an empty dictionary `{}`
- Job name becomes `None`
- Job processing fails due to missing data

### Minimal Reproducible Example

```python
import asyncio
import redis.asyncio as redis
from bullmq import Queue, Worker

async def demonstrate_bug():
    # Create shared Redis connection
    redis_client = redis.from_url('redis://localhost:6379')
    
    # Test data
    test_data = {'value': 42, 'message': 'test'}
    
    # Create queue with shared connection
    queue = Queue('test-queue', {'connection': redis_client})
    job = await queue.add('test-job', test_data)
    print(f"Job added with data: {job.data}")  # Shows correct data
    
    # Create worker with same shared connection
    async def processor(job, token):
        print(f"Worker received data: {job.data}")  # Shows empty {}
        print(f"Worker received name: {job.name}")   # Shows None
        return {'processed': True}
    
    worker = Worker('test-queue', processor, {'connection': redis_client})
    await asyncio.sleep(2)  # Wait for processing
    
    await worker.close()
    await redis_client.aclose()

# Run: asyncio.run(demonstrate_bug())
```

### Workaround
Using Redis URL strings instead of shared connection objects works correctly:

```python
# This works correctly
redis_url = 'redis://localhost:6379'
queue = Queue('test-queue', {'connection': redis_url})
worker = Worker('test-queue', processor, {'connection': redis_url})
```

### How to reproduce.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

mohanrajvenkatesan23-04:
Fixes #3401

### Why
When a user passes the same `redis.asyncio.Redis` instance to both `Queue` and `Worker`, `job.data` arrives as an empty dict and `job.name` as `None` inside the processor. Using Redis URL strings works fine.

Since #3887 unified Python connections on `single_connection_client=True`, `Worker.__init__` calls `RedisConnection(redis_opts)` twice — once for the regular client and once for the blocking client. When `redis_opts` is an already-constructed `redis.Redis`, `RedisConnection` reused that same instance verbatim for both connections, so the Worker's `BZPOPMIN` blocking command monopolised the single underlying socket and interleaved with / corrupted replies of subsequent commands (the `HMGET` that reads the job hash after `moveToActive`). Node avoids this by calling `.duplicate()` on externally-supplied clients (`src/classes/worker.ts`) and creating the blocking `RedisConnection` with `shared: false` (`src/classes/redis-connection.ts`).

### How
- `RedisConnection` now tracks a `shared` flag whenever the caller supplies an already-constructed Redis instance; `close()` and `disconnect()` are no-ops in that case so the caller retains ownership of their client.
- `RedisConnection` gained an `isBlocking` flag. When `isBlocking=True` and the caller supplied a Redis instance, the constructor derives a sibling client via `Redis.client()` so blocking operations run on a dedicated connection from the same pool.
- `Worker` constructs its `blockingRedisConnection` with `isBlocking=True` so `BZPOPMIN` can no longer starve the regular client's replies.

- New `python/tests/shared_connection_test.py` — 6 mock-based regression tests covering the `shared` flag, close/disconnect being no-ops, reuse of the caller's client on the non-blocking path, duplication on the blocking path, and an end-to-end assertion that `worker.client is not worker.bclient` when a shared client is passed.
- `flake8 --select=E9,F63,F7,F82 bullmq tests` is clean; existing non-integration tests still pass.
- No Node / TypeScript source was touched; commit tagged `[python]`.

### DIFF
--- a/python/bullmq/redis_connection.py
+++ b/python/bullmq/redis_connection.py
@@ -76,21 +76,15 @@ class RedisConnection:
         retry_errors = [BusyLoadingError, ConnectionError, TimeoutError]
 
         if isinstance(redisOpts, redis.Redis):
-            # The caller supplied an already-constructed Redis client. We treat
-            # this as a "shared" connection and must not close it in close().
-            # For blocking usage (e.g. the Worker's bclient) we cannot reuse the
-            # shared client directly because blocking commands like BZPOPMIN
-            # would monopolise the single underlying socket and corrupt replies
-            # of subsequent commands issued through the same client (this was
-            # the root cause of empty job.data / None job.name observed when a
-            # single redis.asyncio.Redis instance was passed to both Queue and
-            # Worker). Instead, derive a sibling client bound to a dedicated
-            # connection from the same pool via Redis.client().
-            self.shared = True
+            # Blocking commands need a dedicated socket from the caller's
+            # pool; sharing one would corrupt replies of regular commands.
             if isBlocking and hasattr(redisOpts, "client"):
                 self.conn = redisOpts.client()
             else:
                 self.conn = redisOpts
+            # Only the directly-reused caller client is "shared" — a derived
+            # sibling is owned by us and must be cleaned up in close().
+            self.shared = self.conn is redisOpts
         elif isinstance(redisOpts, dict):
             defaultOpts = {
                 "host": "localhost",
@@ -129,12 +123,9 @@ class RedisConnection:
 
     async def close(self):
         """
-        Close the connection
+        Close the connection.
 
-        When the underlying client was supplied by the caller (shared=True),
-        we must not close it, as it remains owned by the caller. This mirrors
-        the behaviour of the Node implementation's ``shared`` flag on
-        RedisConnection.
+        A shared connection is left open; it remains owned by the caller.
         """
         if self.shared:
             return None
--- a/python/bullmq/redis_connection.py
+++ b/python/bullmq/redis_connection.py
@@ -68,13 +68,29 @@ class RedisConnection:
         "canDoubleTimeout": False
     }
 
-    def __init__(self, redisOpts: Union[dict, str, redis.Redis] = {}):
+    def __init__(self, redisOpts: Union[dict, str, redis.Redis] = {}, isBlocking: bool = False):
         self.version = None
+        self.shared = False
+        self.isBlocking = isBlocking
         retry = Retry(ExponentialBackoff(cap=20, base=1), 20)
         retry_errors = [BusyLoadingError, ConnectionError, TimeoutError]
 
         if isinstance(redisOpts, redis.Redis):
-            self.conn = redisOpts
+            # The caller supplied an already-constructed Redis client. We treat
+            # this as a "shared" connection and must not close it in close().
+            # For blocking usage (e.g. the Worker's bclient) we cannot reuse the
+            # shared client directly because blocking commands like BZPOPMIN
+            # would monopolise the single underlying socket and corrupt replies
+            # of subsequent commands issued through the same client (this was
+            # the root cause of empty job.data / None job.name observed when a
+            # single redis.asyncio.Redis instance was passed to both Queue and
+            # Worker). Instead, derive a sibling client bound to a dedicated
+            # connection from the same pool via Redis.client().
+            self.shared = True
+            if isBlocking and hasattr(redisOpts, "client"):
+                self.conn = redisOpts.client()
+            else:
+                self.conn = redisOpts
         elif isinstance(redisOpts, dict):
             defaultOpts = {
                 "host": "localhost",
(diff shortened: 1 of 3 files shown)


python/bullmq/redis_connection.py
```diff
@@ -68,13 +68,29 @@ class RedisConnection:
...
+    def __init__(self, redisOpts: Union[dict, str, redis.Redis] = {}, isBlocking: bool = False):
         self.version = None
+        self.shared = False
+        self.isBlocking = isBlocking
         retry = Retry(ExponentialBackoff(cap=20, base=1), 20)
         retry_errors = [BusyLoadingError, ConnectionError, TimeoutError]
 
         if isinstance(redisOpts, redis.Redis):
-            self.conn = redisOpts
+            # The caller supplied an already-constructed Redis client. We treat
```
manast: No need for this lengthy comment.

python/bullmq/redis_connection.py
```diff
@@ -107,12 +123,21 @@ def disconnect(self):
...
+        if self.shared:
+            return None
         return self.conn.disconnect()
 
     async def close(self):
         """
         Close the connection
+
+        When the underlying client was supplied by the caller (shared=True),
+        we must not close it, as it remains owned by the caller. This mirrors
```
manast: This sounds more like a comment than actual documentation. It should not use the wording "we must not close it", but instead something along the lines: "it will not be closed, as it remains owned by the caller." or something like that.

python/bullmq/worker.py
```diff
@@ -39,7 +39,14 @@ def __init__(self, name: str, processor: Callable[[Job, str], asyncio.Future], o
         self.opts = final_opts
         redis_opts = opts.get("connection", {})
         self.redisConnection = RedisConnection(redis_opts)
-        self.blockingRedisConnection = RedisConnection(redis_opts)
+        # The blocking client must NEVER share its underlying socket with the
```
manast: unnecessary long comment.

python/tests/shared_connection_test.py
```diff
@@ -0,0 +1,148 @@
+"""
+Regression tests for issue #3401.
+
+When a user passes the SAME ``redis.asyncio.Redis`` instance to both
```
manast: We can have a new suite of tests for shared connections, but we must not have test files specific for every issue resolved as that pollutes the project. Either find an existing test suite where this test would belong to, or create a new test file for these kind of tests if they are general enough.

manast (changes requested):

mohanrajvenkatesan2304-hash:
Thanks for the review @manast! Pushed a follow-up commit addressing every point:

- **Long inline comments trimmed** in `redis_connection.py` (the block at line 79 and the `close()` docstring at line 135) and `worker.py` (line 42).
- **`close()` docstring reworded** to describe behaviour rather than prescribe action — now reads *"A shared connection is left open; it remains owned by the caller."*
- **Test file no longer issue-specific** — class renamed to `TestSharedConnection`, the docstring describes shared-connection behaviour generally, and the issue references are gone. The file is named `shared_connection_test.py` so it can serve as the home for any future shared-connection tests; happy to fold it into another existing suite if you'd prefer one.
- **Ownership fix** for Copilot's concern — only the directly-reused caller client is now flagged as `shared`. The derived blocking sibling has `shared=False`, so `worker.close()` releases it cleanly without touching the caller-owned client. Added an end-to-end assertion that confirms `worker.bclient.aclose()` runs while `caller.aclose` is left alone.

All seven shared-connection tests pass locally. Could you take another look when you have a moment? Happy to iterate on anything that still feels off.

mohanrajvenkatesan2304-hash:
Thanks for the suggestion! Pushed a follow-up commit (`0c1afcd`) adding a real-Redis integration test in `python/tests/shared_connection_test.py`:

- New `TestSharedConnectionIntegration` class (uses `unittest.IsolatedAsyncioTestCase`).
- Builds a single `redis.asyncio.Redis` client and hands it to both `Queue` and `Worker` via `opts["connection"]`.
- Adds a job with a non-trivial name and nested data, awaits the worker's processor (10s timeout), and asserts both `job.name` and `job.data` are intact — directly exercising the original interleaved-reply symptom from #3401.
- Also asserts the caller's shared client is still pingable after `worker.close()` (i.e. the worker only released its derived blocking sibling).
- `flushdb` is run on `asyncSetUp`/`asyncTearDown` so the test is hermetic alongside the existing CI Redis services.

The 7 mock-based unit tests continue to pass locally; the new integration test will run against the `redis@7-alpine`, `valkey@8`, and `dragonflydb@latest` services in CI.

Issue #3699, opened 2026-01-15
[Bug]: --inspect-publish-uid destination can be stderr or http


mdoi2:
### Version

v5.66.5

### Platform

NodeJS

### What happened?

When running queues in a sandbox with Node.js version 24.12.0 or later, the following error occurs.

```
Error [ERR_WORKER_INVALID_EXEC_ARGV]: Initiated Worker with invalid execArgv flags: --inspect-publish-uid destination can be stderr or http
    at new Worker (node:internal/worker:287:13)
    at Child.init (/path/node_modules/.pnpm/bullmq@5.66.5/node_modules/bullmq/dist/cjs/classes/child.js:74:36)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async ChildPool.retain (/path/node_modules/.pnpm/bullmq@5.66.5/node_modules/bullmq/dist/cjs/classes/child-pool.js:38:13)
    at async initChild (/path/node_modules/.pnpm/bullmq@5.66.5/node_modules/bullmq/dist/cjs/classes/sandbox.js:16:33) {
  code: 'ERR_WORKER_INVALID_EXEC_ARGV'
}
```

When checking the values of envArgv during debug execution, they appear to vary significantly depending on the Node.js version.

#### Node.js v24.11.1
```js
[
  "--watch-preserve-output",
]
```

#### Node.js v24.13.0
```js
[
  "--test-concurrency=0",
  "--heap-prof-interval=524288",
  "--max-http-header-size=16384",
  "--stack-trace-limit=10",
  "--report-signal=SIGUSR2",
  "--heapsnapshot-near-heap-limit=0",
[...]

PR #4104, opened 2026-04-24
fix(sandbox): strip node --watch flags from child execArgv

### LINKED ISSUE #1833: [Bug]: Sandboxed workers do not process jobs if running node with `--watch` flag
### Version

v.3.10.4

### Platform

NodeJS

### What happened?

If running node with `--watch` flag sandboxed workers do not pick up jobs. 

I am aware that the `--watch` flag is still experimental and this is probably rather a node issue than a bullmq issue. 
Nevertheless I want to document this behaviour for other users that are running into the same problem.
And maybe you guys have a workaround or an explanation for this issue.

### How to reproduce.

I set up a minimal example with instructions to reproduce.

https://github.com/chaselmann/bullmq-sandboxed-processor-watch-mode

#### Execute example without watch mode
```
> npm test-no-watch
> test-no-watch
> node index.js

completed inline worker
completed sandboxed worker
{ sandboxedCounts: { wait: 0, completed: 1, failed: 0, active: 0 } }
{ notSandboxedCounts: { wait: 0, completed: 1, failed: 0, active: 0 } }

```

#### Execute example with watch mode
```
> npm test-watch
> test-watch
> node --watch index.js

(node:1706604) ExperimentalWarning: Watch mode is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
completed inline worker
{ sandboxedCounts: { wait: 0, completed: 1, failed: 0, active: 1 } }
{ notSandboxedCounts: { wait: 0, completed: 2, failed: 0, active: 0 } }
```

### Relevant log output

_No response_

### Code of Conduct

mohanrajvenkatesan23-04:
Fixes #1833

### Why
When a BullMQ worker is launched with `node --watch`, the `--watch` family of flags is inherited by sandboxed child processes (and worker threads) via `process.execArgv`. Having the sandboxed runtime also enter watch mode interferes with the IPC channel between the parent worker and the processor:

- on older Node.js versions (< 24.15) this leaves jobs stuck in the `active` state (this issue), and
- on newer versions it surfaced as `ERR_WORKER_INVALID_EXEC_ARGV` crashes (now-closed issue #3699 after upstream fix in Node v24.15.0).

The existing `convertExecArgv` filter only stripped `--inspect*`, so `--watch` and friends leaked through.

### How
- Extend `convertExecArgv` in `src/classes/child.ts` with an `isWatchFlag` predicate that matches `--watch`, `--watch=*`, and `--watch-*` and drops those entries before the existing `--inspect` handling.
- Unrelated flags (including `--inspect` with its port-reassignment logic) are untouched.
- Public signature of `convertExecArgv` is unchanged; no new exports.

- New regression assertion in `tests/child-pool.test.ts` pushes the four `--watch*` variants (`--watch`, `--watch-path=...`, `--watch-preserve-output`, `--watch-kill-signal=...`) onto `process.execArgv`, asserts none end up in the forked child's `spawnargs`, and cleans up only the flags it added so neighbouring `execArgv` tests are unaffected.
- New short "Running under `node --watch`" section added to `docs/gitbook/guide/workers/sandboxed-processors.md` warning users about custom `workerForkOptions.execArgv` / `workerThreadsOptions.execArgv` overriding the filter.


mohanrajvenkatesan2304-hash:
Thanks for the review! Pushed a follow-up addressing the cleanup concern:

- The teardown now uses `lastIndexOf` instead of `indexOf` when removing the appended `--watch*` flags. Since each flag is added via `push`, `lastIndexOf` reliably targets only the entry the test added, so a pre-existing `--watch*` from a runner started with `node --watch` is left untouched.

A short comment was added so the rationale isn't lost. Happy to refactor to a snapshot/restore pattern instead if you'd prefer.

manast:

manast:

manast:

PR #4107, opened 2026-04-26
fix(queue): widen upsertJobScheduler id parameter to free string

### LINKED ISSUE #3937: [Bug]: upsertJobScheduler ID parameter is incorrectly constrained to NameType

mohanrajvenkatesan23-04:
### Port Impact Checklist

### Why

Fixes #3937.

`Queue.upsertJobScheduler`'s first parameter `jobSchedulerId` was typed as the queue's `NameType` generic. This created a hard coupling between the **scheduler's unique identifier** and the **union of job names** the queue produces. In real distributed setups the scheduler id is typically a per-instance UUID (`inst_12345`, a database row id, etc.) that has nothing to do with the job-name literal union, so any typed queue rejected the call at compile time:

```ts
type MyJobs = Job<any, any, 'generate-report'> | Job<any, any, 'send-email'>;
const queue = new Queue<MyJobs>('q');

queue.upsertJobScheduler(
  'inst_12345',                                  // ❌ TS error before this PR
  { pattern: '0 * * * *' },
  { name: 'generate-report', data: { foo: 'bar' } },
);
```

The underlying `JobScheduler.upsertJobScheduler` already types the id correctly as `string` (with the job name as a separate `N extends string = string` generic) — the bug was a leak in the public `Queue` wrapper.

### How

- `src/classes/queue.ts`: change the parameter type from `jobSchedulerId: NameType` to `jobSchedulerId: string`. This matches the underlying `JobScheduler.upsertJobScheduler` signature.
- The fallback `jobTemplate?.name ?? jobSchedulerId` is now `jobTemplate?.name ?? (jobSchedulerId as NameType)`. The cast is intentional and preserves the existing runtime behaviour: when the caller does not supply a template name, the resulting job's name is still the scheduler id. Users who want strict type safety on the produced job's name should pass `jobTemplate.name` explicitly (which the issue's example already does).
- No changes to `JobScheduler` itself — the underlying API was already correct.

- New regression test in `tests/job_scheduler.test.ts` (`when the queue has a constrained NameType`) that builds a `Queue<unknown, unknown, 'generate-report' | 'send-email'>` and calls `upsertJobScheduler` with a free-string id. Before the fix this fails to compile; after, it compiles and the runtime asserts the resulting job's name is `'generate-report'`.
- No public API behaviour change — only the type signature is widened. Existing callers keep compiling.
- Scope is purely Node.js / TypeScript types; no Python / Elixir / PHP port work needed.

### DIFF
--- a/src/classes/queue.ts
+++ b/src/classes/queue.ts
@@ -466,7 +466,7 @@ export class Queue<
    * @returns The next job to be scheduled (would normally be in delayed state).
    */
   async upsertJobScheduler(
-    jobSchedulerId: NameType,
+    jobSchedulerId: string,
     repeatOpts: Omit<RepeatOptions, 'key'>,
     jobTemplate?: {
       name?: NameType;
@@ -487,7 +487,7 @@ export class Queue<
     >(
       jobSchedulerId,
       repeatOpts,
-      jobTemplate?.name ?? jobSchedulerId,
+      jobTemplate?.name ?? (jobSchedulerId as NameType),
       jobTemplate?.data ?? <DataType>{},
       { ...this.jobsOpts, ...jobTemplate?.opts },
       { override: true },
--- a/tests/job_scheduler.test.ts
+++ b/tests/job_scheduler.test.ts
@@ -111,6 +111,37 @@ describe('Job Scheduler', () => {
     });
   });
 
+  // Regression for https://github.com/taskforcesh/bullmq/issues/3937:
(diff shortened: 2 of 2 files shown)


manast:

PR #4110, opened 2026-04-26
fix(flow-producer): honour opts.connection when first arg empty [python]

### LINKED ISSUE #3619: [Bug]: FlowProducer doesn't use provided connection string

mohanrajvenkatesan23-04:
### Port Impact Checklist

### Why

Fixes #3619.

The Python `FlowProducer` constructor has a two-arg legacy signature:

```python
def __init__(self, redisOpts: Union[dict, str] = {}, opts: QueueBaseOptions = {}):
```

The first arg is the connection, the second is queue options. Users following the `Queue` / `Worker` convention naturally pass the connection inside `opts`:

```python
FlowProducer({}, {"connection": f"redis://{host}:6379"})
```

The constructor only used the first positional arg and ignored `opts["connection"]`, so `RedisConnection({})` silently routed everything to `localhost:6379`. The existing `#TODO: pass only queueOpts, no need 2 parameters in next breaking change` comment in the constructor body acknowledges the redundancy.

### How

- `python/bullmq/flow_producer.py`: pick the connection from `opts.connection` when `redisOpts` is empty (`{}` or empty string), via `redisOpts or opts.get("connection", {})`. When both are provided the first positional arg still wins, so legacy callers observe no behavioural change.
- A short comment explains the precedence rule in place of the TODO marker (the TODO itself is preserved — it's a separate concern about cleaning up the signature in a breaking release).

- New `TestFlowProducerConstruction` unit-test class in `python/tests/flow_test.py`. It patches `RedisConnection` so the test does not need a Redis server, and asserts the connection arg for the four relevant cases:
  - positional arg only
  - `opts.connection` only (regression for #3619)
  - both supplied — positional wins
  - neither supplied — defaults to `{}`
- All four tests pass locally (`pytest tests/flow_test.py::TestFlowProducerConstruction`). `flake8 --select=E9,F63,F7,F82` is clean.
- Python-only commit; tagged `[python]` so the Node semantic-release channel skips this.

### DIFF
--- a/python/bullmq/flow_producer.py
+++ b/python/bullmq/flow_producer.py
@@ -36,7 +36,13 @@ class FlowProducer:
         """
         Initialize a connection
         """
-        self.redisConnection = RedisConnection(redisOpts)
+        # Allow the connection to be supplied either as the first
+        # positional arg (legacy signature) or via opts["connection"]
+        # (matching the Queue / Worker convention). When redisOpts is
+        # empty we fall back to opts.connection so users who follow the
+        # newer convention are not silently routed to localhost.
+        connection_opts = redisOpts or opts.get("connection", {})
+        self.redisConnection = RedisConnection(connection_opts)
         self.client = self.redisConnection.conn
         self.opts: dict = opts
         self.prefix = opts.get("prefix", "bull")
--- a/python/tests/flow_test.py
+++ b/python/tests/flow_test.py
@@ -284,5 +284,52 @@ class TestJob(unittest.IsolatedAsyncioTestCase):
         await child_queue.obliterate()
         await child_queue.close()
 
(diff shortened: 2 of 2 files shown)


roggervalf:
changed in last breaking change, connection should be provided as part of queue options

PR #4111, opened 2026-04-26
fix(flow-producer): surface aborted transactions instead of silent fail

### LINKED ISSUE #3851: [Bug]: `FlowProducer.add()` can fail silently

mohanrajvenkatesan23-04:
### Port Impact Checklist

### Why

Fixes #3851.

`FlowProducer.add()` and `FlowProducer.addBulk()` swallowed `multi.exec()` returning `null`. ioredis returns `null` from `Pipeline.exec()` whenever the transaction was aborted — for example issued against a READONLY replica, a WATCH conflict, or a pipeline-level error — and the producer would happily resolve with a half-formed `JobNode` whose `job.id` was the locally-generated UUID, pointing at nothing in Redis.

The reporter saw exactly that shape under an Upstash failover window where Redis is briefly read-only:

```
Adding jobs with FlowProducer...
Job added with FlowProducer unexpectedly: 6d82aace-3808-4750-8040-f9feb5149023
.add() did not throw but flow does not exist: undefined
```

`Queue.add()` correctly throws in the same scenario; the bug was in `flow-producer.ts:225` and the parallel block in `addBulk` at line 302.

### How

Two paths needed slightly different treatment because the API contract differs.

- **`add()`** (single flow — atomic by design): throw when `results === null`, and throw the first per-command error if any. Without this, a transaction abort or a per-command failure leaves the user with a JobNode whose UUID id is unrecoverable.
- **`addBulk()`** (partial-success semantics): an existing regression test (`should not corrupt id mapping for successful jobs when some addBulk commands fail`) explicitly relies on `addBulk` letting some root commands fail while others succeed. So we throw **only** on `results === null` (the entire transaction aborted, every flow is dead). Per-command errors continue to leave that root's `job.id` unassigned, and callers detect partial failure via `await queue.getJob(tree.job.id) === undefined` — the same contract that test asserts.

The error message names the cause:

```
Flow could not be added: Redis transaction was aborted
```

- Two new regression tests in `tests/flow.test.ts` use a vitest spy on the next `client.multi()` call to swap in a multi whose `.exec()` returns `null`. Both `add()` and `addBulk()` are asserted to throw `/transaction was aborted/`. The original Pipeline is otherwise pristine — no impact on neighbouring tests.
- The existing partial-success test (`should not corrupt id mapping...`) is preserved unchanged and still passes against the new logic.

### DIFF
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -225,12 +225,26 @@ export class FlowProducer extends EventEmitter {
         const results = (await multi.exec()) as
           | [null | Error, string | number][]
           | null;
-        const [result] = results || [];
-        if (result) {
-          const [err, jobId] = result;
-          if (!err && typeof jobId === 'string') {
-            jobsTree.job.id = jobId;
+
+        // Surface transaction-level and per-command failures instead
+        // of silently returning a half-formed JobNode (issue #3851).
+        // ioredis returns `null` from exec() when the multi was
+        // aborted (READONLY replica, WATCH conflict, pipeline error)
+        // and `[err, value]` per command otherwise.
+        if (!results) {
+          throw new Error(
+            'Flow could not be added: Redis transaction was aborted',
+          );
+        }
+        for (const [err] of results) {
+          if (err) {
+            throw err;
           }
+        }
+
+        const [, jobId] = results[0] ?? [];
+        if (typeof jobId === 'string') {
+          jobsTree.job.id = jobId;
         }
 
         return jobsTree;
@@ -302,8 +316,22 @@ export class FlowProducer extends EventEmitter {
         const results = (await multi.exec()) as
           | [null | Error, string | number][]
           | null;
+
+        // Surface a transaction-level abort (e.g. READONLY replica or
+        // pipeline error) so callers can distinguish "nothing was
+        // added" from "some flows partially failed" (issue #3851).
+        // Per-command errors are NOT thrown here on purpose: callers
+        // can detect partially-failed roots via the returned tree's
+        // missing job id (and there is an existing test that relies
+        // on that semantics for addBulk).
+        if (!results) {
+          throw new Error(
+            'Flows could not be added: Redis transaction was aborted',
+          );
+        }
+
         for (let index = 0; index < jobsTrees.length; ++index) {
-          const result = results?.[index];
+          const result = results[index];
           if (!result) {
             continue;
           }
--- a/src/classes/flow-producer.ts
+++ b/src/classes/flow-producer.ts
@@ -231,7 +231,7 @@ export class FlowProducer extends EventEmitter {
         // ioredis returns `null` from exec() when the multi was
         // aborted (READONLY replica, WATCH conflict, pipeline error)
         // and `[err, value]` per command otherwise.
-        if (!results) {
+        if (results === null) {
           throw new Error(
             'Flow could not be added: Redis transaction was aborted',
           );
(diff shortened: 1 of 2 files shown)


manast:

manast:

mohanrajvenkatesan23-04:
Thanks for the review! Pushed a follow-up commit addressing all four comments:

- `flow-producer.ts` (add and addBulk): switched the aborted-transaction guard to an explicit `results === null` check so an empty `exec()` result is no longer misclassified as an abort.
- `flow.test.ts` (both spots): now use the public `flow.client` getter instead of `(flow as any).connection.client`, dropping the `any` cast and the coupling to internals.

PR #4112, opened 2026-04-26
fix(queue-getters): filter undefined entries from getJobs result

### LINKED ISSUE #3767: [Bug]: `queue.getActive()` can sometimes contain `undefined`

mohanrajvenkatesan23-04:
### Port Impact Checklist

### Why

Fixes #3767.

`Queue#getActive` / `Queue#getWaiting` (and every other thin wrapper around `QueueGetters#getJobs`) ran each id returned by `getRanges` through `Job.fromId`. `Job.fromId` returns `undefined` when the job hash has been deleted between the `getRanges` and `fromId` calls — a real race window any time a worker or another caller cleans up a job concurrently. The gaps were left in the result array, so user code iterating the array saw a `JobBase | undefined` mix even though the TypeScript signature is `Promise<JobBase[]>`.

```ts
const jobs = await queue.getActive();
// runtime can be: [job, job, undefined]
// type says:      [JobBase, JobBase, JobBase]
```

The reporter's example worker called `queue.getActive()` from inside a job processor — a high-contention spot where this race is likely.

### How

`src/classes/queue-getters.ts`: after the `Promise.all(jobIds.map(... fromId ...))`, filter out `undefined` entries with a type-narrowing predicate before returning. The comment captures the why so future readers don't reintroduce the bug.

The fix only widens the runtime guarantee to match the existing type — no public API change. Callers continue to receive `JobBase[]`; they just stop seeing undefined holes when a concurrent cleanup runs.

- New regression test in `tests/getters.test.ts` simulates the race deterministically by deleting one job's hash directly via `client.del` while leaving its id in the waiting list. Asserts `queue.getWaiting()` returns only the surviving job and contains no `undefined`.


mohanrajvenkatesan23-04:
Thanks for the review! Pushed a follow-up commit addressing both comments:

- `queue-getters.ts`: replaced the unsound `as Promise<JobBase>` cast with `this.getJob(jobId)`, which returns `Promise<JobBase | undefined>`. The mapped array now types as `(JobBase | undefined)[]` so the trailing undefined filter narrows correctly. Runtime is unchanged.
- `getters.test.ts`: the deleted hash key now uses `queue.toKey(a.id!)` so it stays resilient to prefix/key-format changes.

manast:
closed in favor of https://github.com/taskforcesh/bullmq/pull/4300

PR #4114, opened 2026-04-26
fix(scheduler): walk past stale slots when re-upserting under same id

### LINKED ISSUE #3063: [Bug]: Removing and upserting a Job Scheduler in close succession does not create a new delayed job

mohanrajvenkatesan23-04:
### Port Impact Checklist

### Why

Fixes #3063.

When a Job Scheduler is removed via `removeJobScheduler` and re-upserted under the same id, the previously-completed jobs from the prior scheduler keep their hash keys in Redis — `removeJobScheduler` only cleans up the *currently-delayed* entry, not historical completed/failed jobs. The next `addJobScheduler` recomputes `nextMillis` and tries to place a new job at `repeat:<id>:<nextMillis>`, which collides with one of those stale hashes.

The old collision recovery walked exactly one slot ahead (`nextMillis + every`); when several consecutive `every` slots were occupied by stale completed jobs the script returned `SchedulerJobSlotsBusy` (`-11`). The reporter saw the symptom: `getJobSchedulers()` shows the scheduler, but `getDelayed()` is empty and the scheduled tasks "never run".

```ts
await queue.upsertJobScheduler('test', { every: 3 * 60_000 }, { name: 'test-job', data: { foo: 'bar' } });
// some time later — multiple iterations have completed
await queue.removeJobScheduler('test');

await queue.upsertJobScheduler('test', { every: 3 * 60_000 }, { name: 'test-job', data: { foo: 'bar' } });
// before this PR: scheduler exists, but no delayed job is created
```

### How

`src/commands/addJobScheduler-11.lua`: replace the single-retry logic in the `every` collision branch with a bounded forward walk:

```lua
local maxSlotScans = 32
local slotScanned = 0
repeat
    nextMillis = nextMillis + every
    jobId = "repeat:" .. jobSchedulerId .. ":" .. nextMillis
    jobKey = prefixKey .. jobId
    slotScanned = slotScanned + 1
until rcall("EXISTS", jobKey) == 0 or slotScanned >= maxSlotScans

if rcall("EXISTS", jobKey) == 1 then
    return -11 -- SchedulerJobSlotsBusy
end
```

The scheduler now lands at the first free `nextMillis + N * every` slot. The error code is preserved for the genuinely-contested case (32 consecutive busy slots is a real anomaly worth surfacing). 32 was chosen empirically — large enough to clear typical stale-slot accumulations but bounded so the script doesn't spin under degenerate input.

The `pattern` (cron) branch is untouched since walking forward through pattern-derived timestamps doesn't have a natural notion of "next slot".

- New regression test in `tests/job_scheduler.test.ts` (`when re-upserting a scheduler with stale completed jobs in adjacent slots`) plants three consecutive stale completed-job hashes under id `stale-test`, calls `upsertJobScheduler('stale-test', { every: 60_000 }, …)`, and asserts:
  - the returned job has an id,
  - the embedded millis is at or beyond `now + 3*every` (i.e. the new job actually skipped past the stale slots),
  - exactly one delayed job exists, and
  - that delayed job is not one of the planted stale ids.
- No JS-side changes; the fix lives entirely in the Lua source-of-truth.
- Touches the scheduler core, so the maintainer may want to confirm the 32-scan bound or swap it for a different limit.


mohanrajvenkatesan23-04:
Thanks for the review! Pushed a follow-up commit addressing both comments:

- `addJobScheduler-11.lua`: the loop's final `EXISTS` result is now captured in a local (`slotExists`) and reused for the busy-slot check, removing the redundant `EXISTS` on the same key.
- `job_scheduler.test.ts`: switched the `hset` call to the object form to match the `IRedisClient` interface and the rest of the suite.

manast:
closing in favor of https://github.com/taskforcesh/bullmq/pull/4299

PR #4121, opened 2026-04-28
fix(flow-producer): reject deduplication and debounce options (#2780)

### LINKED ISSUE #2780: [Bug]: Debounced job ID not being returned on new flow job creation
### Version

v5.12.12

### Platform

NodeJS

### What happened?

I had been using plain queue based jobs in the past, along with the debouncing feature. But recently I needed to create jobs with debouncing + parent/child jobs, so I switched to using Flows.

Now the problem that I'm facing is that when I call flowProducer.add() with the following options:

```
        name: myJobName,
        queueName: myQueueName,
        data: myJobData,
        opts: {
          debounce: { id: myDebounceId },
          removeOnComplete: 200,
          removeOnFail: 200
        },
        children: myChildJobs
      });

// jobNode.job.id here is always unique even for debounced jobs

```

The above call successfully debounces the job, but the returned jobNode.job.id is always a unique UUID identifier. If the job was debounced, then a job with the returned UUID doesn't exist in the queue, but the ID is still returned by the flowProducer.add() function. I am not specifying any job ID in the options, so I am relying on bullmq-generated job IDs.

I've been using similar code for debouncing jobs when using plain queues (instead of flows). When I call queue.add(), if the job is debounced, then the original job's ID is returned, which is very convenient. I would expect flows to behave in a similar manner, for consistency.

Reporting this as a bug, since I assume this isn't intentional behavior.

### How to reproduce.

Create a flow based job using a flow producer using a debounce ID. And then create a second job using the same debounce ID (before the first job completes). The second job created will have a unique job id (based on UUID) even though it has been debounced and no such job has actually been created in the queue.

### Relevant log output

_No response_

### Code of Conduct

mohanrajvenkatesan23-04:
## Why

`FlowProducer.add()` and `addBulk()` silently dropped `opts.deduplication` / `opts.debounce`. A dedup hit returns the *existing* job's id rather than the freshly generated one, which orphans the flow's children — there is no safe way to reconcile the two ids inside the multi-exec transaction.

Per maintainer @roggervalf on #2780:
> *"actually this won't work for a root parent either using flow producer because the parent ids should be known in each addition before running the transaction. Dedup logic can retrieve a different id than the one being generated from the producer."*
> *"this functionality is not supported... We should throw an error in this case."*

Closes #2780.

## What

- `src/classes/flow-producer.ts`: file-private `assertNoDeduplication(flow)` helper called at the top of `add()` and `addBulk()`. Throws a clear error naming the offending queue/job.
- `tests/flow.test.ts`: three regression tests covering `deduplication` on `add()`, `debounce` on `add()`, and `deduplication` mixed into `addBulk()`.

I left the `FlowJob` type alone for now — removing `debounce`/`deduplication` from the root type would be a TypeScript breaking change that warrants a separate decision. The runtime check covers JS users and any TS escape hatch.

## Test plan

- [x] `tests/flow.test.ts` adds three new cases under "when deduplication or debounce is provided to a flow"
- [ ] CI runs the existing flow test suite against `redis@7-alpine`, `valkey@8`, and `dragonflydb@latest`


mohanrajvenkatesan23-04:
Thanks for the review! Pushed a follow-up commit addressing both points:

- `src/classes/flow-producer.ts` — `assertNoDeduplication` now walks the full tree instead of only checking the root. It accepts `FlowJob | FlowChildJob` and recurses into `children`, so a `deduplication` or `debounce` option on any nested node is rejected with the same error. `FlowChildJob` omits both options at the type level, but JavaScript callers (and TS escape hatches) can still set them, and `addChildren()` would otherwise pass them straight through to `Job.addJob`.
- `tests/flow.test.ts` — added an `addBulk()` case for `debounce` mirroring the existing deduplication case, plus a new test covering nested child and grandchild nodes for both options.

matthewshober:
Hi @mohanrajvenkatesan23-04, I originally opened #3761 to request support for deduplication on parent jobs in flows, which was added by the maintainers in #3817. I reviewed your changes and have some feedback below:

Once `assertNoDeduplication` rejects `deduplication`/`debounce` on every flow node up front, it'd be worth revisiting #3817 as part of this PR too, since a few things it added no longer make sense:

- The dedup branch #3817 added to `addParentJob-6.lua` becomes dead code — that script only runs when a node has `children` (`addToWaitingChildren` is set exclusively in `FlowProducer.addNode`), and this PR now rejects dedup on any such node before the backend is ever called, so nothing can reach that branch anymore.
- The `describe('when root parent job has deduplication option', ...)` block in `tests/flow.test.ts` has tests that assert the opposite of what this PR enforces, e.g. "should return deduplicated root job id when flow has no children" and "should return deduplicated ids from addBulk when roots share deduplication id" — both expect dedup on a childless root to succeed and return the same id. Those need to be removed or rewritten to expect rejection.
- The `FlowJob`/`FlowChildJob` typings from #3817 still advertise `deduplication`/`debounce` as valid on the root. Worth updating those too so the types match the new runtime behavior rather than leaning on the runtime check alone.

matthewshober:
> Hi @mohanrajvenkatesan23-04, I originally opened #3761 to request support for deduplication on parent jobs in flows, which was added by the maintainers in #3817. I reviewed your changes and have some feedback below:
> 
> Once `assertNoDeduplication` rejects `deduplication`/`debounce` on every flow node up front, it'd be worth revisiting #3817 as part of this PR too, since a few things it added no longer make sense:
> 
> * The dedup branch [fix(flow): allow deduplication from root parent #3817](https://github.com/taskforcesh/bullmq/pull/3817) added to `addParentJob-6.lua` becomes dead code — that script only runs when a node has `children` (`addToWaitingChildren` is set exclusively in `FlowProducer.addNode`), and this PR now rejects dedup on any such node before the backend is ever called, so nothing can reach that branch anymore.
> * The `describe('when root parent job has deduplication option', ...)` block in `tests/flow.test.ts` has tests that assert the opposite of what this PR enforces, e.g. "should return deduplicated root job id when flow has no children" and "should return deduplicated ids from addBulk when roots share deduplication id" — both expect dedup on a childless root to succeed and return the same id. Those need to be removed or rewritten to expect rejection.
> * The `FlowJob`/`FlowChildJob` typings from [fix(flow): allow deduplication from root parent #3817](https://github.com/taskforcesh/bullmq/pull/3817) still advertise `deduplication`/`debounce` as valid on the root. Worth updating those too so the types match the new runtime behavior rather than leaning on the runtime check alone.

Following up on my earlier comment which was a bit outdated: #4351 already blocks dedup on parent nodes at runtime, so this PR's only new effect is blocking it on a childless root - which works today and is what #3905/#3878 fixed.

PR #4123, opened 2026-04-28
chore(utils): add type hints to utils functions [python]

yogeshwaran-c:
## Summary

Adds missing type hints to three functions in `python/bullmq/utils.py`:

- `isRedisVersionLowerThan(current_version: str | None, minimum_version: str) -> bool`
- `extract_result(job_task: asyncio.Task, emit_callback: Callable[[str, Any], None]) -> Any`
- `get_parent_key(opts: dict[str, str]) -> Optional[str]`

Also adds the necessary imports (`asyncio`, `Callable`, `Optional`).

Note: `isRedisVersionLowerThan` now treats `current_version=None` as "not lower" (returns `False`) instead of raising. This is a small but intentional runtime change: callers like `RedisConnection.version` are typed as `str | None` and previously would have crashed if checked before `getRedisVersion()` ran.

## Motivation

These functions previously lacked type annotations, while the rest of the file (e.g. `parse_json_string_values`, `object_to_flat_array`, `is_redis_cluster`, `get_cluster_nodes`, `get_node_client`) is fully typed. This change brings consistency and improves IDE support / static analysis. Apart from the documented `None`-guard above, no other runtime behavior is altered.

## Test plan

- [ ] No runtime behavior changes apart from the documented `None`-guard in `isRedisVersionLowerThan`
- [ ] Existing Python tests continue to pass

### DIFF
--- a/python/bullmq/utils.py
+++ b/python/bullmq/utils.py
@@ -1,15 +1,16 @@
+import asyncio
 import json
 import traceback
-from typing import Any
+from typing import Any, Callable, Optional
 
 import semver
 
 
-def isRedisVersionLowerThan(current_version, minimum_version):
+def isRedisVersionLowerThan(current_version: str, minimum_version: str) -> bool:
     return semver.Version.parse(current_version).compare(minimum_version) == -1
 
 
-def extract_result(job_task, emit_callback):
+def extract_result(job_task: asyncio.Task, emit_callback: Callable[[str, Any], None]) -> Any:
     try:
         return job_task.result()
     except Exception as e:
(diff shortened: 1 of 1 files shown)


yogeshwaran-c:
Thanks for catching this! You're right that the `current_version is None` short-circuit is a small runtime behavior change, not purely an annotation change. I've updated the PR description to call this out explicitly.

The `None`-handling is intentional defensive behavior that matches the new `current_version: str | None` annotation: callers like `RedisConnection.version` start out as `None` until `getRedisVersion()` populates them, so returning `False` ("not lower than the minimum") is the safest default and avoids a `TypeError` from `semver.Version.parse(None)` if the check happens to run before the version is fetched. Treating an unknown version as "not lower" also matches the optimistic semantics the rest of the codebase assumes when version info isn't yet available.

Happy to split this into two commits (annotations-only + the `None`-guard) if you'd prefer that for review clarity.

manast (approved):

manast:

PR #4124, opened 2026-04-28
docs(queue-meta): add JSDoc to QueueMeta interface and fields

yogeshwaran-c:
## Summary

Adds JSDoc comments to the `QueueMeta` interface and each of its fields in `src/interfaces/queue-meta.ts`. These values mirror the configuration applied via `Queue.setGlobalConcurrency`, `Queue.setGlobalRateLimit`, and the queue's pause/version state, all stored under the queue's Redis `meta` key.

## Changes

- Document the interface as a whole, explaining that it represents queue metadata stored under the Redis `meta` key.
- Document each field (`concurrency`, `max`, `duration`, `maxLenEvents`, `paused`, `version`) with how it is set and what it controls.

## Test plan

- [x] Documentation-only change; no runtime impact.
- [x] TypeScript types unchanged.

### DIFF
--- a/src/interfaces/queue-meta.ts
+++ b/src/interfaces/queue-meta.ts
@@ -27,12 +27,13 @@ export interface QueueMeta {
   duration?: number;
 
   /**
-   * Approximate maximum length of the queue's events stream
-   * (`XADD ... MAXLEN ~ N`). Older events are evicted in FIFO order.
+   * Maximum length of the queue's events stream. Older events are evicted
+   * in FIFO order once the stream grows beyond this size.
    *
-   * Note: this is a best-effort target — Redis's `~` (fast trim) mode
-   * may retain more events than this value. Set without `~` semantics
-   * is not exposed by BullMQ.
+   * Note: `maxLenEvents` is not a guaranteed upper bound. BullMQ relies on
+   * Redis's `XTRIM` "fast trim" (`MAXLEN ~ N`), which is a best-effort
+   * cap — the stream may retain noticeably more events than this value at
+   * any given moment.
    */
   maxLenEvents?: number;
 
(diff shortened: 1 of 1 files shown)


src/interfaces/queue-meta.ts
```diff
@@ -1,8 +1,43 @@
...
+
+  /**
+   * Length of the rate-limit window in milliseconds, paired with `max`.
+   * Set via `Queue.setGlobalRateLimit`.
+   */
   duration?: number;
+
+  /**
+   * Maximum length of the queue's events stream (`MAXLEN ~ N`).
+   * Older events are evicted in roughly-FIFO order when this is exceeded.
```
manast: I think they are evicted in FIFO order not in roughly FIFO order. However the important thing to note is that maxLength is not guaranteed, as it does a "fast trim" and thus it may keep quite many more events than the actual value, it is a best effort done by Redis XTRIM command.

src/interfaces/queue-meta.ts
```diff
@@ -1,8 +1,43 @@
...
+
+  /**
+   * True when the queue has been paused. While paused, workers will not
+   * pick up new jobs.
+   */
   paused?: boolean;
+
+  /**
+   * BullMQ version that produced this queue's data, used for compatibility
+   * checks across upgrades.
```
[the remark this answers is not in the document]
manast: no need to fix this comment

src/interfaces/queue-meta.ts
yogeshwaran-c: Updated the `maxLenEvents` docstring to drop "roughly FIFO" (events are evicted strictly in FIFO order) and to spell out that the value is not a hard cap — BullMQ uses Redis's `XTRIM` fast trim (`MAXLEN ~ N`), so the stream can retain noticeably more events than the configured value at any given moment. Pushed as `b5873ad6c9`. Left the `version` comment untouched per your other note.

manast (approved):

PR #4131, opened 2026-04-30
docs(child-pool): add JSDoc to ChildPool public methods

yogeshwaran-c:
## Summary
- Adds concise JSDoc to public methods of `ChildPool` so downstream users get IDE hints and parameter help.
- Documentation only — no runtime or API changes.

## Test plan
- [x] No code logic changed; only comments added
- [x] `npm run build` passes locally


manast:
This is a private class, and thus we need to mark it as so, otherwise users will get confused believing they can also use this functionality which is strictly internal for BullMQ.

PR #4135, opened 2026-04-30
fix(worker): guard null parent.stdout/stderr in sandboxed Child.init (#2232)

### LINKED ISSUE #2232: [Bug]: Job failed - TypeError: null is not an object (evaluating 'parent.stdout.pipe')
### Version

4.12.3

### Platform

NodeJS

### What happened?

Sandboxed processors fail, non sandboxed work just fine. After migrating from v1 to v4

### How to reproduce.

```
import { Worker, Queue } from "bullmq";
import path from "path";

const processor = path.join(__dirname, "processor.js");

async function bootBullMQ() {
	console.log("Started BULL MQ");

	const sampleName = "TEST_BULL_MQ";
	const connection = {
		host: "localhost",
		port: 6379,
	};

	const worker = new Worker(sampleName, processor, {
		connection,
		useWorkerThreads: true,
	});

	const queue = new Queue(sampleName, {
		connection,
		defaultJobOptions: {
			attempts: 3,
			backoff: {
				type: "exponential",
				delay: 1000,
			},
		},
	});

	await queue.add(
		sampleName,
		{ foo: "bar" },
		{
			repeat: {
				pattern: "*/10 * * * * *",
			},
		}
	);

	await queue.close();

	worker.on("completed", (job) => {
		console.log(`${job.id} completed`);
	});

	worker.on("failed", (job, err) => {
		console.log(`${job.id} failed - ${err.message}`);
		console.log(err);
	});

	worker.on("error", (err) => {
		console.log(`worker error - ${err.message}`);
		console.log(err);
	});
}

export { bootBullMQ };

```

### Relevant log output

```shell
Started BULL MQ
[bun] Warning: worker_threads.Worker option "execArgv" is not implemented.
[bun] Warning: worker_threads.Worker option "stdin" is not implemented.
[bun] Warning: worker_threads.Worker option "stdout" is not implemented.
[bun] Warning: worker_threads.Worker option "stderr" is not implemented.
94 |             this.removeAllListeners();
95 |         });
96 |         parent.on('error', (...args) => this.emit('error', ...args));
97 |         parent.on('message', (...args) => this.emit('message', ...args));
98 |         parent.on('close', (...args) => this.emit('close', ...args));
99 |         parent.stdout.pipe(process.stdout);
            ^
TypeError: null is not an object (evaluating 'parent.stdout.pipe')
      at /Users/monarchmaisuriya/Documents/Projects/sample/node_modules/bullmq/dist/cjs/classes/child.js:99:8
      at processTicksAndRejections (:1:2602)

repeat:79d7789469fb800f1172ca5657645b8a:1697304210000 failed - null is not an object (evaluating 'parent.stdout.pipe')
 96 |         parent.on('error', (...args) => this.emit('error', ...args));
 97 |         parent.on('message', (...args) => this.emit('message', ...args));
 98 |         parent.on('close', (...args) => this.emit('close', ...args));
 99 |         parent.stdout.pipe(process.stdout);
100 |         parent.stderr.pipe(process.stderr);
101 |         await this.initChild();
                                ^
```

### Code of Conduct

maruthang:
### Port Impact Checklist

### Why

Fixes #2232.

`Child.init()` crashes immediately on Bun with:

```
TypeError: null is not an object (evaluating 'parent.stdout.pipe')
```

Bun's `worker_threads.Worker` ignores the `stdin` / `stdout` / `stderr`
options ("`[bun] Warning: worker_threads.Worker option "stdout" is not implemented.`"),
so when BullMQ forks a sandboxed processor the resulting child has `stdout`
and `stderr` set to `null`. The current code then unconditionally calls
`parent.stdout.pipe(...)` and throws, making sandboxed processors
completely unusable under Bun.

The same null-stdio condition can also happen on Node when `stdio: 'ignore'`
is supplied via `workerForkOptions` (an existing supported option), in which
case `child_process.fork` likewise returns a child with null stdio streams
— so the bug is reachable on plain Node too, not just Bun.

### How

Two-line change in `src/classes/child.ts`: replace

```ts
parent.stdout.pipe(process.stdout);
parent.stderr.pipe(process.stderr);
```

with

```ts
parent.stdout?.pipe(process.stdout);
parent.stderr?.pipe(process.stderr);
```

Behavior is unchanged when the streams exist (the common Node default with
`stdio: 'pipe'`). When either stream is null the pipe call is simply
skipped, which is the only sensible thing to do — there is nothing to pipe.
A short comment was added pointing at issue #2232 so future readers know
why the guard is there.

A focused regression test was added at `tests/child_stdio_null.test.ts`.
It mocks `child_process.fork` to return a fake `ChildProcess` with
`stdout = null` and `stderr = null`, mirroring the Bun condition exactly,
and asserts that `child.init()` resolves cleanly. Without the guard this
test fails with the same `TypeError` users see on Bun; with the guard it
passes. The test also pins the inlined `ChildCommand.Init` /
`ParentCommand.InitCompleted` enum values so silent enum drift is caught.

- **Other ports.** This change only touches the Node sandboxed-processor
  glue (`src/classes/child.ts`). The Python, Elixir and PHP ports each
  have their own sandbox implementations and may or may not exhibit the
  same null-stdio bug; reviewing/porting that is out of scope for this PR
  and would warrant separate investigation per port. I have intentionally
  left those checkboxes unchecked.
- **Test runner notes.** The new test (`tests/child_stdio_null.test.ts`)
  uses Vitest, matching the existing test setup in this repo. Direct
  ESLint on the two changed files is clean. The full
  `tests/sandboxed_process.test.ts` suite was not exercised locally on
  Windows due to a pre-existing ESM-loader-on-Windows-paths issue
  unrelated to this change; CI will cover it.
- **Risk.** Very low — the guard only changes behavior when the stream is
  `null`, in which case the previous behavior was an unconditional crash.


maruthang:
CI note on the branch update: the single red job (`testing node@lts/*, dragonflydb@latest`) is the known-flaky `job_cancellation_advanced > should handle rapid cancellations with high concurrency` 10s timeout (817/818 passing) — unrelated to this PR's sandboxed `Child.init` guard. The same test times out intermittently on other PRs and engines.

manast (dismissed):

manast:

manast:

maruthang:
Resolved the conflict and updated the branch (`30e609a4`) — it's mergeable now.

Heads-up for transparency: the `child.ts` null-guard from this PR already landed on master via #4301 (the `parent.stdout?.pipe(...)` / `parent.stderr?.pipe(...)` guards with the same #2232 reference), and #4301 also added a `child_stdio_null.test.ts` covering the same case. The add/add conflict was between the two copies of that test; I took master's version since the assertions are identical and its `vi.mock` setup (`...actual`) is cleaner than mine. As a result the net diff of this branch vs master is now empty.

So there's nothing left to merge here in terms of code — happy to leave it for you to merge for attribution or close as covered by #4301, whichever you prefer. Thanks!

manast:
It seems like these changes have been merged in a different commit, as there are now no changed files...

Issue #2517, opened 2024-04-10
[Bug]: Queue events does not always capture all completed events


bobthekingofegypt:
### Version

v5.7.1

### Platform

NodeJS

### What happened?

We recently updated from the old bull to bullmq in a legacy project.  
With the old bull we used to do some strange wrapping and monitoring of this legacy project from our newer orchestration tool, the tool used the completed callbacks from the queues to understand when this legacy stage was finished processing everything.  This always worked fine.  After switching over to bullmq we have noticed that the orchestrator is not understanding that the queue work is complete.  We found bull had processed all the jobs fine, they have been saved of to our database no problem but the orchestrator process is always waiting for a few callbacks so that the submitted event count equals the completed callback count.  But those callbacks never arrive.  Not sure if we have done something wrong, or if this is expected behaviour.

In summary:
* lots of workers doing the work
* single process monitoring queue for completed event, submits X events, waits till X == completed callback count to stop the workers

### How to reproduce.

https://github.com/bobthekingofegypt/check_bull_complete_count

I uploaded this repo as a minimum test.  It contains a monitor, consumer and producer.  The monitor listens for the complete events, node:cluster starts up loads of workers and producer submits 1million events.  But the monitor doesn't always seem to register 1million callbacks.  I'm testing this on a 20 core machine.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

manast:
I think it is possible this has to do with the max events length, as you are processing very quickly, the Redis stream holding the events may get trimmed before the queue events manage to read the events, you can try to increase this setting to a larger value to see if it improved. https://api.docs.bullmq.io/interfaces/v5.QueueOptions.html#streams Default is 10k, you could try with 100k instead.

bobthekingofegypt:
Tried it with 100k, sadly no difference.

I originally had the reproducible test case running with random sleeps to more match our production machines throughput but when I saw the same issue without them I just removed them for simplicity.  Our production machines don't consume events super quickly; the completion monitor is attached to the end queue of a stream of processors. That queue has the task of saving to postgres, so it's throughput isn't very high.

PR #4136, opened 2026-04-30
docs(events): document QueueEvents best-effort delivery semantics (#2517)

maruthang:
### Port Impact Checklist

### Why

Refs #2517. The reporter expected `QueueEvents` to deliver every `'completed'` event for every job, and the current docs reinforce that expectation by stating that the Redis-streams backing "provides guarantees that the events are delivered and not lost". In practice the events stream is auto-trimmed on every job state transition (`XTRIM MAXLEN ~ <maxLenEvents>`, default `10000`), so under high producer rate or a slow consumer the oldest unread events are evicted by Redis before the consumer reads them. The internal `QueueMeta.maxLenEvents` doc already calls this cap "best-effort"; the user-facing guide should match, so users don't build correctness-critical logic on top of an at-most-once channel.

### How

Surgical, docs-only edit to `docs/gitbook/guide/events/README.md`. No code, no tests, no public API changes.

The two original sentences (the "guarantees that the events are delivered and not lost" paragraph and the one-line danger hint about `streams.events.maxLen`) are replaced with:

1. A revised intro paragraph that scopes the streams advantage over pub-sub to surviving short consumer disconnections, rather than implying durable delivery.
2. An expanded danger hint that:
   - Names the trim mechanism explicitly (`XTRIM MAXLEN ~ <maxLenEvents>`, default `10000`) and notes that the `~` modifier is approximate.
   - States the best-effort guarantee in bold, in line with the existing `QueueMeta.maxLenEvents` internal doc.
   - Documents the `streams.events.maxLen` lever via a link to the `QueueOptions.streams` API doc, and clarifies that increasing it only **reduces** — does not eliminate — the loss risk under sustained back-pressure.
   - Recommends per-`Worker` `'completed'` / `'failed'` listeners as the at-least-once alternative for users who need guaranteed delivery (these are driven by the worker's own active-list bookkeeping and are not subject to stream trimming), and frames `QueueEvents` as appropriate for monitoring, dashboards, and cross-worker aggregation where occasional missed events are acceptable.

- **Documentation-only.** No runtime behavior is changed. A real behavior fix — migrating `QueueEvents` to a Redis Streams consumer group with `XACK` / `XAUTOCLAIM` for at-least-once delivery — is a separate, larger design discussion and is intentionally out of scope here.
- The Python, Elixir, and PHP ports likely carry sibling docs pages with the same misleading wording. Those should be corrected in their own PRs against the respective port repositories; the Port Impact Checklist boxes are deliberately left unchecked because this PR does not touch those ports.
- No new tests; markdown-only change. Verified the rendered output (headings, link, code spans, bold) parses cleanly in the existing GitBook hint block.

### DIFF
--- a/docs/gitbook/guide/events/README.md
+++ b/docs/gitbook/guide/events/README.md
@@ -51,10 +51,14 @@ queueEvents.on(
 );
 ```
 
-The `QueueEvents` class is implemented using [Redis streams](https://redis.io/topics/streams-intro). This has some nice properties, for example, it provides guarantees that the events are delivered and not lost during disconnections such as it would be the case with standard pub-sub.
+The `QueueEvents` class is implemented using [Redis streams](https://redis.io/topics/streams-intro). Compared to standard pub-sub this is more robust across short consumer disconnections, since events buffered in the stream while the consumer is offline can still be read once it reconnects — up to the limits described below.
 
 {% hint style="danger" %}
-The event stream is auto-trimmed so that its size does not grow too much, by default it is \~10.000 events, but this can be configured with the `streams.events.maxLen` option.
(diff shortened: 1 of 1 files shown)


manast:
I am going to close this one because the text is basically slop and will just confuse instead of help.

Issue #2539, opened 2024-04-23
[Bug]: job.data does not return the typed object.


rcfrias:
### Version

v.3.6.2

### Platform

NodeJS

### What happened?

After setting up the handlers, I realize the job.data arrives untyped. This objects are instantiated from classes, so private members become public and functions undefined.

### How to reproduce.

add a job with a custom object.

### Relevant log output

```shell
job.data.getTimestamp is not a function
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rcfrias:
I solved it temporally with a static variable to re-instantiate the object, although this misses the entire point of using .
*I really wish there would be a workaround for this. (nothing to find at the docs)
TypeScript:
static serializeMessage(unserializedMessage: any): Message {
    return new Message(
      unserializedMessage.token,
      unserializedMessage.emitter_name,
      unserializedMessage.msg,
      unserializedMessage.delay,
      unserializedMessage.timestamp,
      unserializedMessage.attachments,
    );
  }

manast:
I don't think there is an issue here, can you provide specific typescript code so that we can understand what you mean?

PR #4137, opened 2026-04-30
docs(job): clarify JSON-serialization caveat for job.data (#2539)

maruthang:
### Port Impact Checklist

### Why

The reporter in #2539 expected `job.data.someMethod()` to work after passing a class instance into the queue, but methods were missing. The current JSDoc and docs do not clearly warn that `job.data` is JSON-serialized through Redis on the way in and JSON-parsed on the way out, which strips prototypes/methods. The underlying behavior is correct and unchanged — this PR is a documentation-only clarification at the canonical user-facing surfaces.

Refs #2539

### How

JSDoc edits on the four entry points where users encounter `data`:

- `Job.data` field — note that data is JSON-serialized when persisted, so only JSON-safe values survive a round trip; class instances lose their prototype/methods.
- `Job.create` — `data` param JSDoc states the same caveat.
- `Queue.add` — `data` param JSDoc states the same caveat.
- `Queue.addBulk` — JSDoc states that each job's `data` is serialized in the same way.

Plus a new section in `docs/gitbook/guide/jobs/job-data.md`:

- "Data is serialized as JSON" — explains the round trip explicitly and shows the recommended workaround pattern: producer sends a plain object, processor re-instantiates the class. Mentions `toJSON()` as an alternative when a custom serialized shape is desired.

No code changes, no test changes.

- Documentation-only PR. No runtime behavior changes.
- Verified runtime behavior in `src/classes/job.ts` (`asJSON()` / `fromJSON()` use plain `JSON.stringify` / `JSON.parse` with no `toJSON` revival hook on the parse side) so the docs match what the code actually does.
- The python/elixir/php ports likely have the same caveat in their respective docs and may warrant follow-up PRs in those directories. Out of scope for this Node-focused PR.


manast (commented):
remove dashes so that the text does not look so obviously written by an LLM.

maruthang:
Closing as superseded — master has since landed an equivalent (and slightly fuller) version of this same clarification: the `job.data` JSDoc now enumerates the JSON-compatible value types (object, array, string, number, boolean, null), and the job-data guide has the same class-reinstantiation example with the import/setup lines included. Both files on this branch are now byte-identical to master, so there's nothing left to add. Thanks!

PR #4164, opened 2026-05-12
chore: several type hints and docs

roggervalf:
### Port Impact Checklist

### Why

  1. Why is this change necessary? In order to group type hint and docs

### How

_Any extra info here._


manast (approved):
LGTM

PR #4167, opened 2026-05-13
fix(build): remove source map generation

roggervalf:
### Port Impact Checklist

### Why

### How

_Any extra info here._

### DIFF
--- a/tests/deduplication.test.ts
+++ b/tests/deduplication.test.ts
@@ -29,7 +29,11 @@ describe('deduplication', () => {
   beforeEach(async () => {
     queueName = `test-${randomUUID()}`;
     queue = new Queue(queueName, { connection, prefix });
-    queueEvents = new QueueEvents(queueName, { connection, prefix });
+    queueEvents = new QueueEvents(queueName, {
+      connection,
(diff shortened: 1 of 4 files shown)


tests/deduplication.test.ts
```diff
@@ -29,7 +29,11 @@ describe('deduplication', () => {
   beforeEach(async () => {
     queueName = `test-${randomUUID()}`;
     queue = new Queue(queueName, { connection, prefix });
-    queueEvents = new QueueEvents(queueName, { connection, prefix });
+    queueEvents = new QueueEvents(queueName, {
+      connection,
+      prefix,
+      blockingTimeout: 500,
```
roggervalf, line 35: this is another attempt to fix flaky tests, current default timeout is 10 seconds, reducing it to 500 milliseconds. QueueEvents polls the stream with XREAD BLOCK . It may be mid-block waiting when the event arrives, so it won't see it until the current block expires and a new read starts

manast (approved):

roggervalf:
ref https://github.com/taskforcesh/bullmq/pull/3720

PR #4200, opened 2026-05-27
feat(rust): initial implementation of rust support

manast:
### Port Impact Checklist

### Why

Rust grows in popularity every passing day, but most importantly it provides benefits for a library like BullMQ.  
First of all, the memory guarantees are a very nice addition for increasing the already rock solid BullMQ stability. Secondly, the Rust compiler is pretty efficient in producing highly optimised code, which shows already in early benchmarks, and also the asynchronous execution model really fits perfectly with BullMQ's concurrency model, to maximise jobs processing concurrently both for IO and CPU bound operations.

Furthermore, we can easily add new language bindings based on Rust in the future for Java, C# or Golang, instead of writing new versions of the library from scratch, and benefit from the same stability and performance as with the pure Rust implementation.

This PR also incorporates follow-up fixes from review feedback so the initial Rust API, manual-processing flows, and worker shutdown/lock behavior are more consistent and usable.

### How

Implemented the initial Rust crate and then tightened a few areas based on review feedback:

- added `Job::token()` so the worker lock token is available from the job for manual-processing flows
- fixed misplaced rustdoc around `Queue::merge_job_options` / `make_script_context`
- hardened `CancellationToken::cancelled()` to avoid missing a cancellation notification
- updated the worker lock-renewal loop to snapshot active job ids/tokens before awaiting Redis calls, avoiding holding the `RwLock` read guard across async work
- clamped `WorkerOptions::effective_lock_renew_time()` so the renewal interval is never `0ms`
- fixed manual `move_to_completed` to store `returnvalue` as JSON, keep the in-memory `Job.returnvalue` aligned with Redis, and pass the configured max attempts to `moveToFinished`
- corrected `Worker::set_concurrency()` rustdoc to match the current implementation
- woke idle worker loops during `close()` so shutdown does not hang while waiting for new jobs
- documented the required `yarn generate:raw:scripts` step before running Rust tests from a fresh checkout

- Simplified `docs/gitbook/rust/changelog.md` to only contain the `# Changelog` heading; semantic-release will populate release entries automatically.
- Fixed field-name consistency in `Job::to_redis_hash` / `from_redis_hash` (both now use `"atm"` for `attempts_made`).
- Corrected Rust code snippets in `retrying-job.md` (correct `retry` signature) and `removing-job.md` (removed erroneous extra `&` reference).
- Added targeted Rust coverage for the lock-renew-time clamp and updated the manual-completion test to assert the JSON-serialized `returnvalue`.
- The Rust workflow generates `rawScripts` before running Cargo commands; after `yarn generate:raw:scripts`, `cargo fmt --all --check`, targeted Rust tests, and `cargo build --all-targets` pass locally.


docs/gitbook/rust/changelog.md
```diff
@@ -0,0 +1,18 @@
+# Changelog
+
+All notable changes to the BullMQ Rust SDK will be documented here.
```
roggervalf: this part is not needed, semantic-release will move it at the end I think

docs/gitbook/rust/changelog.md
```diff
@@ -0,0 +1,18 @@
...
+
+Releases follow [Semantic Versioning](https://semver.org/) and are tagged as `vrsX.Y.Z`.
+
+## [Unreleased]
+
+### Added
+
+- Initial release of BullMQ for Rust
```
roggervalf: these comments can be added later or as part of the squash and merge, semantic release will add a section for v1

manast:
@copilot please address the unresolved comments.

rust/src/job.rs
```diff
@@ -0,0 +1,1013 @@
...
+        f.debug_struct("ScriptContext").finish_non_exhaustive()
+    }
+}
+
+/// Represents a job in the queue.
+///
+/// A `Job` holds the data to be processed and all associated metadata.
+/// Jobs are serialized to Redis as hash fields.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Job {
```
roggervalf, line 40: could you pls point me were token is saved in job class, we should save it at least to have it available for manual processing

manast: @copilot we will need to add something like the following to address this comment: 
```rs
/// The lock token for this job (set by the worker).
pub fn token(&self) -> Option<&str> {
    self.ctx.as_ref().map(|c| c.token.as_str())
}
```
Maybe there are more elegant ways or ways that are closer to what we have on the NodeJS BullMQ version.

manast:
@copilot address the unresolved comments.

roggervalf (dismissed):
Lgtm

manast:
@copilot please address the PR comments.

roggervalf (approved):
Lgtm

PR #4209, opened 2026-06-05
fix(node-redis): accept plain createClient result without cast (#4201)

roggervalf:
### Port Impact Checklist

### Why

This change improves the node-redis adapter typing so users can pass `redis.createClient()` directly into `createNodeRedisClient` without requiring extra casting, while keeping the adapter interface better typed after review feedback.

### How

- Kept `createNodeRedisClient` behavior aligned with accepting plain node-redis clients.
- Updated `NodeRedisRawClient.duplicate()` typing from `any` to `NodeRedisRawClient` to reduce unnecessary `any` usage and improve type safety/documentation.
- Changed `evalSha` and `eval` return types from `Promise<any>` to `Promise<unknown>` to avoid propagating `any` into the adapter surface.
- Changed `NodeRedisRawClient.options` from `Record<string, any>` to `Record<string, unknown>` to remove `any` from the public interface.
- Changed `createNodeRedisClient` parameter from `NodeRedisRawClient | any` (which collapses to `any`) to `unknown`, preventing `any` from leaking into consumers while still accepting plain `createClient()` results.
- Validated with lint/typecheck and final PR validation.

The scope remains Node.js-only and does not introduce runtime behavior changes; this is a typing-focused fix plus follow-up refinement from PR feedback.

### DIFF
--- a/src/classes/node-redis-client.ts
+++ b/src/classes/node-redis-client.ts
@@ -67,7 +67,7 @@ export interface NodeRedisRawTransaction {
 export interface NodeRedisRawClient {
   isReady: boolean;
   isOpen: boolean;
-  options?: Record<string, any>;
+  options?: Record<string, unknown>;
 
   on(event: string, listener: (...args: any[]) => void): this;
   connect(): Promise<unknown>;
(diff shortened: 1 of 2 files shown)


roggervalf:
@copilot address all pending comments

manast (approved):

PR #4234, opened 2026-06-25
refactor(backend)!: decouple Queue/Worker/Job from Redis via IQueueBackend

manast:
### Port Impact Checklist

### Why

Introduce a datastore-agnostic `IQueueBackend` abstraction so the high-level classes (Queue, Worker, Job, FlowProducer, QueueEvents, …) no longer depend on `RedisClient`/`RedisConnection` directly. The backend now owns its connection(s) and is injected through a `BackendFactory` (default `createRedisBackend`), clearing the way for non-Redis backends.

- Add `IQueueBackend` (src/interfaces/queue-backend.ts): every high-level data op plus connection lifecycle (waitUntilReady/close/disconnect/setName, on/once/removeListener for ready|error|close, minimumBlockTimeout, waitForJob, readEvents, disconnect/reconnectBlocking, forQueue, closing).
- Add `BackendFactory` type and inject it into the constructors of QueueBase, Queue, Worker, JobScheduler, QueueEvents, QueueEventsProducer and FlowProducer (defaulting to `createRedisBackend`).
- Rename `Scripts` -> `RedisQueueBackend` (the Redis adapter); it now extends EventEmitter, owns its `RedisConnection`(s), encapsulates the blocking connection, and owns the blocking `waitForJob` (capability-based timeout rounding + stuck-connection watchdog moved out of Worker).
- Add `createRedisBackend` factory (src/utils/create-backend.ts) that builds and owns the main and (optional) dedicated blocking connection.
- FlowProducer spans multiple queues over a single connection via `IQueueBackend.forQueue(queueName)`, returning a sibling backend bound to the node's keys that shares the connection.
- Remove the legacy repeatable-jobs implementation in favour of job schedulers: delete the `Repeat` class and the `repeat` option on `add`/`addBulk`; the internal repeat carrier now lives in `JobSchedulerJobOptions`.
- Slim `MinimalQueue`/`ScriptQueueContext`; the latter is now internal to the Redis adapter only.

Tests: replace removed escape hatches with test helpers (`tests/utils/get-redis-client.ts`: getRedisClient/getBlockingRedisClient/ getRedisVersion/getDatabaseType/getRedisConnection/getBlockingRedisConnection). Stabilize two pre-existing fake-timer scheduler tests (stub `worker.delay`, make the delayed job due before starting the worker / use `autorun: false`).

BREAKING CHANGE: the high-level classes no longer expose Redis internals. The optional `Connection` constructor parameter is replaced by an optional `BackendFactory`. `Queue#client`, `Queue#redisVersion`, `Queue#databaseType`, `Worker#blockingClient` and `FlowProducer#client` are removed; reach the raw Redis client via the concrete `RedisQueueBackend` returned by `getBackend()`. `Worker#waitUntilReady()` now resolves to `void` instead of the client.

Legacy repeatable jobs and their API are no longer supported. The `repeat` option on `Queue#add`/`addBulk` and the `Repeat` class are removed, along with `Queue#getRepeatableJobs`, `Queue#removeRepeatable` and `Queue#removeRepeatableByKey`. Use job schedulers (`Queue#upsertJobScheduler`, `Queue#getJobSchedulers`, `Queue#removeJobScheduler`) instead.


src/classes/job.ts
```diff
@@ -415,24 +411,24 @@ export class Job<
...
 
-    if (json.pb) {
-      job.processedBy = json.pb;
+    if (json.processedBy) {
+      job.processedBy = json.processedBy;
     }
 
-    if (json.nrjid) {
-      job.nextRepeatableJobId = json.nrjid;
+    if (json.nextRepeatableJobId) {
```
roggervalf: now that repeatables are removed, this can be renamed as nextSchedulerJobId

src/interfaces/job-json.ts
```diff
@@ -1,54 +1,30 @@
...
+  priority?: number;
   failedReason: string;
-  stacktrace: string;
+  stacktrace?: string;
   returnvalue: string;
   parent?: ParentKeys;
   parentKey?: string;
   repeatJobKey?: string;
-  nextRepeatableJobKey?: string;
+  nextRepeatableJobId?: string;
```
roggervalf, line 34: same as above

roggervalf (commented):
would be good to include the node-parser upgrade as well

manast:
@copilot address the comments.

roggervalf (approved):

PR #4256, opened 2026-07-08
fix(opts): remove debounce option in favor of deduplication

roggervalf:
### Port Impact Checklist

### Why

### How

_Any extra info here._


tests/deduplication.test.ts
```diff
@@ -45,284 +45,7 @@ describe('deduplication', () => {
     await connection.quit();
   });
 
-  describe('when job is debounced when added again with same debounce id', () => {
```
manast, line 50: do we have tests that cover this functionality and uses cases when using the deduplicationId option instead?

manast (commented):
please do not forget to add the breaking changes section when merging this PR.

tests/deduplication.test.ts
roggervalf: yes we have at the same file

manast:
I renamed the title of this PR as it mistakenly calls it "deprecate" when in fact is a "removal". Deprecation do not break backwards compatibility, it is a recommendation to stop using a feature that will be removed in newer versions.

manast (approved):

PR #4257, opened 2026-07-08
fix(telemetry): no longer optional createGauge

roggervalf:
### Port Impact Checklist

### Why

  1. Why is this change necessary? bullmq-otel has this method since v1.3.0. In order to enforcing the use of the most recent version of bullmq-otel we can remove the todo check of having createGauge as optional. Apart of that, we have another change to do to only keep only syntax for options https://github.com/taskforcesh/bullmq-otel/blob/main/src/bullMQOtel.ts#L218, at the end we will enforce the usage of bullmq-otel v2 since bullmq v6

### How

_Any extra info here._

### DIFF
--- a/src/classes/queue-getters.ts
+++ b/src/classes/queue-getters.ts
@@ -210,7 +210,7 @@ export class QueueGetters<
   }> {
     const counts = await this.getJobCounts(...types);
     const meter = this.opts.telemetry?.meter;
-    if (meter && typeof meter.createGauge === 'function') {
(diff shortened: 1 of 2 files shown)


manast (commented):
It would be great to get an explanation on why this change is needed.

roggervalf:
> It would be great to get an explanation on why this change is needed.

In order to enforce the most recent version of bullmq-otel, also we have a todo in that package for v2

PR #4259, opened 2026-07-08
fix(scheduler): remove missed repeatable references

roggervalf:
### Port Impact Checklist

### Why

### How

_Any extra info here._


manast (commented):
I think we should add a guardrail here so users using legacy repeatable jobs won't just get undefined behaviour if they upgrade to v6, instead they should at a minimum receive a raised exception. Furthermore, this PR should include some instructions on how to upgrade to v6 from v5 if you use legacy repeatable jobs. We should have a new document in docs where we summary all the changes from v5 to v6 and how to upgrade, we can call it "migrate from v5 to v6" or something.

manast:
@copilot this PR is going to land on next major release, so please address the PR comments accordingly.

manast (approved):

manast:
@copilot address PR comments.

src/classes/job-scheduler.ts
```diff
@@ -16,6 +16,40 @@ import { QueueBase } from './queue-base';
...
+  'https://docs.bullmq.io/guide/migrations/migrate-from-v5-to-v6';
+
+/**
+ * Legacy repeatable job keys use the format `name:id:endDate:tz:pattern`.
+ * The suffix after `name:id:endDate:tz:` is either:
+ * - a cron pattern (contains spaces), or
+ * - an `every` interval (purely numeric).
+ */
+export function hasLegacyRepeatableKeyShape(key: string): boolean {
+  const parts = key.split(':');
```
manast: This first smoke test needs to use the fastest way to discriminate non-repeatable keys, I am not sure split(":") is the fastest, please research a bit so that we do not keep some code that potentially slows down all jobs even the ones that are not repeatable.

manast:

manast:

manast:
@copilot in this PR we are modifying lua scripts, this means that all runtimes should be tested, I see that the Rust Test are skipped for some reason. In fact the logic should be: any lua scripts changed -> test all runtimes. If only one runtime has been change, only test that runtime. If only documentation has been changed -> do not test anything.

Issue #4242, opened 2026-06-30
Auto-generated job IDs can be stored as scientific notation by Redis Lua at 300000000


vimeh:
## Summary

BullMQ auto-generated numeric job IDs can be persisted in scientific notation by Redis Lua when the queue counter reaches `300000000`.

This can split queue state: the job hash can be keyed by `300000000`, while dedup/group/event state can point to `3e+8`.

We observed this in a production BullMQ Pro grouped + deduplicated queue with:

- `bullmq@5.46.1`
- `@taskforcesh/bullmq-pro@7.33.0`
- Redis `7.2.4`

The Pro queue made the issue visible through dedup/group state, but the same unsafe auto-ID pattern appears in the OSS BullMQ Lua scripts.

## Redis Lua reproduction

On Redis `7.2.4`:

```sh
docker run --rm -p 6391:6379 redis:7.2.4
```

```sh
redis-cli -p 6391 EVAL "
  redis.call('SET','fmt:id',299999999)
  local jobCounter = redis.call('INCR','fmt:id')
  redis.call('SET','fmt:stored',jobCounter)
  return {
    jobCounter,
    tostring(jobCounter),
    jobCounter .. '',
    redis.call('GET','fmt:stored')
  }
" 0
```

Result:

```text
300000000
300000000
300000000
3e+8
```

So `tostring(jobCounter)` and `jobCounter .. ''` are decimal, but passing the Lua number directly as a Redis command argument stores `3e+8`.

## BullMQ script pattern

In `addStandardJob-8.lua`, the auto-generated id is assigned as a Lua number:

```lua
local jobCounter = rcall("INCR", KEYS[4])

if args[2] == "" then
    jobId = jobCounter
    jobIdKey = args[1] .. jobId
else
    ...
end
```

The script only stringifies at return:

```lua
return jobId .. "" -- convert to string
```

But before that, `jobId` is passed into helpers / Redis commands, including dedup, group/list/zset, events, and job storage paths.

In our observed case:

- real job hash: `bull:<queue>:300000000`
- dedup key value: `3e+8`
- waiter/worker followed `3e+8` and found a job without the expected `data`

## Expected behavior

All internal references to an auto-generated job ID should use the same decimal string representation, e.g. `300000000`.

## Actual behavior

Some internal references can be stored as `3e+8` once the counter reaches `300000000`.

## Suggested fix

Normalize the auto-generated counter immediately after `INCR`, before any Redis command/helper receives it:

```lua
[...]

PR #4262, opened 2026-07-08
fix: fix potential job id integer to string conversion

manast:
### Port Impact Checklist

### Why

Fix for issue #4242

### How

note that this issue does not seem to be reproducible in newer versions of Redis. Nevertheless the fix will solve for the Redis versions where it is relevant.


manast:
@copilot address PR comments.

src/classes/flow-producer.ts
```diff
@@ -234,8 +233,8 @@ export class FlowProducer extends EventEmitter {
           if (typeof jobId === 'number' && jobId < 0) {
             throw this.toFlowError(jobId, parentKey);
           }
-          if (typeof jobId === 'string') {
-            jobsTree.job.id = jobId;
+          if (typeof jobId === 'string' || typeof jobId === 'number') {
```
roggervalf: if we are making sure that counters will always be converted to strings, this number check won't be needed unless we are missing a case for flow producer

manast: I am not sure that would always be the case, maybe a test is failing and thats why this check was added? Btw, what it is definitely wrong is converting jobId to a string if it is already a string!

manast:
@copilot address PR comments

roggervalf (approved):

PR #4284, opened 2026-07-13
fix(sandbox): kill children that fail init and surface refusal errors fixes #4283

### LINKED ISSUE #4283: [Bug]: Sandboxed worker: a child that fails processor init is released back into the free pool and permanently fails every subsequent job with an empty error message
### Version

v5.41.7 (we verified the affected code paths are unchanged on current `master`). We consume bullmq through `@taskforcesh/bullmq-pro`, but every affected file belongs to the base `bullmq` package — all references below point to base bullmq v5.41.7 sources.

### Platform

NodeJS

## TL;DR
If a sandboxed child fails to `import()` the processor file **once** (any transient error: ENOMEM, EACCES, a momentarily missing file...), that child becomes an immortal "zombie":
1. the child survives the failed init (it never exits — by design of `ChildProcessor.init`),
2. `ChildPool.retain()` puts the broken child **back into the free pool** (`release()` instead
   of kill/remove),
3. every subsequent job pops that child, gets refused with`cannot start a not idling child process`, **but the refusal message is lost** due to a payload-key mismatch (`err` vs `value`), so each job fails in ~10-50 ms with an **empty-message Error**.
There is no recovery path: the pool only removes children on the `exit` event, and this child never exits. 
In our production incident, one transient ENOMEM during module load turned into **41,776 failed jobs over ~38 minutes** (the fast-failing slot drains the queue and burns all retry attempts), with an empty `failedReason` that made diagnosis very hard.

### Step 1 — the child fails init but stays alive
`src/classes/child-processor.ts` — `init()`:
```ts
async init(processorFile: string) {
  let processor;
  try {
    const { default: processorFn } = await import(processorFile);  // ← transient throw here
    ...
  } catch (err) {
    this.status = ChildStatus.Errored;          // recorded ONLY inside the child process
    return this.send({
      cmd: ParentCommand.InitFailed,
      err: errorToJSON(err),                    // note: payload key is `err`
    });                                         // ← no process.exit(), no rethrow
  }
  ...
}
```
The exception is swallowed, so it is not an uncaughtException (which *would* exit via `main-base.ts`). The IPC channel plus the `message` listener keep the event loop alive, so the child process lives forever. There is also no in-band way for the parent to terminate it: `ChildCommand.Stop` is a no-op (`main-base.ts`: `case ChildCommand.Stop: break;`).

### Step 2 — the parent releases the broken child into the free pool
`src/classes/child-pool.ts` — `retain()`:
```ts
try {
  await child.init();                    // rejects with the InitFailed error
  ...
} catch (err) {
  console.error(err);
  this.release(child);                   // ← BUG: pushes the broken child into `free`
  throw err;
}

release(child: Child) {
  delete this.retained[child.pid];
  this.getFree(child.processFile).push(child);
}
```
Two aggravating properties of the pool:
* The reuse path performs **no re-init and no health/status check**:
  `let child = this.getFree(processFile).pop(); 
    if (child) { this.retained[child.pid] = child; return child; }`. 
It cannot check status even in principle — `ChildStatus` lives only inside the child process; the parent-side `Child` object has no status field. From the parent's perspective the zombie looks perfectly healthy (`exitCode === null`, not killed).
* The same catch also releases a child that **died** during init (init rejects via the close-handler), pushing an already-dead child into `free`; the next job that pops it fails with `ERR_IPC_CHANNEL_CLOSED`.
### Step 3 — every reuse is refused, and the refusal reason is silently lost
`src/classes/child-processor.ts` — `start()`:
```ts
if (this.status !== ChildStatus.Idle) {
  return this.send({
    cmd: ParentCommand.Error,
    err: errorToJSON(new Error('cannot start a not idling child process')),  // key: `err`
  });
}
```
`src/classes/sandbox.ts` — message handler (the famous line 38):
```ts
case ParentCommand.Failed:
case ParentCommand.Error: {
  const err = new Error();
  Object.assign(err, msg.value);   // ← BUG: `Error` command carries `msg.err`, not `msg.value`
  reject(err);                     //    Object.assign(err, undefined) → Error with message ''
  break;
}
```
`ParentCommand.Failed` uses the `value` key, but `ParentCommand.Error` uses the `err` key — and the sandbox reads `msg.value` for both. So the job is rejected with an Error whose `message` is `''` and whose stack points at the sandbox itself, 

e.g.:
```
Error:
    at Child.msgHandler (.../bullmq/src/classes/sandbox.ts:38:33)
```
### Step 4 — the immortal loop
```
job fetched
  → retain(): free.pop() = Errored shell   (no re-init, no status visible to parent)
  → send Start
  → child: status Errored !== Idle → send { cmd: Error, err: 'cannot start ...' }
  → sandbox: Object.assign(err, msg.value /* undefined */) → empty Error → job fails (~10-50 ms)
  → sandbox finally: child.exitCode === null → childPool.release(child)  → back to free
  → repeat forever
```
The only mechanism that ever removes a child from the pool is the `exit` event (`child.on('exit', this.remove...)`), and this child never exits. Because the broken slot "completes" jobs two orders of magnitude faster than healthy children, it vacuums the waiting queue and exhausts every job's retry attempts. The child never loaded user code, so it emits no logs and no APM signals — from the outside you only see a storm of job failures with an empty `failedReason`.

## Expected behavior
1. A child whose init failed should be killed/removed, never `release()`d into the free pool (same for a child that already exited during init).
2. If a child refuses `Start`, the reason (`cannot start a not idling child process`) should reach the job's `failedReason` — i.e. the sandbox should read the payload from the correct key (`msg.value ?? msg.err`), or `ChildProcessor` should send `Error` under `value`.

## Suggested fix
```ts
// child-pool.ts, retain() catch
catch (err) {
  console.error(err);
  this.kill(child, 'SIGKILL').catch(() => {});   // instead of this.release(child)
  throw err;
}

// sandbox.ts, msgHandler
Object.assign(err, msg.value ?? msg.err);
```

### How to reproduce.

Requires only bullmq and a local Redis at `localhost:6379`.
The key point of the repro: the processor file fails to import **only once** (simulating a transient condition such as ENOMEM). Recovery should be possible from job 2 onward — instead, every subsequent job fails instantly with an empty message.

```js
// processor.js
const { existsSync, unlinkSync } = require('fs');
const path = require('path');
const flag = path.join(__dirname, 'fail-once.flag');

if (existsSync(flag)) {
  unlinkSync(flag); // ONLY the first import fails (simulates transient ENOMEM during require)
  throw new Error('transient module load failure');
}

module.exports = async () => 'ok';
```
```js
// repro.js
const { Queue, Worker } = require('bullmq');
const { writeFileSync } = require('fs');
const path = require('path');

const connection = { host: 'localhost', port: 6379 };
const QUEUE = 'sandbox-zombie-repro';
writeFileSync(path.join(__dirname, 'fail-once.flag'), '1');

async function main() {
  const queue = new Queue(QUEUE, { connection });
  await queue.obliterate({ force: true }).catch(() => {});

  const worker = new Worker(QUEUE, path.join(__dirname, 'processor.js'), {
    connection,
    concurrency: 1,
    useWorkerThreads: false,
  });

  worker.on('completed', (job) => console.log(`job ${job.id} completed`));
  worker.on('failed', (job, err) =>
    console.log(
      `job ${job.id} failed in ${job.finishedOn - job.processedOn}ms,` +
        ` message=${JSON.stringify(err.message)}`,
    ),
  );

  for (let i = 0; i < 5; i++) {
    await queue.add('test', { i }, { attempts: 1 });
    await new Promise((r) => setTimeout(r, 1500));
  }

  await worker.close();
  await queue.close();
}

main();
```

Actual output (bullmq v5.41.7, reproduced on Node.js 22.22.0):

```
job 1 failed in 72ms, message="transient module load failure (simulating ENOMEM during require)"
job 2 failed in 7ms, message=""
job 3 failed in 1ms, message=""
job 4 failed in 3ms, message=""
job 5 failed in 3ms, message=""
```

Every empty-message failure carries this stack (dist/cjs/classes/sandbox.js:28,
which maps to src/classes/sandbox.ts:38):
    at Child.msgHandler (.../bullmq/dist/cjs/classes/sandbox.js:28:53)
    
Expected output: job 1 fails with the transient error (acceptable), jobs 2–5 are processed by a freshly forked child and complete with `'ok'`. If a child ever refuses a Start, the message `cannot start a not idling child process` should be visible in `err.message`.

### Relevant log output

```shell
Production evidence (paths shortened). The original transient failure during module load:

Error: ENOMEM: not enough memory, read
    at Object.readFileSync (node:fs:440:20)
    at Object.readFileSync (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/fs.js:262:57)
    at defaultLoadImpl (node:internal/modules/cjs/loader:1105:17)
    at loadSource (node:internal/modules/cjs/loader:1733:20)
    at Object..js (node:internal/modules/cjs/loader:1832:44)
    at Module.load (node:internal/modules/cjs/loader:1441:32)
    at Function._load (node:internal/modules/cjs/loader:1263:12)

Then 41,776 job failures over ~38 minutes, every one of them carrying this empty error:

Error:
    at Child.msgHandler (/app/node_modules/.pnpm/bullmq@5.41.7/node_modules/bullmq/src/classes/sandbox.ts:38:33)
    at Child.emit (node:events:519:28)
    at ChildProcess.<anonymous> (/app/node_modules/.pnpm/bullmq@5.41.7/node_modules/bullmq/src/classes/child.ts:117:44)
    at ChildProcess.emit (node:events:519:28)
    at emit (node:internal/child_process:949:14)
    at processTicksAndRejections (node:internal/process/task_queues:91:21)
```

### Code of Conduct

manast:
A sandboxed child whose processor init() failed was released back into the free pool instead of being killed. Since the pool only removes children on their 'exit' event and such a child never exits, it became an immortal "zombie": every subsequent job popped it, was refused with 'cannot start a not idling child process', and failed almost instantly.

The refusal reason was also lost because ParentCommand.Error carries the payload under the `err` key while the sandbox message handler only read `msg.value`, producing failures with an empty error message.

- child-pool: kill and remove a child when init() rejects instead of release()-ing it, so a fresh child is forked on the next retain.
- sandbox: read the error payload from `msg.value ?? msg.err` so the refusal/failure reason reaches failedReason.

Adds a regression test covering a transient one-time init failure and recovery on the following job, for both child processes and worker threads.

### Port Impact Checklist

### Why

### How

_Any extra info here._


roggervalf (dismissed):

miridih-jujang:
Hi @manast, thank you for the quick turnaround on #4283!

While reviewing the changes, there are a few points we would appreciate you double-checking for side effects before this ships:

1. this.kill(child, 'SIGKILL').catch(() => {}) silently swallows any kill failure, and the child never exits by itself after sending InitFailed (the open IPC channel keeps its event loop alive). If the parent-side kill ever fails, the process would linger as an invisible orphan outside the pool. A belt-and-braces process.exit() in ChildProcessor.init's catch (after the send resolves) — or at least logging kill failures instead of fully swallowing them — might be worth considering.

2. The new regression test doesn't seem to exercise the sandbox key fix (msg.value ?? msg.err): once init-failed children are killed, the ParentCommand.Error refusal path never fires, so a regression of that line would go unnoticed by the test suite. A small dedicated test (e.g., sending Start to a non-idle child and asserting the failure message) would lock it in.

3. It may be worth confirming the new test also runs under useWorkerThreads: true, since kill() takes the worker.terminate() path there.

Everything else looked correct to us: the synchronous remove() inside kill() cleans the pool maps before the throw propagates, hasProcessExited() covers the already-dead case (which also fixes dead children being pushed into the free pool), and SIGKILL is safe here since the child never loaded any user code.

Thank you again for carefully considering the side effects of this change. If there are any side effects we haven't spotted, we would appreciate you giving them additional consideration as well!

rust/CHANGELOG.md
```diff
@@ -1,6 +1,5 @@
 ## [1.0.1](https://github.com/taskforcesh/bullmq/compare/vrs1.0.0...vrs1.0.1) (2026-07-14)
 
-
 ### Performance Improvements
 
-* **worker:** do fetch next job in same finished roundtrip [rust] ([#4277](https://github.com/taskforcesh/bullmq/issues/4277)) ([677e259](https://github.com/taskforcesh/bullmq/commit/677e2590a04040ea21a0107d6bdcf5a33cf6b5e9))
+- **worker:** do fetch next job in same finished roundtrip [rust] ([#4277](https://github.com/taskforcesh/bullmq/issues/4277)) ([677e259](https://github.com/taskforcesh/bullmq/commit/677e2590a04040ea21a0107d6bdcf5a33cf6b5e9))
```
manast: do not re-format the changelog as it is automatically generated.

manast:
@copilot address my comment regarding formatting of the changelog.

roggervalf (dismissed):

roggervalf (approved):

PR #4288, opened 2026-07-14
feat: idiomatic builder-based ergonomics across the public API [rust]

manast:
Make the Rust port feel first-class for Rustaceans while staying close to the TypeScript DX. Simple paths stay one line; customization is fluent.

- Queue::add now returns an `AddJob` builder implementing `IntoFuture`, so `queue.add(name, data).delay(..).priority(..).await?` works, and the simple `queue.add(name, data).await?` still does. `data` accepts any `impl Serialize` (serialized internally, like TS).
- add_bulk takes `Vec<BulkJob>` with `BulkJob::new(..)`/`with_options(..)` and the same fluent setters.
- JobOptions/QueueOptions/WorkerOptions/RateLimiterOptions gain infallible, chainable builders (public fields kept, so struct-literals still work).
- FlowProducer gains builders: `FlowJob::new(..).options(..).add_child(..)`, plus `GetFlowOptions`, `FlowOptions`, `FlowQueueOptions`, `FlowProducerOptions` builders.
- Unified constructor convention: `new(requiredArgs)` uses defaults, `with_options(requiredArgs, opts)` is explicit (Queue/Worker/QueueEvents/FlowProducer).
- Worker processors accept a bare async closure via a new `IntoProcessor` trait (pre-boxed `ProcessorFn` still accepted).
- Normalized option type names from `Opts` to `Options`.
- Added tests/ergonomics_test.rs covering the new surface end-to-end.

BREAKING CHANGE: several signatures changed.
- `Queue::new(name, opts)` -> `Queue::with_options(name, opts)`; `Queue::new` now takes only the name.
- `Worker::new(name, processor, opts)` -> `Worker::with_options(..)`; `Worker::new(name, processor)` uses default options.
- `QueueEvents::new(name, opts)` -> `QueueEvents::with_options(name, opts)`.
- `FlowProducer::new(opts)` -> `FlowProducer::with_options(opts)`; `FlowProducer::new()` uses defaults.
- `Queue::add(name, data, opts)` -> `Queue::add(name, data)` builder.
- `Queue::add_bulk(Vec<(String, Value, Option<JobOptions>)>)` -> `add_bulk(Vec<BulkJob>)`.
- Renamed types: `ParentOpts`->`ParentOptions`, `FlowOpts`->`FlowOptions`, `GetFlowOpts`->`GetFlowOptions`; methods `add_with_opts`/`add_bulk_with_opts` -> `add_with_options`/`add_bulk_with_options`.

### Port Impact Checklist

### Why

### How

_Any extra info here._


roggervalf (dismissed):

manast:
@copilot can you please help me figure out why many checks are not running and are not allowing us to merge this PR?

manast:
@copilot while it is correct that we shall not run unnecessary tests, we may have some issue in the test.yml file because it would not let us merge as it says some checks haven't completed yet, so the gate for merging is not understanding that we have voluntarily skipped such tests.

roggervalf (approved):

PR #4298, opened 2026-07-16
fix(deps): add bullmq-otel as a peer dependency for minimum version

roggervalf:
### Port Impact Checklist

### Why

### How

_Any extra info here._


.nvmrc
```diff
@@ -0,0 +1 @@
+22.22.1
```
roggervalf, line 1: minimum node version for local, lint-staged required this version as minimum

PR #4326, opened 2026-07-22
fix(worker): emit active event for all jobs moved to active fixes #3911

### LINKED ISSUE #3911: [Bug]: Manual job processing not working as described regarding events and stalled handling
### Version

5.71.1

### Platform

NodeJS

### What happened?

This is regarding 2 separate issues, one of which I discovered while reproducing the first one.
## 1. In manual processing job stalling does not work as expected or described in the docs
In [documentation](https://docs.bullmq.io/patterns/manually-fetching-jobs#checking-for-stalled-jobs) it's explicitly stated:
> This checker is needed to move stalled jobs (whose lock has expired) back to the wait status (***or failed if they have exhausted the maximum number of stalled attempts***

But in reality, the checker moves stalled jobs to waiting status and only sets `deferredFailure` on them. They are moved to failed status only by the automatic job processing, so that never gets done here. So the documentation is incorrect and arguably functionality is too.

## 2. Certain Queue and Worker events never get triggered when they should
1. Queue `waiting` event - get's triggered only on `queue.add()` but not when as in my case job moves back to waiting due to being stalled.
2. Worker `active` event - again it seems it's only triggered in automatic processing mode, so not actually representative of the jobs state transition - `worker.getNextJob()` moves job to active, but doesn't trigger the event.

Everything can be reproduced by running the snippet below.

### How to reproduce.

```ts
import { randomUUID } from 'node:crypto';
import { Queue, Worker } from 'bullmq';
import type Redis from 'ioredis';

// Implement this to connect to Redis
declare function createConnection(): Redis;

const sleep = (interval: number = 60_000) =>
  new Promise((resolve) =>
    setTimeout(async () => {
      resolve(null);
    }, interval),
  );

async function main() {
  const conn1 = createConnection();
  const conn2 = createConnection();

  const queue = new Queue('test-queue', { connection: conn1 });

  queue.on('error', (err) => {
    console.log('QUEUE ERROR', { message: err.message });
  });
  queue.on('waiting', async (job) => {
    // NOTE: gets only called on queue.add, not status transition.
    console.log('QUEUE EVENT: waiting', { jobId: job?.id });
    await sleep(2_000);
    console.log('Attempting to fetch waiting job.');
    const fetchedJob = await worker.getNextJob(randomUUID());
    console.log('Fetched successfully', { jobId: fetchedJob.id });
  });

  const cleaned = await queue.clean(0, 100, 'wait');
  console.log(`Cleaned ${cleaned.length} jobs.`);
  const worker = new Worker('test-queue', async (_job) => {}, {
    connection: conn2,
    autorun: false,
    lockDuration: 5_000,
    stalledInterval: 2_000,
    maxStalledCount: 0,
  });

  worker.on('error', (err) => {
    console.log('WORKER EVENT: error', { message: err.message });
  });
  worker.on('stalled', async (jobId, prev) => {
    console.log('WORKER EVENT: stalled', { jobId, prev });
    await sleep(2_000);
    const counts = await queue.getJobCounts();
    console.log('Job counts', counts);
  });
  worker.on('failed', (job, err) => {
    console.log('WORKER EVENT: failed', {
      jobId: job?.id,
      attemptsMade: job?.attemptsMade,
      message: err?.message,
    });
  });
  worker.on('completed', (job) => {
    console.log('WORKER EVENT: completed', { jobId: job?.id });
  });
  worker.on('active', (job) => {
    // NOTE: never gets called
    console.log('WORKER EVENT: active', { jobId: job?.id });
  });
  worker.on('ready', () => {
    console.log('WORKER EVENT: ready');
  });

  // Start stall checker — this is the mechanism under test
  await worker.startStalledCheckTimer();
  await sleep(2_000);

  // Add job
  const job = await queue.add('test-job', {
    callId: 'test-123',
    payload: 'hello',
  });
  console.log('Job added to queue', { jobId: job.id });

  // Stop after 60 seconds
  await sleep(60_000);
  console.log('Exiting');
  process.exit();
}

main().catch((err) => {
  console.log('Error', err);
});

```

### Relevant log output

```shell
Cleaned 0 jobs.
WORKER EVENT: ready
QUEUE EVENT: waiting { jobId: '1' }
Job added to queue { jobId: '1' }
Attempting to fetch waiting job.
Fetched successfully { jobId: '1' }
WORKER EVENT: stalled { jobId: '1', prev: 'active' }
Job counts {
  active: 0,
  completed: 0,
  delayed: 0,
  failed: 0,
  paused: 0,
  prioritized: 0,
  waiting: 1,
  'waiting-children': 0
}
Exiting
```

### Code of Conduct

manast:
### Port Impact Checklist

### Why

The active emit was not emitted for every job moved to active, because the code was not considering that moveToFinish also moves the next job to active if it can.

### How

Emit the event on nextJobFromJobData.

_Any extra info here._

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -746,10 +746,6 @@ export class Worker<
       }
     }
 
-    if (job) {
-      this.emit('active', job, 'waiting');
-    }
-
     return job;
   }
 
@@ -986,6 +982,9 @@ will never work with more accuracy than 1ms. */
         // Return undefined to indicate no next job is available
         return undefined;
       }
+
+      this.emit('active', job, 'waiting');
+
       return job;
     }
   }
--- a/tests/events.test.ts
+++ b/tests/events.test.ts
@@ -438,6 +438,47 @@ describe('events', () => {
     await worker.close();
   });
 
+  it('should emit an active event for every processed job', async () => {
+    const numJobs = 20;
+
+    let activeCount = 0;
+    let completedCount = 0;
+
+    const worker = new Worker(queueName, async () => {}, {
+      connection,
+      prefix,
+    });
+
+    worker.on('active', () => {
+      activeCount++;
+    });
+
+    const completed = new Promise<void>((resolve, reject) => {
+      worker.on('completed', () => {
+        completedCount++;
+        if (completedCount === numJobs) {
+          resolve();
+        }
+      });
+      worker.on('error', reject);
+    });
+
+    await queue.addBulk(
+      Array.from({ length: numJobs }, (_, i) => ({
+        name: 'test',
+        data: { index: i },
+      })),
+    );
+
+    await completed;
+
+    // The active event must fire once per processed job, including jobs that
+    // are fetched atomically when the previous job is moved to completed.
+    expect(activeCount).toBe(numJobs);
+
+    await worker.close();
+  });
+
   describe('when one job is a parent', () => {
     it('emits waiting-children and waiting event', async () => {
       const worker = new Worker(queueName, async () => {}, {


roggervalf (approved):
Lgtm

Issue #4378, opened 2026-07-27
[Bug]: Worker permanently stops processing when blocking-connection watchdog fires during Sentinel master resolution


Steffen911:
### Version

v5.81.2

### Platform

NodeJS

### What happened (Hypothesis)?

When Redis runs behind Sentinel and the sentinels are briefly unreachable (rolling restart of the Redis statefulset, network partition, or a slow failover), a `Worker`'s blocking connection can be **permanently killed**: it lands in ioredis's terminal `"end"` state, from which ioredis never reconnects and silences all further error events. The worker keeps running (`worker.isRunning()` is `true`), emits no errors, and processes no jobs — forever. Only a process restart recovers. Queue producers on the same topology recover fine.

This is the likely root cause of #3088 and #2964 (both closed without a verified fix), and of langfuse/langfuse#13880.

Regression likely introduced in v5.7.14. Instead of crashing it's failing silently now.

**Mechanism**

1. `Worker.waitForJob` arms a watchdog before each blocking `BZPOPMIN` ([src/classes/worker.ts#L824-L826](https://github.com/taskforcesh/bullmq/blob/master/src/classes/worker.ts#L824-L826)):

   ```ts
   timeout = setTimeout(async () => {
     bclient.disconnect(!this.closing);
   }, blockTimeout * 1000 + 1000);
   ```

2. During a Sentinel outage the blocking client is in ioredis status `"connecting"` **with no live socket** while `SentinelConnector` loops trying to resolve the master. Unlike the standalone/cluster connectors (which create a socket immediately), sentinel mode spends the *entire* outage in this socketless state — on the initial connect and on every reconnect.

3. If the watchdog fires in that window, `Redis.disconnect(true)` calls `connector.disconnect()`, which sets `connecting = false`. There is no socket, so no `"close"` event is ever emitted — and ioredis's `retryStrategy` is only consulted from the `close` handler, so it never runs.

4. The in-flight sentinel resolution loop hits its `if (!this.connecting) throw` check, the pending `connect()` rejects, and ioredis executes `setStatus("end")` — terminal. From `"end"`, ioredis never reconnects, and `silentEmit` drops every subsequent `error` event, which is why the worker goes **silent** rather than noisy.

5. `waitForJob`'s loop keeps calling `bzpopmin` on the dead client; each call rejects immediately with a connection error, which `isNotConnectionError` filters out of the `error` event stream. Jobs accumulate unprocessed.

[...]

PR #4384, opened 2026-07-28
fix(worker): recover blocking client after sentinel disconnect, fixes…

manast:
… #4378

### Port Impact Checklist

### Why

There is an issue where workers stops processing jobs in certain Redis sentinel recovery scenarios.

### How

Force a reconnection if the Redis blocking client has status "end" even though the worker has not been closed or ir closing.

_Any extra info here._


src/classes/redis-connection.ts
```diff
@@ -631,7 +631,30 @@ export class RedisConnection extends EventEmitter {
...
+      ) {
+        return;
+      }
+
+      if (client.status === 'wait' || client.status === 'end') {
+        return client.connect();
+      }
+
+      try {
+        await RedisConnection.waitUntilReady(client);
```
roggervalf, line 647: if RedisConnection.waitUntilReady rejects while status is still connecting, couldn't this loop be CPU intensive. What about adding a delay per iteration?

manast: As far as I can see, "reconnect" is only called in one place in the whole codebase, and thats on the workers "whenCurrentJobsFinished", which is in turn only called from pause or close, so there is no loop that could happen where there would be unbound fast iterations as far as I can see...

manast: Anyway, I pushed a small improvement.

roggervalf (approved):

PR #4441, opened 2026-07-31
fix(events): type completed event returnvalue as deserialized value

### LINKED ISSUE #4147: [Bug]: QueueEventsListener completed event returnvalue incorrectly typed as string
### Version

5.76.5

### Platform

NodeJS

### What happened?

The completed event on QueueEvents types returnvalue as string, but the value passed at runtime is the raw return value of the job - it is already deserialized and can be any type (object, boolean, number, etc.).

## Current interface:
```typescript
completed: (
  args: { jobId: string; prev?: string; returnvalue: string },
  id: string,
) => void
```
## Expected:
returnvalue should be generic, consistent with how Job<DataType, ReturnType> is already typed elsewhere in the library:
```typescript
completed: (
  args: { jobId: string; prev?: string; returnvalue: ReturnType },
  id: string,
) => void
```

### How to reproduce.

```typescript
import { Job, Queue, QueueEvents, Worker } from 'bullmq';

const connection = { host: 'localhost', port: 6379 };

const queue = new Queue("Queue", { connection });
const events = new QueueEvents("Queue", { connection });

new Worker(
  "Queue",
  async (job: Job) => {
    return { foobar: true };
  },
  { connection },
);

events.on('completed', async (job) => {
  console.log(typeof job.returnvalue);  // "object" — not "string"
  console.log(job.returnvalue);
});

await queue.add('test', {});
```

### Relevant log output

```shell
object
{ foobar: true }
```

### Code of Conduct

toufiq-dev:
## Summary

Closes #4147

This PR fixes the `completed` event typing on `QueueEvents`. At runtime, `queue-events.ts` deserializes the job return value with `JSON.parse` before emitting it to listeners, so the `returnvalue` received by event handlers is the actual processor return value (object, number, boolean, etc.) — not the string it was previously typed as.

### Changes

- Make `QueueEventsListener` generic with a `ReturnType = any` type parameter (mirroring the existing `MinimalJob<DataType, ReturnType>` / `WorkerListener` conventions).
- Type the `completed` event as `args: { jobId: string; returnvalue: ReturnType; prev?: string }`.
- Make the `QueueEvents` class generic (`QueueEvents<ReturnType = any>`) and thread the type parameter through `emit`/`off`/`on`/`once`.
- Update the JSDoc for `completed` to reflect that `returnvalue` is deserialized from JSON.
- Update the NestJS guide docs examples that still showed `returnvalue: string`.

### Usage

```ts
const queueEvents = new QueueEvents(queueName);

queueEvents.on<QueueEventsListener<{ result: number }>, "completed">(
  "completed",
  ({ returnvalue }) => {
    // returnvalue is now typed as { result: number }
  },
);
```

### Tests

- Added a regression test in `tests/events.test.ts` (`emits completed global event with the deserialized return value`) that processes a job returning an object and asserts the `completed` event delivers the deserialized object.
- All 17 tests in `tests/events.test.ts` pass.
- `tsc`, `eslint`, and `prettier` are clean.

### DIFF
--- a/docs/gitbook/guide/nestjs/queue-events-listeners.md
+++ b/docs/gitbook/guide/nestjs/queue-events-listeners.md
@@ -12,13 +12,7 @@ import {
 @QueueEventsListener('queueName')
 export class TestQueueEvents extends QueueEventsHost {
   @OnQueueEvent('completed')
-  onCompleted({
-    jobId,
-  }: {
-    jobId: string;
-    returnvalue: string;
-    prev?: string;
-  }) {
+  onCompleted({ jobId }: { jobId: string; returnvalue: any; prev?: string }) {
     // do some stuff
   }
 }
--- a/docs/gitbook/guide/nestjs/queue-events-listeners.md
+++ b/docs/gitbook/guide/nestjs/queue-events-listeners.md
@@ -8,11 +8,19 @@ import {
   QueueEventsHost,
   OnQueueEvent,
 } from '@nestjs/bullmq';
+// Aliased because `QueueEventsListener` from '@nestjs/bullmq' is the decorator
+import type { QueueEventsListener as QueueEventsListenerType } from 'bullmq';
+
+// Use the generic to type the event payload with your job's return value
+type CompletedEvent = QueueEventsListenerType<{ result: number }>['completed'];
+
+type CompletedEventArgs = Parameters<CompletedEvent>[0];
 
 @QueueEventsListener('queueName')
 export class TestQueueEvents extends QueueEventsHost {
   @OnQueueEvent('completed')
-  onCompleted({ jobId }: { jobId: string; returnvalue: any; prev?: string }) {
+  onCompleted({ jobId, returnvalue }: CompletedEventArgs) {
+    // returnvalue is typed as { result: number }
     // do some stuff
   }
 }
--- a/docs/gitbook/bullmq-pro/nestjs/queue-events-listeners.md
+++ b/docs/gitbook/bullmq-pro/nestjs/queue-events-listeners.md
@@ -8,7 +8,7 @@ import {
   QueueEventsHost,
   OnQueueEvent,
 } from '@taskforcesh/nestjs-bullmq-pro';
-// Aliased because `QueueEventsListener` from the decorator package is the decorator
+// Aliased because `QueueEventsListener` from '@taskforcesh/nestjs-bullmq-pro' is the decorator
 import type { QueueEventsListener as QueueEventsListenerType } from 'bullmq';
 
 // Use the generic to type the event payload with your job's return value
--- a/docs/gitbook/bullmq-pro/nestjs/queue-events-listeners.md
+++ b/docs/gitbook/bullmq-pro/nestjs/queue-events-listeners.md
@@ -12,13 +12,7 @@ import {
 @QueueEventsListener('queueName')
 export class TestQueueEvents extends QueueEventsHost {
   @OnQueueEvent('completed')
-  onCompleted({
-    jobId,
-  }: {
-    jobId: string;
-    returnvalue: string;
-    prev?: string;
-  }) {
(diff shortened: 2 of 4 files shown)


docs/gitbook/guide/nestjs/queue-events-listeners.md
```diff
@@ -12,13 +12,7 @@ import {
...
 export class TestQueueEvents extends QueueEventsHost {
   @OnQueueEvent('completed')
-  onCompleted({
-    jobId,
-  }: {
-    jobId: string;
-    returnvalue: string;
-    prev?: string;
-  }) {
+  onCompleted({ jobId }: { jobId: string; returnvalue: any; prev?: string }) {
```
manast: shouldn't we want to use a generic instead of any here?

toufiq-dev: Good point — thanks! I have updated both NestJS guide pages to demonstrate the new generic instead of `any`. The example now pulls the completed event args from `QueueEventsListenerType<{ result: number }>`, so `returnvalue` is properly typed. Committed and pushed.

toufiq-dev:
Heads up on CI: the single failing check (`testing node@lts/-1, redis@7-alpine`) is an unrelated timeout in `tests/job_cancellation_advanced.test.ts > should cancel job that makes external API calls` (test timed out in 10000ms). That test file is untouched by this PR (this PR only touches `src/classes/queue-events.ts`, `tests/events.test.ts`, and two NestJS docs pages). All other jobs — including the main `node@lts/*` ioredis coverage run, smoke, node-redis, bun, valkey, dragonflydb, and postgres — passed.

manast (commented):
@copilot can you verify if this change also needs to be ported to the other runtimes? Python, Elixir, Rust and PHP?

toufiq-dev:
Thanks for the review @manast! I checked all four runtimes in the monorepo to verify whether the same issue needs porting:

- **Python** (`python/bullmq/queue_events.py`): already deserializes `completed.returnvalue` before emitting to listeners via `_JSON_DECODE_FIELDS = {"progress": "data", "completed": "returnvalue"}` + `json.loads(...)`. Since Python is dynamically typed there is no static typing bug to fix — the runtime behavior already matches the corrected Node contract. **No port needed.**

- **Rust** (`rust/src/queue_events.rs`): `QueueEvent::Completed { return_value: String }` deliberately keeps the raw JSON-encoded string (documented as "the raw (JSON-encoded) return value string", and the `parses_completed_event` test asserts the raw string). The typing here is *honest* — it really is a string at runtime — so it does not suffer from the "typed as string but actually deserialized" mismatch. Optionally parsing it into `serde_json::Value` would be a nice parity improvement, but that is a separate enhancement rather than the same bug.

- **Elixir** (`elixir/lib/bullmq/queue_events.ex`): `handle_event(:completed, %{job_id: id, returnvalue: value})` passes the raw string through without JSON decoding, and the typespec (`optional(:returnvalue) => String.t()`) matches the actual runtime value. Again no typing mismatch — a decode-parity follow-up could be considered separately.

- **PHP** (`php/src/`): there is no `QueueEvents` implementation in the PHP runtime at all (only `Job`, `Queue`, `Scripts`, etc.), so there is nothing to port.

So my take: this fix is correctly Node/TypeScript-only. The other runtimes either already deserialize (Python) or genuinely deliver a string with matching types (Rust/Elixir), and PHP has no global events API yet. Happy to open follow-up issues for Rust/Elixir parse parity if you would like.

manast (approved):

toufiq-dev:
Hi @manast  quick heads-up: the 9 CI workflows on this PR are sitting in 'awaiting approval' (they were triggered when the branch was last pushed, but never ran). Since it's a fork PR, only a maintainer can approve them. Could you 'Approve and run' on the checks when you get a chance? Once they pass, it should be ready to merge. Thanks!

PR #4460, opened 2026-08-01
ci: scope renovate prs per port

roggervalf:
### Port Impact Checklist

### Why

  1. Why is this change necessary? Instead of triggering a new version of node every time that renovate pre from other ports are merged, scope them to only affect those ports. Also fixing a character that breaks our docs generation using vite

### How

_Any extra info here._


roggervalf:
@copilot address the pending comments

manast (approved):

PR #4473, opened 2026-08-02
fix(postgres): resolve directory path portably in ESM builds

### LINKED ISSUE #4470: [Bug]: Postgres backend use `__dirname` is ESM build
### Version

v6.0.5

### Platform

NodeJS

### What happened?

When trying to use BullMQ with a Postgres backend in a ESM project it crashes when starting the app because it tries to use `__dirname` instead of `import.meta.dirname`

### How to reproduce.

Build a project using vite + nitro with type module in package.json the output will contains `__dirname` references.

### Relevant log output

Output of nitro:

```javascript
//#region node_modules/bullmq/dist/esm/postgres/sql-loader.js
/**
* Loads a migration's SQL from its `.sql` file — the portable source of truth
* shared with the Elixir/Python ports. Results are cached after the first read.
*
* The `.sql` files live next to this module under `migrations/`. The published
* build copies them alongside the compiled output (a `copy:sql` build step,
* analogous to how the Lua scripts are bundled), so the same relative lookup
* works at runtime.
*/
var MIGRATIONS_DIR = join(__dirname, "migrations");
```

Content of `dist/esm/postgres/sql-loader.js`:

```javascript
import { readFileSync } from 'fs';
import { join } from 'path';
/**
 * Loads a migration's SQL from its `.sql` file — the portable source of truth
 * shared with the Elixir/Python ports. Results are cached after the first read.
 *
 * The `.sql` files live next to this module under `migrations/`. The published
 * build copies them alongside the compiled output (a `copy:sql` build step,
 * analogous to how the Lua scripts are bundled), so the same relative lookup
 * works at runtime.
 */
const MIGRATIONS_DIR = join(__dirname, 'migrations');
```

### Code of Conduct

amandeavor:
## Problem

When using \ullmq\ with the Postgres backend in ES module environments (such as Nitro, Vite, Next.js, or Node.js ESM), starting the application causes a runtime error:

\\\
ReferenceError: __dirname is not defined in ES module scope
\\\

Root cause: \src/postgres/sql-loader.ts\ referenced \__dirname\ directly when resolving \MIGRATIONS_DIR\ and \COMMANDS_DIR\. Because TypeScript emits \__dirname\ verbatim when compiling to ESM (\dist/esm\), importing \sql-loader.js\ in an ESM context throws a \ReferenceError\.

## Solution

- Added a portable \getDirname()\ helper in \src/postgres/sql-loader.ts\ that safely resolves the current directory in both CommonJS (\dist/cjs\) and ES module (\dist/esm\) environments.
- In CommonJS mode, it uses \__dirname\.
- In ES module mode, it resolves the directory path using stack frames and \ileURLToPath\ without triggering TypeScript CJS target restrictions or runtime syntax errors.
- Added comprehensive unit tests in \	ests/postgres/sql_loader.test.ts\.

## Verification

- \
px vitest run --config vitest.postgres.config.ts tests/postgres/sql_loader.test.ts\ (3 tests passed)
- \
pm run tsc:all && npm run copy:sql\ (build succeeded cleanly for both CJS and ESM outputs)
- Verified importing both \dist/cjs/postgres/sql-loader.js\ and \dist/esm/postgres/sql-loader.js\ in Node.js.

Fixes #4470

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -42,7 +42,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@d1ba80a13dd99fba24a470575428917156a28b43 # v4.37.5
+        uses: github/codeql-action/init@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
@@ -53,7 +53,7 @@ jobs:
       # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
       # If this step fails, then you should remove it and run the build manually (see below)
       - name: Autobuild
-        uses: github/codeql-action/autobuild@d1ba80a13dd99fba24a470575428917156a28b43 # v4.37.5
+        uses: github/codeql-action/autobuild@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
 
       # ℹ️ Command-line programs to run using the OS shell.
       # 📚 https://git.io/JvXDl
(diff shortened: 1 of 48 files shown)


manast:
Thanks for the PR. The comment provided by copilot makes sense to me, better an explicit error than failing silently.

manast (dismissed):

manast:

manast:
unfortunately we are not able to resolve this conflicts as only the author of the PR can do it.

roggervalf:
@manast I believe this was handled by copilot, closing this one for now

PR #4492, opened 2026-08-04
ci(renovate): separate dev dependency prs

roggervalf:
### Port Impact Checklist

- [x] **Python** – does this change need to be ported or documented in the Python library?
- [x] **PHP** – does this change need to be ported or documented in the PHP library?
- [x] **Rust** – does this change need to be ported or documented in the Rust library?
- [x] **.NET** – does this change need to be ported or documented in the .NET library?

### Why

  1. Why is this change necessary? Commits not including port names were added in changelog. Replace [^]* with .* in the pattern. The [^] (empty negated character class matching any character including newlines) is a JavaScript-specific extension that breaks in some regex engines used by conventional-changelog. Since commit subjects are single-line, .* is correct and portable. Additionally this pr separates dev dependency upgrades by renovate in their own prs that should not trigger new releases

### How

_Any extra info here._

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,7 @@
               "release": false
             },
             {
-              "header": "**/*\\(dotnet\\)*",
+              "header": "**/*\\[dotnet\\]*",
               "release": false
             }
           ]
@@ -210,7 +210,7 @@
         {
           "preset": "angular",
           "presetConfig": {
-            "ignoreCommits": "(\\[python\\]|\\[elixir\\]|\\[php\\]|\\[rust\\]|\\[dotnet\\]|\\(dotnet\\))"
+            "ignoreCommits": "(\\[python\\]|\\[elixir\\]|\\[php\\]|\\[rust\\]|\\[dotnet\\])"
           }
         }
       ],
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -199,7 +199,7 @@
               "release": false
             },
             {
-              "header": "**/*\\(dotnet\\)*",
+              "header": "**/*\\[dotnet\\]*",
```
roggervalf, line 202: bad notation, should be []

package.json
```diff
@@ -210,7 +210,7 @@
         {
           "preset": "angular",
           "presetConfig": {
-            "ignoreCommits": "(\\[python\\]|\\[elixir\\]|\\[php\\]|\\[rust\\]|\\[dotnet\\]|\\(dotnet\\))"
+            "ignoreCommits": "(\\[python\\]|\\[elixir\\]|\\[php\\]|\\[rust\\]|\\[dotnet\\])"
```
roggervalf, line 213: we shouldn't skip () notation

manast (approved):

PR #4507, opened 2026-08-05
ci(docs): upgrade vitepress to fix docs

roggervalf:
### Port Impact Checklist

### Why

  1. Why is this change necessary? Docs are failing because resolution of vite is incompatible with current version of vitepress, so we need to upgrade vitepress to latest alpha version

### How

_Any extra info here._

### DIFF
--- a/php/package.json
+++ b/php/package.json
@@ -12,8 +12,6 @@
     "semantic-release:debug": "semantic-release --no-ci --dry-run --debug"
   },
   "devDependencies": {
-    "@commitlint/cli": "21.2.1",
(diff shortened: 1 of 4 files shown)


php/package.json
```diff
@@ -12,8 +12,6 @@
     "semantic-release:debug": "semantic-release --no-ci --dry-run --debug"
   },
   "devDependencies": {
-    "@commitlint/cli": "21.2.1",
```
roggervalf, line 15: php doesn't need this dependency as the other ports, we use the one in the root

manast (approved):

PR #4513, opened 2026-08-05
fix: make ioredis optional

### LINKED ISSUE #4511: [Bug]: Using node-redis, ioredis must be installed ?
### Version

v6.0.8

### Platform

NodeJS

### What happened?

Error: Cannot find module 'ioredis'

### How to reproduce.

Force dependency on 'ioredis/built/utils'

### Relevant log output

```shell

```

### Code of Conduct

manast:
fixes #4511

### DIFF
--- a/src/classes/ioredis-client.ts
+++ b/src/classes/ioredis-client.ts
@@ -1,20 +1,5 @@
+import type { Cluster, Redis, ChainableCommander } from 'ioredis';
 import { IRedisClient, IRedisTransaction } from '../interfaces/redis-client';
-
-type RedisTransactionLike = {
-  hset(...args: any[]): any;
-  hscan(...args: any[]): any;
-  sscan(...args: any[]): any;
-  [key: string]: any;
-};
-
-type RedisClientLike = {
-  isCluster?: boolean;
-  options?: any;
-  pipeline(...args: any[]): RedisTransactionLike;
-  multi(...args: any[]): RedisTransactionLike;
-  duplicate(...args: any[]): any;
-  [key: string]: any;
-};
 
 /**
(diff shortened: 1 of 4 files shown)


src/classes/ioredis-client.ts
```diff
@@ -1,4 +1,4 @@
-import { Cluster, Redis, ChainableCommander } from 'ioredis';
+import type { Cluster, Redis, ChainableCommander } from 'ioredis';
```
roggervalf, line 1: Even this type reference will fail if no ioredis is installed when using typescript

manast: I was thinking that maybe we could just install ioredis types, but it is not possible to separate code from types of course...

manast: But since we are loading ioredis dynamically now, maybe this will work actually.

manast: @copilot can you verify this is the case?

manast (commented):
I think the concrete implementation of the redis generic interface should be allowed to import IORedis, same for the node-redid implementation and the BunJS implementation, They need to import the proper types so that they can benefit from typescript type checking, etc. What we probably need to do is to dynamically load the actual implementation when we know which adapter the user is going to use. If this is not possible, then the only way to solve this issue is by having a breaking change where IORedis does not become a default adapter if the user does not provide one, and instead make mandatory the use of the adaptor in the constructors. This would be a breaking change but it is also a bug in my opinion, as v6 should have included this behaviour in v6.0.0.

manast:
@copilot address this comment: https://github.com/taskforcesh/bullmq/pull/4513#pullrequestreview-4882228903

roggervalf (approved):

PR #3909, opened 2026-03-27
feat: add generic ProgressType parameter to Job and Worker

claygeo:
## Problem

The `JobProgress` type is hardcoded as `string | boolean | number | object`, so there is no way to get type-safe progress when calling `updateProgress()` or listening to the `progress` event:

```typescript
// Current behavior — progress is always `string | boolean | number | object`
worker.on('progress', (job, progress) => {
  progress.percent; // TS error: Property 'percent' does not exist on type 'JobProgress'
});
```

Requested in #3721 and discussed in #3184.

## Solution

Add an optional 4th generic type parameter `ProgressType` to `Job`, `MinimalJob`, `Worker`, `WorkerListener`, and `Processor`. It is constrained `extends JobProgress` and defaults to `JobProgress`, making this a **fully backwards-compatible** change.

### Usage

```typescript
type MyProgress = { percent: number; message: string };

const worker = new Worker<InputType, OutputType, string, MyProgress>(
  'myQueue',
  async job => {
    await job.updateProgress({ percent: 50, message: 'halfway' }); // type-safe
    return { result: 'done' };
  },
  { connection },
);

worker.on('progress', (job, progress) => {
  console.log(progress.percent); // number
  console.log(progress.message); // string
});
```

### Backwards compatibility

Existing code using up to 3 generics compiles unchanged — `ProgressType` defaults to `JobProgress`:

```typescript
const worker = new Worker<InputType, OutputType, string>('queue', processor, opts);
```

## Changes

| File | Change |
|------|--------|
| `src/interfaces/minimal-job.ts` | `ProgressType extends JobProgress = JobProgress` 4th generic; typed `progress` + `updateProgress()` |
| `src/classes/job.ts` | `ProgressType` generic on `Job`; typed `progress` field and `updateProgress()` |
| `src/classes/worker.ts` | `ProgressType` threaded through `Worker`, `WorkerListener` (incl. `active`/`completed`/`failed`/`progress`), and the processor flow |
| `src/types/processor.ts` | `ProgressType` 4th generic so the processor's `job.updateProgress()` is typed |
| `tests/worker.test.ts` | Runtime integration test for typed progress (in the worker suite, per review) |

## Design / scope notes

[...]

PR #4529, opened 2026-08-07
feat: add generic ProgressType parameter to Job and Worker

### LINKED ISSUE #3721: [Feature]: add generic type for job progress
Unless I am mistaken, the progress type is hardcoded to `type JobProgress = string | boolean | number | object`. 

It would be very nice to be able to specify the type. Maybe a 4th parameter to the `Worker` class?

```ts
const worker = new Worker<Input, Output, string, { percent: number; msg: string }>(...);

// TypeScript should correctly infer 'progress' here
worker.on('progress', (job, progress) => {
    console.log(progress.msg); // OK
});
```

skrukwa:
The progress type was hardcoded to JobProgress
(string | boolean | number | object), so there was no way to get type-safe progress when calling updateProgress() or when listening to the progress event.

Add an optional ProgressType generic parameter, constrained to JobProgress and defaulting to JobProgress, to MinimalJob, Job, Processor, Worker and WorkerListener. Existing code that passes fewer generics keeps compiling unchanged.

Because ProgressType is constrained to JobProgress, Job<..., P> stays assignable to the internal MinimalJob helpers, which never read or write progress, so the generic does not need to be threaded through them. This also keeps the persistence contract intact, since progress is ultimately serialized before being stored in the backend.

On Worker the parameter is added after the backend generic B so that the existing public signature is not broken.

Sandboxed processors keep JobProgress by design, as progress crosses a process boundary where the static type is lost. Queue side getters such as getJob and Job.fromId also keep the default.

Closes #3721

---

### Port Impact Checklist

TypeScript types only — no runtime or Lua changes, nothing to port.

### Why

`JobProgress` is hardcoded as `string | boolean | number | object`, so progress can't be type-safe:

```typescript
worker.on('progress', (job, progress) => {
  progress.percentage; // TS error: Property 'percentage' does not exist on type 'JobProgress'
});
```

Closes #3721 (also discussed in #3184).

Continues #3909 by @claygeo, closed when the head repo was deleted. Rebased onto current `master` and reworked for the v6 backend generic. @claygeo credited as co-author.

### How

Add an optional `ProgressType` generic to `MinimalJob`, `Job`, `Processor`, `Worker` and `WorkerListener`, constrained `extends JobProgress` and defaulting to `JobProgress`. Existing code with fewer generics compiles unchanged.

```typescript
type MyProgress = { percentage: number; message: string };

const worker = new Worker<MyData, MyReturn, string, RedisQueueBackend, MyProgress>(
  'myQueue',
  async job => {
    await job.updateProgress({ percentage: 50, message: 'halfway' }); // type-safe
  },
  { connection },
);

worker.on('progress', (job, progress) => {
  console.log(progress.percentage, progress.message); // number, string
});
```

Two decisions worth noting:

- **The constraint replaces internal plumbing.** `ProgressType extends JobProgress` keeps `Job<…, P>` assignable to the internal `MinimalJob<…, JobProgress>` helpers, which never touch `progress`. So `Scripts`/`Backoffs` need no threading and are untouched.
- **`ProgressType` sits after `B` on `Worker`.** The 4th slot is taken by the v6 backend generic, so adding it 5th avoids breaking the v6 signature. On `Job`/`MinimalJob`/`Processor` it is purely additive as the 4th.

**Ordering is open.** Cost of the above is that users must name the backend to reach `ProgressType`. Happy to move it to `Worker`'s 4th slot instead if you'll take it as a breaking change.

**Intentionally out of scope:**

- Sandboxed processors keep `JobProgress` — the value crosses a process boundary where the static type is lost.
- `getJob`, `Job.fromId`/`fromJSON` keep the default; threading `ProgressType` through the getter surface can be a follow-up.
- `progress` still defaults to `0`, typed as `ProgressType` with the initial value cast rather than widened to `ProgressType | number`, which would defeat the generic at every read.

**Verified:** `tsc --noEmit` clean on both tsconfigs; eslint, prettier and madge clean; `tests/worker.test.ts` 123/123 and the `job`/`queue`/`events`/`flow` suites 198/198 against Redis 8. Type safety checked with `@ts-expect-error` assertions (narrowed type in the processor and the `progress`/`active`/`completed`/`failed` handlers, `bigint` rejected by the constraint, 0- and 3-generic usage still compiles).

Reproduce with `yarn && yarn pretest && npx vitest run --no-file-parallelism tests/worker.test.ts`.

### DIFF
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -55,13 +55,17 @@ export interface WorkerListener<
   DataType = any,
   ResultType = any,
   NameType extends string = string,
+  ProgressType extends JobProgress = JobProgress,
 > extends IoredisListener {
   /**
    * Listen to 'active' event.
    *
    * This event is triggered when a job enters the 'active' state.
    */
-  active: (job: Job<DataType, ResultType, NameType>, prev: string) => void;
+  active: (
+    job: Job<DataType, ResultType, NameType, ProgressType>,
+    prev: string,
+  ) => void;
 
   /**
    * Listen to 'closed' event.
@@ -83,7 +87,7 @@ export interface WorkerListener<
    * This event is triggered when a job has successfully completed.
    */
   completed: (
-    job: Job<DataType, ResultType, NameType>,
+    job: Job<DataType, ResultType, NameType, ProgressType>,
     result: ResultType,
     prev: string,
   ) => void;
@@ -112,7 +116,7 @@ export interface WorkerListener<
    * reaches the stalled limit and it is deleted by the removeOnFail option.
    */
   failed: (
-    job: Job<DataType, ResultType, NameType> | undefined,
+    job: Job<DataType, ResultType, NameType, ProgressType> | undefined,
     error: Error,
     prev: string,
   ) => void;
@@ -133,8 +137,8 @@ export interface WorkerListener<
    * world.
    */
   progress: (
-    job: Job<DataType, ResultType, NameType>,
-    progress: JobProgress,
+    job: Job<DataType, ResultType, NameType, ProgressType>,
+    progress: ProgressType,
   ) => void;
 
   /**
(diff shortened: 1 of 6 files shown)


src/classes/worker.ts
```diff
@@ -180,12 +184,21 @@ export interface WorkerListener<
  * As soon as the class is instantiated and a connection to Redis is established
  * it will start processing jobs.
  *
+ * @typeParam DataType - The type of the data that the job will process.
+ * @typeParam ResultType - The type of the result of the job.
+ * @typeParam NameType - The type of the name of the job.
+ * @typeParam B - The queue backend used by this worker.
+ * @typeParam ProgressType - The type of the job progress. Defaults to
+ * {@link JobProgress}. Must be JSON-serializable, as progress is persisted
```
roggervalf, line 192: no provided, better to use a real link here and might be better to clarify this at job class method https://github.com/taskforcesh/bullmq/blob/master/src/classes/job.ts#L542

roggervalf (approved):

src/classes/worker.ts
roggervalf: It'll be addressed apart

PR #4535, opened 2026-08-08
feat(queue): support getCountsPerPriority method [elixir] [dotnet]

roggervalf:
### Port Impact Checklist

### Why

### How

_Any extra info here._


rust/src/scripts.rs
```diff
@@ -254,8 +254,8 @@ impl ScriptRegistry {
             ("paginate", 1, include_str!("commands/paginate-1.lua")),
             (
                 "getCountsPerPriority",
-                4,
-                include_str!("commands/getCountsPerPriority-4.lua"),
+                2,
+                include_str!("commands/getCountsPerPriority-2.lua"),
             ),
```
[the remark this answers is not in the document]
roggervalf: on other ports, logic of script doesn't change only removing extra pause key

elixir/lib/bullmq/scripts.ex
```diff
@@ -1456,12 +1456,12 @@ defmodule BullMQ.Scripts do
       Keys.paused(ctx),
       Keys.meta(ctx),
       Keys.prioritized(ctx),
-      Keys.pc(ctx),
```
roggervalf, line 1459: wrong keys being provided

elixir/lib/bullmq/scripts.ex
```diff
@@ -1456,12 +1456,12 @@ defmodule BullMQ.Scripts do
...
+      Keys.events(ctx),
+      Keys.delayed(ctx),
       Keys.marker(ctx),
-      Keys.events(ctx)
     ]
 
-    args = [if(paused?, do: "paused", else: "resumed")]
+    args = [if(paused?, do: "paused", else: "resumed"), '1']
```
roggervalf: missing to emit event

elixir/lib/bullmq/scripts.ex
```diff
@@ -2106,7 +2127,7 @@ defmodule BullMQ.Scripts do
       %{}
       |> maybe_add_opt("del", Map.get(job_opts, :delay) || Map.get(job_opts, "delay") || delay, 0)
       |> maybe_add_opt(
-        "pri",
+        "priority",
```
roggervalf, line 2130: we do not save short version of priority in node

src/commands/getCountsPerPriority-2.lua
```diff
@@ -3,17 +3,14 @@
...
+      KEYS[2] prioritized key
 
       ARGV[1...] priorities
 ]]
 local rcall = redis.call
 local results = {}
 local waitKey = KEYS[1]
-local pausedKey = KEYS[2]
-local prioritizedKey = KEYS[4]
+local prioritizedKey = KEYS[2]
```
manast, line 13: This must have been broken in all ports as it is using an invalid key, are we lacking tests to cover for this functionality?

rust/src/scripts.rs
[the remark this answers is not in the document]
manast: yes, but was this functionality actually working on all the ports?

src/commands/getCountsPerPriority-2.lua
roggervalf: I reviewed all ports:
* dotnet: added missing test cases and missed getCountsPerPriority method
* php: it´s covered by test cases in pause and getCountsPerPriority
* python: same as above
* rust: same as above

manast (approved):

manast (approved):

PR #4550, opened 2026-08-11
fix(worker): add timeout to blockingConnection disconnect to prevent pause() hang

### LINKED ISSUE #4065: `worker.pause()` hangs indefinitely when pausing multiple workers connected to the same Redis — `blockingConnection.disconnect()` never emits `end`
`worker.pause()` hangs indefinitely when pausing multiple workers connected to the same Redis — `blockingConnection.disconnect()` never emits `end`

## Description

When pausing 5 BullMQ workers sequentially (each connected to the same Redis instance), the 4th `worker.pause()` call hangs indefinitely. The first 3 workers pause in ~104ms each. The 4th never resolves.

### Root cause

`worker.pause()` calls `whenCurrentJobsFinished()` (worker.ts:809), which calls `this.blockingConnection.disconnect(true)` (redis-connection.ts:185). The `disconnect(wait=true)` path calls `client.disconnect()` on the ioredis client and then awaits the `end` event:

```js
const disconnecting = new Promise((resolve, reject) => {
    client.once('end', resolve);
    client.once('error', reject);
});
client.disconnect();
await disconnecting;  // <-- hangs here for the 4th connection
```

The ioredis `end` event never fires for the 4th+ connection. There is no timeout on this await, so `pause()` hangs forever.

### Reproduction

- BullMQ v5.73.1, ioredis v5.x, Redis 7.4.8 (Docker container)
- 5 workers, each with their own blocking connection to the same Redis
- All queues have 0 active/waiting/delayed jobs
- Call `worker.pause()` sequentially on each worker
- The 4th call never resolves (which specific worker hangs varies between runs)

```js
// This hangs on the 4th worker
for (const worker of [w1, w2, w3, w4, w5]) {
    await worker.pause(); // 4th call never resolves
}
```

### Diagnostic output

```
pausing worker1... (running=true paused=false waiting=true)
worker1.blockingConnection.disconnect(wait=true) called
worker1.blockingConnection.disconnect resolved (took 2ms)
worker1 paused (took 105ms)

pausing worker2... (same state)
worker2.blockingConnection.disconnect resolved (took 1ms)
worker2 paused (took 105ms)

pausing worker3... (same state)
worker3.blockingConnection.disconnect resolved (took 0ms)
worker3 paused (took 104ms)

pausing worker4... (same state)
worker4.blockingConnection.disconnect(wait=true) called
// ^^^ never resolves — 'end' event never fires
```

### Workaround

Use `worker.pause(true)` (the `doNotWaitActive` flag) to skip `whenCurrentJobsFinished()`. This is safe when no jobs are active.

### Suggested fix

Add a timeout to the `await disconnecting` promise in `RedisConnection.disconnect()`:

```js
const disconnecting = new Promise((resolve, reject) => {
    client.once('end', resolve);
    client.once('error', reject);
    setTimeout(() => resolve(), 5000); // safety timeout
});
client.disconnect();
await disconnecting;
```

### Environment

- BullMQ: 5.73.1
- ioredis: 5.x
- Redis: 7.4.8 (Docker, single instance)
- Node.js: 22.22.0
- OS: Linux (Docker container on macOS via Colima/Rosetta x86_64 emulation)

maheshsingh20:
## Description
Fixes #4065 

When multiple workers call `pause()`, the `disconnect()` method could hang indefinitely waiting for the 'end' event from the Redis client. This particularly occurred with 4+ workers.

## Root Cause
The `disconnect()` method in `redis-connection.ts` waits for an 'end' event that, under certain conditions (especially with multiple blocking connections), may never fire from ioredis, causing the promise to hang indefinitely.

## Solution
Added a 5-second safety timeout to the `disconnect()` method. If the 'end' event doesn't fire within 5 seconds, the disconnect promise resolves anyway, preventing the hang while still allowing graceful disconnection in normal cases.

## Changes
- Modified `src/classes/redis-connection.ts`:
  - Added a 5-second timeout to `disconnect()` method
  - Properly typed resolve/reject callbacks
  - Added conditional cleanup for event listeners

- Added regression test in `tests/pause.test.ts`:
  - Creates 5 workers and pauses them concurrently
  - Verifies no hang occurs within 8 seconds
  - Confirms all workers are properly paused

## Testing
- ✅ All existing pause tests pass (14 tests)
- ✅ New regression test passes
- ✅ Build compiles successfully
- ✅ Tested locally with reproduction scenario

## Notes
The 5-second timeout was chosen to be long enough for normal disconnections but short enough to prevent noticeable hangs in production scenarios.


maheshsingh20:
Hi @manast  and @roggervalf 
This PR is open for review, please have a look here.

PR #4586, opened 2026-08-18
fix(worker): keep blocking client alive when parked in reconnecting

### LINKED ISSUE #4585: [Bug]:  Worker never resumes after a Redis outage > drainDelay on standalone Redis — watchdog disconnect(false) parks the blocking client in "reconnecting" forever
### Version

6.0.10 6.1.2

### Platform

NodeJS

### What happened?

After Redis is unavailable for longer than the blocking timeout (drainDelay + 1 s), a Worker never resumes consuming once Redis is back. isRunning() stays true, no more error events, the non-blocking connection is ready again, queue.add works — but the blocking client stays in ioredis status "reconnecting" forever, active stays 0, wait grows. Only a process restart recovers. Health checks that ping the base connection report healthy.

Not #4378 (that is Sentinel + terminal end) and not #4479 (client comes back to ready) — here the client never leaves "reconnecting".

Mechanism (RedisQueueBackend.waitForJob, same in readEvents):

1. Redis goes down → the blocking client's socket closes → ioredis schedules a retry (status "reconnecting", reconnectTimeout armed).
2. The watchdog fires (roundedTimeout * 1000 + 1000) and calls bclient.disconnect(false).
3. In ioredis, disconnect(false) while "reconnecting" sets manuallyClosing = true, clears reconnectTimeout, and calls connector.disconnect() — which is a no-op because there is no socket. No close event, so closeHandler never runs: the status stays "reconnecting" with no timer → the client is a zombie.
4. finally → reconnectBlocking() → RedisConnection.reconnect(): status is "reconnecting", so it waitUntilReady()s for a ready that can never come. The fetch loop is parked permanently.

5.80.9 called bclient.disconnect(!this.closing) (= disconnect(true)), which keeps ioredis' retry timer — that is why v5 recovers.

### How to reproduce.

**Minimal reproduction**

```javascript
import { Worker, Queue } from "bullmq";
import { Redis } from "ioredis";

const opts = { host: "localhost", port: 6379, maxRetriesPerRequest: null as null,
  retryStrategy: (t: number) => Math.min(t * 200, 5000) };
const conn = new Redis(opts); conn.on("error", () => {});
let processed = 0;
const w = new Worker("repro", async () => { processed++; }, { connection: conn });
w.on("error", () => {});
await w.waitUntilReady();

setInterval(async () => {
  const b: any = (w as any).backend?.blockingConnection;
  const status = b?._client?.status ?? "?";
  console.log(`base=${conn.status} blocking=${status} processed=${processed}`);
  try { const q = new Queue("repro", { connection: opts }); await q.add("t", {}); await q.close(); } catch {}
}, 10000);
```

Run it, then `docker stop redis` for ~90 s, `docker start redis`. Output after Redis is back:

base=ready blocking=reconnecting processed=10   ← stays like this forever, wait grows

Passing connection options instead of an instance behaves the same. With bullmq 5.80.9 the same script resumes within a second (processed keeps increasing).

**Suggested fix**

Only abandon the command when the connection is actually alive; if ioredis is already reconnecting, let it finish on its own — reconnect() in finally then correctly waits for ready:

```javascript
watchdog = setTimeout(() => {
  timedOut = true;
  if (bclient.status === "ready") bclient.disconnect(false);
  resolve(null);
}, roundedTimeout * 1000 + 1000);
```

(same guard for client.disconnect(false) in readEvents). Verified locally as a pnpm patch on 6.0.10: after a 95 s outage all blocking connections return to bzpopmin and newly added jobs are processed. 
Alternatively ioredis could transition to "end" when disconnect() is called reconnect()'s status === "end" → connect() branch fires — but the guard above is the smaller change.

### Relevant log output

```shell
base=reconnecting blocking=reconnecting running=true waiting=true processed=10
base=ready        blocking=reconnecting running=true waiting=true processed=10
base=ready        blocking=reconnecting running=true waiting=true processed=10   (60 s later)
```

### Code of Conduct

manast:
After a Redis outage longer than the blocking timeout, ioredis moves the
blocking client to "reconnecting" (an armed retry timer with no live
socket). The waitForJob/readEvents watchdog called disconnect(false)
unconditionally, which — while reconnecting — clears that retry timer
without emitting a close event, parking the client in "reconnecting"
forever: reconnect() then waits for a ready that never comes and the
worker never resumes consuming.

Only abandon the blocking command when the connection is actually alive
(status === 'ready'); if ioredis is already reconnecting, let it finish
on its own so reconnect() correctly waits for ready. The positive
'ready' check is portable across all client adapters, whereas the
'reconnecting' status is ioredis-specific.

Fixes #4585)


magicknot:
We hit what looks like the same underlying defect in production (bullmq 6.0.11, `QueueEvents`/`readEvents` rather than `Worker`/`waitForJob`), but through a different path. In our case the connection has already recovered when the watchdog fires:

1. the connection drops briefly
2. ioredis noticed the drop and reconnected on its own, before the watchdog firing, i.e., the connection was `ready` again
3. then the watchdog fired anyway

then this happened:

1. watchdog called `client.disconnect(false)` on the now healthy connection
2. the `finally` block calls `connection.reconnect()` in the same tick. Here, ioredis hasn't processed the disconnect yet, so `client.status` is still `ready`. Thus, reconnect is in practice a noop as it sees `ready` and returns
3. on the next tick the connection close lands and never recovers.

I reproduced this locally using Fable, so take this with a grain of salt, but decided to post this here in case it helps finding another blindspot/problematic path.

manast:
@magicknot yes, it is tricky, probably easier to get it fixed in ioredis instead. I will see if we can add cover for this edge case as well without introducing new ones.

roggervalf (dismissed):

redonrxh:
We hit this in production (AWS ElastiCache, standalone node — no Sentinel/cluster): a cache-node restart parked the consume loops of 6 of 8 Workers in one process for 9 hours. All ioredis clients reconnected (so health checks stayed green), but the workers' dedicated blocking clients never re-registered — they were absent from CLIENT LIST entirely — and no error event fired. Confirmed byte-identical behavior on 6.0.8 and 6.3.1.

Deterministic repro (first attempt, every time): start Redis/Valkey in docker, boot one idle Worker (maxRetriesPerRequest: null, unbounded retryStrategy), then docker stop -t 0 the container, hold the outage ~15s (longer than drainDelay + the watchdog's +1s), and docker start. The worker never resumes; a job added afterward is never processed; the blocking client never reappears in CLIENT LIST. A sub-second restart usually recovers cleanly — the outage must outlive the watchdog, matching the disconnect(false)-during-reconnecting analysis in #4586.

Happy to test a canary build against this repro if useful.

manast:

manast:

roggervalf (approved):

manast:
@copilot+claude-opus-4.8 we are going to need to have this fix back ported to the 5.x branch as well, as it seems it is affecting users of that version too.

PR #4594, opened 2026-08-20
test(stalled): add script integration test [elixir]

manast:
The Elixir wrapper passed an extra `failed` key (shifting stalled-check
and every subsequent key), swapped ARGV[1]/ARGV[2], and never sent
ARGV[4] max_check_time. The script then ran `SET stalled-check ts PX nil`,
which errors — and since the worker runs the check fire-and-forget, the
error was swallowed and stalled jobs were never reclaimed.

Send the canonical 9 KEYS and 4 ARGV, using `Keys.key(ctx) <> ":"` for the
queue prefix so job/lock keys resolve correctly, and default max_check_time
to 30_000. Adds an integration regression test.

### DIFF
--- a/elixir/lib/bullmq/scripts.ex
+++ b/elixir/lib/bullmq/scripts.ex
@@ -1412,8 +1412,7 @@ defmodule BullMQ.Scripts do
 
     timestamp = Keyword.get(opts, :timestamp, System.system_time(:millisecond))
 
-    max_check_time =
-      Keyword.get(opts, :stalled_interval, Keyword.get(opts, :max_check_time, 30_000))
+    max_check_time = Keyword.get(opts, :stalled_interval, 30_000)
 
     args = [
       # ARGV[1] max stalled count
--- a/elixir/lib/bullmq/scripts.ex
+++ b/elixir/lib/bullmq/scripts.ex
@@ -1390,27 +1390,38 @@ defmodule BullMQ.Scripts do
           script_result()
   def move_stalled_jobs_to_wait(conn, ctx, max_stalled_count, opts \\ []) do
     keys = [
+      # KEYS[1] stalled
       Keys.stalled(ctx),
+      # KEYS[2] wait
       Keys.wait(ctx),
+      # KEYS[3] active
       Keys.active(ctx),
-      Keys.failed(ctx),
+      # KEYS[4] stalled-check
       Keys.stalled_check(ctx),
+      # KEYS[5] meta
       Keys.meta(ctx),
+      # KEYS[6] paused
       Keys.paused(ctx),
+      # KEYS[7] marker
       Keys.marker(ctx),
+      # KEYS[8] event stream
       Keys.events(ctx),
+      # KEYS[9] repeat
       Keys.repeat(ctx)
     ]
 
     timestamp = Keyword.get(opts, :timestamp, System.system_time(:millisecond))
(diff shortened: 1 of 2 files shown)


elixir/test/bullmq/integration_test.exs
```diff
@@ -675,6 +675,65 @@ defmodule BullMQ.IntegrationTest do
...
+      # guards ARGV[2] (queue prefix): the script derives job-2's lock key as
+      # `<ARGV[2]><jobId>:lock`, so a missing trailing colon would look up the
+      # wrong key, see no lock and wrongly reclaim the running job.
+      {:ok, _} = Redix.command(conn, ["RPUSH", Keys.active(ctx), "job-1", "job-2"])
+      {:ok, _} = Redix.command(conn, ["HSET", Keys.job(ctx, "job-1"), "name", "test"])
+      {:ok, _} = Redix.command(conn, ["HSET", Keys.job(ctx, "job-2"), "name", "test"])
+      {:ok, _} = Redix.command(conn, ["SET", Keys.lock(ctx, "job-2"), "token", "PX", 30_000])
+
+      # Phase 1 only marks lock-less active jobs into the stalled set (grace
+      # period) and returns []. It also arms the stalled-check mutex.
```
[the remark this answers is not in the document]
manast: @copilot Fix the code for all comments in [this review comment](https://github.com/taskforcesh/bullmq/pull/4594#discussion_r3820926892).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

elixir/test/bullmq/integration_test.exs
```diff
@@ -675,6 +675,65 @@ defmodule BullMQ.IntegrationTest do
       {:ok, last_check} = Redix.command(conn, ["GET", Keys.stalled_check(ctx)])
       assert String.to_integer(last_check) == now
     end
+
+    @tag :integration
+    test "move_stalled_jobs_to_wait reclaims lock-less jobs and sets stalled-check TTL (#4587)",
```
roggervalf, line 680: I think that test cases should cover the move to wait from the worker instances too, that way we will make sure that is working on the real flow

manast:

roggervalf (approved):

elixir/lib/bullmq/scripts.ex
```diff
@@ -1390,27 +1390,40 @@ defmodule BullMQ.Scripts do
...
+      # KEYS[8] event stream
       Keys.events(ctx),
+      # KEYS[9] repeat
       Keys.repeat(ctx)
     ]
 
     timestamp = Keyword.get(opts, :timestamp, System.system_time(:millisecond))
 
+    max_check_time =
+      Keyword.get(opts, :stalled_interval, Keyword.get(opts, :max_check_time, 30_000))
```
roggervalf: actually, max_check_time option does not exist in node. We use stalledInterval. For option parity, we should remove this extra reference

roggervalf (changes requested):

elixir/lib/bullmq/scripts.ex
manast: @copilot address comment.

roggervalf (approved):

dbhaasis:
@manast Did a new version get released for elixir? I only see the python and node ones listed above. Also, thanks for looking into the issue!

PR #4623, opened 2026-08-25
feat: make maximumBlockTimeout a backend capability

### LINKED ISSUE #4601: PG backend: maximumBlockTimeout is a Redis-derived constant on a shared path, so an idle worker polls every 10s
## Summary

`maximumBlockTimeout` is a module-level constant in `Worker`, justified by a **Redis** limitation, but it is applied on a code path shared by every backend. With the new PostgreSQL backend it forces a poll every 10 seconds whenever any delayed job exists, which prevents an idle worker from ever going quiet.

`minimumBlockTimeout` is already delegated to the backend. `maximumBlockTimeout` is not, and I think that asymmetry is the bug.

## The code

`src/classes/worker.ts` (v6.2.0):

```ts
// 10 seconds is the maximum time a BZPOPMIN can block.
const maximumBlockTimeout = 10;
```

`BZPOPMIN` is a Redis command, and the reference on line 858 is to #1658, which is about Redis reconnection behaviour.

It is consumed in `getBlockTimeout()`, which lives in the shared `Worker` class:

```ts
protected getBlockTimeout(blockUntil: number): number {
  const opts: WorkerOptions = <WorkerOptions>this.opts;
  if (blockUntil) {                       // when there are delayed jobs
    const blockDelay = blockUntil - Date.now();
    if (blockDelay <= 0) {
      return blockDelay;
    } else if (blockDelay < this.minimumBlockTimeout * 1000) {
      return this.minimumBlockTimeout;
    } else {
      return Math.min(blockDelay / 1000, maximumBlockTimeout);   // <-- here
    }
  } else {
    return Math.max(opts.drainDelay, this.minimumBlockTimeout);  // drainDelay only reachable here
  }
}
```

The result is passed to `this.backend.waitForJob(blockTimeout)`, so the Postgres backend inherits a ceiling derived from a Redis command's semantics.

Meanwhile the *minimum* is already backend-aware:

```ts
get minimumBlockTimeout(): number {
  return this.backend.minimumBlockTimeout;
}
```

## Why it matters in practice

`drainDelay` is only reachable in the `else` branch — when **no** delayed job exists. An application with repeatable/scheduled jobs almost always has one pending, so `drainDelay` is effectively unreachable and the 10s ceiling always applies.

We run 11 repeatable schedules, so a completely idle worker issues its "any work?" queries every 10 seconds indefinitely — roughly 17,280 queries a day that find nothing.

On a serverless Postgres that bills compute time and suspends when idle (Neon, with a 300s idle window), this means **the database can never suspend**. Measured on our own deployment: the database currently sleeps ~7.4h/day, and a Postgres-backed queue would take that to 0, a ~42% increase in the database compute bill for a workload that is genuinely idle.

This is not a complaint about the default value. On Redis the 10s ceiling costs nothing — the connection is already open and `BZPOPMIN` is cheap. It is specifically that a Redis-shaped constraint is applied to a backend with different economics, where an idle poll has a price.

## Suggested fix

Delegate it the same way the minimum already is:

```ts
// IQueueBackend
readonly maximumBlockTimeout: number;

// worker.ts
get maximumBlockTimeout(): number {
  return this.backend.maximumBlockTimeout;
}
```

Redis backend keeps `10` and nothing changes for existing users. The Postgres backend can choose a larger value, or expose it as an option so applications can trade latency for idle cost.

Happy to open a PR if the approach looks right.

## Environment

- bullmq 6.2.0 (source reviewed at tag `v6.2.0`)
- PostgreSQL backend on Neon, `suspend_timeout = 300s`
- 11 repeatable job schedules

0jaspahwa:
## Problem

`maximumBlockTimeout` is a constant in `Worker`, set to `10` because that is the
longest a Redis `BZPOPMIN` can block. It is applied on a path **every backend shares**.

Postgres blocks on `LISTEN`/`NOTIFY`, which has no such limit. Any pending delayed
job puts the worker on that path, so an idle worker queries the database every 10
seconds forever. On serverless Postgres, which sleeps when idle, this alone keeps it awake.

`minimumBlockTimeout` is already backend-owned. The maximum is not.

## Change

Move the limit onto `IQueueBackend`, next to `minimumBlockTimeout`.

| Backend | `maximumBlockTimeout` |
| :-- | :-- |
| Redis | `10`, unchanged, no user impact |
| Postgres | `3600` |
| Not set | falls back to `10` in the worker |

The member is **optional**, so custom backends outside this repo keep compiling.

 **Scope:** this removes one source of idle traffic, not all of it. The stalled-job checker polls every `stalledInterval` (default 30s) whether or not anything is  active, roughly 2,880 queries a day, which is enough on its own to keep a serverless instance awake. Different timer, different call, not addressed here.

## Why a longer wait is safe on Postgres

* `waitForJob` already shortens its timer to the next due delayed job, and *only*
  ever shortens it, so a larger cap cannot push a scheduled job out.
* Every path that adds or re-dates a delayed job fires `pg_notify`, so a job
  created mid-wait still wakes the worker.
* The cap must stay **finite**. `waitForJob` passes it straight to `setTimeout`,
  which fires immediately past ~24.8 days. A test guards this.

## Tests

| Suite | Result |
| :-- | :-- |
| `tests/postgres` | 39 / 39 |
| `test:adapter-conformance` | 66 / 66 |
| default suite | 808 passing |

Failures in `child-pool`, `script_loader` and `sandboxed_process` are pre-existing
on my machine (Windows: child processes and path handling), and identical on clean
`master`.

## Known follow-up

Raising the cap does not create a bug, but it **stops hiding one**. A dead `LISTEN`
client is currently masked by the 10s timer. The worker wakes, runs `moveToActive`
on a pool connection, and jobs keep flowing by de facto polling. At `3600` that same
dead connection stalls the queue for up to an hour.

Nothing replaces a dead client today. `listenClientPromise` is memoized and cleared
only in `close()`, and `ensureListening` re-issues `LISTEN` on that same client.

I have a branch that fixes this, with a test that kills the LISTEN backend via
`pg_terminate_backend` and fails without the fix. Happy to open it as a follow-up,
or fold it in here.

## Ports

Python (`python/bullmq/worker.py:26`) and Elixir (`elixir/lib/bullmq/worker.ex`)
carry the same constant and the same asymmetry. Happy to port if wanted.

Closes #4601

### DIFF
--- a/src/classes/redis-queue-backend.ts
+++ b/src/classes/redis-queue-backend.ts
@@ -347,6 +347,15 @@ export class RedisQueueBackend extends EventEmitter implements IQueueBackend {
   }
 
   /**
+   * Longest block timeout (seconds) to ask `BZPOPMIN` for. Kept at 10s so a
+   * dropped connection is noticed quickly.
+   * reference: https://github.com/taskforcesh/bullmq/issues/1658
+   */
+  get maximumBlockTimeout(): number {
+    return 10;
+  }
+
+  /**
    * Interrupts the in-flight blocking wait by disconnecting the dedicated
    * blocking connection. No-op if there is none.
    */
--- a/src/classes/worker.ts
+++ b/src/classes/worker.ts
@@ -45,8 +45,11 @@ import {
 } from './job-scheduler';
 import { LockManager } from './lock-manager';
 
-// 10 seconds is the maximum time a BZPOPMIN can block.
-const maximumBlockTimeout = 10;
+/**
+ * Used when a backend does not set its own maximum. Matches the Redis backend,
+ * so third-party backends keep the behaviour they had before.
+ */
+const DEFAULT_MAXIMUM_BLOCK_TIMEOUT = 10;
 
 // note: sandboxed processors would also like to define concurrency per process
 // for better resource utilization.
@@ -776,6 +779,10 @@ export class Worker<
     return this.backend.minimumBlockTimeout;
   }
 
+  get maximumBlockTimeout(): number {
+    return this.backend.maximumBlockTimeout ?? DEFAULT_MAXIMUM_BLOCK_TIMEOUT;
+  }
+
   private isRateLimited(): boolean {
     return this.limitUntil > Date.now();
   }
@@ -855,10 +862,9 @@ export class Worker<
       } else if (blockDelay < this.minimumBlockTimeout * 1000) {
         return this.minimumBlockTimeout;
       } else {
-        // We restrict the maximum block timeout to 10 second to avoid
-        // blocking the connection for too long in the case of reconnections
-        // reference: https://github.com/taskforcesh/bullmq/issues/1658
-        return Math.min(blockDelay / 1000, maximumBlockTimeout);
+        // The limit belongs to the backend: Redis must not block for long,
+        // while PostgreSQL's LISTEN-based wait can.
+        return Math.min(blockDelay / 1000, this.maximumBlockTimeout);
       }
     } else {
       return Math.max(opts.drainDelay, this.minimumBlockTimeout);
--- a/src/interfaces/queue-backend.ts
+++ b/src/interfaces/queue-backend.ts
@@ -107,6 +107,15 @@ export interface IQueueBackend {
   readonly minimumBlockTimeout: number;
 
   /**
+   * Longest block timeout (in seconds) supported by the backend's blocking
+   * primitive. Used by workers to bound `waitForJob` when the next delayed job
+   * is due further away than this.
+   *
+   * Optional: backends that do not set it fall back to 10 seconds.
+   */
+  readonly maximumBlockTimeout?: number;
+
+  /**
    * Subscribes to normalized backend lifecycle events (`'ready'`, `'error'`,
    * `'close'`), derived from the underlying connection(s).
    */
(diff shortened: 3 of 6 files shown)


MGrin:
Reporter of #4601. Diff reads right to me - the `?? 10` fallback keeps custom backends compiling, and the finite-cap test is the one I would have asked for.

Two notes for whoever reviews, neither a blocker:

- The suspend benefit in the description is not complete on its own. The stalled checker polls unconditionally - `stalledChecker()` awaits `moveStalledJobsToWait()` then `setTimeout(resolve, this.opts.stalledInterval)` (`worker.ts:1369-1378`), `stalledInterval` defaults to `30000` (`:248`) and must be `> 0` (`:274-279`), and the Postgres implementation is an unconditional round-trip (`postgres-queue-backend.ts:1047-1062`). So an idle worker still hits the database every 30s and a serverless instance with a 300s idle window never suspends. Different timer, separate call - detail in taskforcesh/bullmq#4601.
- The `else` branch is untouched, so with no delayed job at all the cap is `drainDelay`, default `5` (`worker.ts:243`). That one is already user-settable, so nothing to do here - just noting the sibling path is unaffected.

0jaspahwa:
Thanks. Both notes are right, and I have corrected the suspend claim in the description so it does not read as complete. The `else` branch is untouched as you say, so `drainDelay` behaviour is unchanged.

0jaspahwa:
Superseded by ea4543cc78, closing.

PR #4635, opened 2026-08-26
feat(job): add DelayedError and Job.moveToDelayed for processor-controlled delays [python]

davelindley:
### Port Impact Checklist

- [x] **Python** – does this change need to be ported or documented in the Python library?

Python-only: Node.js is unchanged, this brings the Python port up to the existing Node behaviour.

### Why

In Node a processor can re-park its own active job, which is the documented way to wait for something without burning an attempt or spawning a second job ([Process Step Jobs](https://docs.bullmq.io/patterns/process-step-jobs)):

```ts
await job.moveToDelayed(Date.now() + 1000, token);
throw new DelayedError();
```

The Python port already has everything underneath this:

- the `moveToDelayed` Lua script is bundled (`moveToDelayed-12.lua`),
- `Backend.moveToDelayed` is declared abstract and implemented by both `RedisBackend` and `PostgresBackend`,
- `Scripts.moveToDelayedArgs` already supports `skipAttempt`,
- `Worker.processJob` already has the "processor moved the job itself, walk away" arm for `WaitingChildrenError`.

What is missing is the public surface: there is no `DelayedError` and no `Job.moveToDelayed`, so the pattern is unreachable from Python. The only way to defer work today is to add a *new* delayed job and carry the state across yourself.

### How

Four small pieces, each mirroring the Node implementation:

1. `bullmq/custom_errors/delayed_error.py` – `DelayedError`, in the style of the neighbouring `WaitingChildrenError` / `UnrecoverableError`, re-exported from `bullmq` and `bullmq.custom_errors`.
2. `Job.moveToDelayed(timestamp, token)` – computes `now` / `delay` and calls `self.backend.moveToDelayed(..., {"skipAttempt": True})`, same as `Job.moveToDelayed` in `src/classes/job.ts`. `skipAttempt` means a manual delay does not consume an attempt. (Node also calls `recordJobMetrics('delayed')` there; the Python port has no metrics recording, so there is nothing to mirror.)
3. `Worker.processJob` – `except WaitingChildrenError` becomes `except (DelayedError, WaitingChildrenError)`. Node's `handleFailed` treats `DelayedError`, `WaitingError` and `WaitingChildrenError` identically, so `DelayedError` is unconditional here too: it tells the worker not to complete or fail the job, it is the `moveToDelayed` call that parks it.
4. Tests, ported from the Node suite.

### Additional Notes

**Tests** (`python/tests/worker_test.py`, plus a re-export check in `test_imports.py`):

- `test_move_to_delayed_from_processor` – processor moves the job and raises `DelayedError`: the job lands in the delayed set, `active`/`failed`/`completed` stay 0, and no `failed` or `error` event is emitted.
- `test_move_to_delayed_in_one_step_keeping_current_step` – port of the Node case of the same name (`tests/worker.test.ts`, "when moving job to delayed in one step"): the **same job id** is processed twice, `updateData` written before the delay is visible in the second pass, and `attemptsMade == 1` with `attemptsStarted == 2`, which is what `skipAttempt` buys.
- `test_delayed_error_without_moving_job_does_not_finish_job` – guards the semantics from point 3: raising `DelayedError` without calling `moveToDelayed` leaves the job active for the stalled checker; it is never quietly completed or failed.

All three fail without the `worker.py` change, so they actually pin the behaviour down.

**Verification**

- Full Python suite against Redis 7.2 locally: `173 passed, 9 subtests passed`.
- The three new tests also pass against the Postgres backend (`BULLMQ_TEST_BACKEND=postgres`), since `PostgresBackend.moveToDelayed` already honours `skipAttempt`.
- The three new tests also pass with `BULLMQ_TEST_PREFIX={b}` (the DragonflyDB CI configuration).
- Reverting the `worker.py` change (back to `except WaitingChildrenError:`) makes all three new tests fail, so they do pin the behaviour rather than passing incidentally.
- `flake8 ./python --select=E9,F63,F7,F82` clean.
- Node suite per `contributing.md` (`yarn && yarn test`), unaffected but run anyway: `824 passed | 3 skipped (41 files)`.

**One note on timing:** if a v6 release is in flight for the Python package and this is easier to take there, I am happy to retarget it at `next` (or whatever branch you prefer) - it is small and applies cleanly either way.

### DIFF
--- a/python/bullmq/custom_errors/delayed_error.py
+++ b/python/bullmq/custom_errors/delayed_error.py
@@ -1,0 +1,3 @@
+class DelayedError(Exception):
+    "Raised when job is moved to delayed from inside the processor"
+    pass
--- a/python/bullmq/job.py
+++ b/python/bullmq/job.py
@@ -278,6 +278,22 @@ class Job:
             elif self.opts.get("stackTraceLimit"):
                 self.stacktrace = self.stacktrace[-(stackTraceLimit-1):stackTraceLimit]
 
+    async def moveToDelayed(self, timestamp: int, token: str):
+        """
+        Moves the job to the delay set. Only allowed while the job is active, i.e.
+        called from inside a processor, and a DelayedError must be raised afterwards
+        so that the worker does not try to complete or fail the job.
+
+        @param timestamp: timestamp when the job should be moved back to "wait"
+        @param token: token to check job is locked by current worker
+        """
+        now = round(time.time() * 1000)
+        delay = timestamp - now if timestamp > now else 0
+
+        await self.backend.moveToDelayed(self.id, now, delay, token, {"skipAttempt": True})
+
+        self.delay = delay
+
     async def moveToWaitingChildren(self, token, opts:dict) -> bool | None:
         return await self.backend.moveToWaitingChildren(self.id, token, opts)
 
--- a/python/bullmq/__init__.py
+++ b/python/bullmq/__init__.py
@@ -16,4 +16,4 @@ from bullmq.job_scheduler import JobScheduler
 from bullmq.abort_controller import AbortController, AbortSignal, AbortError
 from bullmq.queue_events import QueueEvents
 from bullmq.queue_events_producer import QueueEventsProducer
-from bullmq.custom_errors import WaitingChildrenError, UnrecoverableError
+from bullmq.custom_errors import WaitingChildrenError, UnrecoverableError, DelayedError
--- a/python/bullmq/custom_errors/__init__.py
+++ b/python/bullmq/custom_errors/__init__.py
@@ -1,2 +1,3 @@
+from bullmq.custom_errors.delayed_error import DelayedError
 from bullmq.custom_errors.unrecoverable_error import UnrecoverableError
 from bullmq.custom_errors.waiting_children_error import WaitingChildrenError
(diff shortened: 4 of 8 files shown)


python/bullmq/job.py
```diff
@@ -278,6 +278,22 @@ def updateStacktrace(self):
...
+
+        @param timestamp: timestamp when the job should be moved back to "wait"
+        @param token: token to check job is locked by current worker
+        """
+        now = round(time.time() * 1000)
+        delay = timestamp - now if timestamp > now else 0
+
+        await self.backend.moveToDelayed(self.id, now, delay, token, {"skipAttempt": True})
+
+        self.delay = delay
```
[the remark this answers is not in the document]
davelindley: Agreed, fixed - `token` is now required (86199519).

Node types it as optional (`moveToDelayed(timestamp: number, token?: string)`), but every other public move on the Python `Job` already requires it (`moveToCompleted`, `moveToFailed`, `moveToWaitingChildren`), and the transition is only valid for an active job locked by the current worker, so the default was a footgun with no upside. The method is new, so nothing can break.

python/bullmq/custom_errors/delayed_error.py
```diff
@@ -0,0 +1,3 @@
+class DelayedError(Exception):
+    "Raised when job is moved to delayed from inside the processor"
```
roggervalf, line 2: to be consistent, https://github.com/taskforcesh/bullmq/blob/master/python/bullmq/custom_errors/waiting_children_error.py need to contain same last part of this message

davelindley: Done in 27c99105 - `WaitingChildrenError` now reads "Raised when job is moved to waiting-children from inside the processor". It is raised the same way (after `job.moveToWaitingChildren`), so the two docstrings should match.

roggervalf (approved):

PR #4651, opened 2026-08-28
docs: fix contradictory priority 0 explanation

### LINKED ISSUE #4642: [Docs Bug] Contradictory explanation of Priority 0 in Architecture guide vs Prioritized guide
There is a direct contradiction in the documentation regarding how `priority: 0` works.

1. In the **Architecture** section warning block, it states:
> *"Note that priorities go from 0 to 2^21, where 0 is the highest priority."*

2. However, the state diagram right above it shows that `priority is 0` routes to the standard **Wait** list, while `priority > 0` routes to **Prioritized**.
3. Furthermore, the API interface docs specify that priority ranges from `1` (highest priority) to `2,097,152` (lowest priority), where `0` denotes default/no priority.

---
Image

Image

Image

---

mhinostroza:
Fixes #4642

### Why

The warning block in the Architecture guide stated that "priority: 0" is the highest priority. This contradicts both the state diagram on that same page and the API reference, which correctly indicate that priority: 0 means "no priority" and that those jobs go to the "wait" state.

### How

I confirmed the actual behavior by reviewing storeAndEnqueueJob.lua, where priority > 0 routes to the "prioritized" set, while priority: 0 (or unset) routes to "wait". With that confirmed, I corrected the hint text in architecture.md to be consistent with the diagram and the API: priority values range from 1 to 2^21 (lower number = higher priority), and priority: 0 means no priority.

### Additional Notes

Change limited to documentation (architecture.md), no code or test changes.

### DIFF
--- a/docs/gitbook/guide/architecture.md
+++ b/docs/gitbook/guide/architecture.md
@@ -16,7 +16,7 @@ When a job is added to a queue it can be in one of three states:
 - **“delayed”**: implies that the job is waiting for some timeout or to be promoted for being processed. These jobs are not processed directly, but instead are placed at the beginning of the waiting list, or in a prioritized set, and processed as soon as a worker is idle.
 
 {% hint style="warning" %}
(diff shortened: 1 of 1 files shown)


manast:
Priority 0 means no priority, but also means it will be processed BEFORE any other jobs with priority higher than 0, so I think the documentation is not really wrong in that regard, you can think about priority 0 as highest priority.

mhinostroza:
updated the wording to reflect that, thanks for review

manast (approved):

roggervalf (approved):

PR #4671, opened 2026-09-01
feat(feature parity): add a reusable double blind feature parity check across ports (bun, python, rust tests against nodejs)

Billcountry:
### Port Impact Checklist

Please note, the changes mentioned below do not affect how the port works, they only test it's alignment with the primary implementation.

- [x] **Python** – no port needed, parity scripts added under `python/tests/parity`
- [ ] **Elixir** – parity scripts need to be implemented - no effect on functionality
- [ ] **PHP** – parity scripts need to be implemented - no effect on functionality
- [x] **Rust** – no port needed, parity scripts introduced as a new package under `rust/tests/parity`
- [ ] **.NET** – parity scripts need to be implemented - no effect on functionality
- [x] **Bun** - parity tested as a port since implementation differs a little from the primary node.js implementation

### Why

I've been looking for a lightweight open source queueing system that I can easily move with across various infra offerings and languages. BullMQ offered exactly that and I loved the fact that the library has already been ported across multiple languages.

Going through the codebase, I felt a risk of a feature deployment breaking across ports and the bug might not be obvious on time.

I therefore created more of a feature parity tracker framework that takes each port and tests it against node.js implementation. Each port is tested acting purely as a producer or purely as a consumer.

The framework structure is documented under: `parity/README.md` and the tests can be run by:
```sh
yarn test:parity
```

After the test run is complete, a report is generated under `parity/REPORT.md`. On pull requests this report is also commented on the PR.
For security reasons I don't believe the comment can be on a forked PR, see here on my forked repo: https://github.com/Billcountry/bullmq/pull/1#issuecomment-5496388059

Below the report from the comment:

**----------- REPORT START ------------**

# Feature Alignment Report
- **Last Run**: 2026-09-01T17:58:27.260Z
- **Base Implementation** : Node.JS

### ⚫️  Bun Native - Postgres

**Backend not supported** - this test was skipped

### 🟢  Bun Native - Redis

**All 3 tests passed.**

| Test Id | Name | As Consumer | As Producer | Description |
| :------ | :--- | :---------: | :---------: | :---------- |
| simple-worker-test | Simple Worker | ✓ | ✓ | Tests a simple worker that picks up jobs from a queue and processes them |
| delayed-job-test | Delayed Job | ✓ | ✓ | Tests a delayed job that is added to a queue and processed after a specified delay |
| concurrency-test | Concurrency | ✓ | ✓ | Tests concurrent queue processing |

### ⚫️  Rust - Postgres

**Backend not supported** - this test was skipped

### 🟢  Rust - Redis

**All 3 tests passed.**

| Test Id | Name | As Consumer | As Producer | Description |
| :------ | :--- | :---------: | :---------: | :---------- |
| simple-worker-test | Simple Worker | ✓ | ✓ | Tests a simple worker that picks up jobs from a queue and processes them |
| delayed-job-test | Delayed Job | ✓ | ✓ | Tests a delayed job that is added to a queue and processed after a specified delay |
| concurrency-test | Concurrency | ✓ | ✓ | Tests concurrent queue processing |

### 🟢  Python - Redis

**All 3 tests passed.**

| Test Id | Name | As Consumer | As Producer | Description |
| :------ | :--- | :---------: | :---------: | :---------- |
| simple-worker-test | Simple Worker | ✓ | ✓ | Tests a simple worker that picks up jobs from a queue and processes them |
| delayed-job-test | Delayed Job | ✓ | ✓ | Tests a delayed job that is added to a queue and processed after a specified delay |
| concurrency-test | Concurrency | ✓ | ✓ | Tests concurrent queue processing |

### 🟢  Python - Postgres

**All 3 tests passed.**

| Test Id | Name | As Consumer | As Producer | Description |
| :------ | :--- | :---------: | :---------: | :---------- |
| simple-worker-test | Simple Worker | ✓ | ✓ | Tests a simple worker that picks up jobs from a queue and processes them |
| delayed-job-test | Delayed Job | ✓ | ✓ | Tests a delayed job that is added to a queue and processed after a specified delay |
| concurrency-test | Concurrency | ✓ | ✓ | Tests concurrent queue processing |

**----------- REPORT END ------------**

### How

In summary, the framework:
- introduces a common way of defining parity test cases as json under `parity/definitions.json`
- expects each port to implement two scripts
  - producer script - this reads definitions and creates all the jobs defined, then exits
  - consumer script - this reads definitions and creates all the workers and waits for the jobs from the producer
- the two scripts are run independently and they log events to console in a format that the framework can use to evaluate whether the both the producer and the consumer completed everything successfully
- a test is considered successful when a port is able to pass both as a consumer and as a producer against node.js.

Please find the full details of the implementation under `parity/README.md`

The current tests are designed to test simple features and as a POC of the framework, it should be extended to cover more complex cases like the flow producer since edge cases tend to buildup with complexity.

As new languages and features are added, the framework can be extended accordingly.

For testing:
- For python I used poetry to reduce the complexity of managing environments, you need to have it installed on your system before running the tests.
- Everything else should work with the project as is.


manast:
Thank you for the PR. As parity and compatibility is paramount for us, we guarantee such parity by using the exact same lua scripts (there is only one copy) on all the ports, or the exact same postgreSQL functions. Therefore there is no need to create another layer of tests on top of it, as long as the tests pass for every port, we know they are interoperatible. We have already a very heavy test suite so I am reluctant to add more weight unless it really provides some meaningful value.

Billcountry:
Hi @manast,

Thank you for the honest reply, I'm going to reply in principle, so please don't consider this a push for the PR; rather, a start to a conversation I'm trying to have with the world regarding software testing.

I'm working on approaches that make software QA easier for everyone, and in my journey I might be biased to look at everything as a test. In this I'm working with one basic principle; **〝look at it as your user sees it, test it as your user uses it〞**

You'll see the principle in the way I've written these tests and the way I expect the results to be evaluated based on their effects/outcomes to the world rather than the underlaying logic.

A case that I've encountered more times than I'm comfortable to admit is where we would write the code and tests as developers, hand over the feature to queue for verification, only for the feature to fail on QA despite all the code being tested and reviewed.

I faced this while working on this PR with the bun native client, all tests were passing but when I the bun-consumer implementation my tests were failing.
See this commit https://github.com/taskforcesh/bullmq/pull/4671/changes/f97a2e9367e033ca3db9eef0f0b952d2896e7ee9 where I made some fixes on the bun adapter to correctly run workers; I reverted my changes as the adapter was updated a few days later.

In short, I do think these kind of tests have a place between development and delivery. I'm not necessarily sure that the approach I've taken here is the best, but I think it's a conversation worth having.

If you don't mind continuing this conversation in future, lets connect on linked: https://linkedin.com/in/Billcountry

PR #4691, opened 2026-09-06
fix(worker): calculate next iteration delay from repeat pattern instead of hardcoded 1s [elixir]

mfeitoza:
### Port Impact Checklist

### Why

In the Elixir port, jobs generated by a `JobScheduler` with a cron `pattern` or repeat options executed every 1 second after their first run, rather than following the configured schedule or pattern.

Root causes:
1. `BullMQ.Worker.schedule_next_iteration/4` had hardcoded `next_millis = now + 1000`, causing all subsequent iterations to be scheduled 1,000 ms into the future regardless of the `pattern` or `every` repeat configuration.
2. `BullMQ.JobScheduler.calculate_next_millis/2` only looked for atom keys (e.g. `:pattern`, `:every`), while repeat options decoded from Redis/JSON payloads often arrive with string keys (`"pattern"`, `"every"`, `"prevMillis"`, etc.).
3. `parse_cron_next/3` did not support 6-field cron expressions (which include seconds) and did not advance past `reference_time` if `reference_time` fell precisely on a cron boundary second.

### How

1. **Worker Next Iteration Scheduling (`elixir/lib/bullmq/worker.ex`)**:
   - Replaced hardcoded `now + 1000` with dynamic calculation using `JobScheduler.calculate_next_millis(repeat_opts, base_time)`.
   - Used `base_time = max(now, prev_millis)` to prevent drift and ensure proper progression of successive intervals.
   - Updated `build_scheduler_job_opts/4` to set:
     - `delay`: `max(0, next_millis - now)`
     - `timestamp`: `now`
     - `prevMillis`: `next_millis`
     - `jobId`: `repeat:#{custom_key}:#{next_millis}`
     - `repeatJobKey`: `custom_key`
   - Added `get_job_prev_millis/1` and `parse_opt_int/2` to reliably extract integer timestamps from job options.

2. **JobScheduler Repeat Calculation (`elixir/lib/bullmq/job_scheduler.ex`)**:
   - Enhanced `calculate_next_millis/2` to transparently handle both atom and string keys (`pattern`, `every`, `startDate`, `endDate`, `tz`, `offset`, `prevMillis`).
   - Enabled `extended: true` in `Crontab.CronExpression.Parser.parse!/2` to support 6-field cron expressions.
   - Guarded against boundary matches in `parse_cron_next/3` by advancing candidate time by 1 second if it equals the reference time.

3. **Tests (`elixir/test/bullmq/`)**:
   - Added unit test in `job_scheduler_integration_test.exs` verifying:
     - Repeat option normalization with string keys.
     - 6-field cron pattern evaluation.
     - Strict boundary time advancement.
   - Added processing test in `job_scheduler_processing_test.exs` verifying that after a worker completes a recurring job, the next iteration is delayed according to the pattern (> 50 seconds for an hourly pattern) instead of 1 second.

All existing unit tests and integration tests pass:
- `mix test` passes (187 unit tests, 0 failures).
- `mix test test/bullmq/job_scheduler_processing_test.exs` passes.
- Code formatted with `mix format`.


mfeitoza:
@roggervalf or @manast I think to merge this fix to the refactor PR #4697

manast:
> @roggervalf or @manast I think to merge this fix to the refactor PR #4697

I think it is better to keep it as a separate PR, but sure it would be better to merge it after the refactor PR. Btw it is still marked a draft.

PR #4695, opened 2026-09-08
fix(postgres): serialize addLog to prevent duplicate job_log index

### LINKED ISSUE #4682: BullMQ with postgres has a race condition when doing job.log
# BullMQ Bug Report

---

## Version

```
6.3.4
```

---

## Platform

```
NodeJS
```

---

## What happened?

Using the **PostgreSQL backend** with a **sandboxed (external file) processor**, any job that calls `job.log()` more than once in quick succession fails with a unique constraint violation on `job_log_pkey`, and log lines are silently lost.

Two things combine to cause this:

**1. `add_log.sql` is a non-atomic read-modify-write.**

`dist/*/postgres/commands/add_log.sql` picks the next per-job ordinal with a bare sub-select:

```sql
INSERT INTO job_log (queue, job_id, idx, row)
VALUES (
  $1, $2,
  COALESCE(
    (SELECT MAX(idx) + 1 FROM job_log WHERE queue = $1 AND job_id = $2),
    0
  ),
  $3
)
RETURNING idx;
```

There is no lock, no sequence and no `ON CONFLICT`. Two overlapping calls for the same `(queue, job_id)` both read the same `MAX(idx)`, and the second insert violates `PRIMARY KEY (queue, job_id, idx)` declared in `0001_schema.sql`.

**2. The sandbox does not serialise `addLog` calls.**

In `classes/sandbox.js`, `msgHandler` is an `async` function attached with `child.on('message', msgHandler)`. `EventEmitter` invokes it once per IPC message and discards the returned promise, so the `await job.log(msg.value)` in the `ParentCommand.Log` branch runs concurrently with the handler for the next message. Nothing awaits the previous DB write before starting the next.

**Awaiting in the processor does not avoid this.** The child's `job.log()` is `await send({ cmd: ParentCommand.Log, ... })`, and `send` is `asyncSend` in `utils/index.js`, which resolves on the **IPC write callback** — not on the parent's commit. So a processor that awaits every `job.log()` sequentially still hands the parent a burst of messages it then processes in parallel. The reproduction below is fully sequential in the child and still fails.

This is a regression relative to Bull/Redis, where `addLog` was an atomic `RPUSH` with no client-computed index. It is not specific to custom job IDs — it reproduces with generated IDs.

The Redis backend appears unaffected, since the index is assigned server-side.

Suggested fix: make the index assignment atomic in SQL — e.g. take a `pg_advisory_xact_lock(hashtext($1 || ':' || $2))` before the sub-select, or give `job_log` a per-job sequence — rather than relying on callers not to overlap. Serialising `msgHandler` in the sandbox would fix the sandboxed path but would leave direct concurrent `job.log()` calls from an in-process processor exposed.

---

## How to reproduce.

Requires only a PostgreSQL instance (13+). No Redis, no other dependencies.

`processor.js`:

```js
export default async function (job) {
	for (let i = 0; i < 20; i++) {
		await job.log(`line ${i}`);
	}
	return 'ok';
}
```

`repro.js`:

```js
import { Queue, Worker, createPostgresBackend } from 'bullmq';
import { fileURLToPath } from 'url';
import { dirname, join } from 'path';

const __dirname = dirname(fileURLToPath(import.meta.url));
const connection = { connectionString: process.env.PG_URL, schema: 'bullmq_repro', migrate: true };

const queue = new Queue('repro', { connection }, createPostgresBackend);
const worker = new Worker('repro', join(__dirname, 'processor.js'),
	{ connection, concurrency: 1 }, createPostgresBackend);

worker.on('error', err => console.error('WORKER ERROR:', err.message));

const done = new Promise((resolve, reject) => {
	worker.on('completed', resolve);
	worker.on('failed', (job, err) => reject(err));
});

await queue.add('j', {});
const job = await done;
console.log('logs stored:', (await queue.getJobLogs(job.id)).logs.length, 'of 20 expected');
await queue.obliterate({ force: true });
await worker.close();
await queue.close();
```

Run with `PG_URL=postgresql://user:pass@host/db node repro.js`.

Expected: 20 log lines stored, no errors. Actual: `duplicate key value violates unique constraint "job_log_pkey"` on the worker's `error` event, and fewer than 20 lines stored.

---

## Relevant log output

> This field uses `render: shell`, so GitHub adds the code fence itself. Paste the lines below **without** backticks.

```
error: duplicate key value violates unique constraint "job_log_pkey"
    at /app/node_modules/pg-pool/index.js:45:11
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async PostgresQueueBackend.query (/app/node_modules/bullmq/dist/cjs/postgres/postgres-queue-backend.js:364:20)
    at async PostgresQueueBackend.addLog (/app/node_modules/bullmq/dist/cjs/postgres/postgres-queue-backend.js:953:25)
    at async Child.msgHandler (/app/node_modules/bullmq/dist/cjs/classes/sandbox.js:40:41)
```

---

## Code of Conduct

Tick: **I agree to follow this project's Code of Conduct**

---

juneja-varun:
Fixes #4682.

I had a job calling `job.log()` several times in a row against the Postgres backend, and it kept failing with `duplicate key value violates unique constraint "job_log_pkey"` partway through - losing log lines with no way to avoid it, since even a fully sequential, awaited caller can still race (the sandbox forwards log messages to the parent without waiting for the previous write, so overlapping inserts happen regardless of how the processor itself is written).

`add_log.sql` picks the next index with a plain `COALESCE((SELECT MAX(idx)+1 ...), 0)` - no lock, so two overlapping inserts for the same job can read the same max and collide. An advisory lock taken earlier in the *same* statement doesn't fix this: under READ COMMITTED, a statement's snapshot is fixed when it starts, before the lock wait resolves, so a caller that was blocked still reads stale data once it's unblocked. I moved `addLog` off the single `.sql` file onto a dedicated connection running an explicit transaction - lock, then a fresh read, then the insert - mirroring the advisory-lock pattern the migrator already uses. Verified with a real 20-way concurrent `addLog` test against a local Postgres, which reproduces the exact reported error on unpatched code and passes cleanly with the fix.

One thing worth flagging: `add_log.sql` itself is unchanged and still has the described race - it's also copied verbatim to the Python/Elixir/.NET ports (per the `copy:sql:*` scripts), so those likely have the same bug and would need their own equivalent fix; I didn't touch it here since the Node backend no longer uses it for this operation and I don't have a way to verify a fix against the other ports.

### DIFF
--- a/src/postgres/postgres-queue-backend.ts
+++ b/src/postgres/postgres-queue-backend.ts
@@ -73,6 +73,9 @@ function bigintOrUndefined(value: string | null): number | undefined {
 /** Max events fetched per readEvents round-trip. */
 const EVENT_READ_BATCH = 100;
 
+/** Namespaces the `addLog` advisory lock away from the migrator's own lock key. */
+const ADD_LOG_ADVISORY_LOCK_KEY = 0x4c4f4747; // 'LOGG'
+
 /** Strips `undefined` properties so the JSON shape matches the Redis backend. */
 function removeUndefined<T extends Record<string, any>>(obj: T): T {
   for (const key of Object.keys(obj)) {
@@ -1244,13 +1247,9 @@ export class PostgresQueueBackend
     logRow: string,
     keepLogs?: number,
   ): Promise<number> {
-    let rows: { idx: string }[];
+    let idx: number;
     try {
-      ({ rows } = await this.run<{ idx: string }>('add_log', [
-        this.queueName,
-        jobId,
-        logRow,
-      ]));
+      idx = await this.insertLogAtomically(jobId, logRow);
     } catch (err: any) {
       // 23503 = foreign_key_violation: the job no longer exists.
       if (err && err.code === '23503') {
@@ -1258,7 +1257,7 @@ export class PostgresQueueBackend
       }
       throw err;
     }
-    const count = Number(rows[0].idx) + 1;
+    const count = idx + 1;
 
     if (keepLogs && count > keepLogs) {
       await this.run('trim_logs', [this.queueName, jobId, count - keepLogs]);
@@ -1266,6 +1265,42 @@ export class PostgresQueueBackend
     }
 
     return count;
+  }
+
+  // A lock taken inside the same statement as the MAX(idx) read wouldn't help:
+  // under READ COMMITTED, a statement's snapshot is fixed when it starts, before
+  // the lock wait resolves, so a blocked caller can still read stale data. The
+  // lock and the read need to be separate statements on one connection instead.
+  private async insertLogAtomically(
+    jobId: string,
+    logRow: string,
+  ): Promise<number> {
+    const client = await this.connection.pool.connect();
+    try {
+      await client.query('BEGIN');
+      try {
+        await client.query(
+          "SELECT pg_advisory_xact_lock($1, hashtext($2 || ':' || $3))",
+          [ADD_LOG_ADVISORY_LOCK_KEY, this.queueName, jobId],
+        );
+        const { rows } = await client.query<{ idx: string }>(
+          'SELECT COALESCE(MAX(idx) + 1, 0) AS idx FROM job_log WHERE queue = $1 AND job_id = $2',
+          [this.queueName, jobId],
+        );
+        const idx = Number(rows[0].idx);
+        await client.query(
+          'INSERT INTO job_log (queue, job_id, idx, row) VALUES ($1, $2, $3, $4)',
+          [this.queueName, jobId, idx, logRow],
+        );
+        await client.query('COMMIT');
+        return idx;
+      } catch (err) {
+        await client.query('ROLLBACK');
+        throw err;
+      }
+    } finally {
+      client.release();
+    }
   }
 
   async clearLogs(jobId: string, keepLogs?: number): Promise<void> {
--- a/tests/postgres/operations.test.ts
+++ b/tests/postgres/operations.test.ts
@@ -195,6 +195,29 @@ describe('PostgreSQL backend operations', () => {
     }
   });
 
+  it('assigns a unique idx to every log line under concurrent addLog calls', async () => {
+    const backend = newBackend();
+    try {
+      await backend.waitUntilReady();
+      const id = await backend.addJob(makeJob(), '');
+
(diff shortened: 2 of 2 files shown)


src/postgres/postgres-queue-backend.ts
```diff
@@ -1268,6 +1267,42 @@ export class PostgresQueueBackend
     return count;
   }
 
+  // A lock taken inside the same statement as the MAX(idx) read wouldn't help:
+  // under READ COMMITTED, a statement's snapshot is fixed when it starts, before
+  // the lock wait resolves, so a blocked caller can still read stale data. The
+  // lock and the read need to be separate statements on one connection instead.
+  private async insertLogAtomically(
```
manast, line 1274: The SQL code must be written in a standalone .sql file, please see the rest of the commands implementations. This is so that we can reuse the same code in all backends instead of having runtime specific code as in this case.

juneja-varun: Thanks for the feedback! I tried a standalone-.sql version first (lock + MAX(idx) read in one statement via a CTE), but it doesn't hold up — a statement's snapshot is fixed when it starts, not when the lock wait ends, so the read still sees stale data even after the lock is granted. Verified with a 20-way concurrency test — still got duplicate keys.

A counter-column approach would fix that in one statement too, but it needs a new migration, which your docs mark as a major-version change — didn't want to fold that into a bugfix PR without checking first.

I'd lean toward the schema migration as the appropriate fix, only tradeoff is major-version change — happy to redo it that way if you're on board with the version bump.

manast: Agreed that the lock and insert cannot share a statement snapshot. But our requirement is that SQL lives in shared files, not that the operation must be one statement. Can we put the advisory lock in its own SQL command, then execute it followed by the existing add_log.sql on the same connection within a READ COMMITTED transaction? That preserves shared SQL without a migration or major-version change.

juneja-varun: Good call — done. The lock is now its own `lock_job_log.sql`, executed right before the existing `add_log.sql` on the same connection/transaction. No migration needed, and the concurrency test still passes 20/20.

PR #4723, opened 2026-09-10
fix(bun): allow nested duplicate() on lazy Bun adapter

### LINKED ISSUE #4706: [Bug]: Bun adapter: duplicate() of a duplicate throws TypeError (regression from #4591)
### Version

v6.3.4

### Platform

NodeJS

### What happened?

- bullmq 6.3.4
- Bun 1.4.2 (macOS arm64)
- Bun adapter (`createBunRedisClient`), Redis 8.10.0

#4591 fixed `duplicate()` losing the connection target, and one level of `duplicate()` now works correctly. A second level throws before it sends anything:

```
TypeError: undefined is not an object (evaluating 'src.duplicate')
    at _duplicateRaw (dist/cjs/classes/bun-redis-client.js:391:20)
    at connect (dist/cjs/classes/bun-redis-client.js:223:35)
    at sendCommand (dist/cjs/classes/bun-redis-client.js:522:25)
    at bzpopmin (dist/cjs/classes/bun-redis-client.js:787:33)
```

### How to reproduce.

```ts
import { createBunRedisClient } from "bullmq";

const client = createBunRedisClient(new Bun.RedisClient("redis://localhost:6379"));
const ABSENT_KEY = "repro-4591:absent";

console.log("one duplicate  ->", await client.duplicate().bzpopmin(ABSENT_KEY, 1));
console.log("two duplicates ->", await client.duplicate().duplicate().bzpopmin(ABSENT_KEY, 1));
```

Run with `bun repro.ts` against any reachable Redis.

**Expected:** both lines print `null` — `bzpopmin` on a key that was never written blocks for the timeout and answers `null`.

**Actual:**

```
one duplicate  -> null
two duplicates -> TypeError: undefined is not an object (evaluating 'src.duplicate')
```

`duplicate()` captures the parent's raw client eagerly:

```js
// dist/cjs/classes/bun-redis-client.js:416
const parentRaw = this.raw;
const adapter = new BunRedisAdapter(undefined, {
  rawFactory: () => this._duplicateRaw(parentRaw),
});
```

A duplicate built this way has no `raw` until its own first `connect()` — that is the lazy `rawFactory` design #4591 introduced. So when the duplicate is itself duplicated, `this.raw` is still `undefined`, the grandchild's factory closes over `undefined`, and `_duplicateRaw(undefined)` dereferences `src.duplicate`.

The reconnect path already handles exactly this state:

```js
// dist/cjs/classes/bun-redis-client.js:194
const newRaw = this.raw
  ? await this._duplicateRaw(this.raw)
  : await this.rawFactory();
```

`duplicate()` is the one place missing that guard.

### Relevant log output

```shell
$ bun repro.ts

one duplicate  -> null

386 |      * Its native `duplicate()` is the only reliable way to clone the target and
387 |      * options; we fall back to URL-based reconstruction for exotic raw clients
388 |      * that don't implement it. See #4582.
389 |      */
390 |     async _duplicateRaw(src) {
391 |         if (typeof src.duplicate === 'function') {
                         ^
TypeError: undefined is not an object (evaluating 'src.duplicate')
      at _duplicateRaw (~/.bun/install/cache/bullmq@6.3.4@@@1/dist/cjs/classes/bun-redis-client.js:391:20)
      at connect (~/.bun/install/cache/bullmq@6.3.4@@@1/dist/cjs/classes/bun-redis-client.js:223:35)
      at sendCommand (~/.bun/install/cache/bullmq@6.3.4@@@1/dist/cjs/classes/bun-redis-client.js:522:25)
      at bzpopmin (~/.bun/install/cache/bullmq@6.3.4@@@1/dist/cjs/classes/bun-redis-client.js:787:33)
      at repro.ts:9:71

Bun v1.4.2 (macOS arm64)
```

### Code of Conduct

rome-xi:
`duplicate()` captured `this.raw` into the lazy `rawFactory`. A duplicate has no raw client until first connect, so `client.duplicate().duplicate()` closed over `undefined` and `_duplicateRaw` threw TypeError on `src.duplicate`.

Resolve the parent raw lazily via `_ensureRaw()`, matching the reconnect path that already falls back to `rawFactory` when raw is missing.

Fixes #4706


manast:
Thanks for the PR. If you would address the review comment we could move forward.

rome-xi:
Addressed the Copilot review: concurrent nested `duplicate()` / lazy `rawFactory` now shares an in-flight materialization promise so we don't create two native clients and overwrite an unclosed raw. Added a regression test. Ready for another look.

manast (approved):

PR #4789, opened 2026-09-20
fix(job-scheduler): detect predefined cron aliases in legacy repeatable keys

askalf:
## Summary
- Recognize cron-parser predefined aliases in legacy repeatable-job key suffixes.
- Preserve the existing structural and numeric-key guards; unknown, mixed-case, and embedded aliases remain rejected.
- Add regression coverage for all ten aliases and boundary controls.

## Decisions
An exact alias set matches cron-parser’s table without broadening classification to arbitrary `@`-prefixed scheduler IDs. Redis-backed tests were not run because Redis/Docker are unavailable.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


manast:
in v6 we have deliberately decided to not support legacy keys anymore: https://github.com/taskforcesh/bullmq/compare/v5.81.3...v6.0.0
"Legacy repeatable jobs and their APIs are removed. This includes the repeat option on Queue#add and Queue#addBulk, the Repeat class, Queue#getRepeatableJobs(), Queue#removeRepeatable(), and Queue#removeRepeatableByKey(). Migrate recurring jobs to Job Schedulers."

